PR #8, opened 2022-02-01
Add image build GH action

mikenairn:
Adds a GitHub action triggered on branch/tag creation that builds and pushes the operator, bundle and index images.

OLM Bundle version is always set to 0.0.0 for builds of 'main' branch.
Images built from the 'main' branch are also tagged as 'latest' and all images reference the 'latest' builds i.e. index(latest) ->
bundle(latest) -> operator(latest).

Tested here https://github.com/mikenairn/kuadrant-operator/actions

Example main branch build: https://github.com/mikenairn/kuadrant-operator/actions/runs/1778709855
Example tag build https://github.com/mikenairn/kuadrant-operator/actions/runs/1778795829

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -1,0 +1,137 @@
+name: Build Images
+
+on:
+  push:
+    branches: [ '*' ]
+    tags: [ '*' ]
+
+env:
+  IMG_TAGS: ${{ github.ref_name }}
+  IMG_REGISTRY_HOST: quay.io
+  IMG_REGISTRY_ORG: kuadrant
+  MAIN_BRANCH_NAME: main
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+      - name: Add latest tag
+        if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
+        id: add-latest-tag
+        run: |
+          echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
+      - name: Build Image
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: kuadrant-operator
+          tags: ${{ env.IMG_TAGS }}
+          dockerfiles: |
+            ./Dockerfile
+      - name: Push Image
+        if: ${{ !env.ACT }}
+        id: push-to-quay
+        uses: redhat-actions/push-to-registry@v2
+        with:
+          image: ${{ steps.build-image.outputs.image }}
+          tags: ${{ steps.build-image.outputs.tags }}
+          registry: ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}
+          username: ${{ secrets.IMG_REGISTRY_USERNAME }}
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -282,6 +282,10 @@ endif
 catalog-build: opm ## Build a catalog image.
 	$(OPM) index add --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
 
+.PHONY: catalog-generate
+catalog-generate: opm ## Generate a catalog/index Dockerfile.
+	$(OPM) index add --generate --container-tool docker --mode semver --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
```
mikenairn, line 287: This generates an index.Dockerfile rather than triggering a docker build so we can instead build it using the same process as the others.

.github/workflows/build-images.yaml
```diff
@@ -0,0 +1,137 @@
...
+        id: add-latest-tag
+        run: |
+          echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
+      - name: Run make bundle
+        if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
+        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.ref_name }}
+      - name: Run make bundle (main)
+        if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
+        run: make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=latest VERSION=0.0.0
+      - name: Git diff
```
eguzki, line 70: should the workflow fail if git diff is not empty?

.github/workflows/build-images.yaml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Build Image
+        id: build-image
+        uses: redhat-actions/buildah-build@v2
+        with:
+          image: kuadrant-operator-bundle
+          tags: ${{ env.IMG_TAGS }}
+          dockerfiles: |
+            ./bundle.Dockerfile
+      - name: Push Image
+        if: ${{ !env.ACT }}
```
eguzki, line 81: what is the purpose of this condition? what is `env.ACT`?

eguzki (approved):
good work! :medal_military:

.github/workflows/build-images.yaml
mikenairn: Not for main, but for anything else, maybe/probably. We are modifying the VERSION to always be "0.0.0" and forcing the latest tag when building main, so the git diff should show changes. I put this here so i could check what it was doing, but you are right, we might want to consider failing it if there are unexpected changes for release tags or branches.

.github/workflows/build-images.yaml
mikenairn: I've been using [act](https://github.com/nektos/act) to test these actions locally. It adds this env var by default, and it's just to stop it doing any push locally when testing.

tbh, act didn't really work well for this action because of the buildah/podman dependencies (wasted a lot of time trying to get it to work), but it works fine for testing the test action, you can use the make command to try it `make act-pull-request-jobs`

PR #17, opened 2022-05-11
Istio ext authz service config

eguzki:
### what

Authorino set up in Istio as a Auth GRPC service 

Fixes https://github.com/Kuadrant/kuadrant-controller/issues/149

* add Gateway API CRDs dependency
* istio install makefile targets using the operator
* Upgrade installed Istio to 1.13.2. Isito 1.12.X has an [issue](https://github.com/istio/istio/issues/36114).
* Implemented in the controller:
  * setup logger: support for log levels (`debug` →  `info` →  `error`) and configurable log modes: `production`, `development`
  * status conditions: `Ready` condition
  * reconcile Istio mesh configuration for the kuadrant external authorizer. When the kuadrant CR is deleted, the configuration is cleaned up. Currently, the external authorizer address is hardcoded to 
```yaml
- envoyExtAuthzGrpc:
    port: 50051
    service: authorino-authorino-authorization.kuadrant-system.svc.cluster.local
  name: kuadrant-authorization
```

  * The name/NS of the IstioOperator CR used to configure external authorizer is read from en var

|   | Environemnt Variable Name | Default Value |
| --- | --- | --- | 
|  IstioOperator name | `ISTIOOPERATOR_NAME` | `istiocontrolplane` |
|  IstioOperator namespace | `ISTIOOPERATOR_NAMESPACE` | `istio-system` |

### verification steps

create kuadrant cluster locally with kind

```
```

Check istio config does not have any kuadrant's external authorizer. For that, the `istiocontrolplane` named IstioOperator CR is checked:
```
k get istiooperator istiocontrolplane -n istio-system -o jsonpath='{.spec.meshConfig.extensionProviders}'
```
The result should be empty.

Create empty Kuadrant CR object

```yaml
kubectl apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

Check istio config **does** have kuadrant¡s external authorizer. For that, the `istiocontrolplane` named IstioOperator CR is checked:
```yaml
k get istiooperator istiocontrolplane -n istio-system -o jsonpath='{.spec.meshConfig.extensionProviders}'
- envoyExtAuthzGrpc:
    port: 50051
    service: authorino-authorino-authorization.kuadrant-system.svc.cluster.local
  name: kuadrant-authorization
```

Delete the kuadrant CR
```
k delete kuadrant kuadrant-sample
```
Check istio config does no longer have any kuadrant's external authorizer. For that, the `istiocontrolplane` named IstioOperator CR is checked:
```
k get istiooperator istiocontrolplane -n istio-system -o jsonpath='{.spec.meshConfig.extensionProviders}'
[]
```
The result should be empty (or empty array)


eguzki:

pkg/log/log.go
```diff
@@ -0,0 +1,142 @@
...
+	"k8s.io/klog/v2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+var (
+	// Log is the base logger
+	Log logr.Logger = ctrllog.NullLogger{}
+)
```
rahulanand16nov, line 20: I don't think we require an upgrade here, right?

rahulanand16nov (approved):
LGTM!

controllers/kuadrant_controller.go
```diff
@@ -36,24 +48,248 @@ type KuadrantReconciler struct {
...
+	}
+
+	//meshConfig:
+	//    extensionProviders:
+	//      - envoyExtAuthzGrpc:
+	//          port: POST
+	//          service: AUTHORINO SERVICE
+	//        name: kuadrant-authorization
+
+	if iop.Spec.MeshConfig == nil {
```
guicassolato, line 205: Technically, the IstioOperator `Spec` can be `nil`. I know we control the resource, but would this panic if somebody deleted the spec and the operator tried to reconcile it back?

controllers/kuadrant_controller.go
```diff
@@ -36,24 +48,248 @@ type KuadrantReconciler struct {
...
+	// IstioOperator
+	//
+	//meshConfig:
+	//    extensionProviders:
+	//      - envoyExtAuthzGrpc:
+	//          port: POST
+	//          service: AUTHORINO SERVICE
+	//        name: kuadrant-authorization
+
+	extensionProvidersObj, ok := iop.Spec.MeshConfig["extensionProviders"]
```
guicassolato, line 266: My previous [comment](https://github.com/Kuadrant/kuadrant-operator/pull/17/files#r875762922) actually applies better here.

controllers/kuadrant_status.go
```diff
@@ -0,0 +1,85 @@
...
+
+const (
+	ReadyConditionType string = "Ready"
+)
+
+func (r *KuadrantReconciler) reconcileStatus(ctx context.Context, kObj *kuadrantv1beta1.Kuadrant, specErr error) (ctrl.Result, error) {
+	logger := logr.FromContext(ctx)
+	newStatus := r.calculateStatus(kObj, specErr)
+
+	equalStatus := kObj.Status.Equals(newStatus, logger)
```
guicassolato, line 26: I suppose this is to prevent unnecessary updates on the status, where a minimum spottable diff would be the value of `LastTransitionTime` of a particular condition, is that right?

Is it worth the cost of checking while:
1. leader election is enabled and therefore only one replica at a time will reconcile the resource?
2. status update is an idempotent operation?

Please, no need to change anything. I'm just trying to understand the flow.

controllers/kuadrant_status.go
```diff
@@ -0,0 +1,85 @@
...
+	meta.SetStatusCondition(&newStatus.Conditions, *availableCond)
+
+	return newStatus
+}
+
+func (r *KuadrantReconciler) readyCondition(specErr error) *metav1.Condition {
+	cond := &metav1.Condition{
+		Type:    ReadyConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  "KReady",
```
guicassolato, line 74: Typo?

```suggestion
		Reason:  "Ready",
```

controllers/kuadrant_status.go
```diff
@@ -0,0 +1,85 @@
...
+
+	availableCond := r.readyCondition(specErr)
+
+	meta.SetStatusCondition(&newStatus.Conditions, *availableCond)
+
+	return newStatus
+}
+
+func (r *KuadrantReconciler) readyCondition(specErr error) *metav1.Condition {
+	cond := &metav1.Condition{
```
guicassolato, line 71: `metav1.Condition.LastTransitionTime` is annotated as `+kubebuilder:validation:Required`. Do we need to add it here?

https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

main.go
```diff
@@ -18,37 +18,63 @@ package main
...
+		log.WriteTo(os.Stdout),
+	).WithName("kuadrant-operator")
+	log.SetLogger(logger)
+}
+
+func printControllerMetaInfo() {
+	setupLog := log.Log
+
+	setupLog.Info(fmt.Sprintf("go version: %s", runtime.Version()))
+	setupLog.Info(fmt.Sprintf("go os/arch: %s/%s", runtime.GOOS, runtime.GOARCH))
```
guicassolato, line 69: Nice touch! I might steel this idea for [kuadrant/authorino](https://github.com/kuadrant/authorino) now that we're building for multi-arch. 🙂

pkg/log/log.go
eguzki: we will when upgrading logr to 1.X

controllers/kuadrant_controller.go
eguzki: nope. Spec is not a pointer in the Go struct. So it always exists

```
Spec   KuadrantSpec   `json:"spec,omitempty"`
```

eguzki: Just for the case, I checked it. 

I have deleted the spec. The k8s API serves without the spec:
```yaml
kind: Kuadrant
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kuadrant.kuadrant.io/v1beta1","kind":"Kuadrant","metadata":{"annotations":{},"name":"kuadrant-sample","namespace":"default"},"spec":{}}
  creationTimestamp: "2022-05-18T13:47:47Z"
  finalizers:
    - kuadrant.kuadrant.io/finalizer
  generation: 2
  name: kuadrant-sample
  namespace: default
  resourceVersion: "5880"
  uid: 4d26b8ff-5669-48de-a9a5-5487cc079495
status:
  conditions:
    - lastTransitionTime: "2022-05-18T13:47:47Z"
      message: Kuadrant is ready
      reason: KReady
      status: "True"
      type: Ready
  observedGeneration: 2
```

But when it is read by the controller, spec object has been initialized with zero value. This is a json serialization of the go struct:
```json
{
  "kind": "Kuadrant",
  "apiVersion": "kuadrant.kuadrant.io/v1beta1",
  "metadata": {
    "name": "kuadrant-sample",
    "namespace": "default",
    "uid": "4d26b8ff-5669-48de-a9a5-5487cc079495",
    "resourceVersion": "6681",
    "generation": 2,
    "creationTimestamp": "2022-05-18T13:47:47Z",
    "annotations": {
      "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"kuadrant.kuadrant.io/v1beta1\",\"kind\":\"Kuadrant\",\"metadata\":{\"annotations\":{},\"name\":\"kuadrant-sample\",\"namespace\":\"default\"},\"spec\":{}}\n",
      "something": "jdjd"
    },
    "finalizers": [
      "kuadrant.kuadrant.io/finalizer"
    ],
  },
  "spec": {},
  "status": {
    "observedGeneration": 2,
    "conditions": [
      {
        "type": "Ready",
        "status": "True",
        "lastTransitionTime": "2022-05-18T13:47:47Z",
        "reason": "KReady",
        "message": "Kuadrant is ready"
      }
    ]
  }
}
```

main.go
eguzki: :police_officer:

controllers/kuadrant_controller.go
eguzki: updated

controllers/kuadrant_controller.go
eguzki: updated

guicassolato (approved):

controllers/kuadrant_status.go
eguzki: To prevent unnecessary updates on the status, specifically related to the `LastTransitionTime` field, conditions are first copied from the existing CR and then desired values applied. After that, they are compared. The status is only updated when there are changes. 

> leader election is enabled and therefore only one replica at a time will reconcile the resource?

I do not fully understand this point. The controller does not expect anyone to change the status. But even if it changes, the controller will set its desired state.

> status update is an idempotent operation?

Yes it is. Identical status requests should always turn into a identical status objects in the CR.

guicassolato: > leader election is enabled and therefore only one replica at a time will reconcile the resource?
>> I do not fully understand this point. The controller does not expect anyone to change the status. But even if it changes, the controller will set its desired state.

When leader election is enabled in a manager, no matter how many replicas of the operator are running, only the leader will perform the reconciliation tasks under this manager; therefore, also the status update. IOW, I think there's no concurrency to worry about. (Unless reconcilers are invoked in concurrent goroutines. Are they? 🤔)

> status update is an idempotent operation?
>> Yes it is.

So why carrying checking for changes in the subresource? Because sorting the array and marshalling 2 JSON objects cost less than sending the update request? If the update happens in the background (because we don't have to wait until it finishes to return from reconciling, right?), wouldn't it be better to just trust idempotence and skip the check?

guicassolato: In [Authorino](https://github.com/Kuadrant/authorino/blob/20ed001f50544ec25e00b9a62e15073b46798274/controllers/auth_config_status_updater.go#L74) and in [Authorino Operator](https://github.com/Kuadrant/authorino-operator/blob/2d97c6044b82c7b70e850404e382bba6ebf60cfa/controllers/authorino_controller.go#L739) we went with a different approach, where we trust idempotence of the status update operation and leave the rest to the API server.

But it's your call.

> one update request has much more cost (I would say several orders of magnitude) than serializing in memory struct objects which are not expected to be "big"

Maybe I wasn't clear before. You don't have to wait for the state update to finish. You can fork a goroutine to handle the update and return immediately. So, from the perspective of freeing the reconciler/closing the connection, performing the update without the check is theoretically faster.

On the other hand, I totally understand if you don't want to handle this in the background (it could fail!)

And, as you said, sorting two arrays of 5-6 elements each + serializing two small structs as JSON, we're talking nanoseconds might not worth saving.

Anyway, the PR LGTM. As I said before, I was just trying to understand the flow. Not suggesting changing anything.

PR #18, opened 2022-05-23
Deploy limitador authorino and kuadrant-controller

eguzki:
### What

Fixes https://github.com/Kuadrant/kuadrant-controller/issues/148

- [x] Deploy Limitador in the same namespace as the kuadrant CR
- [x] kuadrant controller and kuadrant API (AuthPolicy and RatelimitPolocy CRDs) are no longer deployed as part of the operator deployment. A kuadrant CR is required to deploy the kuadrant control plane and kuadrant API. 
- [x] Deploy authorino in the same namespace as the kuadrant CR
- [x] Kuadrant CR status conditions
- [x] Clean up when Kuadrant CR is deleted

### Verification Steps
* Deploy kuadrant operator (and operator dependencies) manually
```
```

* Deploy empty kuadrant CR
```yaml
kubectl apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```
wait until kuadrant CR status reports it is ready
```yaml
k wait --for=condition=ready kuadrant/kuadrant-sample --timeout=-1s
kuadrant.kuadrant.kuadrant.io/kuadrant-sample condition met

k get kuadrant kuadrant-sample -o yaml | yq_k8s_clean 
kind: Kuadrant
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kuadrant.kuadrant.io/v1beta1","kind":"Kuadrant","metadata":{"annotations":{},"name":"kuadrant-sample","namespace":"default"},"spec":{}}
  creationTimestamp: "2022-05-23T19:38:38Z"
  finalizers:
    - kuadrant.kuadrant.io/finalizer
  generation: 1
  name: kuadrant-sample
  namespace: default
  resourceVersion: "1670"
  uid: f6f0e2c8-824a-4053-8891-adeea22d26a3
spec: {}
status:
  conditions:
    - lastTransitionTime: "2022-05-23T19:38:59Z"
      message: Kuadrant is ready
      reason: Ready
      status: "True"
      type: Ready
  observedGeneration: 1
```
Check limitador, authorino and the kuadrant controller are running in the namespace where the kuadrant CR was deployed
```
k get pods 
NAME                                           READY   STATUS    RESTARTS   AGE
authorino-897fb4b87-5m77m                      1/1     Running   0          3m2s
kuadrant-controller-manager-7f59f745c5-6v8d6   1/1     Running   0          3m3s
limitador-6c89f69875-wf2bn                     1/1     Running   0          3m4s
```

Check that the CRDs have also been deployed:

```
k get crds | grep apim.kuadrant.io
authpolicies.apim.kuadrant.io                 2022-05-23T19:38:38Z
ratelimitpolicies.apim.kuadrant.io            2022-05-23T19:38:38Z
```

Dev Notes: 
* Followed same approach as the Istio operator. With the just the operator deployed, the control plane (istiod) and Istio API are not available. The control plane and the API are available after a istio installation via the IstioOperator CR


config/dependencies/gateway-api/gateway/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+---
+# Adds namespace to all resources.
+namespace: istio-system
+resources:
+- gateway.yaml
```
guicassolato, line 5: Have you considered the following?

```suggestion
- github.com/kubernetes-sigs/gateway-api/config/crd?ref=master
```

config/dependencies/gateway-api/gateway/gateway.yaml
```diff
@@ -0,0 +1,19 @@
+---
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: Gateway
+metadata:
+  labels:
+    istio: ingressgateway
+  name: istio-ingressgateway
```
guicassolato, line 7: Does it have to be called "istio-*"? Now that the Gateway API resource is being used, it looks like Istio is an implementation detail that could perhaps be abstracted at this level. WDYT?

eguzki: In the addresses, the value `value: istio-ingressgateway.istio-system.svc.cluster.local` is a must because this deployment relies on istio ingress gateway. 

I agree that the name can be anything, like kuadrant-gateway. I would not say, though, that Istio is an implementation detail. I do not think that kuadrant can, or should, abstract entirely Istio. I see kuadrant more of a Istio extension rather than a abstraction layer. This is out of context :) But dropping here anyway.

Anyway, this is not part of the kuadrant-operator release. This file belongs to the testing environment setup configuration. So I think it is ok as it is.

guicassolato: Fair enough. Thanks for explaining, @eguzki.

guicassolato (commented):
Minor thing...

After `make kind-create-kuadrant-cluster`, because the image exported is `quay.io/kuadrant/kuadrant-operator:dev`, the following uncommitted change remains:

```
$ git diff
diff --git a/config/manager/kustomization.yaml b/config/manager/kustomization.yaml
index b038c37..2550fb7 100644
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -13,4 +13,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: quay.io/kuadrant/kuadrant-operator
-  newTag: latest
+  newTag: dev
```

Maybe the `deploy` target could roll back the image tag to the default `latest` after `kustomize build`?

```mk
deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
	$(KUSTOMIZE) build config/deploy | kubectl apply -f -
	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMAGE_TAG_BASE}:latest
```

eguzki:
> Maybe the `deploy` target could roll back the image tag to the default `latest` after `kustomize build`?
> 
> ```makefile
> deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
> 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
> 	$(KUSTOMIZE) build config/deploy | kubectl apply -f -
> 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMAGE_TAG_BASE}:latest
> ```

Thanks for the tip! I will add the image restore statement

guicassolato (approved):

PR #22, opened 2022-05-31
Propagating Limitador's env vars

didierofrivia:
This PR makes it possible to inject the `Name` and `Namespace` to the _Kuadrant Controller_ Deployment object.

It also patches the `RoleBinding` and `ClusterRoleBinding` matching the _Kuadrant_ namespace.

Closes https://github.com/Kuadrant/kuadrant-controller/issues/157
Part of https://github.com/Kuadrant/kuadrant-controller/issues/108

### PR, verification steps

Deploy kuadrant operator (and operator dependencies) manually
```bash
```

Deploy empty Kuadrant CR
```bash
kubectl apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

Wait until Kuadrant CR status reports it is ready
```bash
kubectl wait --for=condition=ready kuadrant/kuadrant-sample --timeout=-1s
kuadrant.kuadrant.kuadrant.io/kuadrant-sample condition met
```
Deploy Service
```bash
kubectl apply -f - <<EOF
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  selector:
    matchLabels:
      app: toystore
  template:
    metadata:
      labels:
        app: toystore
    spec:
      containers:
        - name: toystore
          env:
            - name: PORT
              value: "3000"
          ports:
            - containerPort: 3000
              name: http
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: toystore
spec:
  selector:
    app: toystore
  ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 3000
EOF

```
```
deployment.apps/toystore created
service/toystore created
```

Create HTTPRoute to configure routing to the toustore service
```bash
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80

EOF
```
```
httproute.gateway.networking.k8s.io/toystore created
```
Port forward the istio gateway service in order to reach it
```bash
kubectl port-forward -n istio-system service/istio-ingressgateway 8080:80 &
```

Check if you can reach the toystore service
```bash
curl -H 'Host: api.toystore.com' http://127.0.0.1:8080/toy

Handling connection for 8080
{
  "method": "GET",
  "path": "/toy",
  "query_string": null,
  "body": "",
  "headers": {
    "HTTP_HOST": "api.toystore.com",
    "HTTP_USER_AGENT": "curl/7.79.1",
    "HTTP_ACCEPT": "*/*",
    "HTTP_X_FORWARDED_FOR": "10.244.0.10",
    "HTTP_X_FORWARDED_PROTO": "http",
    "HTTP_X_ENVOY_INTERNAL": "true",
    "HTTP_X_REQUEST_ID": "8e7bbf90-3edd-4cbf-8bb7-7594a9d7933b",
    "HTTP_X_ENVOY_DECORATOR_OPERATION": "toystore.default.svc.cluster.local:80/toy((\\/).*)?",
    "HTTP_X_ENVOY_PEER_METADATA": "ChQKDkFQUF9DT05UQUlORVJTEgIaAAoaCgpDTFVTVEVSX0lEEgwaCkt1YmVybmV0ZXMKGQoNSVNUSU9fVkVSU0lPThIIGgYxLjEzLjIKvwMKBkxBQkVMUxK0AyqxAwodCgNhcHASFhoUaXN0aW8taW5ncmVzc2dhdGV3YXkKEwoFY2hhcnQSChoIZ2F0ZXdheXMKFAoIaGVyaXRhZ2USCBoGVGlsbGVyCjYKKWluc3RhbGwub3BlcmF0b3IuaXN0aW8uaW8vb3duaW5nLXJlc291cmNlEgkaB3Vua25vd24KGQoFaXN0aW8SEBoOaW5ncmVzc2dhdGV3YXkKGQoMaXN0aW8uaW8vcmV2EgkaB2RlZmF1bHQKMAobb3BlcmF0b3IuaXN0aW8uaW8vY29tcG9uZW50EhEaD0luZ3Jlc3NHYXRld2F5cwohChFwb2QtdGVtcGxhdGUtaGFzaBIMGgo3ZmQ1NjhmYzk5ChIKB3JlbGVhc2USBxoFaXN0aW8KOQofc2VydmljZS5pc3Rpby5pby9jYW5vbmljYWwtbmFtZRIWGhRpc3Rpby1pbmdyZXNzZ2F0ZXdheQovCiNzZXJ2aWNlLmlzdGlvLmlvL2Nhbm9uaWNhbC1yZXZpc2lvbhIIGgZsYXRlc3QKIgoXc2lkZWNhci5pc3Rpby5pby9pbmplY3QSBxoFZmFsc2UKGgoHTUVTSF9JRBIPGg1jbHVzdGVyLmxvY2FsCi8KBE5BTUUSJxolaXN0aW8taW5ncmVzc2dhdGV3YXktN2ZkNTY4ZmM5OS1oaG0ydAobCglOQU1FU1BBQ0USDhoMaXN0aW8tc3lzdGVtCl0KBU9XTkVSElQaUmt1YmVybmV0ZXM6Ly9hcGlzL2FwcHMvdjEvbmFtZXNwYWNlcy9pc3Rpby1zeXN0ZW0vZGVwbG95bWVudHMvaXN0aW8taW5ncmVzc2dhdGV3YXkKFwoRUExBVEZPUk1fTUVUQURBVEESAioACicKDVdPUktMT0FEX05BTUUSFhoUaXN0aW8taW5ncmVzc2dhdGV3YXk=",
    "HTTP_X_ENVOY_PEER_METADATA_ID": "router~10.244.0.10~istio-ingressgateway-7fd568fc99-hhm2t.istio-system~istio-system.svc.cluster.local",
    "HTTP_X_ENVOY_ATTEMPT_COUNT": "1",
    "HTTP_X_B3_TRACEID": "db0b968ee02ef57fcaf0ad1bc7448d00",
    "HTTP_X_B3_SPANID": "caf0ad1bc7448d00",
    "HTTP_X_B3_SAMPLED": "0",
    "HTTP_VERSION": "HTTP/1.1"
  },
  "uuid": "7de70ba8-3447-4750-a7e9-6ffabf861b0f"
}%

```

Apply a RLP
```bash
kubectl apply -f - <<EOF
---
apiVersion: apim.kuadrant.io/v1alpha1
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
        - paths: ["/toy"]
          methods: ["GET"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: get-operation
                descriptor_value: "1"
      limits:
        - conditions:
            - "get-operation == 1"
          maxValue: 2
          seconds: 5
          variables: []
EOF
```

Rollout Limitador. // This is needed for now until a fix is merged in Limitador repo
```bash
kubectl rollout restart deployment/limitador
```

Try again curling the service, you should be ratelimited to 2 request per 5 seconds
```bash
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H "Host: api.toystore.com" http://localhost:8080/toy | egrep --color "\b(429)\b|$”;
```
Voilá! Profit!
```
200
200
429
429
429
200
200
429
429
429
```

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -21,13 +21,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-
 	"github.com/go-logr/logr"
 	authorinov1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
 	istioapiv1alpha1 "istio.io/api/operator/v1alpha1"
 	iopv1alpha1 "istio.io/istio/operator/pkg/apis/istio/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -43,8 +43,14 @@ import (
 )
 
 const (
-	kuadrantFinalizer = "kuadrant.kuadrant.io/finalizer"
-	extAuthorizerName = "kuadrant-authorization"
+	kuadrantFinalizer     = "kuadrant.kuadrant.io/finalizer"
+	extAuthorizerName     = "kuadrant-authorization"
+	envLimitadorNamespace = "LIMITADOR_NAMESPACE"
+	envLimitadorName      = "LIMITADOR_NAME"
+)
+
+var (
+	limitadorName = common.FetchEnv(envLimitadorName, "limitador")
 )
 
 // KuadrantReconciler reconciles a Kuadrant object
@@ -344,7 +350,7 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
 			APIVersion: "limitador.kuadrant.io/v1alpha1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "limitador",
+			Name:      limitadorName,
 			Namespace: kObj.Namespace,
 		},
 		Spec: limitadorv1alpha1.LimitadorSpec{},
@@ -389,14 +395,28 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj 
 			return err
 		}
 
-		k8sObjKind := k8sObj.DeepCopyObject().GetObjectKind()
+		var newObj client.Object
+		newObj = k8sObj
 
-		err = r.Client().Create(ctx, k8sObj)
-		logger.V(1).Info("create resource", "GKV", k8sObjKind.GroupVersionKind(), "name", k8sObj.GetName(), "error", err)
+		switch obj := k8sObj.(type) {
+		case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars
+			obj.Spec.Template.Spec.Containers[0].Env = append(
+				obj.Spec.Template.Spec.Containers[0].Env,
+				v1.EnvVar{Name: envLimitadorNamespace, Value: kObj.Namespace},
+				v1.EnvVar{Name: envLimitadorName, Value: limitadorName},
+			)
+			newObj = obj
+		default:
+		}
+		newObjCloned := newObj.DeepCopyObject()
+		err = r.Client().Create(ctx, newObj)
+
+		k8sObjKind := newObjCloned.GetObjectKind()
+		logger.V(1).Info("create resource", "GKV", k8sObjKind.GroupVersionKind(), "name", newObj.GetName(), "error", err)
 		if err != nil {
 			if apierrors.IsAlreadyExists(err) {
 				// Omit error
-				logger.Info("Already exists", "GKV", k8sObjKind.GroupVersionKind(), "name", k8sObj.GetName())
+				logger.Info("Already exists", "GKV", k8sObjKind.GroupVersionKind(), "name", newObj.GetName())
 			} else {
 				return err
 			}
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -28,6 +28,7 @@ import (
 	iopv1alpha1 "istio.io/istio/operator/pkg/apis/istio/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -405,6 +406,25 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj 
 				v1.EnvVar{Name: envLimitadorNamespace, Value: kObj.Namespace},
 				v1.EnvVar{Name: envLimitadorName, Value: limitadorName},
 			)
+			newObj = obj
+		// TODO: DRY the following 2 case switches
+		case *rbacv1.RoleBinding:
+			if obj.Name == "kuadrant-leader-election-rolebinding" {
+				for i, subject := range obj.Subjects {
+					if subject.Name == "kuadrant-controller-manager" {
+						obj.Subjects[i].Namespace = kObj.Namespace
+					}
+				}
+			}
+			newObj = obj
+		case *rbacv1.ClusterRoleBinding:
+			if obj.Name == "kuadrant-manager-rolebinding" {
+				for i, subject := range obj.Subjects {
+					if subject.Name == "kuadrant-controller-manager" {
+						obj.Subjects[i].Namespace = kObj.Namespace
+					}
+				}
+			}
 			newObj = obj
 		default:
 		}


controllers/kuadrant_controller.go
```diff
@@ -386,9 +393,19 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
 			return err
 		}
 
+		switch obj := k8sObj.(type) {
+		case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars
+			obj.Spec.Template.Spec.Containers[0].Env = []v1.EnvVar{
+				{Name: "LIMITADOR_NAMESPACE", Value: kObj.Namespace, ValueFrom: &v1.EnvVarSource{}},
```
eguzki, line 399: you can leave `ValueFrom` empty

controllers/kuadrant_controller.go
```diff
@@ -386,9 +393,19 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
...
+				{Name: envLimitadorServiceName, Value: limitadorServiceName, ValueFrom: &v1.EnvVarSource{}},
+				{Name: envLimitadorServiceGrpcPort, Value: limitadorServiceGrpcPort, ValueFrom: &v1.EnvVarSource{}},
+			}
+			err = r.Client().Create(ctx, obj)
```
eguzki, line 403: not a big deal, but can you try to call create one time? For me it feels like objects are conditionally updated and then created. So something like this would work?

```go
var newObj client.Object                                                                    
// should not be copied, just another reference                                             
newObj = k8sObj                                                                             
                                                                                            
switch obj := k8sObj.(type) {                                                               
case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars         
    obj.Spec.Template.Spec.Containers[0].Env = []v1.EnvVar{                                 
        {Name: "LIMITADOR_NAMESPACE", Value: kObj.Namespace, ValueFrom: &v1.EnvVarSource{}},
        {Name: envLimitadorServiceName, Value: limitadorServiceName, ValueFrom: &v1.EnvVarSource{}},
        {Name: envLimitadorServiceGrpcPort, Value: limitadorServiceGrpcPort, ValueFrom: &v1.EnvVarSource{}},
    }                                                                                       
    newObj = obj                                                                            
default:                                                                                    
}                                                                                           
                                                                                            
k8sObjKind := newObj.DeepCopyObject().GetObjectKind()                                       
err = r.Client().Create(ctx, newObj)                                                        
logger.V(1).Info("create resource", "GKV", k8sObjKind.GroupVersionKind(), "name", k8sObj.GetName(), "error", err)

```

eguzki (commented):
Looking super nice

controllers/kuadrant_controller.go
```diff
@@ -386,14 +394,27 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
...
 		}
 
-		k8sObjKind := k8sObj.DeepCopyObject().GetObjectKind()
+		var newObj client.Object
+		newObj = k8sObj
+
+		switch obj := k8sObj.(type) {
+		case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars
+			obj.Spec.Template.Spec.Containers[0].Env = []v1.EnvVar{
+				{Name: envLimitadorNamespace, Value: kObj.Namespace},
```
eguzki, line 403: Should we consider to pass only Limitador CR name/NS?? According to [this](https://github.com/Kuadrant/limitador-operator/issues/21) and if we implement that, the kuadrant controller (the control plane not the operator) only needs to know the location of the Limitador CR. It will be the Limitador Operator the one responsible of injecting  limits into pods.

controllers/kuadrant_controller.go
```diff
@@ -386,14 +392,26 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
 			return err
 		}
 
-		k8sObjKind := k8sObj.DeepCopyObject().GetObjectKind()
+		var newObj client.Object
+		newObj = k8sObj
+
+		switch obj := k8sObj.(type) {
+		case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars
+			obj.Spec.Template.Spec.Containers[0].Env = []v1.EnvVar{
```
eguzki, line 400: What about appending instead if assigning?

```go
obj.Spec.Template.Spec.Containers[0].Env = append(obj.Spec.Template.Spec.Containers[0].Env, {Name: envLimitadorNamespace, Value: kObj.Namespace},{Name: envLimitadorName, Value: limitadorName})
```
Even if `obj.Spec.Template.Spec.Containers[0].Env` is `nil`, `append` operation works as expected.

controllers/kuadrant_controller.go
```diff
@@ -386,14 +392,26 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
...
+				{Name: envLimitadorName, Value: limitadorName},
+			}
+			newObj = obj
+		default:
+		}
+		err = r.Client().Create(ctx, newObj)
 
-		err = r.Client().Create(ctx, k8sObj)
-		logger.V(1).Info("create resource", "GKV", k8sObjKind.GroupVersionKind(), "name", k8sObj.GetName(), "error", err)
+		k8sObjKind := newObj.DeepCopyObject().GetObjectKind()
```
eguzki, line 409: deep copy before creating the object. For an unknown reason, the go client removes type info of the object being passed to the `Create` method.

eguzki (commented):
have you tested manually e2e?

* Deploy kuadrant with the operator. 
* Deploy RLP
* Rate limit is happening

controllers/kuadrant_controller.go
```diff
@@ -386,14 +392,28 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
...
+		case *appsv1.Deployment: // If it's a Deployment obj, it adds the required env vars
+			obj.Spec.Template.Spec.Containers[0].Env = append(
+				obj.Spec.Template.Spec.Containers[0].Env,
+				v1.EnvVar{Name: envLimitadorNamespace, Value: kObj.Namespace},
+				v1.EnvVar{Name: envLimitadorName, Value: limitadorName},
+			)
+			newObj = obj
+		default:
+		}
+		newObj.DeepCopyObject()
```
eguzki, line 408: you need to store the cloned object

```go
cloned := newObj.DeepCopyObject()
```

then

```
k8sObjKind := cloned.GetObjectKind()
```

controllers/kuadrant_controller.go
```diff
@@ -389,14 +396,47 @@ func (r *KuadrantReconciler) createOnlyInKuadrantNSCb(ctx context.Context, kObj
 			return err
 		}
 
-		k8sObjKind := k8sObj.DeepCopyObject().GetObjectKind()
+		var newObj client.Object
+		newObj = k8sObj
```
alexsnaps, line 400: I must be missing something here, but there is a [lot](https://github.com/Kuadrant/kuadrant-operator/pull/22/files#diff-53e20feafdf0308bc6f15e87b2dc63acd05e732a6bf70bc5247745edd59b1e4cR409) [of](53e20feafdf0308bc6f15e87b2dc63acd05e732a6bf70bc5247745edd59b1e4cR419) [copying](53e20feafdf0308bc6f15e87b2dc63acd05e732a6bf70bc5247745edd59b1e4cR419) going on here. 
Is the `switch obj := k8sObj.(type)` a copy itself? Trying to understand… but I wonder why not use the `default:` if a "default" copy has to happen? That's if they are all different… I guess I need to refresh my golang

alexsnaps (approved):

controllers/kuadrant_controller.go
didierofrivia: I know, it's a bit misleading the `default` case, its block will be executed if none of the other cases match.

didierofrivia: `switch obj := k8sObj.(type)` will only live within the switch block, so we need to copy it to a different `var`. 

I'd love the swith case to include a `finally` clause, that will be executed at the very end, thus we could avoid a lot of repetition

PR #37, opened 2022-08-29
Automate CSV generation

didierofrivia:
This PR makes it possible to automate the generation of the ClusterServiceVersion image, bundle version and its tag. It also updates the maintainers list.

### DIFF
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -14,8 +14,12 @@ metadata:
         }
       ]
     capabilities: Basic Install
+    categories: Integration & Delivery
+    containerImage: quay.io/kuadrant/kuadrant-operator:latest
     operators.operatorframework.io/builder: operator-sdk-v1.22.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
+    repository: https://github.com/Kuadrant/kuadrant-operator
+    support: kuadrant
   name: kuadrant-operator.v0.0.0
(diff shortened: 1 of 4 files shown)


bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -414,9 +418,19 @@ spec:
...
+    name: Guilherme Cassolato
+  - email: didier@redhat.com
+    name: Didier Di Cesare
   maturity: alpha
+  minKubeVersion: 1.8.0
   provider:
     name: Red Hat
     url: https://github.com/Kuadrant/kuadrant-operator
-  version: 0.0.0
+  version: 0.1.0
```
eguzki, line 436: I think that in `main` branch, we should be using `0.0.0` for the version and `latest` for the image tags

eguzki (changes requested):

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
guicassolato: > I think that in `main` branch, we should be using `0.0.0` for the version and `latest` for the image tags

@eguzki's suggestion matches what we do in https://github.com/Kuadrant/authorino-operator/blob/7e4022a92e6859a18765377d26fd4d0d88bd37b5/Makefile#L24-L25

didierofrivia: Makes sense

eguzki (approved):

PR #41, opened 2022-09-09
wasm shim image from env var

eguzki:
### what

In order to pin the wasm shim image from the manifest bundle, the image needs to be exposed in the manifests and read from an env var by the controller

Additionally, the operator configures the kuadrant core controller (aka kuadrant-controller) with the `WASM_FILTER_IMAGE` env var.

### verification steps

run dev env

```
```

Create kuadrant CR

```yaml
k apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```
Check that the kuadrant core controller has the wasm shim image url specified in the deployment object

```yaml
k get deployments kuadrant-controller-manager -o jsonpath='{.spec.template.spec.containers[].env}' | yq_pretty
- name: LIMITADOR_NAMESPACE
  value: default
- name: LIMITADOR_NAME
  value: limitador
- name: WASM_FILTER_IMAGE
  value: oci://quay.io/kuadrant/wasm-shim:latest
```

### DIFF
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -338,6 +338,9 @@ spec:
                 - --leader-elect
                 command:
                 - /manager
+                env:
+                - name: RELATED_IMAGE_WASMSHIM
+                  value: oci://quay.io/kuadrant/wasm-shim:latest
                 image: quay.io/kuadrant/kuadrant-operator:latest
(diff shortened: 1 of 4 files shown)


eguzki:
the kuadrant controller manifest test is fixed here https://github.com/Kuadrant/kuadrant-operator/pull/42

eguzki:
integration test fixed in https://github.com/Kuadrant/kuadrant-operator/pull/45

PR #44, opened 2022-09-14
update kuadrant core controller manifests

eguzki:


didierofrivia (approved):

eguzki:
integration test fixed in https://github.com/Kuadrant/kuadrant-operator/pull/45

PR #52, opened 2022-10-04
Fixing dependencies

didierofrivia:
This PR fixes the operator image build that was broken in Github workflows.

To make this fix, it was needed to adjust dependencies and the Go version, since it was installing wrong dependencies even when the `go.mod` stated otherwise, as well as bumping the Go version to `1.18`

It also removes the kuadrantmanifests directory since now are included in the config.


eguzki (approved):
In the go.sum there are `indirect` deps that still needs to be fixed. We can do that in coming PR's

PR #54, opened 2022-10-12
Kuadrant reconciling

didierofrivia:
This PR is the first one that brings the reconciling process for **RateLimitPolicies** and **AuthPolicies**. It also includes the necessary make targets for dev testing and many fixes to it's API and config.

**Verification Steps**

1. Setup a local Kuadrant cluster
```bash
make local-setup
```

2. Forward istio gateway port
```bash
kubectl port-forward -n istio-system service/istio-ingressgateway 8080:80 &
```

3. Install toystore service:
```bash
 kubectl apply -f - <<EOF
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  selector:
    matchLabels:
      app: toystore
  template:
    metadata:
      labels:
        app: toystore
    spec:
      containers:
        - name: toystore
          env:
            - name: PORT
              value: "3000"
          ports:
            - containerPort: 3000
              name: http
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: toystore
spec:
  selector:
    app: toystore
  ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 3000
EOF
```

4. Add a HTTPRoute for the toystore service
```bash
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
EOF
```

5. Check if the service is running
```bash
curl -v -H 'Host: api.toystore.com' http://localhost:8080/toy
```
_It should return a `200`_

6. Apply a RateLimit Policy
```bash
kubectl apply -f - <<EOF
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - rules:
        - hosts: ["rate-limited.toystore.com"]
      configurations:
        - actions:
            - generic_key:
                descriptor_key: "limited"
                descriptor_value: "1"
      limits:
        - conditions:
            - "limited == 1"
          maxValue: 5
          seconds: 10
          variables: []
EOF
```

7. Now hitting the rate limited endpoint of toystore service should allow only 5 every 10 seconds. If you do 6 consecutive curls, the last one should return a `429`
```bash
curl -v -H 'Host: rate-limited.toystore.com' http://localhost:8080/toy
```

8. Profit!


Makefile
```diff
@@ -187,6 +226,12 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/deploy | kubectl delete -f -
 
+deploy-dependencies: kustomize ## Deploy dependencies to the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/dependencies | kubectl apply -f -
+	kubectl -n "$(KUADRANT_NAMESPACE)" wait --timeout=300s --for=condition=Available deployments --all
+	kubectl apply -n "$(KUADRANT_NAMESPACE)" -f utils/local-deployment/authorino.yaml
```
eguzki, line 232: this is not part of the env setup, this is the job of the kuadrant-controller. 

Do the files `utils/local-deployment/authorino.yaml` and `utils/local-deployment/limitador.yaml` make any sense in this repo?

bundle/manifests/kuadrant.kuadrant.io_authpolicies.yaml
```diff
@@ -2,21 +2,21 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.8.0
```
eguzki, line 5: downgrading??? CAn we keep v0.8.0? OI even better, we could upgrade to v0.9.2 ?? maybe v0.10.0?

bundle/manifests/kuadrant.kuadrant.io_authpolicies.yaml
```diff
@@ -2,21 +2,21 @@ apiVersion: apiextensions.k8s.io/v1
...
+    controller-gen.kubebuilder.io/version: v0.7.0
   creationTimestamp: null
   labels:
     app: kuadrant
-  name: authpolicies.apim.kuadrant.io
+  name: authpolicies.kuadrant.kuadrant.io
 spec:
-  group: apim.kuadrant.io
+  group: kuadrant.kuadrant.io
```
eguzki, line 11: what about having the group: `kuadrant.io` ?

`authpolicies.kuadrant.io`

`ratelimitpolicies.kuadrant.io`

`kuadrants.kuadrant.io`

main.go
```diff
@@ -81,16 +92,34 @@ func main() {
...
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&configFile, "config", "",
+		"The operator will load its initial configuration from this file. "+
+			"Omit this flag to use the default configuration values. "+
+			"Command-line flags override configuration from this file.")
 	flag.Parse()
 
+	options := ctrl.Options{Scheme: scheme}
```
eguzki, line 114: `options` is being used? 

This is a mix of config file driven  configuration with cli params. I think we should pick one of the options. I tend to prefer the config file driven configuration, but I would also accept the other approach.

Makefile
```diff
@@ -155,6 +157,43 @@ clean-cov: ## Remove coverage report
 test: clean-cov manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out -v -timeout 0
 
+.PHONY: namespace
+namespace: ## Creates a namespace where to deploy Kuadrant Operator
+	kubectl create namespace $(KUADRANT_NAMESPACE)
+
+.PHONY: local-setup
+local-setup: export IMG := $(IMAGE_TAG_BASE):dev
+local-setup: ## Deploy locally kuadrant operator from the current code
```
eguzki, line 166: I like the `make local-setup`, it is kind of standard in kuadrant projects. 

This target replaces the `kind-create-kuadrant-cluster` make target in `make/kind.mk` file. It should be deleted.

eguzki (changes requested):

main.go
didierofrivia: Indeed, `options` were not applied. Now they are after https://github.com/Kuadrant/kuadrant-operator/pull/54/commits/6dc60a2d0de69a6422f0d9b6af10c7e76a09b9ac.

I don't have any particular preference, maybe we could discuss it with the rest and refactor accordingly? WDYT?

eguzki (approved):

PR #55, opened 2022-10-20
Fixing tests

didierofrivia:
TODO

Note: Tests should pass when merged.


Makefile
```diff
@@ -154,8 +160,13 @@ clean-cov: ## Remove coverage report
 	rm -rf cover.out
 
 .PHONY: test
-test: clean-cov manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out -v -timeout 0
+test: test-unit test-integration ## Run all tests
+
+test-integration: clean-cov generate fmt vet manifests envtest ## Run Integration tests.
```
eguzki, line 165: run them on CI? we are not running them in GH actions

controllers/ratelimitpolicy_controller_test.go
```diff
@@ -64,7 +64,7 @@ func testBuildBasicHttpRoute(routeName, gwName, ns string) *gatewayapiv1alpha2.H
 		},
 		Spec: gatewayapiv1alpha2.HTTPRouteSpec{
 			CommonRouteSpec: gatewayapiv1alpha2.CommonRouteSpec{
-				ParentRefs: []gatewayapiv1alpha2.ParentRef{
```
eguzki, line 67: wow, this is unexpected. They broke the API from 0.4 to 0.5. I guess this kind of things happens when working with 0.X versions.

didierofrivia: Exactly this, I was a bit surprised too... but then, well, same conclusion.

eguzki (approved):

eguzki:
tests failing.. should be passing

controllers/authpolicy_controller_test.go
```diff
@@ -28,16 +28,21 @@ const GatewayName = "istio-ingressgateway"
 var _ = Describe("AuthPolicy controller", func() {
 	var testNamespace string
 
-	BeforeEach(CreateNamespaceCallback(&testNamespace))
+	beforeEachCallback := func() {
+		CreateNamespace(&testNamespace)
+		ApplyKuadrantCR(testNamespace)
```
eguzki, line 33: should it be `kuadrantDefaultNamespace`?

eguzki (approved):

controllers/authpolicy_controller_test.go
didierofrivia: in this test is not particularly needed, since the assertions made and Authorino OP and service are not setup via ENVs.

PR #104, opened 2022-11-14
Only reconcile Authorino if Istio

### LINKED ISSUE #103: Kuadrant requiring IstioOperator even if used for ratelimiting only
As far as I know you should be able to use Kuadrant for ratelimiting only even if you do not have access to `IstioOperator`.
CR:
```
kind: Kuadrant
metadata:
  name: kuadrant-sample
  namespace: kuadrant
spec: {}
status:
  conditions:
    - lastTransitionTime: '2022-11-01T13:34:57Z'
      message: >-
        no matches for kind "IstioOperator" in version
        "install.istio.io/v1alpha1"
      reason: ReconcilliationError
      status: 'False'
      type: Ready
```

alexsnaps:
So I absolutely do not think this is the way to fix #103 on the long term, but this could help us make progress until we test OSSM

### What's wrong?

 - The `Status` needs to reflect this.
 - The `IstioOperator` could have disappeared, while an `Authorino` was already running… I don't even know what this means to be honest!
 - I wouldn't be surprised if this breaks more stuff… started with the tests 🤦 

### So what?!

We need to figure things out like these I think:

- [ ] Should it be lazy and only happen when a `Policy` requiring something appears? 
- [ ] Be lazy always, yet make sure we have all dependencies resolvable, and report on that? 
- [ ] Expose it through the CR as to what is lazily or eagerly deployed?
- [ ] Should for now add a way to disable auth and/or rl entirely?
- [ ] … Others?

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -299,10 +299,6 @@ func hasKuadrantAuthorizer(extensionProviders []*istiomeshv1alpha1.MeshConfig_Ex
 	//          service: AUTHORINO SERVICE
 	//        name: kuadrant-authorization
 
-	if len(extensionProviders) == 0 {
-		return false
-	}
-
 	for _, extensionProvider := range extensionProviders {
 		if extensionProvider.Name == extAuthorizerName {
 			return true
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -258,16 +258,24 @@ func (r *KuadrantReconciler) registerExternalAuthorizer(ctx context.Context, kOb
 }
 
 func (r *KuadrantReconciler) reconcileSpec(ctx context.Context, kObj *kuadrantv1beta1.Kuadrant) (ctrl.Result, error) {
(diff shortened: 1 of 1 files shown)


controllers/kuadrant_controller.go
```diff
@@ -258,16 +258,24 @@ func (r *KuadrantReconciler) registerExternalAuthorizer(ctx context.Context, kOb
 }
 
 func (r *KuadrantReconciler) reconcileSpec(ctx context.Context, kObj *kuadrantv1beta1.Kuadrant) (ctrl.Result, error) {
+	var reconcileAuthorino = true
+
 	if err := r.registerExternalAuthorizer(ctx, kObj); err != nil {
-		return ctrl.Result{}, err
+		if apierrors.IsNotFound(err) {
```
eguzki, line 264: The error is not a `Not Found`. OSSM does not deploy IstioOperator CRD, hence the error is:
```
no matches for kind \"IstioOperator\" in version \"install.istio.io/v1alpha1\"
```

I will fix it

Issue #69, opened 2022-11-04
Kuadrant instance only works in the `kuadrant-system` namespace


eguzki:
When the PR https://github.com/Kuadrant/kuadrant-operator/pull/48 is merged, the kuadrant as a system will not be working unless the namespace where kuadrant is installed is `kuadrant-system`. This issue comes from the fact that the policy controller are no longer deployed as components of a kuadrant instance.  Instead, the controllers live at the kuadrant's operator pod and they are up&running even if there is no kuadrant instance running in the cluster. 

The very source issue of this "side effect" is the design about how backend services were wired with the Ratelimit/Auth policies. As it was designed, it allowed kuadrant instance to be installed in any namespace, however, the design only allowed one kuadrant instance to be running in the cluster. When the controllers were moved to the operator's pod, one of the design's requirements (the controllers know at deploy time with env vars where the kuadrant's backend services live) was unmet, thus causing the issue.

This issue is to track the fix, as it seems desirable that kuadrant can be installed in any namespace. 

Related but not in the scope of this issue is the decision about whether kuadrant supports multiple instances in a single cluster or not. That would require a definition of how policies are wired to the backend services (limitador for the rate limiting and authorino for the authN/authZ services). Once it is clear how policies are linked to the backend services, the fix of this issue should not be hard.

Issue #56, opened 2022-10-27
Kuadrant services settings


didierofrivia:
Currently there are some services, I.E: _Limitador_ that the Operator tells the Gateway where it's been deployed (Name, Namespace, Service port, etc...) and many others that are relying on defaults and hardcoded settings.
Ideally one can have an unified interface (Kuadrant CR, params, ENV, etc..) and avoid any hardcoded values or split ways of configuration.

PR #48, opened 2022-09-26
Kuadrant Merge

didierofrivia:
This PR includes every single step, introduced by previous PRs, accomplishing the merge of the Kuadrant Controller repo and it's features within the Kuadrant Operator.
 
Closes https://github.com/Kuadrant/kuadrant-operator/issues/65
Closes https://github.com/Kuadrant/kuadrant-operator/issues/66

## Verification Steps
Every user guide should work as expected. Any reviewer should be able to reproduce them:

* [Simple rate limit for API owners](https://github.com/Kuadrant/kuadrant-operator/blob/582e038540395be8993e05059a26a77867a8cafa/doc/user-guides/simple-rl-for-api-owners.md)
* [Gateway rate limit for cluster operators](https://github.com/Kuadrant/kuadrant-operator/blob/582e038540395be8993e05059a26a77867a8cafa/doc/user-guides/gateway-rl-for-cluster-operators.md)
* And our best seller [Authenticated rate limit for API Owners](https://github.com/Kuadrant/kuadrant-operator/blob/582e038540395be8993e05059a26a77867a8cafa/doc/user-guides/authenticated-rl-for-api-owners.md)

Also, you could try this simple rate limit example:

1. Setup a local Kuadrant cluster
```bash
make local-setup
```

2. Apply Kuadrant object in order to deploy required dependencies (this will change in the future most probably)

```bash
kubectl -n kuadrant-system apply -f - <<EOF
---
apiVersion: kuadrant.io/v1beta1
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

3. Forward istio gateway port
```bash
kubectl port-forward -n istio-system service/istio-ingressgateway 8080:80 &
```

4. Install toystore service:
```bash
 kubectl apply -f - <<EOF
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  selector:
    matchLabels:
      app: toystore
  template:
    metadata:
      labels:
        app: toystore
    spec:
      containers:
        - name: toystore
          image: quay.io/3scale/authorino:echo-api
          env:
            - name: PORT
              value: "3000"
          ports:
            - containerPort: 3000
              name: http
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: toystore
spec:
  selector:
    app: toystore
  ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 3000
EOF
```

[...]

PR #109, opened 2022-11-22
Fix dependencies namespace propagation

didierofrivia:
This PR aims to solve the [limitation](https://github.com/Kuadrant/kuadrant-operator/issues/69) introduced by https://github.com/Kuadrant/kuadrant-operator/pull/48. Noting that this will not be the final behaviour since a RFC is still being defined.

Closes https://github.com/Kuadrant/kuadrant-operator/issues/69
Addresses partially https://github.com/Kuadrant/kuadrant-operator/issues/56

**A few considerations and limitations adopted:**

1. There's only one _Kuadrant_ instance (CR applied) in the cluster.
2. When the Kuadrant CR is applied, it will manage every single Gateway resource. (Will tag them)
3. New Gateways are not managed by the existing Kuadrant.

**Diagram of how everything will be setup in the verification steps:**
custom-namespace

##### Namespaces
* **istio-system**: This one is where istio and its gateway is installed. It also includes the Istio Authentication Policies.
* **kuadrant-system**: This namespace is where the operators are installed, it could be setup with an env var.
* **kuadrant-test**: An admin chosen namespace where the Kuadrant instance (Kuadrant CR and services) are gonna be installed. It also will host the Authorino AuthConfigs.
* **default**: Or any other chosen namespace where the app developer will install its services, define Kuadrant policies, own networking objects, etc.

**Verification Steps:**

1. Setup a local Kuadrant cluster
```bash
make local-setup
```

5. Create a test namespace

```bash
kubectl create namespace kuadrant-test
```

6. Apply Kuadrant object in order to deploy required dependencies (this will change in the future most probably)

```bash
kubectl -n kuadrant-test apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

7. Forward istio gateway port
```bash
kubectl port-forward -n istio-system service/istio-ingressgateway 9080:80 &
```

8. Install toystore service:
```bash
 kubectl apply -f - <<EOF
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  selector:
    matchLabels:
      app: toystore
  template:
    metadata:
      labels:
        app: toystore
    spec:
      containers:
        - name: toystore
          env:
            - name: PORT
              value: "3000"
          ports:
            - containerPort: 3000
              name: http
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: toystore
spec:
  selector:
    app: toystore
  ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 3000
EOF
```

9. Add a HTTPRoute for the toystore service
```bash
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
EOF
```

10. Check if the service is running. It might take up a couple of minutes for the deployment to be ready.
```bash
curl -v -H 'Host: api.toystore.com' http://localhost:9080/toy
```
_It should return a `200`_

11.  Create API keys for user `Bob` and `Alice`

```yaml
kubectl apply -f -<<EOF
---
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: bob-key
  labels:
    app: toystore
stringData:
  api_key: IAMBOB
type: Opaque
---
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: alice-key
  labels:
    app: toystore
stringData:
  api_key: IAMALICE
type: Opaque
EOF
```

12. Create Kuadrant's `AuthPolicy` to configure API key based authentication

```yaml
kubectl apply -f - <<EOF
---
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
  - hosts: ["*.toystore.com"]
    paths: ["/toy*"]
  authScheme:
    identity:
    - name: friends
      apiKey:
        allNamespaces: true
        selector:
          matchLabels:
            app: toystore
      credentials:
        in: authorization_header
        keySelector: APIKEY
    response:
    - json:
        properties:
          - name: userID
            value: null
            valueFrom:
      name: rate-limit-apikey
      wrapper: envoyDynamicMetadata
      wrapperKey: ext_auth_data
EOF
```

13. Now the endpoint should be protected

```bash
curl -v -H 'Host: api.toystore.com' http://localhost:9080/toy
```

_It should return a `401 Unauthorized`_ 

You should be able to use either Alice or Bob API Key like:

```bash
curl -v -H 'Authorization: APIKEY IAMBOB' -H 'Host: api.toystore.com' http://localhost:9080/toy
```

14. Create RateLimitPolicy to rate limit per API key basis

| User | Rate Limits |
| ------------- | -----: |
| `Bob` | **2** reqs / **10** secs (0.2 rps) |
| `Alice` | **5** reqs / **10** secs (0.5 rps) |

```yaml
kubectl apply -f -<<EOF
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
  - configurations:
      - actions:
          - metadata:
              descriptor_key: "userID"
              default_value: "no-user"
              metadata_key:
                key: "envoy.filters.http.ext_authz"
                path:
                  - segment:
                      key: "ext_auth_data"
                  - segment:
                      key: "userID"
    limits:
      - conditions:
          - "userID == bob"
        maxValue: 2
        seconds: 10
        variables: []
      - conditions:
          - "userID == alice"
        maxValue: 5
        seconds: 10
        variables: []
EOF
```
15. Validating the rate limit policy

Only 2 requests every 10 allowed for Bob.

```bash
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H 'Authorization: APIKEY IAMBOB' -H 'Host: api.toystore.com'  -X GET http://localhost:9080/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

Only 5 requests every 10 allowed for Alice.

```bash
while :; do curl --write-out '%{http_code}' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com'  -X GET http://localhost:9080/toy | egrep --color "\b(429)\b|$"; sleep 1; done
```

16. Profit!


pkg/common/common.go
```diff
@@ -28,15 +29,22 @@ import (
...
 const (
-	KuadrantNamespace                    = "kuadrant-system"
 	KuadrantRateLimitClusterName         = "kuadrant-rate-limiting-service"
 	HTTPRouteKind                        = "HTTPRoute"
 	KuadrantRateLimitPolicyRefAnnotation = "kuadrant.io/ratelimitpolicies"
 	RateLimitPolicyBackRefAnnotation     = "kuadrant.io/ratelimitpolicy-direct-backref"
 	AuthPolicyBackRefAnnotation          = "kuadrant.io/authpolicy-backref"
+	KuadrantNamespaceLabel               = "kuadrant.io/namespace"
 	NamespaceSeparator                   = '/'
+	LimitadorName                        = "limitador"
```
alexsnaps, line 39: Did you mean to remove the ability to redefine this thru a environment variable?

controllers/kuadrant_controller.go
```diff
@@ -142,6 +141,10 @@ func (r *KuadrantReconciler) Reconcile(eventCtx context.Context, req ctrl.Reques
 		}
 	}
 
+	if gwErr := r.reconcileClusterGateways(ctx, kObj); gwErr != nil {
```
alexsnaps, line 144: What drove this? This is new, right?

pkg/common/common.go
didierofrivia: Yes, I could rollback to env var with default too.

controllers/kuadrant_controller.go
didierofrivia: Yes, this is basically the key step to communicate the Kuadrant instance namespace to the policies. It will annotate the Gateways with its namespace, thus the policies, who knows about the GW or at least the HTTPRoute pointing to the Gateway will have access to this annotation

didierofrivia:
Previously it was the same, limitador and authorino were installed in the hardcoded `kuadrant-system` and previously that, to the ENV that kuadrant-controller was.

pkg/common/common.go
didierofrivia: I just thought if there's only one limitador and authorino per kuadrant, there was no need to make a difference

didierofrivia:
I'll merge, then I could do a following PR addressing any future comments

PR #110, opened 2022-11-23
disable status check for authorino when not istio

eguzki:
### what

When OSSM (maybe better described as not Istio), the status tries to check the availability of authorino. As expected, if authorino is not deployed, the status check fails:

```yaml
status:
  conditions:
  - lastTransitionTime: "2022-11-18T13:12:43Z"
    message: Authorino.operator.authorino.kuadrant.io "authorino" not found
    reason: AuthorinoNotReady
    status: "False"
    type: Ready
  observedGeneration: 1
```

This PR conditionally checks for authorino availability when Istio is found

### how

Uses the context to carry the flag `istioEnabled`

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -282,7 +282,7 @@ func (r *KuadrantReconciler) reconcileSpec(prevCtx context.Context, kObj *kuadra
 		logger.Info("external auth disabled")
 	}
 
-	ctx := newContextWithIstioEnabled(prevCtx, istioEnabled)
+	ctx := newContextWithIstioInfo(prevCtx, istioEnabled)
 
 	if err := r.registerExternalAuthorizer(ctx, kObj); err != nil {
 		return ctrl.Result{}, err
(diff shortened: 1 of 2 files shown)


alexsnaps (approved):

controllers/kuadrant_controller.go
```diff
@@ -370,14 +402,47 @@ func (r *KuadrantReconciler) reconcileAuthorino(ctx context.Context, kObj *kuadr
...
+		return false, err
+	}
+
+	return true, nil
+}
+
+// extAuthEnabled is how we find external auth enable flag in a context.Context.
+type extAuthEnabled struct{}
+
+func newContextWithIstioEnabled(ctx context.Context, istioEnabled bool) context.Context {
```
guicassolato, line 434: The name of this function makes it look like that when invoked Istio _is_ enabled, while this is not always the case. Maybe `newContextWithIstioInfo`?

eguzki: done

guicassolato (approved):

PR #123, opened 2022-12-21
Fix conditions in examples

alexsnaps:
Tho they work, they rely on the lenient mode in limitador

### DIFF
--- a/examples/toystore/ratelimitpolicy_gateway.yaml
+++ b/examples/toystore/ratelimitpolicy_gateway.yaml
@@ -16,10 +16,10 @@ spec:
       configurations:
         - actions:
             - generic_key:
-                descriptor_key: expensive-op
+                descriptor_key: expensive_op
                 descriptor_value: "1"
       limits:
-        - conditions: ["expensive-op == 1"]
+        - conditions: ["expensive_op == '1'"]
           maxValue: 2
           seconds: 30
           variables: []
--- a/examples/toystore/ratelimitpolicy_httproute.yaml
+++ b/examples/toystore/ratelimitpolicy_httproute.yaml
@@ -15,10 +15,10 @@ spec:
       configurations:
         - actions:
             - generic_key:
-                descriptor_key: get-toy
+                descriptor_key: get_toy
                 descriptor_value: "yes"
       limits:
-        - conditions: ["get-toy == yes"]
+        - conditions: ["get_toy == 'yes'"]
           maxValue: 2
           seconds: 30
           variables: []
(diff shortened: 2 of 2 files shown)


guicassolato (approved):

examples/toystore/ratelimitpolicy_gateway.yaml
```diff
@@ -19,7 +19,7 @@ spec:
                 descriptor_key: expensive-op
                 descriptor_value: "1"
       limits:
-        - conditions: ["expensive-op == 1"]
+        - conditions: ["expensive-op == '1'"]
```
eguzki, line 22: shouldn't be `expensive_op`?

examples/toystore/ratelimitpolicy_httproute.yaml
```diff
@@ -18,7 +18,7 @@ spec:
                 descriptor_key: get-toy
                 descriptor_value: "yes"
       limits:
-        - conditions: ["get-toy == yes"]
+        - conditions: ["get-toy == 'yes'"]
```
eguzki, line 21: shouldn't be `get_toy` ?

examples/toystore/ratelimitpolicy_httproute.yaml
```diff
@@ -42,14 +42,14 @@ spec:
...
+            - "admin == 'yes'"
+            - "user-id == 'bob'"
           maxValue: 2
           seconds: 30
           variables: []
         - conditions:
-            - "admin == yes"
-            - "user-id == alice"
+            - "admin == 'yes'"
+            - "user-id == 'alice'"
```
eguzki, line 52: shouldn't be `user_id` ?

examples/toystore/ratelimitpolicy_httproute.yaml
```diff
@@ -42,14 +42,14 @@ spec:
                         key: "user-id"
       limits:
         - conditions:
-            - "admin == yes"
-            - "user-id == bob"
+            - "admin == 'yes'"
+            - "user-id == 'bob'"
```
eguzki, line 46: shouldn't be `user_id` ?

eguzki (changes requested):

alexsnaps:
Good catch @eguzki !

eguzki (approved):

eguzki:
by the way, do we have this limitation documented? I cannot find in https://github.com/Kuadrant/limitador/blob/main/doc/migrations/conditions.md

alexsnaps:
>by the way, do we have this limitation documented? 
Yeah, no... it's not in there. Or say not directly. Hard to document what really has changed, might be better to just document what makes a valid identifier?

Issue #126, opened 2023-01-04
AuthorizationPolicy is not cleaned up after AuthPolicy removal


pehala:
After removing of `AuthPolicy` the `AuthorizationPolicy` is **sometimes** not correctly cleaned up. I see in the logs:
```
{"level":"info","ts":1672843350.9243772,"logger":"kuadrant-operator.authpolicy","msg":"Reconciling AuthPolicy","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta"}
{"level":"info","ts":1672843350.924472,"logger":"kuadrant-operator.authpolicy","msg":"Removing Istio's AuthorizationPolicy","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta"}
{"level":"info","ts":1672843350.9245126,"logger":"kuadrant-operator.authpolicy","msg":"Removing Authorino's AuthConfigs","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta"}
{"level":"info","ts":1672843350.924537,"logger":"kuadrant-operator.authpolicy","msg":"delete object","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta","kind":"v1beta1.AuthConfig","name":"ap-authorino-policy-phala-kudrnt-te-c4ta","namespace":"kuadrant"}
{"level":"info","ts":1672843350.9392066,"logger":"kuadrant-operator.authpolicy","msg":"update object","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta","kind":"v1beta1.AuthPolicy","name":"policy-phala-kudrnt-te-c4ta","namespace":"authorino"}
{"level":"info","ts":1672843350.9534516,"logger":"kuadrant-operator.authpolicy","msg":"Reconciling AuthPolicy","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta"}
{"level":"info","ts":1672843350.9535172,"logger":"kuadrant-operator.authpolicy","msg":"no AuthPolicy found","AuthPolicy":"authorino/policy-phala-kudrnt-te-c4ta"}
```
Based on the logs I can see it tried to remove `AuthorizationPolicy` but probably didn't find it as I cant see the actual deletion logged and I can still see the `AuthorizationPolicy` in the `istio-system`. Might be cause of #124

guicassolato:
**Steps to reproduce:**

Spin off the cluster, install Kuadrant, deploy the app, and create the HTTPRoute:
[...]

PR #134, opened 2023-01-11
[refactor] GW utils for all types of policies

guicassolato:
Preparation for #126.

- Moves the functions to compute gw diffs and reconcile gateway back refs from the `RateLimitPolicyReconciler` to the base `TargetRefReconciler`
- Moves the functions (`NewGateways`, `SameGateways` and `LeftGateways`) and `GatewayWrapper` type of the Gateway utils to `pkg/common`
- Renames the `GatewayWrapper` type as `TargetGateway`
- Makes all the above generic to it can be used with all types of policies


eguzki (approved):

pkg/reconcilers/targetref_reconciler.go
```diff
@@ -239,11 +232,110 @@ func (r *TargetRefReconciler) DeleteTargetBackReference(ctx context.Context,
...
+// Returns:
+// * list of gateways to which the policy applies for the first time
+// * list of gateways to which the policy no longer applies
+// * list of gateways to which the policy still applies
+func (r *TargetRefReconciler) ComputeGatewayDiffs(ctx context.Context, policy common.KuadrantPolicy, policyRefsConfig common.PolicyRefsConfig) (*GatewayDiff, error) {
+	logger, _ := logr.FromContext(ctx)
+
+	gwKeys, err := r.TargetedGatewayKeys(ctx, policy.GetTargetRef(), policy.GetNamespace())
+	if err != nil {
+		return nil, err
```
guicassolato, line 253: I think we should only return an error here is it's other than not found.

eguzki: right. On NotFound error, it should return an empty list.

guicassolato: I'll make the change.

PR #139, opened 2023-01-13
fix: `ComputeGatewayDiffs` when missing target HTTPRoute

### LINKED ISSUE #138: RateLimitPolicy cleanup fails when the HTTPRoute is missing
A few places in the code known to fail when the target HTTPRoute is missing, instead of properly cleaning up/setting the status of policy:

1. `controllers.RateLimitPolicyReconciler.validateRuleHosts`
https://github.com/Kuadrant/kuadrant-operator/blob/fa824534a0d53d8ab6c45cbcdc8cac09489d5498/controllers/ratelimitpolicy_controller.go#L203-L206

2. `reconcilers.ReconcileTargetBackReference`
https://github.com/Kuadrant/kuadrant-operator/blob/fa824534a0d53d8ab6c45cbcdc8cac09489d5498/pkg/reconcilers/targetref_reconciler.go#L179-L183

3. `reconcilers.ComputeGatewayDiffs` - see https://github.com/Kuadrant/kuadrant-operator/pull/134#discussion_r1069308944
https://github.com/Kuadrant/kuadrant-operator/blob/fa824534a0d53d8ab6c45cbcdc8cac09489d5498/pkg/reconcilers/targetref_reconciler.go#L251-L254

---

Possibly duplicate of #125

guicassolato:
Related to #138


pkg/reconcilers/targetref_reconciler.go
```diff
@@ -249,7 +249,7 @@ func (r *TargetRefReconciler) ComputeGatewayDiffs(ctx context.Context, policy co
 	logger, _ := logr.FromContext(ctx)
 
 	gwKeys, err := r.TargetedGatewayKeys(ctx, policy.GetTargetRef(), policy.GetNamespace())
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
```
eguzki, line 252: I would implement that inside the `TargetedGatewayKeys` method. That way it can be reused. The contract of the method would be "if the targeted network resource is an HTTPRoute and this resource does not exists, the returned targeted Gateway keys will be the empty list". Wdyt?

eguzki: Currently, at this stage of the reconciliation process, when the target ref is HTTPRoute, the resource must exist. Otherwise it would had failed in a previous stage: reconciling the direct back reference. Anyway, I think that we should handle the NotFound the error here as well. 

Ok, let's leave the NotFound handling to the caller.

In this particular context of the `ComputeGatewayDiffs` method, when the HTTPRoute is not found, the policy should not be applied in any gateway and if there is configuration in place, it should be removed. For that, we need `gwKeys` to be `nil` or empty. What about this?

```go
gwKeys, err := r.TargetedGatewayKeys(ctx, policy.GetTargetRef(), policy.GetNamespace())   
if err != nil {  
   if !apierrors.IsNotFound(err) {                                                           
     return nil, err                                                                       
   }
   // when the network resource is not found, the policy should not be applied and if there is configuration in place, 
  // it should be removed. gwKeys being an empty list would also work
   gwKeys = nil
}                                                                                         

```

guicassolato: You are absolutely right, @eguzki. In fact, I was trying to reproduce the bug and, as you said, it fails first at https://github.com/Kuadrant/kuadrant-operator/blob/73c8d089e7e60a3aac4e8e4ed6d94c1b017c50f7/controllers/ratelimitpolicy_controller.go#L202-L206

guicassolato: Even before reconciling the direct back reference. It fails when trying to validate the rules.

guicassolato: There's no need to set `gwKeys` when `apierrors.IsNotFound(err)`. It will be `nil` already.

guicassolato:
Flagging this as WIP so we can fix as well other points where reconciliation fails due to missing the targeted HTTPRoute.

guicassolato:
Let's make this about `ComputeGatewayDiffs` only (so it unblocks the fix for #126), and let #138 and related issues cover the rest of the work needed to fix the reconciliation of RLP upon deletion of target HTTPRoutes.

pkg/reconcilers/targetref_reconciler.go
eguzki: You cannot tell from the caller perspective.  Better to be assertive

didierofrivia (approved):
👍🏼

didierofrivia (approved):

PR #144, opened 2023-01-27
Simplify RateLimitPolicy

### LINKED ISSUE #133: Improvements to RateLimitPolicy
Currently, the minimal working example (At least what I could do) looks like this. However, this could be much easier for the simple limit I want to enforce.
```
kind: RateLimitPolicy
metadata:
  name: gw-limit
  namespace: istio-system
spec:
  rateLimits:
    - configurations:
        - actions:
            - generic_key:
                descriptor_key: limited
                descriptor_value: '1'
      limits:
        - conditions: []
          maxValue: 5
          seconds: 10
          variables: []
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
```
Notable problems:

* `configurations` cannot be empty or the rate limit won't work, and it is not always needed
   * **Solution 1**: If it is empty, Kuadrant could generate a static action key.
   * **Solution 2**: Make Kuadrant throw an error if it is empty.
* `conditions` and `variables` are not required but need to be specified at least as an empty list:
   * **Solution**: Make them optional and if missing, assign an empty list. 

Please let me know what you think. These are just what I see as problems from my perspective.

eguzki:
### What

Make some fields of the RateLimitPolicy optional to make the required CR object smaller and simpler. After this changes, the simplest RLP instance looks like this:

```yaml
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rateLimits:
    - limits:
        - maxValue: 5
          seconds: 10
```
Only `targetRef` and one limit is required. Furthermore, the limit object does not require `conditions` and `variables` fields. 

**[ALMOST BREAKING CHANGE]** Technically this is breaking the current v1beta1 API. However, I do not expect issues as all the changes are switching required fields to optional. Thus, any existing "old" RLP instance will still be valid with the "new" CRD. No need to conversion either. 

TODO:
- [ ] unittests and e2e tests
- [ ] Doc

Fixes #133 

### Verification steps


eguzki:

maleck13:
@eguzki I don't feel qualified to approve the PR but I love this change and take a look through it looks reasonable

pkg/rlptools/limit_index.go
```diff
@@ -129,12 +129,23 @@ func (l *LimitIndex) ToLimits() []limitadorv1alpha1.RateLimit {
...
+				// the fields "Variables" and "Conditions" are required
+				variables := limitList[idx].Variables
+				if variables == nil {
+					variables = make([]string, 0)
+				}
+				conditions := limitList[idx].Conditions
+				if conditions == nil {
+					conditions = make([]string, 0)
+				}
+
```
didierofrivia, line 142: You might be able to get rid of these 8 lines by using `GetDefaultIfNil` fn from our `common` package.

pkg/rlptools/limit_index.go
```diff
@@ -129,12 +129,23 @@ func (l *LimitIndex) ToLimits() []limitadorv1alpha1.RateLimit {
...
 				limits = append(limits, limitadorv1alpha1.RateLimit{
 					Namespace:  common.MarshallNamespace(gwKey, domain),
 					MaxValue:   limitList[idx].MaxValue,
 					Seconds:    limitList[idx].Seconds,
-					Variables:  limitList[idx].Variables,
-					Conditions: limitList[idx].Conditions,
+					Variables:  variables,
+					Conditions: conditions,
```
didierofrivia, line 148: Like this:

```suggestion
					Variables: GetDefaultIfNil(limitList[idx].Variables, make([]string,0)),
					Conditions: GetDefaultIfNil(limitList[idx].Conditions, make([]string,0)),
```

Makefile
```diff
@@ -219,8 +219,11 @@ test: test-unit test-integration ## Run all tests
 test-integration: clean-cov generate fmt vet envtest ## Run Integration tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) $(ARCH_PARAM) use $(ENVTEST_K8S_VERSION) -p path)" USE_EXISTING_CLUSTER=true go test ./... -coverprofile $(PROJECT_PATH)/cover.out -tags integration -ginkgo.v -ginkgo.progress -v -timeout 0
 
+ifdef TEST_NAME
+test-unit: TEST_PATTERN := --run $(TEST_NAME)
+endif
 test-unit: clean-cov generate fmt vet ## Run Unit tests.
-	go test ./... -coverprofile $(PROJECT_PATH)/cover.out -tags unit -v -timeout 0
+	go test ./... -coverprofile $(PROJECT_PATH)/cover.out -tags unit -v -timeout 0 $(TEST_PATTERN)
```
didierofrivia, line 226: ❤️

pkg/rlptools/limit_index.go
eguzki: `Variables` is not a pointer, it's a slice.

TRying to compile it complains:
```
pkg/rlptools/limit_index.go:134:41: type []string of limitList[idx].Variables does not match *T (cannot infer T)
```
I can try defining a new method for slices

didierofrivia:
The verification steps are working 👍🏼

pkg/rlptools/limit_index.go
eguzki: done

pkg/common/common.go
```diff
@@ -105,6 +112,19 @@ func SliceCopy[T any](s1 []T) []T {
 	return s2
 }
 
+func ReverseSlice[T any](input []T) []T {
```
guicassolato, line 115: Where is it used?

pkg/rlptools/limit_index.go
guicassolato: Will it fail without the `GetEmptySliceIfNil` at all?

If `Variables` and `Conditions` are slices rather than pointers, then both `len` and `range` should work even when the values were omitted in the CR, no?

pkg/rlptools/wasm_utils.go
```diff
@@ -26,6 +26,23 @@ type GatewayAction struct {
...
+func DefaultGatewayConfiguration(key client.ObjectKey) []kuadrantv1beta1.Configuration {
+	return []kuadrantv1beta1.Configuration{
+		{
+			Actions: []kuadrantv1beta1.ActionSpecifier{
+				{
+					GenericKey: &kuadrantv1beta1.GenericKeySpec{
+						DescriptorValue: key.String(),
+						// using default value as specified in Envoy spec
+						// https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action-generickey
+						DescriptorKey: &[]string{"generic_key"}[0],
```
guicassolato, line 38: I wonder if there's any difference from the compiler's POV between defining a simple `string` var vs. a fixed length-1 `[]string`. In the end, what we want here is the pointer to a string, right?

guicassolato, line 38: What about the following instead?

```suggestion
						DescriptorKey: &[]string{"ratelimitpolicy"}[0],
```

pkg/rlptools/limit_index.go
eguzki: sorry, I do not follow. You mean using the `GetDefaultIfNil` func?

pkg/common/common.go
eguzki: tests! But as it is quite generic, I decided to include in the common package. 

I can move thatto a test only code if you do not want code unused in "production"

pkg/rlptools/wasm_utils.go
eguzki: it's a way to specify an address of a string literal. Let me know if there is a better way

eguzki: does not make any diff and I was using the envoys RLS default value. It could be anything, it should not conflict as it is namespaced per gateway *and* domain

guicassolato: > does not make any diff 

At least it means something for one who reads yielded config. "generic_key" means nothing IMO.

pkg/common/common.go
guicassolato: Sorry. I don't see it used. Not even in the tests.

pkg/rlptools/limit_index.go
guicassolato: I mean if we do

```go
Variables: limitList[idx].Variables,
Conditions: limitList[idx].Conditions,
```

Will it fail for the minimum RLP that does not include neither `variables` nor `conditions`? I'm guessing it will not, but I may be wrong.

pkg/common/common.go
eguzki: here https://github.com/Kuadrant/kuadrant-operator/pull/144/files#diff-b4522ffe746a4ae4e75451963ec5b9d946b1ea0c5524df85a3e897384c816b8cR75

pkg/rlptools/limit_index.go
eguzki: Yeah, it cannot be `nil`. Hence that change in the code. Limitador's CRD does not allow it to be empty, it is required to be empty at least

guicassolato (approved):
Thanks for addressing the comments. I think [this](https://github.com/Kuadrant/kuadrant-operator/pull/144/files#r1104364277) is only still relevant, but even so I'm guessing it won't matter for too long.

Verification steps work. Approved.

pkg/rlptools/limit_index.go
didierofrivia: Maybe you could've used

```suggestion
					Variables:  common.GetDefaultIfNil(&limitList[idx].Variables, make([]string, 0)),
					Conditions: common.GetDefaultIfNil(&limitList[idx].Conditions, make([]string, 0)),
```

That should compile, or I'm missing something?

pkg/rlptools/wasm_utils.go
eguzki: done

guicassolato (approved):

Issue #146, opened 2023-02-08
Generate a tag with fixed image on each merge


pehala:
# Problem

Currently, only the `latest` tag is built on each merge, which logically uses the `latest` tag for all associated images. This, however, complicates things for QE because images can change under QE hands and will change depending on when it was installed (e.g. a new image of Limitador might have dropped and will automatically be used).

# Solution

Generate a new tag (e.g. based on a commit sha), and in that tag, pin the images to a specific digest, similar to how Authorino and its related repositories already do that.

## Side-effects

* This will certainly generate a lot of "junk" in quay repositories
   * Can be addressed when it actually becomes a problem with periodic cleaning
* Hard to propagate changes from Authorino and Limitador to Kuadrant, meaning there would have to be some triggers that would trigger build when Authorino or Limitador changes.
   * Can be done later. For the initial solution, a Kuadrant-only trigger will suffice (e.g. when something in the kuadrant-operator changes, generate a build) 
* It is almost like doing a release (without docs, changelog etc.) on every merge, so creating a pipeline now will help with the release process later

PR #149, opened 2023-03-02
Schedule build images with git sha reference

didierofrivia:
This PR introduces a series of changes that makes it possible to:

1. Build (and push) images and tag them with it's [SHA commit reference](https://git-scm.com/book/en/v2/Git-Internals-Git-References)
2. Schedules a daily workflow that collects the dependencies ([limitador-operator-bundle](https://quay.io/repository/kuadrant/liimtador-operator-bundle?tab=tags), [authorino-operator-bundle](https://quay.io/repository/kuadrant/authorino-operator-bundle?tab=tags) and [wasm-shim](https://quay.io/repository/kuadrant/wasm-shim?tab=tags))latest SHA references  and stores it in this repo's [action variables](https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28)
3. Schedules a weekly build, providing the `make bundle` target the dependencies stored variables already collected.
4. It also provides a `workflow_dispatch` event, so the user can manually trigger the whole bundle build with specific versions

Closes https://github.com/Kuadrant/kuadrant-operator/issues/146

Other PRs that complements this one:
* https://github.com/Kuadrant/limitador-operator/pull/63
* https://github.com/Kuadrant/wasm-shim/pull/30


.github/workflows/build-images.yaml
```diff
@@ -4,12 +4,28 @@ on:
...
+  workflow_call:
+  workflow_dispatch:
+    inputs:
+      authorinoOperatorBundleVersion:
+        description: Authorino Operator bundle version
+        required: true
+        default: latest
+      limitadorOperatorBundleVersion:
+        description: Limitador Operator bundle version
+        required: true
```
pehala, line 16: Why are these vars required when they are fetched in update-store-dependencies?

didierofrivia: Those are required if triggered by _workflow_dispatch_, say you want to build a specific bundle with a unique combination of tags, you'll need to provide each one. The ones fetched from _action variables_ are used when a _workflow_call_ was triggered.

.github/workflows/build-images.yaml
```diff
@@ -66,17 +95,54 @@ jobs:
...
+      - name: Set Operator version
+        id: operator-version
+        run: |
+          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
+          echo "VERSION=${GITHUB_REF_NAME/\//-}" >> $GITHUB_ENV
+          else
+          echo "VERSION=${{ github.sha }}" >> $GITHUB_ENV
+          fi
+      - name: Add SHA tags if triggered by scheduled workflow
+        if: ${{ github.event_name == "workflow_call" }}
```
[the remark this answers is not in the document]
didierofrivia: The conditional steps setup the envs passed to the build bundle, if nothing matches, goes to defaults in the Makefile... or the particular event will set the desired variables. I found it the least verbose, but it could be more explicit if moved to a different job altogether and triggered conditionally. I could do this change if needed or maybe we could elaborate a bit more on your proposition? let me know :)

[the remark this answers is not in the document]
didierofrivia: Makes total sense, thanks for the suggestion. Take a look and let me know your thoughts

.github/workflows/build-images.yaml
```diff
@@ -66,17 +102,41 @@ jobs:
...
+      - name: Run make bundle
+        run: |
+          if [[ ${github.event_name} == "workflow_call" || ${github.event_name} == "workflow_dispatch" ]]; then
+          echo "Building bundle via workflow call"
+          make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} VERSION=${TAG_NAME/v/} \
+            AUTHORINO_OPERATOR_VERSION=${github.event.inputs.authorinoOperatorBundleVersion} \
+            LIMITADOR_OPERATOR_VERSION=${github.event.inputs.limitadorOperatorBundleVersion} \
+            WASM_SHIM_VERSION=${github.event.inputs.wasmShimVersion}
+          else
+          make bundle REGISTRY=${{ env.IMG_REGISTRY_HOST }} ORG=${{ env.IMG_REGISTRY_ORG }} IMAGE_TAG=${{ github.ref_name }}
```
pehala, line 133: Is the special treatment for on_push needed? Wouldn't it also work if you supplied it with "latest" tags in `AUTHORINO_OPERATOR_VERSION`?

didierofrivia: Actually, I forgot to remove that conditional... but the default values in the Makefile target are set to latest.

pehala (approved):

eguzki:
what is the value of the weekly build?

eguzki:
If one of the dependencies is updated, the kuadrant operator generates a new bundle to update the dependency?

eguzki:
Do we still build operator bundle with references to latest?? would be good to have them as well.

didierofrivia:
@eguzki thanks for the review! in response to your comments:

> what is the value of the weekly build?

The weekly build will satisfy the needs of QE of having a "snapshot" of kuadrant operator with every dependency pointing to their sha tag stored at the time.


> If one of the dependencies is updated, the kuadrant operator generates a new bundle to update the dependency?

Current solution doesn't support this behaviour, it relies only on the weekly build with the dependencies version stored in action variables. These vars are updated on a daily job requesting the latest reference (commit sha)

> Do we still build operator bundle with references to latest?? would be good to have them as well.

We still add the _latest_ tag when the push event targets the main branch.

PR #150, opened 2023-03-08
Update limitador api to 0.4.0

eguzki:
### what
* Update limitador api to 0.4.0
* nil and empty array are equal for conditions and variables


didierofrivia (approved):
🥐

PR #153, opened 2023-03-23
Update gateway-api module to v0.6.2

mikenairn:
Update gateway-api module to v0.6.2

- Since v0.6.0 GatewayClass, Gateway and HTTPRoute have moved to the v1beta1 version of the API and support in v1alpha2 will be removed eventually see https://github.com/kubernetes-sigs/gateway-api/blob/main/CHANGELOG.md#deprecations. All references to these APIS are now updated to use v1beta1. PolicyTargetReference are still v1alpha2.

Update controller runtime
Use github.com/onsi/ginkgo/v2 which is used by the latest version if controller-runtime (envtest).
Set go version 1.19


mikenairn:
We are currently adding the rate limiting policy into the control plane for the multi cluster ~~traffic~~ gateway controller https://github.com/Kuadrant/multicluster-gateway-controller/issues/116, but hit an issue with a dependency version mismatch for `sigs.k8s.io/gateway-api` when trying to add the kuadrant operator as a dependency (This defines the rate limit policy so easiest for us to add this as a dependency and add the scheme from here). We are currently using 0.6.0 and the kuadrant-operator is on  0.5.1. We could downgrade if needs be, but preferably we would just update everything to the latest version which is currently 0.6.2.

Any issues with doing this here? What level of testing needs to be done here to ensure this is all still working as expected, or are the integration tests enough?

/cc @eguzki @guicassolato @maleck13

eguzki:
until we reach v1 and become more conservative, it is safe to upgrade whenever we need.

eguzki:
changes look good. But there is an issue with the e2e tests.

mikenairn:
Thanks @eguzki. Some of the constants we are using in our code seem to only be available in v1beta1 now. I missed one of them in the integration test, should hopefully pass now.

pkg/common/gatewayapi_utils.go
```diff
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
```
eguzki, line 12: why keep using v1alpha2?

mikenairn: Not everything seems to have moved to v1beta1, so i just stuck with the minimal changes to make it work with the latest version.

david-martin:
This PR came up on the Kuadrant call today.
There's an outstanding action to do a release from main with the current version of gateway-api first(@pehala and @didierofrivia have taken an action there).

After that, upgrading beyond that version may lose compatibility with Openshift Service Mesh. But I think that's OK as we need to move forward, and OSSM will catch up in time.
Plus people can use the older version of kuadrant for targetting OSSM if needed.

There's a separate action from @alexsnaps to try out the latest rc of 0.7.0 and eventually update to that version.
So the order of play is:

- release from main with 0.5.1 of gateway api
- merge this bump to 0.6.2 (and optionally release that too)
- merge a future PR to bump to 0.7.x

pkg/common/gatewayapi_utils.go
guicassolato: It looks like `PolicyTargetReference` is now the only type we use that still requires v1alpha2.

mikenairn: I believe so yes

mikenairn:
This lgtm. If you are merging, can you hold off deleting the branch as it will break the MGC local-setup https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/config/kuadrant/kustomization.yaml#L2. I'll update that after merge and delete this branch then.

eguzki (approved):
moving forward!

PR #157, opened 2023-04-03
Inheriting all secrets from caller workflow

didierofrivia:
Aiming to fix the following error when pushing to Quay:

```
Error: writing blob: initiating layer upload to /v2/kuadrant/kuadrant-operator/blobs/uploads/ in quay.io: unauthorized: access to the requested resource is not authorized
```

It should be fixed using [secrets: inherit](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idsecretsinherit)

Apparently the only secret available on the called workflow is the `secrets.GITHUB_TOKEN`

Note: It's not possible to check this out using act.

### DIFF
--- a/.github/workflows/scheduled-build.yaml
+++ b/.github/workflows/scheduled-build.yaml
@@ -11,6 +11,7 @@ jobs:
   workflow-build:
     name: Calls build-images workflow
     uses: ./.github/workflows/build-images.yaml
+    secrets: inherit
     with:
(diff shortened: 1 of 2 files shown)


pehala (approved):

alexsnaps:
Not really related, but it'd be great if the push to quay only happens if secrets are actually set. When forking, the actions fails as obviously my personal account doesn't have these (but that's also true for anyone wanting to work on a fork)

alexsnaps (approved):

PR #158, opened 2023-04-04
Fix bundle generation

didierofrivia:
This PR aims to fix the bundle image generation, which currently is listing wrong versions, then makes it fail the catalog generation:

https://github.com/Kuadrant/kuadrant-operator/actions/runs/4598777582/jobs/8123707273

```
[ERROR] Referenced Limitador version is , but found 0.0.0 in the bundle 
make[1]: *** [make/catalog.mk:26: /home/runner/work/kuadrant-operator/kuadrant-operator/catalog/kuadrant-operator-catalog/operator.yaml] Error 1
make[1]: Leaving directory '/home/runner/work/kuadrant-operator/kuadrant-operator'
make: *** [make/catalog.mk:35: catalog] Error 2
```

The following will be applied:

* Version will be either _semantic_ or _github.sha_
* Tag will be `latest` when main branch, `semantic` when release and `github.sha` as default
* The final images will be tagged with a combo of sha, latest, branch and semantic as applicable

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -107,12 +107,17 @@ jobs:
         if: ${{ github.ref_name == env.MAIN_BRANCH_NAME }}
         id: add-latest-tag
         run: |
+          echo "TAG_NAME=latest" >> $GITHUB_ENV
           echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Add release tag
+      - name: Add tag based on branch
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         id: add-branch-tag
         run: |
+          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
           TAG_NAME=${GITHUB_REF_NAME/\//-}
+          else
+          TAG_NAME=${{ github.sha }}
+          fi
           echo "TAG_NAME=${TAG_NAME}" >> $GITHUB_ENV
           echo "IMG_TAGS=${TAG_NAME} ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
       - name: Set Operator version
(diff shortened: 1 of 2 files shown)


.github/workflows/build-images.yaml
```diff
@@ -107,12 +107,17 @@ jobs:
...
         id: add-latest-tag
         run: |
+          echo "TAG_NAME=latest" >> $GITHUB_ENV
           echo "IMG_TAGS=latest ${{ env.IMG_TAGS }}" >> $GITHUB_ENV
-      - name: Add release tag
+      - name: Add tag based on branch
         if: ${{ github.ref_name != env.MAIN_BRANCH_NAME }}
         id: add-branch-tag
         run: |
+          if [[ ${GITHUB_REF_NAME/\//-} =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ ]]; then
```
eguzki, line 116: What about preventing all this ugly conditionals with a new `build-image-on-tag.yaml` with 
```yaml
name: Publish in GitHub Package Registry on tag push

on:  
  push:
    tags:
      - 'v*'
```

Makefile
```diff
@@ -60,12 +60,6 @@ ORG ?= kuadrant
 # quay.io/kuadrant/kuadrant-operator-bundle:$VERSION and quay.io/kuadrant/kuadrant-operator-catalog:$VERSION.
 IMAGE_TAG_BASE ?= $(REGISTRY)/$(ORG)/kuadrant-operator
 
-ifeq (0.0.0,$(VERSION))
```
eguzki, line 63: Why removing this? this is pretty and neat way to set docker image tag when not `0.0.0`

eguzki (changes requested):

.github/workflows/build-images.yaml
didierofrivia: I reckon that in order to get rid of all the conditionals we might need either specific steps or workflows to the following options for not only versions and tags, but dependencies too:

1. Push: main, pull request regular branch, release and sha
2. Scheduled: sha builds 
3. Workflow dispatch: custom

Might be good to address the cleaner way including all our use cases in a following PR?

Makefile
didierofrivia: I was thinking the use case is covered here: https://github.com/Kuadrant/kuadrant-operator/pull/158/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L37-L41

Basically if the `VERSION` is 0.0.0 would enter the first conditional and set the tag to `latest`. If it's semantic, would set it to the `vx.y.z` and the default would be the default or the one passed.

Otherwise with the removed code it was setting the `IMAGE_TAG` to `v$(VERSION)` either when `VERSION` was a _sha_

didierofrivia:
Still not fixing the main issue, will do more debugging

didierofrivia:
Fixed for regular bundle builds, once merged, will do a test for `workflow_dispatch`

eguzki (changes requested):
You need to move the SHELL makefile var to the top

```
make bundle
/bin/sh: 1: Syntax error: "(" unexpected
```

Issue #159, opened 2023-04-05
Kind Cluster. Automatic port forwarding is broken on linux


art-tapin:
When following the [Simple Rate Limit For API Owners](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/simple-rl-for-api-owners.md) guide, in the 'Check toystore HTTPRoute works' section it states in the Note that `curl` works 'out-of-the-box' only on linux environments. When running `curl -v -H 'Host: api.toystore.com' http://localhost:9080/toy` command on Fedora 37, I have the following response:
```bash
*   Trying 127.0.0.1:9080...
* connect to 127.0.0.1 port 9080 failed: Connection refused
*   Trying ::1:9080...
* connect to ::1 port 9080 failed: Connection refused
* Failed to connect to localhost port 9080 after 0 ms: Connection refused
* Closing connection 0
curl: (7) Failed to connect to localhost port 9080 after 0 ms: Connection refused
```

After forwarding the ports by running `kubectl port-forward -n istio-system service/istio-ingressgateway 9080:80 &` it worked as expected.

PR #160, opened 2023-04-06
Update kind-cluster config

art-tapin:
Description:

- Update kind-cluster.yaml with additional port mappings to expose Istio ingress gateway on the host machine
- Bump the version of Kubernetes image for kind-cluster to v1.24.12

Testing:

- Verified that the command `curl -v -H 'Host: api.toystore.com' http://localhost:9080/toy` works out of the box on Linux environment (Fedora 37) without needing to run kubectl port-forward
- While testing was also done with more recent stable releases of Kubernetes (1.25 and 1.26),  these versions did not work properly with the Kuadrant Operator, returning `curl: (52) Empty reply from server`. The latest stable release that is compatible with our codebase  is 1.24.12.

Closes https://github.com/Kuadrant/kuadrant-operator/issues/159


eguzki:
looking good

eguzki (approved):

didierofrivia (approved):
🏅

PR #161, opened 2023-04-07
Refactoring Github workflows

didierofrivia:
TODO


.github/workflows/build-images-release.yaml
```diff
@@ -0,0 +1,20 @@
+name: Build images for release
+
+on:
+  push:
+    tags: ['*']
+
+jobs:
+  workflow-build:
+    if: ${{ github.ref_name =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-.+)?$ }}
```
eguzki, line 9: This snippet does not work??

```yaml
on:
  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

didierofrivia: Some how it wasn't triggering in Act, that's why I change it to the job conditional. I could do that way, it might work with Github

eguzki (approved):

PR #166, opened 2023-04-17
Istio (v1.17.2) and Gateway API (v0.6.2) version bump

guicassolato:
- [x] Bumps the version of Istio used for local dev and testing to v1.17.2.
- [x] Bumps the version of Gateway API to v0.6.2

<br/>

**Other co-dependencies updated:**
- sigs.k8s.io/controller-runtime (v0.14.6)
- go.uber.org/zap (v1.24.0)


guicassolato:
This is failing because of this: https://github.com/istio/istio.io/issues/12630#issuecomment-1426402223

I.e. the second gateway created by the integration test case at https://github.com/Kuadrant/kuadrant-operator/blob/5af4c71fcd41f14ba51b6045a35989d7fd3929d3/controllers/authpolicy_controller_test.go#L42-L43

won't be ready, thus failing at this check that is performed by the controller when the `targetRef.Kind == "Gateway"`: https://github.com/Kuadrant/kuadrant-operator/blob/5af4c71fcd41f14ba51b6045a35989d7fd3929d3/pkg/reconcilers/targetref_reconciler.go#L55-L57

eguzki (approved):

mikenairn:
This seems to be making the same change to the gateway-api module as https://github.com/Kuadrant/kuadrant-operator/pull/153, but with far fewer changes. I think it;s becaue you haven;t change the api version to v1beta1 https://github.com/Kuadrant/kuadrant-operator/pull/153#discussion_r1146261371. DO we not want to do that?

guicassolato:
> This seems to be making the same change to the gateway-api module as #153, but with far fewer changes. I think it's because you haven't changed the api version to v1beta1 [#153 (comment)](https://github.com/Kuadrant/kuadrant-operator/pull/153#discussion_r1146261371). DO we not want to do that?

@mikenairn

Focus here started on upgrading Istio. I ended bringing GWAPI v0.6 together due to dependencies.

With many CRDs now promoted to beta, I believe we do want to upgrade them in our code base as well. Perhaps an obvious consecutive step then would be rebasing and merging #153, right after this PR? Or you'd say it better done all at once as part of the same iteration? Either way works for me.

To have in mind that having Istio up to date is also important before moving to GWAPI v0.6.2. And related to that, this PR also fixes the status check of Gateway objects – from using `Ready` to now `Programmed`.

mikenairn:
> With many CRDs now promoted to beta, I believe we do want to upgrade them in our code base as well. Perhaps an obvious consecutive step then would be rebasing and merging #153, right after this PR? Or you'd say it better done all at once as part of the same iteration? Either way works for me.

Yeah, lets just merge this and I'll update my PR to just update the api version we are using.

mikenairn (approved):

guicassolato:
I've tested the changes in this PR with a custom build of the operator, bundle and catalog images pushed to quay.io:
1. in a local Kubernetes cluster ([instructions](http://kuadrant.io/docs/kuadrant-operator/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.html), with step 1 replaced by `make local-setup`), and
2. in an OSD cluster with OpenShift Service Mesh 2.3.3 ([instructions](https://cloud.redhat.com/blog/rate-limiting-and-authnz-for-openshift-service-mesh-with-kuadrant-developer-preview), with Kuadrant install replaced by `make deploy-catalog CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:gwapi-v0.6.2`)

The changes worked in both scenarios.

In OpenShift, to workaround errors in OSSM related to the GWAPI version, I had to first install an older version of GWAPI and then upgrade it to v0.6.2 after OSSM had successfully started `istiod-controlplane`.

Issue #167, opened 2023-04-17
Improve test coverage for package common


didierofrivia:
Currently the test coverage it's 52.0% of statements, ideally the percentage should be the closest to 100% we can achieve.

The following functions are currently not tested:

* [common/apimachinery_status_conditions.go](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/apimachinery_status_conditions.go)
  * CopyConditions
  * ConditionMarshal
* [common/authorino_conditions.go](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/authorino_conditions.go)
  * FindAuthorinoStatusCondition
* [common/hostname.go](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/hostname.go):
  * IsWildCarded
  * String
* [common/common.go](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/authorino_conditions.go)
  * FetchEnv
  * GetDefaultIfNil
  * GetEmptySliceIfNil
  * NamespacedNameToObjectKey
  * Contains
  * Map
  * SliceCopy
  * ReverseSlice
  * MergeMapStringString
  * UnMarshallLimitNamespace
  * MarshallNamespace
  * UnMarshallObjectKey
  * HostnamesToStrings
* [common/k8s_utils](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/common/k8s_utils.go)
  * ObjectInfo
  * ReadAnnotationsFromObject
  * TagObjectToDelete
  * IsObjectTaggedToDelete
  * StatusConditionsMarshalJSON
  * IsOwnedBy
  * GetServicePortNumber
  * ObjectKeyListDifference
  * ContainsObjectKey
  * FindObjectKey
  * FindDeploymentStatusCondition
* [common/yaml_decoder.go](https://github.com/Kuadrant/kuadrant-operator/tree/main/pkg/common)/yaml_decoder.go
  * DecodeFile

PR #176, opened 2023-04-28
[test] Optimizations, improvements, and unit tests for common/common.go (part 1 of 3)

art-tapin:
This pull request contains optimizations, improvements, and unit-tests for several functions in the common/common.go file. The changes made are as follows:

- Refactored the NamespacedNameToObjectKey function to optimize its performance and added unit-tests for it.
- Simplified the GetEmptySliceIfNil function for improved performance and added unit-tests for it.
- Added comments for the FetchEnv and GetDefaultIfNil functions.
- Added unit-tests for the FetchEnv and GetDefaultIfNil functions.

Note that this is only the first of three pull requests, and we have only tested 1/3 of the untested functions in the common/common.go file, so there is more work to be done.

This pull request partly closes issue #167.

Coverage: 59.2% of statements

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -47,6 +47,10 @@ type KuadrantPolicy interface {
 	GetRulesHostnames() []string
 }
 
+// FetchEnv fetches the value of the environment variable with the specified key,
+// or returns the default value if the variable is not found or has an empty value.
+// If an error occurs during the lookup, the function returns the default value.
+// The key and default value parameters must be valid strings.
 func FetchEnv(key string, def string) string {
 	val, ok := os.LookupEnv(key)
 	if !ok {
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -60,6 +60,7 @@ func FetchEnv(key string, def string) string {
 	return val
 }
 
+// GetDefaultIfNil returns the value of a pointer argument, or a default value if the pointer is nil.
 func GetDefaultIfNil[T any](val *T, def T) T {
 	if reflect.ValueOf(val).IsNil() {
 		return def
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -68,6 +68,7 @@ func GetDefaultIfNil[T any](val *T, def T) T {
 	return *val
 }
 
+// GetEmptySliceIfNil returns a provided slice, or an empty slice of the same type if the input slice is nil.
 func GetEmptySliceIfNil[T any](val []T) []T {
 	if val == nil {
 		return make([]T, 0)
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -79,11 +79,10 @@ func GetEmptySliceIfNil[T any](val []T) []T {
 // NamespacedNameToObjectKey converts <namespace/name> format string to k8s object key.
 // It's common for K8s to reference an object using this format. For e.g. gateways in VirtualService.
 func NamespacedNameToObjectKey(namespacedName, defaultNamespace string) client.ObjectKey {
-	split := strings.Split(namespacedName, "/")
-	if len(split) == 2 {
-		return client.ObjectKey{Name: split[1], Namespace: split[0]}
+	if i := strings.IndexRune(namespacedName, '/'); i >= 0 {
+		return client.ObjectKey{Namespace: namespacedName[:i], Name: namespacedName[i+1:]}
 	}
-	return client.ObjectKey{Namespace: defaultNamespace, Name: split[0]}
+	return client.ObjectKey{Namespace: defaultNamespace, Name: namespacedName}
 }
 
 func Contains(slice []string, target string) bool {
--- a/pkg/common/common_test.go
+++ b/pkg/common/common_test.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"reflect"
 	"testing"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func TestValidSubdomains(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/common/common.go
```diff
@@ -56,28 +60,29 @@ func FetchEnv(key string, def string) string {
...
 func GetDefaultIfNil[T any](val *T, def T) T {
 	if reflect.ValueOf(val).IsNil() {
 		return def
 	}
 	return *val
 }
 
+// GetEmptySliceIfNil returns a provided slice, or an empty slice of the same type if the input slice is nil.
 func GetEmptySliceIfNil[T any](val []T) []T {
-	if val == nil {
```
eguzki, line 67: This refactor is not wrong, per se, because it does not change behavior. So my question is why the change. I might be missing something. 

If the input slice is empty, it will enter the `if` branch and it will return a new empty slice. While in the "old" implementation, the method would return the same empty input slice.

pkg/common/common.go
```diff
@@ -56,28 +60,29 @@ func FetchEnv(key string, def string) string {
...
 	return val
 }
 
 // NamespacedNameToObjectKey converts <namespace/name> format string to k8s object key.
 // It's common for K8s to reference an object using this format. For e.g. gateways in VirtualService.
 func NamespacedNameToObjectKey(namespacedName, defaultNamespace string) client.ObjectKey {
-	split := strings.Split(namespacedName, "/")
-	if len(split) == 2 {
-		return client.ObjectKey{Name: split[1], Namespace: split[0]}
+	if i := strings.IndexRune(namespacedName, '/'); i >= 0 {
```
eguzki, line 82: Is there any difference here comparing to the "old" implementation?

pkg/common/common.go
art-tapin: The change from 'val == nil' to 'len(val) == 0' in the if statement really doesn't affect the behavior of the code. I think, I wanted to change only the return statement, but left the len-check because this was the last thing I was also benchmarking and they were almost identical. So, you are right, thanks for noticing that! Do you want me to revert this change in if-statement?

pkg/common/common.go
art-tapin: Yes, indeed it is! According to results of the benchmark I did, this new implementation is approximately 12-13 times faster than the previous implementation.

strings.IndexRune() is a more lightweight operation than strings.Split(). When you call strings.Split() on a string, the Go runtime must do a significant amount of work to split the string into an array. In contrast, when you call strings.IndexRune(), the runtime only needs to find the index of the first occurrence of a single character in the string.

Hope, that answeres your question, @eguzki! :angel:

pkg/common/common.go
didierofrivia: Yes, revert the change. By rule of thumb, never change code while writing tests unless you find a bug ;)

pkg/common/common.go
didierofrivia: As I mentioned before, always better to not include changes when writing tests... however, since this benchmark is considerable faster than the other, I'd say to leave it...

pkg/common/common_test.go
```diff
@@ -67,3 +71,192 @@ func TestFind(t *testing.T) {
...
+		val := "info"
+		os.Setenv(key, val)
+		defer os.Unsetenv(key)
+
+		result := FetchEnv(key, defaultVal)
+
+		if result != val {
+			t.Errorf("Expected %v, but got %v", val, result)
+		}
+	})
```
didierofrivia, line 101: I'd ditch this assertion, since it's not meaningful and would only add time to the whole battery of tests

pkg/common/common_test.go
```diff
@@ -67,3 +71,192 @@ func TestFind(t *testing.T) {
...
+		defaultVal := "production"
+		val := ""
+		os.Setenv(key, val)
+		defer os.Unsetenv(key)
+
+		result := FetchEnv(key, defaultVal)
+
+		if result != val {
+			t.Errorf("Expected %v, but got %v", val, result)
+		}
```
didierofrivia, line 113: This is an interesting case, when it's empty... would we want this behaviour @eguzki , wdyt?

pkg/common/common_test.go
```diff
@@ -67,3 +71,192 @@ func TestFind(t *testing.T) {
...
+	t.Run("when key is empty or invalid then return the default value", func(t *testing.T) {
+		key := ""
+		defaultVal := "production"
+
+		result := FetchEnv(key, defaultVal)
+
+		if result != defaultVal {
+			t.Errorf("Expected %v, but got %v", defaultVal, result)
+		}
+	})
```
didierofrivia, line 134: This one can be removed too, we're not expecting an _empty_ env var name

pkg/common/common_test.go
```diff
@@ -67,3 +71,192 @@ func TestFind(t *testing.T) {
...
+		key := "MY_VAR"
+		defaultVal := ""
+
+		result := FetchEnv(key, defaultVal)
+
+		if result != defaultVal {
+			t.Errorf("Expected %v, but got %v", defaultVal, result)
+		}
+	})
+}
```
didierofrivia, line 167: This is the same as https://github.com/Kuadrant/kuadrant-operator/pull/176/files#diff-7b84e51157fdc20b9d781ec3c17f206f4c5c72ae17e90da9658a2669b0e0aeb4R115

pkg/common/common_test.go
```diff
@@ -67,3 +71,192 @@ func TestFind(t *testing.T) {
...
+
+		result := NamespacedNameToObjectKey(namespacedName, defaultNamespace)
+
+		expected := client.ObjectKey{Name: "test-name", Namespace: "default"}
+		if !reflect.DeepEqual(result, expected) {
+			t.Errorf("Expected %v, but got %v", expected, result)
+		}
+	})
+
+	t.Run("when an empty string is provided, then return an ObjectKey with default namespace and empty name", func(t *testing.T) {
```
didierofrivia, line 251: I'm wondering if we should also return error, since it's mandatory to have a _name_ for querying k8s API... wdyt @eguzki ?

didierofrivia (changes requested):
Looking good! a few bits to change and others to decide if we want to address them in this PR or not

pkg/common/common_test.go
eguzki: I would say so. The env var is set. Empty value is still a value.

pkg/common/common_test.go
eguzki: having `NamespacedNameToObjectKey` return an error for the case the name is empty seems overcomplicating. The function does what it says and if you provide an empty string, you would get empty name (plus the default namespace). I think that the error triggering should lay on the callers, not on this method. But just my opinion...

pkg/common/common.go
art-tapin: Next time I'll create a separate pull request! 🫡🫡🫡

pkg/common/common_test.go
```diff
@@ -67,3 +71,159 @@ func TestFind(t *testing.T) {
...
+		key := "LOG_MODE"
+		defaultVal := "production"
+
+		result := FetchEnv(key, defaultVal)
+
+		if result != defaultVal {
+			t.Errorf("Expected %v, but got %v", defaultVal, result)
+		}
+	})
+	t.Run("when a runtime error occurs during execution then return the default value", func(t *testing.T) {
```
eguzki, line 112: where is the runtime error?

pkg/common/common_test.go
```diff
@@ -67,3 +71,159 @@ func TestFind(t *testing.T) {
...
+		var val *string
+		def := "default"
+
+		result := GetDefaultIfNil(val, def)
+
+		if result != def {
+			t.Errorf("Expected %v, but got %v", def, result)
+		}
+	})
+	t.Run("when value is not a pointer type and default value is provided then return default value", func(t *testing.T) {
```
eguzki, line 157: this test is somewhat a duplication. The value is non nil pointer, hence the value of `val` (0 in the test) must be returned.

pkg/common/common_test.go
art-tapin: I double-checked and yes the previous test-case covers that. Unsetting the env value leads to the same state as 'env value does not exist', so you are right, it is redundant! I am going to get rid of this test-case
Thanks, @eguzki !

pkg/common/common_test.go
art-tapin: I see, it is alrady initialised with a default value (in this case 0), so technically it is an initialised non-nil value. I'll delete this one as well then!

eguzki (approved):

didierofrivia (approved):
🎖️  and extra 🌮  for documenting the functions!

Issue #156, opened 2023-03-30
Implement RLPv2 Direct Policy Attachment


guicassolato:
- [x] #179
- [x] #200 (Kuadrant/kuadrant-operator#204)
- [x] #218 (reconciliation of Limitador CR)

PR #180, opened 2023-05-05
RLP v2 CRD

### LINKED ISSUE #179: New RLP CRD version
From [RFC 0001](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md)

Group | Version | Kind
-- | -- | --
kuadrant.io | v2beta1 | RateLimitPolicy

Example with all fields:

```yaml
apiVersion: kuadrant.io/v2beta1
kind: RateLimitPolicy
metadata: {}
spec:
  targetRef:
    kind: HTTPRoute
    name: <HTTPROUTE-NAME>
  limits:
    <limit-name>:
      routeSelectors:           
      - hostnames:
         - "*example.com"
        matches:                # []HTTPRouteMatch
        - path:                      # HTTPPathMatch 
            type: PathPrefix
            value: "/toys"
          method: GET          # HTTPMethod 
          headers:                # []HTTPHeaderMatch 
          - type: Exact
            name: "X-Canary"
            value: "1"
          queryParams:      # []HTTPQueryParamMatch 
          - type: Exact
            name: "q"
            value: "1"
      rates:
      - limit: 50
        duration: 1
        unit: minute
      counters:
      - auth.identity.username
      - context.request.http.path
      when:
      - selector: auth.identity.group
        operator: neq
        value: admin

```

eguzki:
Fixes #179

From [RFC 0001](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md)

Group | Version | Kind
-- | -- | --
kuadrant.io | v2beta1 | RateLimitPolicy

Example with all fields:

```yaml
apiVersion: kuadrant.io/v2beta1
kind: RateLimitPolicy
metadata: {}
spec:
  targetRef:
    kind: HTTPRoute
    name: <HTTPROUTE-NAME>
  limits:
    <limit-name>:
      routeSelectors:           
      - hostnames:
         - "*example.com"
        matches:                # []HTTPRouteMatch
        - path:                      # HTTPPathMatch 
            type: PathPrefix
            value: "/toys"
          method: GET          # HTTPMethod 
          headers:                # []HTTPHeaderMatch 
          - type: Exact
            name: "X-Canary"
            value: "1"
          queryParams:      # []HTTPQueryParamMatch 
          - type: Exact
            name: "q"
            value: "1"
      rates:
      - limit: 50
        duration: 1
        unit: minute
      counters:
      - auth.identity.username
      - context.request.http.path
      when:
      - selector: auth.identity.group
        operator: neq
        value: admin

```


config/manager/manager.yaml
```diff
@@ -25,33 +25,33 @@ spec:
...
-          initialDelaySeconds: 5
-          periodSeconds: 10
-        resources:
-          limits:
-            cpu: 200m
-            memory: 300Mi
-          requests:
-            cpu: 200m
-            memory: 200Mi
+        - command:
```
eguzki, line 28: :shrug:

eguzki:
this PR targets https://github.com/Kuadrant/kuadrant-operator/tree/rlp-v2-base

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+	// Duration defines the time uni
+	// Possible values are: "second", "minute", "hour", "day"
+	Unit TimeUnit `json:"unit"`
+}
+
+// RouteSelector defines semantics for matching an HTTP request based on conditions
+// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteSpec
+type WhenCondition struct {
+	// Selector defines one item from the well known selectors
+	// TODO Document properly "Well-known selector" https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md#well-known-selectors
```
eguzki, line 63: Those *well known selectors* need to be properly documented

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ContextSelector defines one item from the well known selectors
+// TODO Document properly "Well-known selector" https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md#well-known-selectors
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type ContextSelector string
+
+// +kubebuilder:validation:Enum:=eq;neq;incl;excl;matches
```
guicassolato, line 40: For RLP, I thought we were going with only `eq` and `neq` for now, so nothing needs to change in Limitador.

```suggestion
// +kubebuilder:validation:Enum:=eq;neq
```

eguzki: Well, not necessarily `when` conditions are going to be applied in Limitador. The wasm-shim can  enforce those rules as well to save Limitador's traffic. 

But It's ok to start with only `eq` and `neq` for now. We can extend those later.

guicassolato: Probably a debate for another forum, when discussing the attributions of the wasm-whim rather...

You are right about the wasm-shim technically being capable of enforcing all kinds of user-defined conditions (route related or non-route related) if we want it to. After all, it's all data the wasm-shim will need to request anyway to build the descriptors for the RL request.

I personally have always thought on route-related conditions evaluated by the wasm-shim and translated into one single descriptor to be matched by one single Limitador condition, which is [the one with the identifier of the limit](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md#artificial-limitador-condition-for-routeselectors); and then conditions not related to route attributes, transparently passed along by the wasm-shim in the RL request and evaluated only by Limitador. But it doesn't have to be this way.

What I think you're suggesting as alternative is a bit bolder and could/would make the wasm-shim the solely evaluator of user-defined conditions effectively. Perhaps saving requests to Limitador is enough reason to do that.

Of course, requests to Limitador can only be saved when none of the limits have matching conditions, after all it takes only one single limit definition to have a matching condition for the wasm-shim to have to call Limitador. But perhaps the wasm-shim could be more selective and inject _less_ descriptors when building requests to Limitador – i.e., only the ones for matching limit identifiers, because other conditions were all evaluated upfront by the wasm-shim itself and the limits are known to have to be triggered?

So less requests to Limitador _and_ less conditions evaluated by Limitador in the cases where a RL request is unavoidable. 🤔

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ContextSelector defines one item from the well known selectors
+// TODO Document properly "Well-known selector" https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md#well-known-selectors
+// +kubebuilder:validation:MinLength=1
+// +kubebuilder:validation:MaxLength=253
+type ContextSelector string
```
didierofrivia, line 38: If we already have some identified, maybe we could `enum` them ?

eguzki: the list is too big and we also do not want to limit to a set of options, do we? It is all about a context with attributes that can be added dynamically.

Right, @guicassolato ? BTW, do we know if this is documented? do we have a link?

guicassolato: It is documented, but it isn't. Perhaps the closest thing we've got is https://github.com/Kuadrant/authorino/blob/main/docs/architecture.md#the-authorization-json, combined with https://www.envoyproxy.io/docs/envoy/latest/api-v3/service/auth/v3/external_auth.proto#envoy-v3-api-msg-service-auth-v3-checkrequest. What's not explicitly there is because it's really an open struct. But we do need a proper doc explaining it, with examples and all.

In the case of Authorino, the selectors include this as well: https://github.com/Kuadrant/authorino/blob/main/docs/features.md#common-feature-json-paths-valuefromauthjson

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+
+// RouteSelector defines semantics for matching an HTTP request based on conditions
+// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteSpec
+type WhenCondition struct {
+	// Selector defines one item from the well known selectors
+	// TODO Document properly "Well-known selector" https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md#well-known-selectors
+	Selector ContextSelector `json:"selector"`
+
+	// The binary operator to be applied to the content fetched from the selector
+	// Possible values are: "eq" (equal to), "neq" (not equal to), "incl" (includes; for arrays), "excl" (excludes; for arrays), "matches" (regex)
```
guicassolato, line 67: I think some of these were removed.

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+	Operator WhenConditionOperator `json:"operator"`
+
+	// The value of reference for the comparison.
+	// If used with the "matches" operator, the value must compile to a valid Golang regex.
+	Value string `json:"value"`
```
guicassolato, line 72: Authorino conditions support only string comparisons too. [I've been wondering](https://github.com/Kuadrant/authorino/issues/357#issuecomment-1502904440) lately if we shouldn't try to support any JSON/YAML value type. In the future, this may be helpful for numeric operators (e.g. greater than, lower than, etc), array lookups, etc. WDYT?

api/v2beta1/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// RouteSelector defines semantics for matching an HTTP request based on conditions
+// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteSpec
+type RouteSelector struct {
+	// Hostnames defines a set of hostname that should match against the HTTP Host header to select a HTTPRoute to process the request
+	// https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRouteSpec
+	// +optional
+	Hostnames []gatewayapiv1alpha2.Hostname `json:"hostnames,omitempty"`
```
guicassolato, line 81: This means a validation (which I like!) against the following pattern:
```
^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$
```

Only to highlight that the example provided in the description of the PR (`*example.com`) is **not** valid.

api/v2beta1/ratelimitpolicy_types.go
eguzki: The Envoy's `Ext-Authz CheckRequest.AttributeContext type` context may not be the same a wasm filter (or any filter) may access to. 

This is promising doc: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes

api/v2beta1/ratelimitpolicy_types.go
eguzki: Good point. For now, as the condition operator is either `eq` or `neq`, I would stick to literal (string).

api/v2beta1/ratelimitpolicy_types.go
eguzki: good catch! I will change it.

eguzki:
ready for a new review @guicassolato

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+	Status RateLimitPolicyStatus `json:"status,omitempty"`
+}
+
+func (r *RateLimitPolicy) Validate() error {
+	if r.Spec.TargetRef.Group != gatewayapiv1alpha2.Group("gateway.networking.k8s.io") {
+		return fmt.Errorf("invalid targetRef.Group %s. The only supported group is gateway.networking.k8s.io", r.Spec.TargetRef.Group)
+	}
+
+	if r.Spec.TargetRef.Kind != gatewayapiv1alpha2.Kind("HTTPRoute") && r.Spec.TargetRef.Kind != gatewayapiv1alpha2.Kind("Gateway") {
+		return fmt.Errorf("invalid targetRef.Kind %s. The only supported kind types are HTTPRoute and Gateway", r.Spec.TargetRef.Kind)
```
didierofrivia, line 176: Oh, so, we will still only cover the usage of `HTTPRoute` and `Gateway`. I thought for this v2 we were thinking on adding `GatewayClass`.. not sure about the `VirtualService`

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	if r.Spec.TargetRef.Namespace != nil && string(*r.Spec.TargetRef.Namespace) != r.Namespace {
+		return fmt.Errorf("invalid targetRef.Namespace %s. Currently only supporting references to the same namespace", *r.Spec.TargetRef.Namespace)
```
didierofrivia, line 180: Would be informative if we show in the error the valid namespace

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,217 @@
...
+		logger.V(1).Info("ObservedGeneration not equal", "difference", diff)
+		return false
+	}
+
+	// Marshalling sorts by condition type
+	currentMarshaledJSON, _ := common.ConditionMarshal(s.Conditions)
+	otherMarshaledJSON, _ := common.ConditionMarshal(other.Conditions)
+	if string(currentMarshaledJSON) != string(otherMarshaledJSON) {
+		diff := cmp.Diff(string(currentMarshaledJSON), string(otherMarshaledJSON))
+		logger.V(1).Info("Conditions not equal", "difference", diff)
```
guicassolato, line 146: Worth saving the cost of `cmp.Diff` if `!logger.V(1).Enabled()`?

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,217 @@
...
+		return false
+	}
+
+	// TODO(eastizle): reflect.DeepEqual does not work well with lists without order
+	if !reflect.DeepEqual(s.GatewaysRateLimits, other.GatewaysRateLimits) {
```
guicassolato, line 151: I wonder if order matters. If it does matter, do we want to enforce same order? I'm guessing "yes, we do", right?

guicassolato (approved):

api/v1beta2/ratelimitpolicy_types.go
eguzki: Currently, it is only HTTPRoute. https://github.com/Kuadrant/kuadrant-operator/issues/156

api/v1beta2/ratelimitpolicy_types.go
eguzki: It's the current one. The rate limit policy's namespace.

Anyway, this is only about the CRD we can improve that in coming PR's with actual implementation

api/v1beta2/ratelimitpolicy_types.go
eguzki: I have removed the `GatewaysRateLimits` struct from the status section of the CRD. I guess this is still a valid feature, but needs to be refactored entirely. We can add items to the status later when we implement the feature. Does it sound good to you?

PR #186, opened 2023-05-10
[test] Unit-tests for common/common.go (part 3 of 3)

art-tapin:
This pull request contains updated code and unit tests for several functions in the `common/common.go` file. Specifically:

- Added unit tests for the `HostnamesToStrings`, `UnMarshallObjectKey`, `MarshallNamespace`, and `UnMarshallLimitNamespace` functions.
- Add comment to the `UnMarshallLimitNamespace`.
- Added two **FAILING** test cases for the `UnMarshallLimitNamespace` function to test error handling when the gateway or domain name is missing. Here, for more information please look at commit [test: Add two FAILING unit-tests for UnMarshallLimitNamespace](https://github.com/Kuadrant/kuadrant-operator/commit/8b044acac8050acfdd2103665315c2d5ff18fcd7). Do we consider these two scenarios?

This is the last part (3 of 3) of testing the common package. Please note that this pull request only partially closes issue #167, and there may be further updates needed in the future.

Coverage: 68.5%

PR #187, opened 2023-05-10
Optimize common package functions for improved performance

art-tapin:
This pull request contains several optimizations to improve the performance of the functions in the `common package` as a part of working on #167. The commits included in this pull request are:

1. Optimized the `UnMarshallObjectKey` function to reduce unnecessary allocations and improve validity checks for input strings. This resulted in an average performance improvement of **17.3%** in average across 3 benchmarks with input sizes of 1000, 100, and 10. Additionally, this change partly fixes one of two failing test-case in [Add two FAILING unit-tests for UnMarshallLimitNamespace (https://github.com/Kuadrant/kuadrant-operator/issues/167)](https://github.com/Kuadrant/kuadrant-operator/commit/8b044acac8050acfdd2103665315c2d5ff18fcd7), the "when namespace has no gateway name then return an error" specifically.

2. Optimized the `UnMarshallLimitNamespace` function to reduce string allocations and use a more efficient search method for delimiter characters. This resulted in an average performance improvement of **52%** in average across 3 benchmarks with input sizes of 1000, 100, and 10.

3. Optimized the `HostnamesToStrings` function by pre-allocating the output slice and reducing the number of string allocations required. This resulted in an average performance improvement of **69%** in average across 3 benchmarks with input sizes of 1000, 100, and 10.

These optimizations will help to improve the performance of the common package when working with large volumes of Kubernetes objects.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -17,7 +17,6 @@ limitations under the License.
 package common
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"reflect"
@@ -158,7 +157,7 @@ func MergeMapStringString(existing *map[string]string, desired map[string]string
 func UnMarshallLimitNamespace(ns string) (client.ObjectKey, string, error) {
 	delimIndex := strings.IndexByte(ns, '#')
 	if delimIndex == -1 {
-		return client.ObjectKey{}, "", errors.New("failed to split on #")
+		return client.ObjectKey{}, "", fmt.Errorf("failed to split on #")
 	}
 
 	gwSplit := ns[:delimIndex]
@@ -180,7 +179,7 @@ func MarshallNamespace(gwKey client.ObjectKey, domain string) string {
 func UnMarshallObjectKey(keyStr string) (client.ObjectKey, error) {
 	namespaceEndIndex := strings.IndexByte(keyStr, NamespaceSeparator)
 	if namespaceEndIndex < 0 || len(keyStr)-namespaceEndIndex-1 < 1 {
-		return client.ObjectKey{}, errors.New(fmt.Sprintf("failed to split on %s %s", string(NamespaceSeparator), keyStr))
+		return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s %s", string(NamespaceSeparator), keyStr))
 	}
 
 	return client.ObjectKey{Namespace: keyStr[:namespaceEndIndex], Name: keyStr[namespaceEndIndex+1:]}, nil
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -155,7 +155,7 @@ func MergeMapStringString(existing *map[string]string, desired map[string]string
 
 // UnMarshallLimitNamespace parses limit namespace with format "gwNS/gwName#domain"
 func UnMarshallLimitNamespace(ns string) (client.ObjectKey, string, error) {
-	delimIndex := strings.IndexByte(ns, '#')
+	delimIndex := strings.IndexRune(ns, '#')
 	if delimIndex == -1 {
 		return client.ObjectKey{}, "", fmt.Errorf("failed to split on #")
 	}
@@ -177,7 +177,7 @@ func MarshallNamespace(gwKey client.ObjectKey, domain string) string {
 }
 
 func UnMarshallObjectKey(keyStr string) (client.ObjectKey, error) {
-	namespaceEndIndex := strings.IndexByte(keyStr, NamespaceSeparator)
+	namespaceEndIndex := strings.IndexRune(keyStr, NamespaceSeparator)
 	if namespaceEndIndex < 0 {
 		return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr))
 	}
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -178,8 +178,8 @@ func MarshallNamespace(gwKey client.ObjectKey, domain string) string {
 
 func UnMarshallObjectKey(keyStr string) (client.ObjectKey, error) {
 	namespaceEndIndex := strings.IndexByte(keyStr, NamespaceSeparator)
-	if namespaceEndIndex < 0 || len(keyStr)-namespaceEndIndex-1 < 1 {
-		return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s %s", string(NamespaceSeparator), keyStr))
+	if namespaceEndIndex < 0 {
+		return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr))
 	}
 
 	return client.ObjectKey{Namespace: keyStr[:namespaceEndIndex], Name: keyStr[namespaceEndIndex+1:]}, nil
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -188,9 +188,9 @@ func UnMarshallObjectKey(keyStr string) (client.ObjectKey, error) {
 
 // HostnamesToStrings converts []gatewayapi_v1alpha2.Hostname to []string
 func HostnamesToStrings(hostnames []gatewayapiv1alpha2.Hostname) []string {
-	hosts := []string{}
-	for idx := range hostnames {
-		hosts = append(hosts, string(hostnames[idx]))
+	hosts := make([]string, len(hostnames))
+	for i, h := range hostnames {
+		hosts[i] = string(h)
 	}
 	return hosts
 }
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -156,19 +156,20 @@ func MergeMapStringString(existing *map[string]string, desired map[string]string
 
 // UnMarshallLimitNamespace parses limit namespace with format "gwNS/gwName#domain"
 func UnMarshallLimitNamespace(ns string) (client.ObjectKey, string, error) {
-	split := strings.Split(ns, "#")
-	if len(split) != 2 {
+	delimIndex := strings.IndexByte(ns, '#')
+	if delimIndex == -1 {
 		return client.ObjectKey{}, "", errors.New("failed to split on #")
 	}
 
-	domain := split[1]
+	gwSplit := ns[:delimIndex]
+	domain := ns[delimIndex+1:]
 
-	gwKey, err := UnMarshallObjectKey(split[0])
+	objKey, err := UnMarshallObjectKey(gwSplit)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/common/common.go
```diff
@@ -172,19 +172,19 @@ func MarshallNamespace(gwKey client.ObjectKey, domain string) string {
 }
 
 func UnMarshallObjectKey(keyStr string) (client.ObjectKey, error) {
-	keySplit := strings.Split(keyStr, string(NamespaceSeparator))
-	if len(keySplit) < 2 {
-		return client.ObjectKey{}, fmt.Errorf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr)
+	namespaceEndIndex := strings.IndexByte(keyStr, NamespaceSeparator)
+	if namespaceEndIndex < 0 || len(keyStr)-namespaceEndIndex-1 < 1 {
```
eguzki, line 176: The function is returning an error if the `NamespaceSeparator` is not found or it is the last char. 

What if it is the first one? I think it is enough to know there is at least one and then read both sides of the first separator. Which, btw, any of them could be empty, yes.

art-tapin: I see! I added two test-cases for that and going to push them now to the PR (https://github.com/Kuadrant/kuadrant-operator/pull/186):
1. when valid namespace with empty object (ends with '/') then return ObjectKey with namespace only
2. when valid name with empty namespace (starts with '/') then return ObjectKey with name only

Regarading changes, I think, this simplification of the condition includes the cases you described:

```
if namespaceEndIndex < 0 {
    return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr))
}
```

This implementation passes the unit-tests, which are in the PR (https://github.com/Kuadrant/kuadrant-operator/pull/186) + two new test-cases mentioned

What do you think @eguzki?

art-tapin: I made the changes and pushed them here: https://github.com/Kuadrant/kuadrant-operator/pull/187/commits/159b206bf9820b3aa4d8d4a23a2f41d83621a7f5

Thanks again, @eguzki :angel:

pkg/common/common.go
```diff
@@ -151,19 +150,20 @@ func MergeMapStringString(existing *map[string]string, desired map[string]string
 
 // UnMarshallLimitNamespace parses limit namespace with format "gwNS/gwName#domain"
 func UnMarshallLimitNamespace(ns string) (client.ObjectKey, string, error) {
-	split := strings.Split(ns, "#")
-	if len(split) != 2 {
-		return client.ObjectKey{}, "", errors.New("failed to split on #")
+	delimIndex := strings.IndexByte(ns, '#')
```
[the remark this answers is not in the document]
eguzki: What about [IndexRune](https://pkg.go.dev/strings#IndexRune)?

pkg/common/common.go
```diff
@@ -172,19 +172,19 @@ func MarshallNamespace(gwKey client.ObjectKey, domain string) string {
...
-	keySplit := strings.Split(keyStr, string(NamespaceSeparator))
-	if len(keySplit) < 2 {
-		return client.ObjectKey{}, fmt.Errorf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr)
+	namespaceEndIndex := strings.IndexByte(keyStr, NamespaceSeparator)
+	if namespaceEndIndex < 0 {
+		return client.ObjectKey{}, fmt.Errorf(fmt.Sprintf("failed to split on %s: '%s'", string(NamespaceSeparator), keyStr))
 	}
 
-	return client.ObjectKey{Namespace: keySplit[0], Name: keySplit[1]}, nil
+	return client.ObjectKey{Namespace: keyStr[:namespaceEndIndex], Name: keyStr[namespaceEndIndex+1:]}, nil
```
eguzki, line 180: I would like to see tests with:

* keyStr -> `#` which makes `namespaceEndIndex` to be `0` and then what happens with `keyStr[namespaceEndIndex+1:]` ?

* keyStr -> `something#` which makes `namespaceEndIndex` to be the last char and then what happens with `keyStr[namespaceEndIndex+1:]` ?

eguzki:
The tests regarding `TestUnMarshallLimitNamespace` should be included in this PR to be part of the same change.

art-tapin:
> The tests regarding `TestUnMarshallLimitNamespace` should be included in this PR to be part of the same change.

@eguzki, could you please clarify this moment? I talked to DD and the idea was to split "Tests" and "Improvements" to different PR's. I think, despite the changes made in the codebase, we still planning to keep all new unit-tests that are created right now, because they are relevant to the functionality, not just particular implementation. If I am wrong, please, correct me!
 
So, why should we include the tests regarding `TestUnMarshallLimitNamespace` to this PR, instead of dedicated separate one? (see [[test] Unit-tests for common/common.go (part 3 of 3)](https://github.com/Kuadrant/kuadrant-operator/pull/186))

eguzki:
> So, why should we include the tests regard

If you do not change the code, then it is right to have a dedicated PR for testing to increase coverage. However, when you open a PR with changes in the code, I also expect some tests in the same PR making sure the **new** code works as expected.

pkg/common/common.go
art-tapin: Could you please check out my PR with the test-cases for UnMarshallObjectKey? 

I have the following test-case there, which includes the two cases you mentioned:
```
{
    name: "when '#' separator used instead of '/' then return an error",
        input: "default#object1",
        expectedOutput: client.ObjectKey{},
        expectedError:  fmt.Errorf("failed to split on %s: 'default#object1'", string(NamespaceSeparator)),
},
```

When there is no `NamespaseSeparator', which is defined by constant, then it immediately returns an error in all cases. Here, a non-default separator '#' is used, so it won't get to the last return statement with slicing.

But I'll porbably rename it, by adding word 'default separator', because separator may possibly be changed? Anyway, it will make tests less dependant on implementation, I think.

art-tapin:
> If you do not change the code, then it is right to have a dedicated PR for testing to increase coverage. However, when you open a PR with changes in the code, I also expect some tests in the same PR making sure the **new** code works as expected.

Thanks @eguzki :angel: 
I will add new test-cases regarding new code here

pkg/common/common.go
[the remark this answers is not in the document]
art-tapin: Fantastic suggestion @eguzki! Thank you for coming up with this simple fix. It really solves the issue with with non-ascii chars! :medal_sports:

pkg/common/common_test.go
```diff
@@ -411,3 +413,249 @@ func TestMergeMapStringString(t *testing.T) {
...
+			expectedDomain: "",
+			expectedError:  true,
+		},
+		{
+			name:           "when namespace contains more than one '#' then return an error",
+			namespace:      "exampleNS/exampleGW#domain.com#extra",
+			expectedKey:    client.ObjectKey{},
+			expectedDomain: "",
+			expectedError:  true,
+		},
```
art-tapin, line 445: @eguzki, @didierofrivia Are we considering this case? 
For this, I have a quick fix, adding `strings.ContainsRune(ns[delimIndex+1:], '#')` to the first if-clause:

```
func UnMarshallLimitNamespaceMultipleHashtags(ns string) (client.ObjectKey, string, error) {
	delimIndex := strings.IndexRune(ns, '#')
	if delimIndex == -1 || strings.ContainsRune(ns[delimIndex+1:], '#') {
		return client.ObjectKey{}, "", fmt.Errorf("failed to split on #")
	}
        // rest of the code
}
```

pkg/common/common_test.go
```diff
@@ -411,3 +413,249 @@ func TestMergeMapStringString(t *testing.T) {
...
+		{
+			name:           "when namespace is invalid (no '#domain') then return an error",
+			namespace:      "exampleNS/exampleGW",
+			expectedKey:    client.ObjectKey{},
+			expectedDomain: "",
+			expectedError:  true,
+		},
+		{
+			name:           "when namespace contains more than one '#' then return an error",
+			namespace:      "exampleNS/exampleGW#domain.com#extra",
```
art-tapin, line 441: @eguzki, @didierofrivia Are we considering this case? 
For this, I have a quick fix, adding `strings.ContainsRune(ns[delimIndex+1:], '#')` to the first if-clause:

```
func UnMarshallLimitNamespaceMultipleHashtags(ns string) (client.ObjectKey, string, error) {
	delimIndex := strings.IndexRune(ns, '#')
	if delimIndex == -1 || strings.ContainsRune(ns[delimIndex+1:], '#') {
		return client.ObjectKey{}, "", fmt.Errorf("failed to split on #")
	}
        // rest of the code
}
```

pkg/common/common_test.go
```diff
@@ -411,3 +413,249 @@ func TestMergeMapStringString(t *testing.T) {
...
+			expectedKey:    client.ObjectKey{},
+			expectedDomain: "",
+			expectedError:  true,
+		},
+		{
+			name:           "when namespace has no domain name then return an error",
+			namespace:      "exampleNS/exampleGW#",
+			expectedKey:    client.ObjectKey{},
+			expectedDomain: "",
+			expectedError:  true,
```
art-tapin, line 458: @eguzki @didierofrivia This is the second test-case that fails. Here I can suggest this chagne to `UnMarshallLimitNamespaceTwoFallingTestCases`, to add a check right after the domain is parsed (see the last lines):
```
func UnMarshallLimitNamespaceTwoFallingTestCases(ns string) (client.ObjectKey, string, error) {
	delimIndex := strings.IndexRune(ns, '#')
	if delimIndex == -1 || strings.ContainsRune(ns[delimIndex+1:], '#') {
		return client.ObjectKey{}, "", fmt.Errorf("failed to split on #")
	}

	gwSplit := ns[:delimIndex]
	domain := ns[delimIndex+1:]

	if domain == "" {
		return client.ObjectKey{}, "", fmt.Errorf("empty domain")
	}
        // rest of the code
```
Here, we'll need to put another error message, I guess? If so, I'll need to refactor tests not only to check the expectancy of the error, but the specific type of it. 

What do you think?

didierofrivia: That could solve it, beware of the if case that needs to be negated when checking the existence of the rune

didierofrivia:
Rebase with origin/main

pkg/common/common_test.go
art-tapin: After negotiating, this test-case was decided to be deleted, because it is on Kubernetes to validate it.
Thanks, @eguzki and @didierofrivia :angel:

pkg/common/common_test.go
art-tapin: According to our implementation, this is a valid case, where domain name should be empty and the ObjectKey is a valid struct with correct values.
Thanks, @eguzki and @didierofrivia :angel:

didierofrivia (approved):
✌🏼

PR #188, opened 2023-05-10
fix e2e tests: do not use istio-ingressgateway in tests

eguzki:
### What 

Tests should be reproducible and independent from external factors such as the environment or running order. This PR removes the usage of the shared `istio-ingressgateway` Gateway in the namespace `istio-system`. The integration tests only work with resources in the temporary namespace created for each test. 

Additionally, new RLP integration test is added where the policy targets the custom gateway directly

### DIFF
--- a/pkg/common/gatewayapi_utils.go
+++ b/pkg/common/gatewayapi_utils.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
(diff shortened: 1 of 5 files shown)


pkg/common/gatewayapi_utils.go
```diff
@@ -447,3 +449,37 @@ func GetGatewayWorkloadSelector(ctx context.Context, cli client.Client, gateway
...
+		return false
+	}
+
+	// Check HTTProute parents (gateways) in the status object
+	// if any of the current parent gateways reports not "Admitted", return false
+	for _, parentRef := range httpRoute.Spec.CommonRouteSpec.ParentRefs {
+		routeParentStatus := func(pRef gatewayapiv1alpha2.ParentReference) *gatewayapiv1alpha2.RouteParentStatus {
+			for idx := range httpRoute.Status.RouteStatus.Parents {
+				if reflect.DeepEqual(pRef, httpRoute.Status.RouteStatus.Parents[idx].ParentRef) {
+					return &httpRoute.Status.RouteStatus.Parents[idx]
```
didierofrivia, line 468: What about returning false if not deepEqual here? and then in line 472 return true

eguzki: The method returns the `RouteParentStatus` reference because later the caller needs to access the conditions field to see if it has been accepted.

didierofrivia (approved):

PR #190, opened 2023-05-19
[test] Unit-tests for common/k8s_utils.go (part 1 of 3)

art-tapin:
This pull request includes code refactorings and the addition of unit tests for multiple functions in common/k8s_utils.go. Comments and unit tests were added to functions:

- IsObjectTaggedToDelete
- TagObjectToDelete
- ReadAnnotationsFromObject
- ObjectInfo

Note: This is the part 1 of 3 of testing the k8s_utils.go. This pull request only partially closes issue https://github.com/Kuadrant/kuadrant-operator/issues/167, and there may be further updates needed in the future.

Coverage: 71.1%

### DIFF
--- a/pkg/common/k8s_utils.go
+++ b/pkg/common/k8s_utils.go
@@ -42,6 +42,8 @@ func ObjectInfo(obj client.Object) string {
 	return fmt.Sprintf("%s/%s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
 }
 
+// ReadAnnotationsFromObject reads the annotations from a Kubernetes object
+// and returns them as a map. If the object has no annotations, it returns an empty map.
 func ReadAnnotationsFromObject(obj client.Object) map[string]string {
 	annotations := obj.GetAnnotations()
 	if annotations == nil {
--- a/pkg/common/k8s_utils.go
+++ b/pkg/common/k8s_utils.go
@@ -52,6 +52,8 @@ func ReadAnnotationsFromObject(obj client.Object) map[string]string {
 	return annotations
 }
 
+// TagObjectToDelete adds a special 'delete' tag to the object's annotations.
+// If the object's annotations are nil, it first initializes the Annotations field with an empty map.
 func TagObjectToDelete(obj client.Object) {
 	// Add custom annotation
 	annotations := obj.GetAnnotations()
--- a/pkg/common/k8s_utils.go
+++ b/pkg/common/k8s_utils.go
@@ -35,6 +35,9 @@ const (
 	ReadyStatusConditionType = "Ready"
 )
 
+// ObjectInfo generates a string representation of the provided Kubernetes object, including its kind and name.
+// The object must implement the client.Object interface, which provides access to the object's metadata.
+// The generated string follows the format: "kind/name".
 func ObjectInfo(obj client.Object) string {
 	return fmt.Sprintf("%s/%s", obj.GetObjectKind().GroupVersionKind().Kind, obj.GetName())
 }
--- a/pkg/common/k8s_utils.go
+++ b/pkg/common/k8s_utils.go
@@ -52,7 +52,7 @@ func ReadAnnotationsFromObject(obj client.Object) map[string]string {
 	return annotations
 }
 
-// TagObjectToDelete adds a special 'delete' tag to the object's annotations.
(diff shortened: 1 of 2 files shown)


pkg/common/k8s_utils.go
```diff
@@ -35,10 +35,15 @@ const (
 	ReadyStatusConditionType = "Ready"
 )
 
+// ObjectInfo generates a string representation of the provided Kubernetes object, including its kind and name.
+// The object must implement the client.Object interface, which provides access to the object's metadata.
```
eguzki, line 39: nitpick: this is a personal taste. Comments are usually not in excess. The more comments, the better. But I tend to add comments only for "not obvious" things.

pkg/common/k8s_utils_test.go
```diff
@@ -137,3 +140,241 @@ func TestGetServiceWorkloadSelector(t *testing.T) {
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			TagObjectToDelete(tc.input)
+
+			annotations := tc.input.GetAnnotations()
+			if annotations == nil {
+				t.Fatal("Expected annotations to be not nil, but got nil")
+			}
+
+			for key, expectedValue := range tc.expectedTags {
```
eguzki, line 309: `reflect.DeepEqual` is not good enough?

eguzki:
looking good. Left few comments

pkg/common/k8s_utils_test.go
```diff
@@ -137,3 +140,241 @@ func TestGetServiceWorkloadSelector(t *testing.T) {
...
 }
+
+func TestObjectInfo(t *testing.T) {
+	testCases := []struct {
+		name     string
+		input    client.Object
+		expected string
+	}{
+		{
+			name: "when given Kuadrant Limitador object then return formatted string",
```
guicassolato, line 151:
```suggestion
			name: "when given a Kubernetes object then return formatted string",
```

pkg/common/k8s_utils_test.go
```diff
@@ -137,3 +140,241 @@ func TestGetServiceWorkloadSelector(t *testing.T) {
...
+					Kind: "Limitador",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-limitador",
+				},
+			},
+			expected: "Limitador/test-limitador",
+		},
+		{
+			name: "when given k8s Pod object then return formatted string",
```
guicassolato, line 163: How is this test different from previous? Is the idea here to highlight that the function works with kind of Kubernetes resource? If so, I'd say not needed because no finite set of examples will suffice.

pkg/common/k8s_utils_test.go
```diff
@@ -137,3 +140,241 @@ func TestGetServiceWorkloadSelector(t *testing.T) {
...
+					Kind: "Pod",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-pod",
+				},
+			},
+			expected: "Pod/test-pod",
+		},
+		{
+			name: "when given k8s Service object with empty Kind then return formatted string",
```
guicassolato, line 175:
```suggestion
			name: "when given a Kubernetes object with empty Kind then return formatted string",
```

pkg/common/k8s_utils_test.go
```diff
@@ -137,3 +140,241 @@ func TestGetServiceWorkloadSelector(t *testing.T) {
...
+			input: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-service",
+				},
+			},
+			expected: "/test-service",
+		},
+		{
+			name: "when given k8s Namespace object with empty Name then return formatted string",
```
guicassolato, line 184:
```suggestion
			name: "when given a Kubernetes object with empty Name then return formatted string",
```

pkg/common/k8s_utils_test.go
art-tapin: Good point! The intention of this test was indeed to emphasize that the function works with Kuadrant resources (Limitador as example). However, you're right @guicassolato that providing a finite set of examples may not be sufficient.  I am going to get rid of the redundant one!

Thanks for pointing that out @guicassolato!

didierofrivia (approved):
🍶

Issue #168, opened 2023-04-17
Update to remove the need to install istio via the istio operator 


R-Lawton:
Currently to be able to avail of the rate limit and auth policy in Kuadrant you need to have an istio operator cr. To be able to have an istio operator cr you have to install istio operator which is now discouraged by the istio team.  Istioctl and Helm installation methods are recommended [instead](https://istio.io/latest/docs/setup/install/operator/).

pehala:
> Personally, I don’t think that’s the direction we should be going, and neither should Istio IMHO.

While I agree, Istio is already going that way as IstioOperator will not have new features and is discouraged (https://istio.io/latest/docs/setup/install/operator/)

maleck13:
When you use the istioctl or helm approach, is there no configuration resource created that kuadrant could hook into if it was told where that resource was? I guess my question is where is the config stored if not when not in the operator resource?

guicassolato:
@maleck13 

There is a ConfigMap that by default looks like this:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: istio-system
  name: istio
  labels:
    install.operator.istio.io/owning-resource: istiocontrolplane
    install.operator.istio.io/owning-resource-namespace: istio-system
    istio.io/rev: default
    operator.istio.io/component: Pilot
    operator.istio.io/managed: Reconcile
    operator.istio.io/version: 1.13.2
    release: istio
data:
  mesh: |-
    defaultConfig:
      discoveryAddress: istiod.istio-system.svc:15012
      proxyMetadata: {}
      tracing:
        zipkin:
          address: zipkin.istio-system:9411
    enablePrometheusMerge: true
    rootNamespace: istio-system
    trustDomain: cluster.local
  meshNetworks: 'networks: {}'
```

The ConfgiMap can be patched in the following way to inject the extension provider info:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: istio-system
  name: istio
  labels: […]
data:
  mesh: |-
    defaultConfig:
      discoveryAddress: istiod.istio-system.svc:15012
      proxyMetadata: {}
      tracing:
        zipkin:
          address: zipkin.istio-system:9411
    enablePrometheusMerge: true
    extensionProviders:
    - envoyExtAuthzGrpc:
        port: 50051
        service: authorino-authorino-authorization.kuadrant-system.svc.cluster.local
      name: kuadrant-authorization
    rootNamespace: istio-system
[...]

PR #192, opened 2023-05-23
Istio external authorizer available not only for IstioOperator

didierofrivia:
This PR aims to fix the registration of our Authorino as an external authorizer using Istio when it's not installed via its Operator.

Closes https://github.com/Kuadrant/kuadrant-operator/issues/168

**Notes**
* Reconciles also the Istio ConfigMap with the external authorizers
* If the IstioOperator CR returns an error, it continues reconciling CM

**Verification Steps**

❶ Create a cluster with Gateway API and Istio:

```sh
make kind-create-cluster gateway-api-install istioctl
```
```sh
./bin/istioctl install -f config/dependencies/istio/istio-operator.yaml
```
```sh
make deploy-gateway
```

❷ Deploy Kuadrant dependencies and a local build, then apply its CR:

Create the namespace for kuadrant control plane
```sh
kubectl create namespace kuadrant-system 

```
Then install dependencies and a local kuadrant image

```sh
make deploy-dependencies local-deploy

```

Then apply Kuadrant CR
```sh
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

❸ Apply toystore example service and create an HTTPRoute
```sh
```

```sh
kubectl apply -f - <<EOF
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
EOF
```

❹ Create API Key as Secrets for user Bob
```sh
kubectl apply -f -<<EOF
---
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: bob-key
  labels:
    app: toystore
stringData:
  api_key: IAMBOB
type: Opaque
EOF
```

❺ Apply Authorino AuthPolicy
```sh
kubectl apply -f - <<EOF
---
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
  - paths: ["/toy*"]
  authScheme:
    identity:
    - name: friends
      apiKey:
        allNamespaces: true
        selector:
          matchLabels:
            app: toystore
      credentials:
        in: authorization_header
        keySelector: APIKEY
    response:
    - json:
        properties:
          - name: userID
            value: null
            valueFrom:
      name: rate-limit-apikey
      wrapper: envoyDynamicMetadata
      wrapperKey: ext_auth_data
EOF
```

❻ Send requests:

```sh
curl -v -H 'Host: api.toystore.com' http://localhost:9080/toy # 401
```
```sh
curl -v -H 'Authorization: APIKEY IAMBOB' -H 'Host: api.toystore.com' http://localhost:9080/toy # 200
```

TODO:
* [x] Update the unregister function with the new functionality
* [x] Refactor the istio functions into pkg/common
* [x] Tests
* [x] Verification steps


Makefile
```diff
@@ -266,9 +266,8 @@ test-unit: clean-cov generate fmt vet ## Run Unit tests.
 namespace: ## Creates a namespace where to deploy Kuadrant Operator
 	kubectl create namespace $(KUADRANT_NAMESPACE)
 
-.PHONY: local-setup
-local-setup: $(KIND) ## Deploy locally kuadrant operator from the current code
-	$(MAKE) local-env-setup
+.PHONY: local-deploy
+local-deploy: ## Deploy Kuadrant Operator in the cluster pointed by KUBECONFIG
```
didierofrivia, line 270: This is not part of the fix, but an easy way to separate the targets and use them for testing other ways of setup

pkg/common/mesh_config.go
```diff
@@ -0,0 +1,159 @@
+package common
```
guicassolato, line 1: Why moving these functions to pkg/common?

didierofrivia: I reckoned they could be encapsulated and unit tested since are not dependent on client. pkg/common might not be the best destination, thought of _pkg/istio_ but the mesh config also belongs to OSSM... what do you have in mind?

guicassolato: I think those functions are better kept flagged as private and the least opened to be reused at other places as possible. We don't want other things messing with the mesh config when this is responsibility of the Kuadrant controller.

Perhaps others who might be thinking on different (future) ways to manage instance of Kuadrant (e.g. CLI tool, add-on) may disagree. But at this stage I see the change as more than how we organise the code.

As for the unit tests, you can still test those functions in the controllers package even if they are not exported.

didierofrivia: I see your point and it makes sense. However, these functions are stateless and without the proper context of the Kuadrant controller reconciler, little they can mess with the actual mesh config. Also, even tho there's no planning on having a CLI or add-on,  we also have some functions that are strictly a responsibility of the kuadrant_controller, like manipulating the gateways. That being said, I'm fine with moving them back to the controller (or having them in any other package, maybe _mesh_config_?), since I might minimizing the scope of this change :)

pkg/common/mesh_config.go
```diff
@@ -0,0 +1,159 @@
...
+	return
+}
+
+func UpdateMeshConfig(configObject client.Object, updateFunc func(meshConfig *istiomeshv1alpha1.MeshConfig) bool) (bool, error) {
+	meshConfig, err := getMeshConfig(configObject)
+	if err != nil {
+		return false, err
+	}
+	if updateFunc(meshConfig) {
+		switch config := configObject.(type) {
```
guicassolato, line 92: What do you think about having wrapper structs for the IstioOperator, for the ConfigMap and for ServiceMeshControlPlane resources that implement an interface to abstract these type-specific functions (`getMeshConfig`, `UpdateMesgConfig`) that otherwise require the type checking in a `switch` clause?

controllers/kuadrant_controller.go
```diff
@@ -181,44 +181,40 @@ func (r *KuadrantReconciler) unregisterExternalAuthorizer(ctx context.Context) e
...
+	iop := &iopv1alpha1.IstioOperator{}
 	iopKey := client.ObjectKey{Name: controlPlaneProviderName(), Namespace: controlPlaneProviderNamespace()}
 	if err := r.Client().Get(ctx, iopKey, iop); err != nil {
 		logger.V(1).Info("failed to get istiooperator object", "key", iopKey, "err", err)
-		return err
+		if apimeta.IsNoMatchError(err) {
+			// return err only if there's no istiooperator CRD
+			return err
+		}
+		// otherwise, we assume that the control plane is istio but no installed by its operator
```
guicassolato, line 194: IDK if I understand returning only if IsNoMatchError 🤔  This will cause the reconciliation to try to proceed before other API errors, such as a 429 (occasionally raised by the [API Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) filter). Do we really want this?

controllers/kuadrant_controller.go
```diff
@@ -181,44 +181,40 @@ func (r *KuadrantReconciler) unregisterExternalAuthorizer(ctx context.Context) e
...
-			meshConfigStruct, err := meshConfigToStruct(meshConfig)
-			if err != nil {
+	for _, config := range configsToUpdate {
+		if needsUpdate, err := config.UpdateConfig(
+			istio.RemoveKuadrantAuthorizerFromConfig,
+		); err != nil {
+			return err
+		} else if needsUpdate {
+			logger.Info("remove external authorizer from meshconfig")
+			if err := r.Client().Update(ctx, config.GetConfigObject()); err != nil {
```
guicassolato, line 213: I'm guessing your intention was of keeping like before, to not change too much. So just pointing out we have this alternative from the BaseReconciler:

```suggestion
			if err := r.UpdateResource(ctx, config.GetConfigObject()); err != nil {
```

controllers/kuadrant_controller.go
```diff
@@ -297,37 +293,40 @@ func (r *KuadrantReconciler) registerExternalAuthorizer(ctx context.Context, kOb
...
-		return err
-	}
-
-	if iop.Spec == nil {
-		iop.Spec = &istioapiv1alpha1.IstioOperatorSpec{}
+		if apimeta.IsNoMatchError(err) {
+			// if it's a no match error (CRD not installed), return the error
+			// otherwise continue with the fetching of the configmap
+			return err
+		}
```
guicassolato, line 306: https://github.com/Kuadrant/kuadrant-operator/pull/192/files#r1235056662

controllers/kuadrant_controller.go
```diff
@@ -297,37 +293,40 @@ func (r *KuadrantReconciler) registerExternalAuthorizer(ctx context.Context, kOb
...
+	} else {
+		// if there is no error, add the iop to the list of configs to update
+		configsToUpdate = append(configsToUpdate, istio.NewOperatorWrapper(iop))
 	}
 
-	meshConfig, err := meshConfigFromStruct(iop.Spec.MeshConfig)
-	if err != nil {
+	istioConfigMap := &corev1.ConfigMap{}
+	if err := r.Client().Get(ctx, client.ObjectKey{Name: "istio", Namespace: controlPlaneProviderNamespace()}, istioConfigMap); err != nil {
```
guicassolato, line 313: https://github.com/Kuadrant/kuadrant-operator/pull/192/files#r1221733307

pkg/istio/mesh_config.go
```diff
@@ -0,0 +1,223 @@
...
+
+const (
+	ExtAuthorizerName = "kuadrant-authorization"
+)
+
+// ConfigWrapper wraps the IstioOperator CRD or ConfigMap
+type ConfigWrapper interface {
+	GetConfigObject() client.Object
+	GetConfig() (*istiomeshv1alpha1.MeshConfig, error)
+	UpdateConfig(updateFunc func(meshConfig *istiomeshv1alpha1.MeshConfig) bool) (bool, error)
```
guicassolato, line 23: This function seems to to have been thought to receive either `RemoveKuadrantAuthorizerFromConfig` or the return value of `AddKuadrantAuthorizerToConfig(string)`. I was thinking... These are the kind of abstraction that this interface can provide.

E.g.

The following are high-level abstractions that don't have to belong to the `istio` package IMO:

```go
type KuadrantAuthorizer interface {
  GetNamespace() string
}

type ConfigWrapper interface {
  HasKuadrantAuthorizer(KuadrantAuthorizer) bool
  RegisterKuadrantAuthorizer(KuadrantAuthorizer) error
  UnregisterKuadrantAuthorizer(KuadrantAuthorizer) error
  GetConfigObject() client.Object
}
```

Then the `istio` package implements the interface – for the `IstioOperator`, `IstioConfigMap` and `OpenShiftServiceMeshControlPlane` specific types. The following become all details that the KuadrantReconciler doesn't need to know about:
- `MeshConfigToStruct(config *istiomeshv1alpha1.MeshConfig) (*structpb.Struct, error)`
- `MeshConfigFromString(config string) (*istiomeshv1alpha1.MeshConfig, error)`
- `ExtensionProvidersFromMeshConfig(config *istiomeshv1alpha1.MeshConfig) (extensionProviders []*istiomeshv1alpha1.MeshConfig_ExtensionProvider)`
- `CreateKuadrantAuthorizer(namespace string) *istiomeshv1alpha1.MeshConfig_ExtensionProvider`

Maybe even:
- `HasKuadrantAuthorizer(extensionProviders []*istiomeshv1alpha1.MeshConfig_ExtensionProvider) bool`

WDYT?

controllers/kuadrant_controller.go
```diff
@@ -181,44 +180,48 @@ func (r *KuadrantReconciler) unregisterExternalAuthorizer(ctx context.Context, k
...
+		}
+		// otherwise, we assume that the control plane is istio but no installed by its operator
+	} else {
+		configsToUpdate = append(configsToUpdate, iop)
 	}
 
-	meshConfig, err := meshConfigFromStruct(iop.Spec.MeshConfig)
-	if err != nil {
+	istioConfigMap := &corev1.ConfigMap{}
+	if err := r.Client().Get(ctx, client.ObjectKey{Name: "istio", Namespace: controlPlaneProviderNamespace()}, istioConfigMap); err != nil {
```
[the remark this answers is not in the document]
didierofrivia: That particular ENV, among other from the `pilot-agent` are marked as experimental... Was thinking that maybe is not a good idea to rely on those, and use them if are requested and/or istio decides to mark them as stable.

controllers/kuadrant_controller.go
```diff
@@ -246,23 +242,23 @@ func (r *KuadrantReconciler) unregisterExternalAuthorizerOSSM(ctx context.Contex
 		if err != nil {
 			return err
 		}
-		meshConfig, _ = meshConfigFromStruct(meshConfigStruct)
+		meshConfig, _ = istio.MeshConfigFromStruct(meshConfigStruct)
```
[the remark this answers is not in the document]
didierofrivia: Yup, still have some commits pending hehe

pkg/istio/mesh_config.go
didierofrivia: I agree! from the beginning was a bit troubled by the scope of each. Thanks!

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
guicassolato: Maybe define ours then? Hard-coding the name won't give much options to the users.

controllers/kuadrant_controller.go
```diff
@@ -163,13 +161,13 @@ func (r *KuadrantReconciler) Reconcile(eventCtx context.Context, req ctrl.Reques
 	return ctrl.Result{}, nil
 }
 
-func (r *KuadrantReconciler) unregisterExternalAuthorizer(ctx context.Context) error {
+func (r *KuadrantReconciler) unregisterExternalAuthorizer(ctx context.Context, kObj *kuadrantv1beta1.Kuadrant) error {
 	logger, _ := logr.FromContext(ctx)
 
-	err := r.unregisterExternalAuthorizerIstio(ctx)
+	isIstioInstalled, err := r.unregisterExternalAuthorizerIstio(ctx, kObj)
```
guicassolato, line 167: Maybe in a next iteration `unregisterExternalAuthorizerIstio` and `unregisterExternalAuthorizerOSSM` can become one, and similarly `registerExternalAuthorizerIstio` and `registerExternalAuthorizerOSSM`. Fundamentally, and given the new abstractions now introduced with this PR, these pairs of functions are not very different between them – if any different at all.

`getIstioConfigObjects` could become simply `getConfigObjects` (or perhaps `getGatewayProviderConfigObjects` to be more literal with the name), and return all gateway provider config objects that need to be patched.

Istio or OSSM, it doesn't really matter. In the end, it all reduces to iterating over all configs, passing them into the calls to `common.UnregisterKuadrantAuthorizer` and `common.RegisterKuadrantAuthorizer`. Thanks to this PR, the details of how to patch each kind of config is decoupled from the caller, but abstracted by each config wrapper itself.

controllers/kuadrant_controller.go
```diff
@@ -343,46 +295,56 @@ func (r *KuadrantReconciler) registerExternalAuthorizerOSSM(ctx context.Context,
...
+	var configsToUpdate []common.ConfigWrapper
+
+	iop := &iopv1alpha1.IstioOperator{}
+	iopKey := client.ObjectKey{Name: controlPlaneProviderName(), Namespace: controlPlaneProviderNamespace()}
+	if err := r.GetResource(ctx, iopKey, iop); err != nil {
+		logger.V(1).Info("failed to get istiooperator object", "key", iopKey, "err", err)
+		if apimeta.IsNoMatchError(err) {
+			// return nil and nil if there's no istiooperator CRD, means istio is not installed
+			return nil, nil
+		} else if err.Error() != fmt.Sprintf("IstioOperator.install.istio.io \"%s\" not found", controlPlaneProviderName()) {
```
guicassolato, line 333: Can we use [`apierrors.IsNotFound`](https://pkg.go.dev/k8s.io/apimachinery@v0.26.1/pkg/api/errors#IsNotFound) here?

```suggestion
		} else if !apierrors.IsNotFound(err) {
```

guicassolato (approved):
Nice job @didierofrivia!

I've left a couple additional comments, which, if you agree, can be addressed in this PR still or in another iteration.

Verification steps succeeded. Additionally, after step ❷:

```
❯ kubectl get istiooperator/installed-state-istiocontrolplane -n istio-system -o yaml | grep authorino
❯ kubectl get configmap/istio -n istio-system -o yaml | grep authorino
        service: authorino-authorino-authorization.kuadrant-system.svc.cluster.local
```

controllers/kuadrant_controller.go
didierofrivia: Makes total sense now, I'll keep in mind this for the next refactor, that most possibly will be the plugin way of using GWs

guicassolato (approved):

PR #194, opened 2023-05-25
Building the catalog with the replaces directive

didierofrivia:
The `replaces` version needs to be included in the catalog, in the previous change introduced by https://github.com/Kuadrant/kuadrant-operator/commit/0d9b7f64bc3900209b95454acb4809d6e17b0f6b it was only addressing the bundle


pehala (approved):

eguzki (approved):

utils/generate-catalog.sh
```diff
@@ -12,9 +12,10 @@ CHANNEL_NAME=preview
 OPM="${1?:Error \$OPM not set. Bye}"
 YQ="${2?:Error \$YQ not set. Bye}"
 BUNDLE_IMG="${3?:Error \$BUNDLE_IMG not set. Bye}"
-LIMITADOR_OPERATOR_BUNDLE_IMG="${4?:Error \$LIMITADOR_OPERATOR_BUNDLE_IMG not set. Bye}"
-AUTHORINO_OPERATOR_BUNDLE_IMG="${5?:Error \$AUTHORINO_OPERATOR_BUNDLE_IMG not set. Bye}"
-CATALOG_FILE="${6?:Error \$CATALOG_FILE not set. Bye}"
+REPLACES_VERSION="${4?:Error \$REPLACES_VERSION not set. Bye}"
```
guicassolato, line 15: Almost worth using a `getopt` loop here to retrieve named args. Future improvement idea.

PR #204, opened 2023-06-22
[rlp-v2] WasmPlugin reconciliation 

### LINKED ISSUE #200: [rlp-v2] Reconciliation of route selectors
The rules for reconciling RLP v2's `routeSelectors` (see [RFC 0001](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md)) are:
- A route selector selects 1+ HTTPRouteRules
- For RLPs that target a HTTPRoute, all HTTPRouteRules in the HTTPRoute must be scanned for possible selection
- For RLPs that target a Gateway, all HTTPRouteRules in all HTTPRoutes pointing the Gateway as parent must be scanned for possible selection
- A HTTPRouteRule is selected when it contains at least one HTTPRouteMatch whose attributes include totally the attributes of the route selector
- For a set of attributes of a HTTPRouteMatch to include the attributes of a route selector, all attributes of the route selector must be explicitly and identically declared as the attributes of the HTTPRouteMatch
- While selecting a HTTPRouteRule, the HTTPRouteMatch may contain attributes that are not explicitly declared amongst the attributes of the route selectors – i.e., `attrs(route selector) ⊆ attrs(HTTPRouteMatch)`, though not necessarily `attrs(route selector) = attrs(HTTPRouteMatch)`
- For RLPs that target a HTTPRoute, the set of attributes of any selectable HTTPRouteMatch implicitly includes the `hostnames` specified in the targeted HTTPRoute and all the `hostname`s of all Gateways pointed as parents of the HTTPRoute
- For RLPs that target a Gateway, the set of attributes of any selectable HTTPRouteMatch implicitly includes the `hostname` of the targeted Gateway and all `hostnames` of all HTTPRoutes pointing the Gateway as parent
- Hostname attributes specified in a route selector must as well be explicitly and identically declared in any of the instances above (HTTPRoute `hostnames` or Gateway `hostname`) that applies
- A route selector that finds no HTTPRouteRule with at least one HTTPRouteMatch including its attributes is considered invalid

---

Related to: https://github.com/Kuadrant/architecture/issues/13, #156
Follows-up on: #199.

guicassolato:
Reconciliation of the Istio WasmPlugin CR with wasm rules generated for each limit definition from the affecting RLPs – triggers based on `routeSelectors` and `when` conditions, and RL data (descriptors) to activate the limits in Limitador and additional counter qualifiers.

Implements [Direct Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/#direct-policy-attachment) of RLPs targeting a HTTPRoute or a Gateway (WasmPlugin part).

### Highlights

- [x] Reconciliation of HTTPRoute RLPs → Wasm rule conditions generated out of the limit's `when` conditions and HTTPRouteRules selected by the `routeSelectors`, and data that activates the limit in Limitador plus additional counter qualifiers
- [x] Fix pattern expression for hostnames with the right operator (`endswith`)
- [x] Compute the domain intersection between the hostnames specified in the HTTPRoute and the hostnames specified in the Gateway, so it doesn’t generate wasm rules for the wrong gateway/hostname
- [x] Stop repeating RLPs for each hostname within the wasm config (see https://github.com/Kuadrant/wasm-shim/issues/35#issuecomment-1607140659)
- [x] Reconciliation of Gateway RLPs → Fetch all HTTPRoutes accepted by the Gateway that do not have a RLP of its own, and then fake a HTTPRoute with all rules from those HTTPRoutes, to to generate sets of wasm rules (conditions and data) for each limit of the gateway RLP
- [x] Prevent use of `routeSelectors` in RLPs that target a Gateway
- [x] Stop merging gateway RLP rules into the route RLP configs – otherwise we would be implementing (partially) [Inherited Policy Attachment](https://gateway-api.sigs.k8s.io/references/policy-attachment/#inherited-policy-attachment-its-all-about-the-defaults-and-overrides) instead.
- [x] Do not generate conditions for gateway policy wasm rules from HTTPRoutes that have a RLP of its own
- [x] Avoid building a wasm plugin config if there are no rules to apply

### Verification steps

Topology:

```
          *                       *.internal
          │                           │
          ▼                           ▼
┌──────────────────────┐    ┌───────────────────┐
│     (Gateway)        │    │     (Gateway)     │
│ istio-ingressgateway │    │ internal-gateway  │
└─────────┬────────────┘    └─────────┬─────────┘
          │                           │
          └─────────────┬─────────────┘
                        │
                  ┌─────┴───────┐
                  │ (HTTPRoute) │
                  │  toystore   │
                  └─────────────┘
                         ▲
                         │
             ┌───────────┴───────────┐
             │  (RateLimitPolicy)    │
             │ toystore-per-endpoint │
             └───────────────────────┘
```

① Setup:

```sh
make local-setup

kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF

```

② Create a second Gateway:

```sh
kind: Gateway
metadata:
  name: internal-gateway
  annotations:
    kuadrant.io/namespace: kuadrant-system
    networking.istio.io/service-type: ClusterIP
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    port: 80
    protocol: HTTP
    hostname: '*.internal'
    allowedRoutes:
      namespaces:
        from: All
EOF
```

③ Create a HTTPRoute:

(Notice the route specifies 2 gateway parents and 3 host names, where all 3 host names bind the route to the `istio-ingressgateway` gateway and one of them to the `internal-gateway` gateway.)

```sh
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  - name: internal-gateway
    namespace: istio-system
  hostnames:
  - "*.toystore.acme.com"
  - "*.toystore.io"
  - "*.toystore.internal"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: "/toys"
      method: GET
    - path:
        type: PathPrefix
        value: "/toys"
      method: POST
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/assets/"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

④ Create the RateLimitPolicy:

(Notice one of the limits applies to only one host name.)

```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: toystore-per-endpoint
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    toys:
      rates:
      - limit: 50
        duration: 1
        unit: minute
      counters:
      - auth.identity.username
      routeSelectors:
      - matches: # matches the 1st HTTPRouteRule (i.e. GET or POST to /toys*)
        - path:
            type: PathPrefix
            value: "/toys"
        hostnames:
        - "*.toystore.acme.com"
      when:
      - selector: auth.identity.group
        operator: neq
        value: admin
    assets:
      rates:
      - limit: 5
        duration: 1
        unit: minute
      - limit: 100
        duration: 12
        unit: hour
      routeSelectors:
      - matches: # matches the 2nd HTTPRouteRule (i.e. /assets/*)
        - path:
            type: PathPrefix
            value: "/assets/"
EOF
```

⑤ Verify the Limitador CR was reconciled correctly:

```sh
kubectl get limitador limitador -n kuadrant-system -o yaml
```

<details>
  <summary>Expected output</summary>

  ```yaml
  apiVersion: limitador.kuadrant.io/v1alpha1
  kind: Limitador
  metadata:
    name: limitador
    namespace: kuadrant-system
    …
  spec:
    limits:
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/istio-ingressgateway#*.toystore.acme.com
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/istio-ingressgateway#*.toystore.acme.com
      seconds: 43200
      variables: []
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/istio-ingressgateway#*.toystore.acme.com
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/istio-ingressgateway#*.toystore.io
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/istio-ingressgateway#*.toystore.io
      seconds: 43200
      variables: []
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/istio-ingressgateway#*.toystore.io
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/istio-ingressgateway#*.toystore.internal
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/istio-ingressgateway#*.toystore.internal
      seconds: 43200
      variables: []
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/istio-ingressgateway#*.toystore.internal
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/internal-gateway#*.toystore.acme.com
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/internal-gateway#*.toystore.acme.com
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/internal-gateway#*.toystore.acme.com
      seconds: 43200
      variables: []
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/internal-gateway#*.toystore.io
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/internal-gateway#*.toystore.io
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/internal-gateway#*.toystore.io
      seconds: 43200
      variables: []
    - conditions:
      - default/toystore-per-endpoint/toys == "1"
      max_value: 50
      namespace: istio-system/internal-gateway#*.toystore.internal
      seconds: 60
      variables:
      - auth.identity.username
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 5
      namespace: istio-system/internal-gateway#*.toystore.internal
      seconds: 60
      variables: []
    - conditions:
      - default/toystore-per-endpoint/assets == "1"
      max_value: 100
      namespace: istio-system/internal-gateway#*.toystore.internal
      seconds: 43200
      variables: []
  status:
    conditions:
    - lastTransitionTime: "2023-07-06T13:36:32Z"
      message: Limitador is ready
      reason: Ready
      status: "True"
      type: Ready
    observedGeneration: 2
    service:
      host: limitador-limitador.kuadrant-system.svc.cluster.local
      ports:
        grpc: 8081
        http: 8080
  ```
</details>

⑥ Verify the WasmPlugin CRs were reconciled correctly:

```sh
kubectl get wasmplugin kuadrant-istio-ingressgateway -n istio-system  -o yaml
```

<details>
  <summary>Expected output</summary>

  ```yaml
  apiVersion: extensions.istio.io/v1alpha1
  kind: WasmPlugin
  metadata:
    name: kuadrant-istio-ingressgateway
    namespace: istio-system
    …
  spec:
    phase: STATS
    pluginConfig:
      failureMode: deny
      rateLimitPolicies:
      - domain: istio-system/istio-ingressgateway#*.toystore.acme.com
        hostnames:
        - '*.toystore.acme.com'
        - '*.toystore.io'
        - '*.toystore.internal'
        name: default/toystore-per-endpoint
        rules:
        - conditions:
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /toys
            - operator: eq
              selector: request.method
              value: GET
            - operator: endswith
              selector: request.host
              value: .toystore.acme.com
            - operator: neq
              selector: auth.identity.group
              value: admin
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /toys
            - operator: eq
              selector: request.method
              value: POST
            - operator: endswith
              selector: request.host
              value: .toystore.acme.com
            - operator: neq
              selector: auth.identity.group
              value: admin
          data:
          - static:
              key: default/toystore-per-endpoint/toys
              value: "1"
          - selector:
              selector: auth.identity.username
        - conditions:
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /assets/
          data:
          - static:
              key: default/toystore-per-endpoint/assets
              value: "1"
        service: kuadrant-rate-limiting-service
    selector:
      matchLabels:
        app: istio-ingressgateway
        istio: ingressgateway
    url: oci://quay.io/kuadrant/wasm-shim:latest
  ```
</details>

```sh
kubectl get wasmplugin kuadrant-internal-gateway -n istio-system  -o yaml
```

<details>
  <summary>Expected output</summary>

  ```yaml
  apiVersion: extensions.istio.io/v1alpha1
  kind: WasmPlugin
  metadata:
    name: kuadrant-internal-gateway
    namespace: istio-system
    …
  spec:
    phase: STATS
    pluginConfig:
      failureMode: deny
      rateLimitPolicies:
      - domain: istio-system/internal-gateway#*.toystore.internal # <==== this will have to change or it will activate different limits/increase different counters
        hostnames:
        - '*.toystore.internal'
        name: default/toystore-per-endpoint
        rules:
        - conditions:
          - allOf:
            - operator: startswith
              selector: request.url_path
              value: /assets/
          data:
          - static:
              key: default/toystore-per-endpoint/assets
              value: "1"
        service: kuadrant-rate-limiting-service
    selector:
      matchLabels:
        istio.io/gateway-name: internal-gateway
    url: oci://quay.io/kuadrant/wasm-shim:latest
  ```
</details>

---

Closes #200 
Relates to #156


pkg/common/gatewayapi_utils.go
```diff
@@ -91,6 +91,117 @@ func RulesFromHTTPRoute(route *gatewayapiv1beta1.HTTPRoute) []HTTPRouteRule {
...
+		for _, param := range s.QueryParams {
+			if _, found := Find(ruleMatch.QueryParams, func(otherParam gatewayapiv1beta1.HTTPQueryParamMatch) bool {
+				return reflect.DeepEqual(param, otherParam)
+			}); !found {
+				return false
+			}
+		}
+
+		// method
+		if s.Method != nil && !reflect.DeepEqual(s.Method, ruleMatch.Method) {
```
didierofrivia, line 128: Maybe we can hoist this comparison after path, so we can return early?

api/v1beta2/route_selectors.go
```diff
@@ -0,0 +1,47 @@
...
+
+// SelectRules returns, from a HTTPRoute, all HTTPRouteRules that either specify no HTTRouteMatches or that contain at
+// least one HTTRouteMatch whose statements expressly include (partially or totally) the statements of at least one of
+// the matches of the selector. If the selector does not specify any matches, then all HTTPRouteRules are selected.
+//
+// Additionally, if the selector specifies a non-empty list of hostnames, a non-empty intersection between the literal
+// hostnames of the selector and set of hostnames specified in the HTTPRoute must be exist. Otherwise, the function
+// returns nil.
+func (s *RouteSelector) SelectRules(route *gatewayapiv1beta1.HTTPRoute) []gatewayapiv1beta1.HTTPRouteRule {
+	rulesIndices := make(map[int]gatewayapiv1beta1.HTTPRouteRule, 0) // using a map is an easy way to avoid repeated rules but it may not preserve the order
```
didierofrivia, line 31: So it could happen that the rules are repeated, thus we need to filter them... Maybe if we found ourselves dealing with repetition and order, we could introduce an `OrderedMap` generic, wrapping an ordinary map and a slice of keys for the order.

pkg/common/common.go
```diff
@@ -135,6 +188,16 @@ func ReverseSlice[T any](input []T) []T {
 	return output
 }
 
+func MapValues[T comparable, U any](m map[T]U) []U {
```
didierofrivia, line 191: The existence of this fn is because we need the ordered values of a map, right?

pkg/reconcilers/targetref_reconciler.go
```diff
@@ -94,6 +94,35 @@ func (r *TargetRefReconciler) FetchValidTargetRef(ctx context.Context, targetRef
 	return nil, fmt.Errorf("FetchValidTargetRef: targetRef (%v) to unknown network resource", targetRef)
 }
 
+// FetchAcceptedGatewayHTTPRoutes returns the list of HTTPRoutes that have been accepted as children of a gateway.
+func (r *TargetRefReconciler) FetchAcceptedGatewayHTTPRoutes(ctx context.Context, gwKey client.ObjectKey) (routeKeys []gatewayapiv1beta1.HTTPRoute) {
```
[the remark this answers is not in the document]
guicassolato: Absolutely yes, @alexsnaps! Not doing it now just because this change goes to the RLPv2 feature branch first and we haven't moved `main` to using the gateway-api-machinery yet.

api/v1beta2/route_selectors.go
guicassolato: These rules (HTTPRouteRules) can be repeated because it's possible to have more than one route selector in a policy that selects the same HTTPRouteRule of the HTTPRoute. You are right about better methods to control the repetition here that avoid messing with the order.

Other examples where we also fail to guarantee the order, but that might require its own different method to tackle, is when we list objects from the cluster (via k8s server API) and then iterate over the items to generate config.

In both cases, simply by going through lists of objects in a different order than before can cause a resource to be flagged as different from its previous generation and issue an update, even if those 2 versions of the resources means effectively the same thing.

eguzki:
High level, reviewed the verification steps and they are what we should expect (I have not run them yet). Always under the assumption of `when a policy targets a HTTPRoute, that policy fully replaces any other policies that might exist targeting the HTTPRoute's parents Gateways.`

One nitpick: For the rlp-4 applied to the route-2, the empty `conditions` (also absence of the `conditions` field which is options) works like "catch all" and matches all requests for the given hostname in the wasm-shim. No need to add 

```yaml
 - conditions:
        - allOf: # from route-2
          - operator: startswith
            selector: request.url_path
            value: /
```

api/v1beta2/route_selectors_test.go
```diff
@@ -0,0 +1,211 @@
...
+					{
+						Path: &gatewayapiv1beta1.HTTPPathMatch{
+							Type:  &[]gatewayapiv1beta1.PathMatchType{gatewayapiv1beta1.PathMatchExact}[0],
+							Value: &[]string{"/toy"}[0],
+						},
+					},
+				},
+			},
+			route:    route,
+			expected: nil,
```
eguzki, line 148: If the route says `/toys*`, then the route selector `/toy` should match, shouldn't it?

pkg/reconcilers/targetref_reconciler.go
```diff
@@ -94,6 +94,35 @@ func (r *TargetRefReconciler) FetchValidTargetRef(ctx context.Context, targetRef
 	return nil, fmt.Errorf("FetchValidTargetRef: targetRef (%v) to unknown network resource", targetRef)
 }
 
+// FetchAcceptedGatewayHTTPRoutes returns the list of HTTPRoutes that have been accepted as children of a gateway.
+func (r *TargetRefReconciler) FetchAcceptedGatewayHTTPRoutes(ctx context.Context, gwKey client.ObjectKey) (routes []gatewayapiv1beta1.HTTPRoute) {
+	logger, _ := logr.FromContext(ctx)
+	logger = logger.WithName("FetchAcceptedGatewayHTTPRoutes").WithValues("gateway", gwKey)
+
+	routeList := &gatewayapiv1beta1.HTTPRouteList{}
+	err := r.Client().List(ctx, routeList)
```
eguzki, line 103: Without any criteria, this could be over-killing. Can't we use Gateway status or the annotations to get this list?

pkg/common/common_test.go
```diff
@@ -250,6 +250,274 @@ func TestContains(t *testing.T) {
...
+
+func TestIntersection(t *testing.T) {
+	testCases := []struct {
+		name     string
+		slice1   []string
+		slice2   []string
+		expected []string
+	}{
+		{
+			name:     "when slice1 and slice2 have one common item then return true",
```
eguzki, line 467: nitpick: the return value is not a `bool`

api/v1beta2/route_selectors_test.go
guicassolato: It should not.

pkg/common/common_test.go
guicassolato: Nice catch. It was copied and pasted from the tests written for another function, `Intersect`. I'll fix it.

pkg/reconcilers/targetref_reconciler.go
guicassolato: IIRC, common options for filtering results here are:
- labels
- resource fields ([not all fields are supported](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-fields))
- namespace

I didn't know we could use annotations.

If the HTTPRoute allows us to use `parentRef` and/or `status.parents` for filtering, that would be great. I'll look into it.

api/v1beta2/route_selectors_test.go
eguzki: We should document this very clearly. For me it is counter intuitive.

didierofrivia: `/toys*` should not include `/toy`, but `/toy*` would

pkg/rlptools/wasm_utils.go
```diff
@@ -20,7 +21,7 @@ var (
 	WASMFilterImageURL = common.FetchEnv("RELATED_IMAGE_WASMSHIM", "oci://quay.io/kuadrant/wasm-shim:latest")
 )
 
-// WasmRules computes WASM rules from the policy and the targeted Route (which can be nil when a gateway is targeted)
+// WasmRules computes WASM rules from the policy and the targeted route.
 func WasmRules(rlp *kuadrantv1beta2.RateLimitPolicy, route *gatewayapiv1beta1.HTTPRoute) []wasm.Rule {
```
eguzki, line 25: This method should return `error` as well. One of the use cases I found is when the route selector does not match any HTTPRoute rule (route)

eguzki (changes requested):

pkg/rlptools/wasm_utils.go
eguzki: As an alternative, the route selectors validity could be evaluated in a early stage of the reconcilliation loop to report in the status when the selector does not select any route. With this approach, the `WasmRules` only returns rules

pkg/reconcilers/targetref_reconciler.go
eguzki: `s/annotations/labels/g`

Indeed, I meant labels. Anyway, enforcing users to add some kuadrant labels to the HTTPRoute resources only the controller can query the cluster is not very cool.

eguzki:
verification steps working

eguzki:
Tested removing httproutes and rlp's to see how reconciliation handles and works as expected :muscle:

eguzki:
I think now I (fully) understand the route selector concept. We really need to document it clearly.  If `routeSelector` are thought as traffic qualifiers for rate limiting, that may lead to unexpected behavior. `routeSelectors` are (surprise) nothing else than a HTTPRoute rule (route) selector. And the traffic qualifiers for rate limiting are the (selected) HTTPRoute rules. 

Example:

* HTTPRoute rules: `GET /v1*` or `POST /v1*`
* RLP routeSelector: `GET /v1*`

The traffic to be rate limited will be: `GET /v1*` or `POST /v1*`. One (like me) might expect to be `GET /v1*`

api/v1beta2/route_selectors_test.go
guicassolato: Neither `/toy`, nor `/toy*` in a route selector match selects a HTTPRouteRule that specifies a single HTTPRouteMatch `/toys*`.

For a route selector to select a HTTPRouteRule that specifies a single HTTPRouteMatch `/toys*`, the route selector MUST specify `/toys*`. I.e. the fields of route selector's HTTPRouteMatch must be identical to the same fields in HTTPRouteRule's HTTPRouteMatch. Otherwise, we're back to having to deal with the sets and subsets problem all over again.

<br/>

Here's the semantics to have in mind for each pair of route selector HTTPRouteMatch and HTTPRouteRule's HTTPRouteMatch to decide whether the route selector selects the HTTPRouteRule or not:

1. A route selector selects a HTTPRouteRule if the HTTPRouteRule contains at least one HTTPRouteMatch that specifies fields that are literally identical to all the fields specified by at least one HTTPRouteMatch of the route selector.
2. A HTTPRouteMatch within a HTTPRouteRule may include other fields that are not specified in a route selector match and yet the route selector match selects the HTTPRouteRule if rule all fields of the route select match are identically included in the HTTPRouteRule's HTTPRouteMatch; the opposite is NOT true.
3. Each field `path` of a HTTPRouteMatch, as well as each field `method` of a HTTPRouteMatch, as well as each element of the fields `headers` and `queryParams` of a HTTPRouteMatch, is atomic – this is true for the HTTPRouteMatches within a HTTPRouteRule, as well as for HTTPRouteMatches of a route selector.

This semantics is described through examples in [RFC 0001](https://github.com/Kuadrant/architecture/blob/main/rfcs/0001-rlp-v2.md).

pkg/rlptools/wasm_utils.go
guicassolato: Right now, it's simply skipping a route selector that does not select anything.

Instead of raising a error here, I'd advocate for a combination of the following:
1. the alternative you suggest – i.e. checking for route selector mismatches early on; I'd say in a validating webhook;
2. reporting in the status of the RLP about all limits that will not be applied due to route selector mismatch. This one because this reconciliation can be triggered by changes in the HTTPRoutes as well.

didierofrivia:
Verification steps working, both guides! 

BTW, super useful to extract to a user guide 📖

pkg/common/common_test.go
guicassolato: Done

eguzki (approved):

Issue #216, opened 2023-07-17
user-guide: authorization using k8s service account returning 403 instead of 200


KevFan:
**WHAT**
Following the [Rate-limiting and protecting an API with JSON Web Tokens (JWTs) and Kubernetes authnz using Kuadrant](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md) user guide, at [Step 6](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md), the following `GET` request using the K8s service account token should return a `200` status code:
```
curl -H "Authorization: Bearer $SA_TOKEN" -H 'Host: api.toystore.com' http://localhost:9080/toy -i
# HTTP/1.1 200 OK
```
However, it insteads return a `403` still:
```
curl -H "Authorization: Bearer $SA_TOKEN" -H 'Host: api.toystore.com' http://localhost:9080/toy -i
# HTTP/1.1 403 Forbidden
# x-ext-auth-reason: Not authorized: unknown reason
# date: Mon, 17 Jul 2023 12:58:04 GMT
# server: istio-envoy
# content-length: 0
```
Due to this error, in [Step 7](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md#try-the-api-rate-limited) of the user guide, it is not possible to validate the Rate limit using the service account also.

**DONE**
* The guide is working as expected

guicassolato:
This is a regression cause by https://github.com/Kuadrant/authorino/issues/394 and more specifically by https://github.com/Kuadrant/authorino/pull/403.

Here's a PR that fixed the Authorino docs with an updated example that can be used to fix this user guide: https://github.com/Kuadrant/authorino/pull/408.

Because the AuthPolicy includes 2 sources of identity (Keycloak realm and Kubernetes SAs), you may want to rely on Authorino's [`extendedProperties`](https://github.com/Kuadrant/authorino/blob/main/docs/features.md#extra-identity-extension-extendedproperties) feature to ensure `auth.identity.sub` is set for the `k8s-service-accounts` identity objects was well.

PR #217, opened 2023-07-17
docs: minor improvements

KevFan:
General improvements in docs:

* Fix some typos
* Fix general linking to resources
* Consistent spacing for tables so markdowns can be read easier
* Add code type to code blocks for syntax highlighting
* Some consitency in docs / guides for creating `Kuadrant` instance
* Fix some spacing for proper display in the website (potentially)
* Consitent heading sizes in user guides
* Consistent inital setup in user guides
* Closes https://github.com/Kuadrant/kuadrant-operator/issues/216

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The Operator to install and manage the lifecycle of the [Kuadrant](https://githu
       * [If you are an <em>API Provider</em>](#if-you-are-an-api-provider)
       * [If you are a <em>Cluster Operator</em>](#if-you-are-a-cluster-operator)
 * [User guides](#user-guides)
-* [<a href="/doc/rate-limiting.md">Kuadrant Rate Limiting</a>](#kuadrant-rate-limiting)
+* [<a href="doc/rate-limiting.md">Kuadrant Rate Limiting</a>](#kuadrant-rate-limiting)
 * [Documentation](#documentation)
 * [Contributing](#contributing)
 * [Licensing](#licensing)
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ The Operator to install and manage the lifecycle of the [Kuadrant](https://githu
       * [If you are an <em>API Provider</em>](#if-you-are-an-api-provider)
       * [If you are a <em>Cluster Operator</em>](#if-you-are-a-cluster-operator)
 * [User guides](#user-guides)
-* [<a href="/doc/rate-limiting.md">Kuadrant Rate Limiting</a>](#kuadrant-rate-limiting)
+* [<a href="doc/rate-limiting.md">Kuadrant Rate Limiting</a>](#kuadrant-rate-limiting)
```
eguzki, line 21: Just for curiosity, is `/doc/rate-limiting.md` wrong? the link seems to be working (on Linux/Firefox)

README.md
```diff
@@ -125,12 +126,13 @@ kubectl create namespace kuadrant
 
 Apply the `Kuadrant` custom resource:
 
-```yaml
-kubectl apply -n kuadrant -f -<<EOF
+```sh
```
eguzki, line 129: nitpick: technically it is a shell script, but I would be highlighting the yaml content

doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md
```diff
@@ -49,12 +41,6 @@ Create the `HTTPRoute`:
 kubectl apply -f examples/toystore/httproute.yaml
 ```
 
-Expose the API:
```
eguzki, line 52: this is needed when not running in Linux

eguzki (approved):
after some nitpicking, LGTM

README.md
guicassolato: Plus, if you keep it as `sh`, you can run it in vscode with https://marketplace.visualstudio.com/items?itemName=guicassolato.tothom 😜

doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md
guicassolato: On macOS at least, it's not needed anymore.

README.md
guicassolato: The leading slash breaks it in https://docs.kuadrant.io/kuadrant-operator/. Not sure if using the simple relative path will solve it tho.

KevFan: It works but it is broken in the website for the above in the table of contents ^^ . I am also not sure will this change fix it since it links to https://docs.kuadrant.io/doc/rate-limiting.md vs https://docs.kuadrant.io/kuadrant-operator/doc/rate-limiting/ which seems to be the correct url instead 🤷 

Though this change is also to give consitency to references to other docs in the repo. In most other links, we use `docs/x` instead of `/docs/x` 🤔

README.md
KevFan: Yeah, this is a hard one 🤔  TBH previously, we are not consistent in this use for code blocks. In some places, we use `yaml` vs `sh`. 

While highlighting for yaml is nice, I've went with `sh` since it is a shell script and as Gui hinted for vscode, in Goland IDE there is a similar function to run commands directly from markdowns for shell commands also, which is also a nice feature

image

KevFan:
I'll push a commit for fixing https://github.com/Kuadrant/kuadrant-operator/issues/216 as part of this PR

doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md
KevFan: I've added the following note a bit further down the guide instead:

```
It should return `200 OK`.

**Note**: This only works out of the box on linux environments. If not on linux,
you may need to forward ports

kubectl port-forward -n istio-system service/istio-ingressgateway 9080:80 &
```

This keeps it consistent with the other user guides which specifies this note instead of explicity requiring the user to port forward as a guide step 🤔

README.md
eguzki: take my money

PR #228, opened 2023-08-10
codecov: do not fail ci on error

KevFan:
* Prevent codecov from failing ci on error due to flakiness with codecov upload

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -40,7 +40,7 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           flags: unit
-          fail_ci_if_error: true
+          fail_ci_if_error: false
           verbose: true
 
   integration-tests:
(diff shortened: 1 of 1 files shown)


adam-cattermole (approved):
Looks good to me

art-tapin (changes requested):
Hey @KevFan! 
The CodeCov uploads the results two times: for unit tests and for integration tests separately. Please, check out the line 90 in test.yaml to change the `true` condition to `false` for integration tests too:
 ```
       - name: Upload integration-test coverage reports to CodeCov # more at https://github.com/codecov/codecov-action
        uses: codecov/codecov-action@v3
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
          flags: integration
          fail_ci_if_error: true
          verbose: true

KevFan:
Thanks @art-tapin ! Good catch, I've pushed that change just now 👍

art-tapin (approved):

PR #230, opened 2023-08-11
RLP v1beta2

### LINKED ISSUE #156: Implement RLPv2 Direct Policy Attachment

alexsnaps:
I did a merge myself, and came up with some differences, will leave comments inline

---

Closes #156


pkg/rlptools/wasm_utils_test.go
```diff
@@ -3,56 +3,26 @@
...
 import (
-	"reflect"
 	"testing"
 
-	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
-	"github.com/kuadrant/kuadrant-operator/pkg/common"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"github.com/google/go-cmp/cmp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
```
alexsnaps, line 10:
```	
gatewayapiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
```

?

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -0,0 +1,247 @@
...
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/google/go-cmp/cmp"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
```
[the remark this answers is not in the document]
KevFan: I believe we do for `TargetRef gatewayapiv1alpha2.PolicyTargetReference`:
- https://github.com/KevFan/kuadrant-operator/blob/rlp-v2-rebase/api/v1beta2/ratelimitpolicy_types.go#L119

I dont believe this is available in v1beta1 unless I'm missing something 🤔

pkg/rlptools/wasm_utils_test.go
KevFan: Good catch 👍

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -106,7 +106,7 @@ type Limit struct {
 }
 
 func (l Limit) CountersAsStringList() []string {
-	var ret []string
+	ret := make([]string, 0)
```
guicassolato, line 109: These 2 ways to declare the slice are NOT equivalent. Even though I think we don't usually rely on null check for slices, but rather on `len(s) == 0`, effectively this change alters the behaviour of the function and should be avoided.

If the goal is to improve memory allocation and performance, then I would do:

```go
size := len(l.Counters)
if (size == 0) {
	return nil
}
ret := make([]string, len)
// we cannot use copy(ret, l.Counters) due to different types
for idx := range l.Counters {
	ret[idx] = string(l.Counters[idx])
}
return ret
```

Or, in this particular case:

```go
size := len(l.Counters)
if (size == 0) {
	return nil
}
return common.Map(l.Counters, func(counter ContextSelector) string { return string(counter) })
```

guicassolato (approved):
I've left a minor comment although I don't think it's critical.
In general, LGTM. [Verification steps](https://github.com/Kuadrant/kuadrant-operator/pull/218#issue-1808230441) also working fine.
Thanks for putting this together!

guicassolato:
To others interested in revieweing this PR:

```sh
git diff remotes/origin/rlp-v2-update-main..remotes/KevFan/rlp-v2-rebase
```

api/v1beta2/ratelimitpolicy_types.go
KevFan: Thanks ! I've pushed this change 👍

guicassolato: The pattern is also applied in https://github.com/Kuadrant/kuadrant-operator/pull/230/commits/4bf4198540c75fc37d44ebf8d1f080ec1a6762d0#diff-3565ea71e4b2b9f8f80ea1c57659cc4bf19888115b8ccca64958f797eb39bcd4R28

KevFan: Should the slice size check be applied also  ?

```go
	if len(rlpRefs) == 0 {
		return nil
	}
```
In this instance the logic from `r.reconcileLimitador(ctx, rlp, rlpRefsWithoutRLP)` will be skipped ? 🤔 
- https://github.com/KevFan/kuadrant-operator/blob/rlp-v2-rebase/controllers/ratelimitpolicy_limits.go#L35

I guess the `Filter` function should be used instead ? Though it looks to set the capacity also to `make([]T, 0)`
- https://github.com/KevFan/kuadrant-operator/blob/rlp-v2-rebase/pkg/common/common.go#L159-L168

Something like:
```go
	rlpRefsWithoutRLP := common.Filter(rlpRefs, func(rlpRef client.ObjectKey) bool {
		return rlpRef.Name != rlp.Name && rlpRef.Namespace != rlp.Namespace
	})
```

guicassolato: > Should the slice size check be applied also ?

I would say not. Let it pass `nil` into `reconcileLimitador` so we do not change the behaviour.

<br/>

> I guess the `Filter` function should be used instead ? Though it looks to set the capacity also to `make([]T, 0)`

Yes, but watch out for the logic though. The `Filter` function receives a predicate function that appends the item to the resulting slice when true. In this case, we want to append all `rlpRef`s but the one whose name and namespace match `rlp`'s name and namespace. ∴

```go
func (r *RateLimitPolicyReconciler) deleteLimits(ctx context.Context, rlp *kuadrantv1beta2.RateLimitPolicy) error {
	rlpRefs, err := r.GetAllGatewayPolicyRefs(ctx, &common.KuadrantRateLimitPolicyRefsConfig{})
	if err != nil {
		return err
	}
	rlpRefsWithoutRLP := common.Filter(rlpRefs, func(rlpRef client.ObjectKey) bool {
		return rlpRef.Name != rlp.Name || rlpRef.Namespace != rlp.Namespace
	})
	return r.reconcileLimitador(ctx, rlp, rlpRefsWithoutRLP)
}
```

KevFan: I see, thanks ! Updated 👍

guicassolato (approved):
🏆

PR #232, opened 2023-08-22
default dynamic wasm rule data to empty string

guicassolato:
Sets the empty string as the default value for the counter qualifiers ("wasm data items") in the wasm config, to prevent the selectors that fail to select a value to invalidate the limit.

Ref.: https://github.com/Kuadrant/wasm-shim/pull/39#discussion_r1301663487

### DIFF
--- a/pkg/rlptools/wasm_utils.go
+++ b/pkg/rlptools/wasm_utils.go
@@ -245,7 +245,7 @@ func dataFromLimt(limitIdentifier string, limit *kuadrantv1beta2.Limit) (data []
 	data = append(data, wasm.DataItem{Static: &wasm.StaticSpec{Key: limitIdentifier, Value: "1"}})
 
 	for _, counter := range limit.Counters {
-		data = append(data, wasm.DataItem{Selector: &wasm.SelectorSpec{Selector: counter}})
+		data = append(data, wasm.DataItem{Selector: &wasm.SelectorSpec{Selector: counter, Default: common.Ptr("")}})
 	}
 
 	return data
--- a/controllers/ratelimitpolicy_controller_test.go
+++ b/controllers/ratelimitpolicy_controller_test.go
@@ -457,6 +457,7 @@ var _ = Describe("RateLimitPolicy controller", func() {
 					{
 						Selector: &wasm.SelectorSpec{
 							Selector: kuadrantv1beta2.ContextSelector("auth.identity.username"),
+							Default:  common.Ptr(""),
 						},
 					},
 				},
--- a/pkg/rlptools/wasm_utils_test.go
+++ b/pkg/rlptools/wasm_utils_test.go
@@ -10,6 +10,7 @@ import (
 	gatewayapiv1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
 	"github.com/kuadrant/kuadrant-operator/pkg/rlptools/wasm"
 )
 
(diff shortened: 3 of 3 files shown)


pkg/rlptools/wasm_utils.go
```diff
@@ -245,7 +245,7 @@ func dataFromLimt(limitIdentifier string, limit *kuadrantv1beta2.Limit) (data []
 	data = append(data, wasm.DataItem{Static: &wasm.StaticSpec{Key: limitIdentifier, Value: "1"}})
 
 	for _, counter := range limit.Counters {
-		data = append(data, wasm.DataItem{Selector: &wasm.SelectorSpec{Selector: counter}})
+		data = append(data, wasm.DataItem{Selector: &wasm.SelectorSpec{Selector: counter, Default: common.Ptr("")}})
```
eguzki, line 248: What if I set as counter `request.headers.My-Custom-Header-01` ?  All the requests not having that header will contribute to a single (limitador's internal hit counter)

guicassolato: Request that do not have that header are treated as requests that have that header with an empty value, therefore sharing the same counter. If what you want is to trigger or not trigger a limit based on the presence of a header, use a condition instead; not a counter qualifier.

eguzki: Indeed, what you say is correct, but I do not fully agree on this UX: if you set a counter based on a header, add a condition on the header existence. 

For me, adding a counter implicitly tells I **only** want to rate limit traffic that is qualified by the counter. 

Let's see what @didierofrivia and @alexsnaps have to say about this

guicassolato: @eguzki, coincidently, this was exactly the example that made us[^1] want to default to an empty string in the wasm-shim (what https://github.com/Kuadrant/wasm-shim/pull/39 wants to revert now).

[^1]: Discussed offline with @alexsnaps and @didierofrivia on 2023-07-27, and [implemented](https://github.com/Kuadrant/wasm-shim/pull/37/commits/59033b0b4f89290391c8eee9acf00fde51fe6654) by @alexsnaps on 2023-07-28.

eguzki: https://github.com/Kuadrant/wasm-shim/pull/39 does not revert entirely [that commit](https://github.com/Kuadrant/wasm-shim/pull/37/commits/59033b0b4f89290391c8eee9acf00fde51fe6654). Only reverts the behavior when the descriptor is not found in the module generating RLS descriptors. For the conditions, https://github.com/Kuadrant/wasm-shim/pull/39 keeps the changes to allow operators to be meaningful.

guicassolato: > [Kuadrant/wasm-shim#39](https://github.com/Kuadrant/wasm-shim/pull/39) does not revert entirely [that commit](https://github.com/Kuadrant/wasm-shim/pull/37/commits/59033b0b4f89290391c8eee9acf00fde51fe6654).

I understand that. But, most importantly, I also understand that setting the default value to `""`, so some RL is enforced, is one approach. In itself, it is nor better, nor worse than setting no default at all. If most people prefers the contrary of what's proposed here, I'm good with that too obviously.

guicassolato:
So, after some discussion we've decided NOT to move on with this change. The reasoning:
1. Setting an empty string loses information; by neglecting the distinction between empty and null, we're dismissing the meaning of missing data – changing that in the future may be hard
2. Missing a value is an exception case, therefore users should NOT design with that in mind; instead, the behaviour for such case MUST be different – e.g. define other limits for that case, using the conditions, etc
3. Defaulting to a general RL case cannot be the solution to protect against failures (e.g. typos), but to improve the visibility instead (e.g. better logging, debugging tools, etc)
4. Consistency with Limitador - in Limitador, a limit is only enforced when the required data is complete; the same principle can apply to the wasm-shim. While the `default` field of the wasm config is not exposed to users, completeness in the case is controlled by the RLP controller. Let's not make "complete" something that was not set as complete by the user
5. In the future, we can provide an option (e.g. expose the `default` field to users) for users to define their own desired (safe) intended behaviours when data is missing, e.g. blocking the request altogether (in contrast to sending the traffic to Limitador and let Limitador decide)

Issue #203, opened 2023-06-22
Move to using kuadrant/gateway-api-machinery


guicassolato:
Move to using [Kuadrant Gateway API Machinery](https://github.com/kuadrant/gateway-api-machinery) in the implementation of the controllers.

Requires upgrading [controller-runtime](https://github.com/Kuadrant/kuadrant-operator/blob/3f21f814c05241ddbb51126d928f1673cd592cb9/go.mod#LL27C2-L27C32) and to Golang v1.20.

guicassolato:
Superseded by #806

PR #240, opened 2023-08-28
Reconcile resources

### LINKED ISSUE #234: Configurations not recociled on but Kuadrant operator
When kuadrant is configured using an `AuthPolicy` and `RateLimitPolicy` CRs, a user can still edit the configurations in the `AuthConfig` and `Limitador` CRs. These edits are not reverted by the kuadrant operator and leads to a misalignment in expect configurations.

Solution here would be for the kuadrant operator to monitor changes to the `AuthConfig` & `Limitador` CRs and reconcile configuration changes back to what is stated in the  `AuthPolicy` & `RateLimitPolicy`.

Related Issue: 
* https://github.com/Kuadrant/kuadrant-operator/issues/163
* https://github.com/Kuadrant/kuadrant-operator/issues/164
* https://github.com/Kuadrant/architecture/pull/16

Boomatang:
closes #234 

Reconcile changes made to the AuthConfig and Limitador CRs to reflect the configuration in AuthPolicy and RateLimitPolicy. Multiple rate limit policies can be configured in the on Limitador CR but it is only one rate limit policy per resource (http Route & gateway)

# Verification 

## Setup
* Install kuadrant operator using `make local-setup`
* Set up to gateway, routes and etc, etc
```sh
echo '
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  selector:
    matchLabels:
      app: toystore
  template:
    metadata:
      labels:
        app: toystore
    spec:
      containers:
        - name: toystore
          env:
            - name: PORT
              value: "3000"
          ports:
            - containerPort: 3000
              name: http
  replicas: 1
---
apiVersion: v1
kind: Service
metadata:
  name: toystore
spec:
  selector:
    app: toystore
  ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 3000
---
kind: HTTPRoute
metadata:
  name: toystore
  labels:
    app: toystore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/toy"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
---
kind: HTTPRoute
metadata:
  name: dollstore
  labels:
    app: dollstore
spec:
  parentRefs:
    - name: istio-ingressgateway
      namespace: istio-system
  hostnames: ["*.toystore.com"]
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/doll"
          method: GET
      backendRefs:
        - name: toystore
          port: 80
---
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: bob-key
  labels:
    app: toystore
stringData:
  api_key: IAMBOB
type: Opaque
---
apiVersion: v1
kind: Secret
metadata:
  annotations:
  name: alice-key
  labels:
    app: toystore
stringData:
  api_key: IAMALICE
type: Opaque 
---
' | kubectl apply -f -

```

## AuthPolicy Reconcile
* create authPolicy
```sh
echo '
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
  - paths: ["/toy*"]
  authScheme:
    identity:
    - name: friends
      apiKey:
        allNamespaces: true
        selector:
          matchLabels:
            app: toystore
      credentials:
        in: authorization_header
        keySelector: APIKEY
    response:
    - json:
        properties:
          - name: userID
            value: null
            valueFrom:
      name: rate-limit-apikey
      wrapper: envoyDynamicMetadata
      wrapperKey: ext_auth_data
' | kubectl apply -f -
```
* wait for reconcile to complete
```
watch kubectl get AuthPolicy toystore -o jsonpath='{.status.conditions[].status}'
```
Expect: True
* change part of the `AuthConfig` spec that was define as part of the AuthPolicy. Good example is `spec.identity.apikey.name = friend`
```sh 
kubectl edit AuthConfig ap-default-toystore  
```
* Check that the changes have being reverted. 
```
kubectl get AuthConfig ap-default-toystore -o yaml
```
* Expect changes should be reverted.
* Also not in the annotations a back reference to the authPolicy has being set.

### RateLimitPolicy at gateway level
* Create a RateLimitPolicy on the gateway
```sh
echo "
kind: RateLimitPolicy
metadata:
  name: toystore
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  limits:
    toystore:
      rates:
        - limit: 5
          duration: 10
          unit: second
      counters:
        - context.request.http.path
" | kubectl apply -f -

```
* wait for reconcile to complete
```
watch kubectl get RateLimitPolicy toystore -o jsonpath='{.status.conditions[].status}' -n istio-system
```
Expect: True
* Modify part of the limitador CR, example change the `max_value` from 5 to 1.
```sh
kubectl edit limitador limitador -n default
```
* Expect the value to be reverted back 
* Note the a reference added to the annotations for the rateLimitPolicy
### RateLimitPolicy at http Route level
* Create RateLimitPolicies at the route level
```sh
echo '
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    toystore:
      routeSelectors:
        - hostnames:
          - "*.toystore.com"
          matches:
            - path:
                type: PathPrefix
                value: "/toy"
              method: GET
      rates:
        - limit: 5
          duration: 10
          unit: second
      counters:
        - context.request.http.path
---
kind: RateLimitPolicy
metadata:
  name: dollstore
spec:
  targetRef:
    kind: HTTPRoute
    name: dollstore
  limits:
    toystore:
      routeSelectors:
        - hostnames:
          - "*.toystore.com"
          matches:
            - path:
                type: PathPrefix
                value: "/dolls"
              method: GET
      rates:
        - limit: 1
          duration: 5
          unit: second
      counters:
        - context.request.http.path
' | kubectl apply -f - 
```
* once again view the limitador CR
* Notice the different reference in the annotations.
* Edit some parts of the spec 
* Expect the edits to be reverted.
* Start removing the ratelimitPolicies
* Expect: the annotations should only list the current RateLimitPolicies.
* Once all RateLimitPolicies have being removed from the cluster the reference annotation should be fully removed. 
  * NOTE: the limitador CR will not be removed. This needs to be discussed and is out of scope for this work.

### DIFF
--- a/controllers/authpolicy_controller.go
+++ b/controllers/authpolicy_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+
 	authorinoapi "github.com/kuadrant/authorino/api/v1beta1"
 
 	"github.com/go-logr/logr"
--- a/controllers/authpolicy_controller.go
+++ b/controllers/authpolicy_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	authorinoapi "github.com/kuadrant/authorino/api/v1beta1"
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -199,6 +200,10 @@ func (r *AuthPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Logger: r.Logger().WithName("gatewayEventMapper"),
 	}
 
+	authConfigEventMapper := &AuthConfigEventMapper{
+		Logger: r.Logger().WithName("authConfigEventMapper"),
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&api.AuthPolicy{}).
 		Watches(
@@ -207,5 +212,7 @@ func (r *AuthPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		).
 		Watches(&source.Kind{Type: &gatewayapiv1beta1.Gateway{}},
 			handler.EnqueueRequestsFromMapFunc(gatewayEventMapper.MapToAuthPolicy)).
+		Watches(&source.Kind{Type: &authorinoapi.AuthConfig{}},
+			handler.EnqueueRequestsFromMapFunc(authConfigEventMapper.MapToAuthPolicy)).
 		Complete(r)
 }
--- a/controllers/limitador_eventmapper.go
+++ b/controllers/limitador_eventmapper.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"encoding/json"
+
 	"github.com/go-logr/logr"
 	"github.com/kuadrant/kuadrant-operator/pkg/common"
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
@@ -37,6 +38,7 @@ func (m LimitadorEventMapper) MapToRateLimitPolicy(obj client.Object) []reconcil
 		requests = append(requests, reconcile.Request{
 			NamespacedName: ref,
 		})
+		break
 	}
 	return requests
 }
--- a/controllers/limitador_eventmapper.go
+++ b/controllers/limitador_eventmapper.go
@@ -1,0 +1,42 @@
+package controllers
+
+import (
+	"encoding/json"
+	"github.com/go-logr/logr"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
+	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type LimitadorEventMapper struct {
+	Logger logr.Logger
+}
+
+func (m LimitadorEventMapper) MapToRateLimitPolicy(obj client.Object) []reconcile.Request {
+	limitador, ok := obj.(*limitadorv1alpha1.Limitador)
+	if !ok {
+		return []reconcile.Request{}
+	}
+
+	objAnnotations := limitador.GetAnnotations()
+	val, ok := objAnnotations[common.RateLimitPoliciesBackRefAnnotation]
+	if !ok {
+		return []reconcile.Request{}
+	}
+
+	var refs []client.ObjectKey
+	err := json.Unmarshal([]byte(val), &refs)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+
+	requests := make([]reconcile.Request, 0)
+	for _, ref := range refs {
+		m.Logger.V(1).Info("MapRateLimitPolicy", "ratelimitpolicy", ref)
+		requests = append(requests, reconcile.Request{
+			NamespacedName: ref,
+		})
+	}
+	return requests
+}
--- a/controllers/ratelimitpolicy_controller.go
+++ b/controllers/ratelimitpolicy_controller.go
@@ -19,8 +19,8 @@ package controllers
 import (
 	"context"
 	"encoding/json"
-
 	"github.com/go-logr/logr"
+	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -186,7 +186,7 @@ func (r *RateLimitPolicyReconciler) reconcileResources(ctx context.Context, rlp 
 		return err
 	}
 
-	// set annotation of policies afftecting the gateway - should be the last step, only when all the reconciliation steps succeed
+	// set annotation of policies affecting the gateway - should be the last step, only when all the reconciliation steps succeed
 	return r.ReconcileGatewayPolicyReferences(ctx, rlp, gatewayDiffObj)
 }
 
@@ -216,7 +216,13 @@ func (r *RateLimitPolicyReconciler) deleteResources(ctx context.Context, rlp *ku
 		}
 	}
 
-	// update annotation of policies afftecting the gateway
+	// remove direct back ref from limitador CR
+	err = r.deleteLimitadorBackReference(ctx, rlp)
+	if err != nil {
+		return err
+	}
+
+	// update annotation of policies affecting the gateway
 	return r.ReconcileGatewayPolicyReferences(ctx, rlp, gatewayDiffObj)
 }
 
@@ -241,6 +247,11 @@ func (r *RateLimitPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Logger: r.Logger().WithName("gatewayRateLimitPolicyEventMapper"),
 		Client: r.Client(),
 	}
+
+	limitadorEventMapper := &LimitadorEventMapper{
+		Logger: r.Logger().WithName("limitadorEventMapper"),
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kuadrantv1beta2.RateLimitPolicy{}).
 		Watches(
@@ -257,6 +268,9 @@ func (r *RateLimitPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(
 			&source.Kind{Type: &kuadrantv1beta2.RateLimitPolicy{}},
 			handler.EnqueueRequestsFromMapFunc(gatewayRateLimtPolicyEventMapper.MapRouteRateLimitPolicy),
+		).
+		Watches(&source.Kind{Type: &limitadorv1alpha1.Limitador{}},
+			handler.EnqueueRequestsFromMapFunc(limitadorEventMapper.MapToRateLimitPolicy),
 		).
 		Complete(r)
 }
--- a/controllers/ratelimitpolicy_controller.go
+++ b/controllers/ratelimitpolicy_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+
 	"github.com/go-logr/logr"
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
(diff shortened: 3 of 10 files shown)


controllers/authpolicy_controller.go
```diff
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	authorinoapi "github.com/kuadrant/authorino/api/v1beta1"
```
guicassolato, line 7:
```suggestion

	authorinoapi "github.com/kuadrant/authorino/api/v1beta1"
```

controllers/ratelimitpolicy_controller.go
```diff
@@ -216,7 +216,13 @@ func (r *RateLimitPolicyReconciler) deleteResources(ctx context.Context, rlp *ku
 		}
 	}
 
-	// update annotation of policies afftecting the gateway
+	// remove direct back ref from limitador CR
+	err = r.deleteLimitadorBackReference(ctx, rlp)
```
guicassolato, line 220: Hmm. So `deleteLimits` → `reconcileLimitador` → `reconcileLimitadorBackRef` adds the RLP ref to the back-ref annotation and performs an update on the `Limitador` resource, then we call `deleteLimitadorBackReference`, which removes the same RLP ref from the annotation and updates the `Limitador` resource again?

controllers/ratelimitpolicy_limits.go
```diff
@@ -108,3 +113,72 @@ func (r *RateLimitPolicyReconciler) buildRateLimitIndex(ctx context.Context, rlp
...
+	}
+	objAnnotations[common.RateLimitPoliciesBackRefAnnotation] = string(serialized)
+	limitador.SetAnnotations(objAnnotations)
+	return true, nil
+}
+
+func (r *RateLimitPolicyReconciler) deleteLimitadorBackReference(ctx context.Context, policy client.Object) error {
+	policyKey := client.ObjectKeyFromObject(policy)
+
+	limitadorList, err := r.listLimitadorByNamespace(ctx, policyKey.Namespace)
```
guicassolato, line 147: The `Limitador` CR is not necessarily in the same namespace as the RLP: https://github.com/Kuadrant/kuadrant-operator/blob/6f3b3fdee2d0213da3331265e14e72a2523cba4f/controllers/ratelimitpolicy_limits.go#L48

guicassolato (changes requested):
Adding and removing a policy ref from the back-ref annotation of a Kubernetes resource is a pattern we (will) repeat multiple times if we want to map all events related to these resources back to triggering the reconciliation of possibly affected policies.

In https://github.com/Kuadrant/gateway-api-machinery, we tried to define types that are a bit more generic, so they can be reused to manage back-ref annotations for different kinds of policies (e.g. Gateway, HTTPRoute). Perhaps we could try going one step further there and also generalise for the cases where any/multiple arbitrary kinds of resource are annotated with back refs of a same (generic) kind of policy (e.g. Limitador, WasmPlugin, and EnvoyFilter for the RateLimitPolicy).

Either way, I see an opportunity here to apply patterns from https://github.com/Kuadrant/gateway-api-machinery, to coordinate with changes to that other repo if needed, while addressing https://github.com/Kuadrant/kuadrant-operator/issues/203.

cc @KevFan

controllers/ratelimitpolicy_controller.go
Boomatang: Yes, that is correct and I should explain my reasoning. The `reconcileLimitador` function runs no matter what action is being carried out let it be create, update or delete. The label removal logic could be in the `reconcileLimitadorBackRef` but doing this would mean having recreate the logic for checking if the resource has being deleted. As it is the `deleteResources` --> `deleteLimitadorBackReference` only get called when a RLP is deleted there was no need to any more checks in the `deleteLimitadorBackReference`. I am aware there are more api calls with what I have done but as its only on deletion and not every reconcile I felt the trade off was worth it. 

Do you want me to move the label deletion logic to the `reconcileLimitadorBackRef`?

guicassolato: > Do you want me to move the label deletion logic to the `reconcileLimitadorBackRef`?

I would wait. As I said in my other [comment](https://github.com/Kuadrant/kuadrant-operator/pull/240#pullrequestreview-1598759818), if we're going to watch and reconcile changes on all derived resources – the `Limitador` CR is just one of then! –, I see more of this pattern of annotating/de-annotating ahead. I think we can aim for more generic functions and types that we can reuse in more places.

Boomatang: From the community call my understanding is we wont be using the  https://github.com/Kuadrant/gateway-api-machinery till after MVP. So I thinking I address the other issue raised now and create an issue to follow up with a refactor of the labelling once we can use the gateway-api-machinery. Would you be happy with this approach?

eguzki (changes requested):
I think that adding annotations with all the RLPs to the Limitador CR is something we will be doing the sooner or later. Nothing against that. Totally in favor of that. 

However, monitoring the Limtiador CR and generating a request event for each RLP instance is not necessary and that rings some bell. I think that the source issue is not the implementation of this PR but the implementation of the reconciliation of the Limitador CR's limits. That raises an out of scope discussion. 

I think that a bigger refactor is needed. But meanwhile, being less that sub-optimal, I cannot come up with a better approach. 


Last, but not least, what about the other manager resources like EnvoyFilter, WasmPlugin, AuthorizationPolicy from Istio??

controllers/limitador_eventmapper.go
```diff
@@ -0,0 +1,42 @@
...
+
+	var refs []client.ObjectKey
+	err := json.Unmarshal([]byte(val), &refs)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+
+	requests := make([]reconcile.Request, 0)
+	for _, ref := range refs {
+		m.Logger.V(1).Info("MapRateLimitPolicy", "ratelimitpolicy", ref)
```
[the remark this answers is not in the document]
Boomatang: I many explore this. What your saying is making sense but my mind is not allowing me to accept it as true. :smiley: I will add any changes as needed.

controllers/ratelimitpolicy_limits.go
```diff
@@ -68,8 +69,12 @@ func (r *RateLimitPolicyReconciler) reconcileLimitador(ctx context.Context, rlp
 		return err
 	}
 
-	// return if limitador is up to date
-	if rlptools.Equal(rateLimitIndex.ToRateLimits(), limitador.Spec.Limits) {
+	updated, err := r.reconcileLimitadorBackRef(limitador, rlp)
```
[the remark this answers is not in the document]
Boomatang: Ok, I see how this is would be wrong from comparing `existing` and `desired`. In my mind the reconcile loops were about eventually consistency which requires adding and rescheduling. This comes from how the RHOAM operator was working. I am happy to refactor this to match better with the comparing of `existing` and `desired` approach.

controllers/ratelimitpolicy_limits.go
```diff
@@ -68,8 +69,12 @@ func (r *RateLimitPolicyReconciler) reconcileLimitador(ctx context.Context, rlp
...
+	if err != nil {
+		return err
+	}
+	// return if limitador is up-to-date
+	if rlptools.Equal(rateLimitIndex.ToRateLimits(), limitador.Spec.Limits) && !updated {
```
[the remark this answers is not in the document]
Boomatang: Happy to refactor

controllers/ratelimitpolicy_limits.go
```diff
@@ -108,3 +113,72 @@ func (r *RateLimitPolicyReconciler) buildRateLimitIndex(ctx context.Context, rlp
 
 	return rateLimitIndex, nil
 }
+
+func (r *RateLimitPolicyReconciler) reconcileLimitadorBackRef(limitador *limitadorv1alpha1.Limitador, policyKey client.Object) (updated bool, err error) {
+	policy := client.ObjectKeyFromObject(policyKey)
```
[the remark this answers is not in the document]
Boomatang: I did see the function in the gatewayapi_utits. At the time I was in two minds about doing the refactor. The original method is tied a struct and not reusable as is,  and the method logic is straight forward. Personal I have no problem with repeating code twice. More than that I do think it should be generalized. 

I do think this labelling procedure will be moved to the gateway-api-machinery but that most likely will not happen before MVP at this stage.

Yes, naming is hard. `AddBackrefToLimitador` is  a better name. I will look to rename this while doing the refactor to `Existing` and `Desired` design mention earlier.

controllers/ratelimitpolicy_limits.go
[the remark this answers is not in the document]
Boomatang: On the name space first, Gui mention this and I have the change locally. As to why list the instances. I have being burned by hard-coded values changing to dynamic values a few times. So now I believe it to be more robust to use list over get. I am happy to refactor if you think it should but was my reasoning.

Boomatang:
> I think that a bigger refactor is needed. But meanwhile, being less that sub-optimal, I cannot come up with a better approach.

I tend to agree with this. The use of `watches` and `owns` can spiral out of control. I am not a fan of controllers and I think the use of informers is the better approach. As you said this is a large discussion which is out of scope here.

controllers/ratelimitpolicy_controller.go
guicassolato: This won't be complete anyway until we cover for all kinds of resources that, once modified directly, can break with the source of truth. I understand that whereas one is a refactoring, while the other prevents what could be perceived as a bug, I still think those two things are intimately related and better if tackled together. I'm also afraid of the explosion of annotations while [others seem to be addressing the issue differently](https://github.com/kubernetes-sigs/gateway-api/pull/2128). IOW, let's not amend it quickly due to release pressure and do this the right way. The fix can be part of a post-release patch if we want.

controllers/ratelimitpolicy_limits.go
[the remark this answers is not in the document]
eguzki: we can fix all DRY issues when depending on gateway-api-machinery lib

PR #242, opened 2023-08-30
Remove rlp watcher for gateway rlps

eguzki:
This even mapper for RLP's targeting a gateway is no longer necessary. When a RLP targeting a Gateway is updated, all rate limit required resources are updated correcly


guicassolato:
What about when the Gateway is modified (e.g. new listener added)?

eguzki:
That change will be picked by [this watcher](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimitpolicy_controller.go#L252-L255)

guicassolato (approved):

PR #243, opened 2023-08-30
Limitador cluster EnvoyFilter controller

eguzki:
### What

Specific controller to manage rate limit's EnvoyFilter used to add Limitador's service cluster to the Gateway config. 

RateLimitPolicy CRD controller is responsible of adding/deleting RLP refs to the gateway.

EnvoyFilter controller watches gateways. When one of them has `kuadrant.io/ratelimitpolicies` annotation, it means rate limiting should happen in that gateway, thus a EnvoyFilter resource is created for that gateway.

When there is no annotation or the list of RLP refs is empty, the EnvoyFilter resource is deleted (if exists).

The EnvoyFilter resource has owner ref to the gateway associated. The controller also watches for changes in the EnvoyFilter resource and reverts them back (if spec is changed).

### Verification steps

Setup the environment:

```sh
make local-setup
```

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

Create a Kuadrant `RateLimitPolicy` to configure rate limiting:

```sh
kind: RateLimitPolicy
metadata:
  name: gw-rlp
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
    namespace: istio-system
  limits:
    "global":
      rates:
      - limit: 5
        duration: 10
        unit: second
EOF
```

Wait for RLP to be available 

```sh
kubectl wait --for=condition=available ratelimitpolicy/gw-rlp -n istio-system --timeout=10s
```

Check for the Gateway's references to RLPs

```sh
kubectl get gateways istio-ingressgateway -n istio-system -o jsonpath='{.metadata.annotations.kuadrant\.io/ratelimitpolicies}' | yq e -P
```
It should output something like
```yaml
- Namespace: istio-system
  Name: gw-rlp
```

Check for the Limitador's Cluster EnvoyFilter resource

```sh
kubectl get envoyfilter kuadrant-ratelimiting-cluster-istio-ingressgateway -n istio-system -o yaml
```

It should output something like
```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  creationTimestamp: "2023-08-31T17:36:09Z"
  generation: 1
  name: kuadrant-ratelimiting-cluster-istio-ingressgateway
  namespace: istio-system
  ownerReferences:
  - apiVersion: gateway.networking.k8s.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: Gateway
    name: istio-ingressgateway
    uid: 043732fa-e588-4c3b-980f-7d9497e3342a
  resourceVersion: "2000"
  uid: 91b89285-eca3-4c0b-a7e4-84f6a7c4aeae
spec:
  configPatches:
  - applyTo: CLUSTER
    match:
      cluster:
        service: limitador-limitador.kuadrant-system.svc.cluster.local
    patch:
      operation: ADD
      value:
        connect_timeout: 1s
        http2_protocol_options: {}
        lb_policy: ROUND_ROBIN
        load_assignment:
          cluster_name: kuadrant-rate-limiting-service
          endpoints:
          - lb_endpoints:
            - endpoint:
                address:
                  socket_address:
                    address: limitador-limitador.kuadrant-system.svc.cluster.local
                    port_value: 8081
        name: kuadrant-rate-limiting-service
        type: STRICT_DNS
  workloadSelector:
    labels:
      app: istio-ingressgateway
      istio: ingressgateway
```
Deleting the RLP should also delete the EnvoyFilter object as there should not be RLP refs in the gateway

```sh
kubectl delete ratelimitpolicy/gw-rlp -n istio-system
```

Check for the Gateway's references to RLPs

```sh
kubectl get gateways istio-ingressgateway -n istio-system -o jsonpath='{.metadata.annotations.kuadrant\.io/ratelimitpolicies}' | yq e -P
```
It should output empty array
```yaml
[]
```

Check for the Limitador's Cluster EnvoyFilter resource

```sh
kubectl get envoyfilter kuadrant-ratelimiting-cluster-istio-ingressgateway -n istio-system -o yaml
```

It should `NotFound` error
```
Error from server (NotFound): envoyfilters.networking.istio.io "kuadrant-ratelimiting-cluster-istio-ingressgateway" not found
```


eguzki:
Looking for early review

guicassolato (commented):
Looking good, @eguzki! I know you'll probably add some tests still, but leaving my 👍 here since now 'cause I like the direction this is going. Thanks!

eguzki:
Test script for macos failing :shrug: 

```
Run make operator-sdk
./utils/install-operator-sdk.sh /Users/runner/work/kuadrant-operator/kuadrant-operator/bin/operator-sdk v1.28.1
Downloading https://github.com/operator-framework/operator-sdk/releases/download/v1.28.1/operator-sdk_darwin_amd64
gpg: directory '/Users/runner/.gnupg' created
gpg: keyserver receive failed: No route to host
make: *** [/Users/runner/work/kuadrant-operator/kuadrant-operator/bin/operator-sdk] Error 2
Error: Process completed with exit code 2.
```

`https://github.com/operator-framework/operator-sdk/releases/download/v1.28.1/operator-sdk_darwin_amd64` can be downloaded correctly

guicassolato:
> Test script for macos failing

It's been for a while, including in `main`.

I think the GHA fails to resolve `keyserver.ubuntu.com` from the [`gpg` command](https://github.com/Kuadrant/kuadrant-operator/blob/7dcbc3dabbe13d2092adeaf018b2aaccb0eee23a/utils/install-operator-sdk.sh#L19). We could try `keys.openpgp.org` instead, but I read somewhere that it could also be a problem of the Docker environment that runs the action, on macos.

controllers/limitador_cluster_envoyfilter_controller.go
```diff
@@ -0,0 +1,170 @@
...
+	kuadrantistioutils "github.com/kuadrant/kuadrant-operator/pkg/istio"
+	"github.com/kuadrant/kuadrant-operator/pkg/reconcilers"
+)
+
+// LimitadorClusterEnvoyFilterReconciler reconciles a EnvoyFilter object with limitador's cluster
+type LimitadorClusterEnvoyFilterReconciler struct {
+	*reconcilers.BaseReconciler
+}
+
+//+kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;update;patch;delete
```
guicassolato, line 47: I know it's redundant but do you think it's worth specifying here other permissions that this reconciler requires, 
but just happen to be granted by other kubebuilder markers at other controller files?

E.g.: 

```suggestion
//+kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;update;patch;delete
//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get;list;watch;update;patch
```

controllers/limitador_cluster_envoyfilter_controller.go
```diff
@@ -0,0 +1,170 @@
...
+		if apierrors.IsNotFound(err) {
+			logger.Info("no gateway found")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "failed to get gateway")
+		return ctrl.Result{}, err
+	}
+
+	if logger.V(1).Enabled() {
+		jsonData, err := json.MarshalIndent(gw, "", "  ")
```
guicassolato, line 67: Why `MarshalIndent`? I wonder whether it's suitable for when using structured logging, aka `production` log mode.

guicassolato (approved):
Left a couple comments, but nothing that would block the PR.
In general, excellent work, @eguzki!

controllers/limitador_cluster_envoyfilter_controller.go
eguzki: AFAIK, from the `go-logr` interface I can only check logger level, nothing about the format (defined by development or production mode). I see that this `MarshalIndent` call is not entirely correct as logger could be configured with structured format and debug level. By default (no `LOG_*` env vars), the logger is configured with `info` level and `production` level. When running with `make run`. it is `debug` and `development`. Thus, no issue with this so far. 

Any better approach you are considering? Not for this PR, but considering that (almost) all controllers have this code in the beginning of the `Reconcile` method, it would be nice to have all of them fixed.

guicassolato: I didn't mean checking here what the log mode setting is. I just assumed that log level and log mode were two separate things, like in [Authorino](https://docs.kuadrant.io/authorino/docs/user-guides/observability/#log-levels-and-log-modes). But it looks like log mode and log level are somehow coupled together.

IMO, it does not have to be like that.

In Authorino, all 4 combinations are allowed, i.e.:
- `level: info, mode: production`
- `level: debug, mode: production`
- `level: info, mode: development`
- `level: debug, mode: production`

(Maybe the naming is a bit unfortunate, because `production` and `development` mean actually "structured" and "human-readable", respectively.)

Because it felt weird printing indented JSON when log mode is set to `production`, I'd probably have used `json.Marshal` instead. Not a big deal though.

eguzki: This project's log system is exactly the same, with those 4 combinations allowed. 

> Because it felt weird printing indented JSON when log mode is set to production, I'd probably have used json.Marshal instead. Not a big deal though.

Check my previous comment out again. You are right. However, if I change to `Mashal`, when I run in debug + development mode when I run with `make run`, it is not human readable. I need to indent it :)

So looking for alternatives that makes a good dev experience as well.

Issue #237, opened 2023-08-24
dep: upgrade controller runtime


KevFan:
As a pre-req to https://github.com/Kuadrant/kuadrant-operator/issues/203, we should upgrade the controller runtime version
* https://github.com/kubernetes-sigs/controller-runtime/releases

The next minor versions of controller runtime introduces several breaking changes that would need to be addressed from this upgrade.

This should be completed after the upgrade to Go 1.20 https://github.com/Kuadrant/kuadrant-operator/issues/236 is complete

KevFan:
Controller runtime v0.16.1 is used as part of the gatway-api v0.8.0 release
- https://github.com/kubernetes-sigs/gateway-api/blob/78ed7340dfdb1def89a950d98c90905e872575d9/go.mod#L16

KevFan:
Gateway api v1.0.0 has been released using controller runtime v0.16.3
- https://github.com/kubernetes-sigs/gateway-api/blob/v1.0.0/go.mod#L17

PR #246, opened 2023-08-31
refactor: controller-runtime v0.16.3

KevFan:
Resolves: https://github.com/Kuadrant/kuadrant-operator/issues/237


KevFan:
Blocked / Kept as draft as decided to wait for MVP release before making these changes

main.go
```diff
@@ -100,23 +102,25 @@ func main() {
...
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
 	flag.Parse()
 
-	options := ctrl.Options{Scheme: scheme}
-
-	if configFile != "" {
-		options, err = options.AndFrom(ctrl.ConfigFile().AtPath(configFile))
```
KevFan, line 115: `ComponentConfig` has been deprecated (https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0). We can either use a custom implementation that sets Manager.Options directly or move back to configuration via arguments. 

Moving back to configuration using arguments as it's the easier option and we do not have many configurations in general

KevFan:
Gateway api v1.0.0 has been released using controller runtime v0.16.3
- https://github.com/kubernetes-sigs/gateway-api/blob/v1.0.0/go.mod#L17

adam-cattermole (approved):
These changes look good - I opted for a near identical implementation when I started to look at gwapi v1 and needed a newer controller-runtime. Nice work!

KevFan:
Rebased to resolve conflicts with main

main.go
```diff
@@ -100,23 +103,25 @@ func main() {
...
+		metricsAddr          string
+		enableLeaderElection bool
+		probeAddr            string
+		err                  error
 	)
-	flag.StringVar(&configFile, "config", "",
-		"The operator will load its initial configuration from this file. "+
-			"Omit this flag to use the default configuration values. "+
-			"Command-line flags override configuration from this file.")
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
```
eguzki, line 111: I wonder if the deployment of the operator should change after this change. Or defaults are good enough?

KevFan: I think these defaults are good. These are the same as https://github.com/Kuadrant/kuadrant-operator/commit/52b2cdb62e3594f39b353665428cf5531cce574c which was before it was moved to loading config using file.

Only difference I see between defaults and deployment options is that `leacterElection` is `true` only for deployment but `false` by default 🤔 :
- https://github.com/Kuadrant/kuadrant-operator/pull/246/files#diff-ef2df99e0c13ecb3c10ae6aa1923d7892fdccfa82e8f8777faebdaa4528c7568L1-L11
- https://github.com/Kuadrant/kuadrant-operator/pull/246/files#diff-319f8dc9adf6d477651de0cb44f85c968909535fc7c3a61be091a08b68298732R12

eguzki: But the deployment is setting it explicitly, so we should be good.

eguzki (approved):

Issue #244, opened 2023-08-31
Create new stable channel for "blessed" releases of kuadrant service protection


maleck13:
**What**

Rather than using just one channel for our images, ideally we would have the existing "alpha" channel for nightly / master builds and a separate channel for our "blessed" releases. This should also help when having blessed upgrad paths etc

PR #250, opened 2023-09-07
[workflow] Adding channels input for bundle creation

didierofrivia:
This PR addresses partially issue https://github.com/Kuadrant/kuadrant-operator/issues/244, it will need to be applied also in Authorino and Limitador operators.

The workflow dispatch input _channels_ accepts comma separated values, i.e.: "stable,release". The Makefile target could be executed as:

```sh
make bundle REGISTRY=quay.io ORG=kuadrant \
    VERSION=0.3.1 \
    AUTHORINO_OPERATOR_VERSION=0.7.0 \
    LIMITADOR_OPERATOR_VERSION=0.5.0 \
    WASM_SHIM_VERSION=0.2.0 \
    REPLACES_VERSION=0.3.0 \
    CHANNELS=stable,release
```

And it will generate, among other changes, the following:

```sh
diff --git a/bundle.Dockerfile b/bundle.Dockerfile
index d1171ce..c6b110f 100644
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
-LABEL operators.operatorframework.io.bundle.channels.v1=alpha
+LABEL operators.operatorframework.io.bundle.channels.v1=stable,release

diff --git a/bundle/metadata/annotations.yaml b/bundle/metadata/annotations.yaml
index 34b5dfc..3f734c1 100644
--- a/bundle/metadata/annotations.yaml
+++ b/bundle/metadata/annotations.yaml
@@ -4,7 +4,7 @@ annotations:
-  operators.operatorframework.io.bundle.channels.v1: alpha
+  operators.operatorframework.io.bundle.channels.v1: stable,release
```

Notes:
- **ONLY WORKS WITH A SINGLE CHANNEL for the catalog creation**
- Default value chosen is _preview_ because it's the default channel previously chosen in the catalog file generator https://github.com/Kuadrant/kuadrant-operator/blob/main/utils/generate-catalog.sh#L8-L9

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -170,7 +170,8 @@ jobs:
             AUTHORINO_OPERATOR_VERSION=${{ inputs.authorinoOperatorVersion }} \
             LIMITADOR_OPERATOR_VERSION=${{ inputs.limitadorOperatorVersion }} \
             WASM_SHIM_VERSION=${{ inputs.wasmShimVersion }} \
-            REPLACES_VERSION=${{ inputs.replacesVersion }}
+            REPLACES_VERSION=${{ inputs.replacesVersion }} \
+            CHANNELS=${{ inputs.channels }}
       - name: Install qemu dependency
         run: |
           sudo apt-get update
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -27,6 +27,10 @@ on:
         description: Kuadrant Operator replaced version
         default: 0.0.0-alpha
         type: string
+      channels:
+        description: Bundle and catalog channels, comma separated
+        default: preview
+        type: string
   workflow_dispatch:
     inputs:
       kuadrantOperatorTag:
@@ -52,6 +56,10 @@ on:
       replacesVersion:
         description: Kuadrant Operator replaced version
         default: 0.0.0-alpha
+        type: string
+      channels:
+        description: Bundle and catalog channels, comma separated
+        default: preview
         type: string
 
 env:
@@ -119,7 +127,8 @@ jobs:
             AUTHORINO_OPERATOR_VERSION=${{ inputs.authorinoOperatorVersion }} \
             LIMITADOR_OPERATOR_VERSION=${{ inputs.limitadorOperatorVersion }} \
             WASM_SHIM_VERSION=${{ inputs.wasmShimVersion }} \
-            REPLACES_VERSION=${{ inputs.replacesVersion }}
+            REPLACES_VERSION=${{ inputs.replacesVersion }} \
+            CHANNELS=${{ inputs.channels }}
       - name: Build Image
         id: build-image
         uses: redhat-actions/buildah-build@v2
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -20,12 +20,13 @@ $(CATALOG_FILE): $(OPM) $(YQ)
 	@echo REPLACES_VERSION               = $(REPLACES_VERSION)
 	@echo LIMITADOR_OPERATOR_BUNDLE_IMG  = $(LIMITADOR_OPERATOR_BUNDLE_IMG)
 	@echo AUTHORINO_OPERATOR_BUNDLE_IMG  = $(AUTHORINO_OPERATOR_BUNDLE_IMG)
+	@echo CHANNELS  					 = $(CHANNELS)
 	@echo "************************************************************"
 	@echo
 	@echo Please check this matches your expectations and override variables if needed.
 	@echo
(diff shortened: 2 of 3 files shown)


adam-cattermole (commented):
These changes here look good and the labels do get set on the created bundle. However do we also need to ensure that the channels are set appropriately in the catalog? I see that the catalog sets everything by default to `preview` https://github.com/Kuadrant/kuadrant-operator/blob/main/utils/generate-catalog.sh#L9.

If I create a bundle (setting the channels to `stable,release`) and `-build` and `-push`, then create a catalog from it I see the bundle channels still set to preview:
```
docker run --rm -it -p 50051:50051 quay.io/acatterm/kuadrant-operator-catalog:v0.3.1

grpcurl -plaintext localhost:50051 api.Registry/ListBundles | jq '{packageName: .packageName, name: .csvName, version: .version, bundlePath: .bundlePath, replaces: .replaces, skipRange: .skipRange, skips: .skips, channelName: .channelName}'
{
  "packageName": "authorino-operator",
  "name": "authorino-operator.v0.7.0",
  "version": "0.7.0",
  "bundlePath": "quay.io/kuadrant/authorino-operator-bundle:v0.7.0",
  "replaces": null,
  "skipRange": null,
  "skips": null,
  "channelName": "preview"
}
{
  "packageName": "kuadrant-operator",
  "name": "kuadrant-operator.v0.3.1",
  "version": "0.3.1",
  "bundlePath": "quay.io/acatterm/kuadrant-operator-bundle:v0.3.1",
  "replaces": "kuadrant-operator.v0.3.0",
  "skipRange": null,
  "skips": null,
  "channelName": "preview"
}
{
  "packageName": "limitador-operator",
  "name": "limitador-operator.v0.5.0",
  "version": "0.5.0",
  "bundlePath": "quay.io/kuadrant/limitador-operator-bundle:v0.5.0",
  "replaces": null,
  "skipRange": null,
  "skips": null,
  "channelName": "preview"
}
```

didierofrivia:
@adam-cattermole Thanks for the review and noting the catalog channel. In the [OLM docs](https://olm.operatorframework.io/docs/reference/file-based-catalogs/) for the catalog creation, they just mention in the `olm init` command that the flag `--default-channel`, which is the one receiving the constant it's mentioned in the description and also in your comment, that  _subscriptions will default to it if unspecified_... However it's interesting to see it's not changed in the generated catalog in your testing. We might need to also explicitly change it in the catalog generation

adam-cattermole:
One additional note - I added the catalog as a `CatalogSource` in an OpenShift cluster and during install from the catalog I witnessed the same behaviour - only one channel option; `preview`

adam-cattermole:
These changes look good and update the catalog to the specific channel. It's worth noting though that the comma-separated list of channels works for the labels on a bundle, but does not work for the file-based catalog and is treated as a string:

```
{
  "packageName": "authorino-operator",
  "name": "authorino-operator.v0.7.0",
  "version": "0.7.0",
  "bundlePath": "quay.io/kuadrant/authorino-operator-bundle:v0.7.0",
  "replaces": null,
  "skipRange": null,
  "skips": null,
  "channelName": "stable,release"
}
{
  "packageName": "kuadrant-operator",
  "name": "kuadrant-operator.v0.3.1",
  "version": "0.3.1",
  "bundlePath": "quay.io/acatterm/kuadrant-operator-bundle:v0.3.1",
  "replaces": "kuadrant-operator.v0.3.0",
  "skipRange": null,
  "skips": null,
  "channelName": "stable,release"
}
{
  "packageName": "limitador-operator",
  "name": "limitador-operator.v0.5.0",
  "version": "0.5.0",
  "bundlePath": "quay.io/kuadrant/limitador-operator-bundle:v0.5.0",
  "replaces": null,
  "skipRange": null,
  "skips": null,
  "channelName": "stable,release"
}

```

didierofrivia:
@adam-cattermole Yes, OPM basically picks the first one from the comma separated list to set the name of the channel. Either we create the different channels out of the list or we leave it as it is, what do you reckon?

didierofrivia:
At the time being, we might be working with a single channel at a time, maybe we could address the multiple channel subscriptions in the future (?)

adam-cattermole:
> @adam-cattermole Yes, OPM basically picks the first one from the comma separated list to set the name of the channel. Either we create the different channels out of the list or we leave it as it is, what do you reckon?

For the catalog it doesn't pick one - it's setting the channelName to be the entire string `stable,release` - see in the screenshot from OLM.

I agree though we could just have this work for single channel for now and worry about multi-channel later

adam-cattermole (approved):
I'll approve based on supporting single channel - nice work on these changes

Issue #248, opened 2023-08-31
Add `routeSelectors` fields to the AuthPolicy


alexsnaps:
- Simple route selectors
  - Top level conditions
  - Evaluator/config level conditions
- Add query string support as a when condition on the authconfig

PR #256, opened 2023-09-19
[authpolicy-v2] route selectors

guicassolato:
Closes:
- #248

### TODOs

- [x] Merge the hostnames of HTTPRoute (direct or inherited from the Gateway) into all Istio AuthorizationPolicy rules that do not include hostnames already built from the route selectors
- [x] Fix Authorino conditions built out of multiple HTTPRouteMatches → matches should be OR'ed together instead of AND'ed → requires https://github.com/Kuadrant/authorino/issues/424
- [x] Trigger reconciliation of Gateway-targeting policies after finishing reconciling HTTPRoute ones - to workaround AuthorizationPolicies and AuthConfigs generated under race condition
- [x] Implement or skip rule matches based on regexp types when building Authorino conditions and Istio AuthorizationPolicies rules
  - Authorino conditions for query string params from GW-API HTTPRouteMatches with `QueryParamMatchRegularExpression` type
  - Istio AuthorizationPolicies rules for path and headers from GW-API HTTPRouteMatches with `PathMatchRegularExpression` and `HeaderMatchRegularExpression` types respectively
- [ ] ~Report (in the logs, status, etc) about route selectors that match no HTTPRouteRule~ (separate PR)

#### Setup 

①  Create the cluster, install & deploy Kuadrant:

```sh
make local-setup
```

②  Request an instance of Kuadrant:

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

③  Deploy a sample API:

```sh
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/cars"
    - method: GET
      path:
        type: PathPrefix
        value: "/dolls"
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

(Optional) Test the API unprotected:

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/cars -i
# HTTP/1.1 200 OK
```

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/dolls -i
# HTTP/1.1 200 OK
```

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/admin -i
# HTTP/1.1 200 OK
```

#### Create AuthPolicies

④  Create a trivial AuthPolicy:

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
EOF
```

⑤  Create an AuthPolicy with auth rules:

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
EOF
```

⑥  Create an AuthPolicy with top-level route selectors:

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  routeSelectors:
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
EOF
```

⑦  Create an AuthPolicy with config route selectors:

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
        routeSelectors:
        - matches:
          - path:
              type: PathPrefix
              value: "/admin"
EOF
```

<br/>

After each step ④  →  ⑦  above,

❶  Check the state of the Istio AuthorizationPolicy:

```sh
kubectl get authorizationpolicy/on-istio-ingressgateway-using-toystore -n istio-system -o yaml
```

❷  Check the state of the Authorino AuthConfig:

```sh
kubectl get authconfig/ap-default-toystore -o yaml
```

❸  Send requests to the 3 endpoints of the API:

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/cars -i
```

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/dolls -i
```

```sh
curl -H 'Host: api.toystore.com' http://localhost:9080/admin -i
```


Boomatang (approved):
Verification steps work and nothing in the code is standing.

adam-cattermole (approved):
Initially I was not 100% on the intended behaviour of step 7 of the verification as it didn't seem entirely intuitive but after some more investigation and discussion with @Boomatang and @alexsnaps I'm happy that it's expected. Code all looks good as far as I can tell also.

controllers/authpolicy_controller.go
```diff
@@ -190,6 +198,24 @@ func (r *AuthPolicyReconciler) deleteNetworkResourceDirectBackReference(ctx cont
...
+		return err
+	}
+	mapper := HTTPRouteParentRefsEventMapper{
+		Logger: logger,
+		Client: r.Client(),
+	}
+	requests := mapper.MapToAuthPolicy(route)
+	for i := range requests {
+		request := requests[i]
+		go r.Reconcile(context.Background(), request)
```
maleck13, line 214: This seems odd. Are we not dropping errors and reconcile requests being returned from this this way? Should it be done as part of a Watches in the `NewControllerManagedBy` ?

We have something "like this " 

```
Watches(&source.Kind{Type: &workv1.ManifestWork{}}, handler.EnqueueRequestsFromMapFunc(func(o client.Object) []reconcile.Request {
			log.V(3).Info("enqueuing gateways based on manifest work change ", "work namespace", o.GetNamespace())
			requests := []reconcile.Request{}
			annotations := o.GetAnnotations()
			if annotations == nil {
				log.V(3).Info("no parent or anotations on manifest work ", "work ns", o.GetNamespace(), "name", o.GetName())
				return requests
			}
			key := annotations["kuadrant.io/parent"]
			ns, name, err := cache.SplitMetaNamespaceKey(key)
			if err != nil {
				log.Error(err, "failed to parse namespace and name from maifiest work")
				return requests
			}
			log.Info("requeuing gateway ", "namespace", ns, "name", name)
			requests = append(requests, reconcile.Request{
				NamespacedName: types.NamespacedName{Namespace: ns, Name: name},
			})
			return requests
		}), builder.OnlyMetadata).

```

maleck13 (commented):
I have given this a review as best I can without knowing the code base well. I have one question about how we are triggering a reconcile other than that it looks reasonable

controllers/authpolicy_controller.go
guicassolato: What kind of resource should we watch and map to AuthPolicies, @maleck13?

The intent here is, after reconciling an AuthPolicy attached to a HTTPRoute, to trigger the reconciliation of all AuthPolicies attached to Gateways that are parents of the HTTPRoute.

A few constraints are:
1. Neither HTTPRoute nor the Gateways are touched during any of those reconciliation loops. Only the internal custom resources might (i.e. `AuthorizationPolicy`, `AuthConfig`.)
2. We do not want to rely on back-ref annotations for this, as order matters here and therefore we want the latest state possible.

maleck13: Right. Prob wrong place to start figuring this out here. It would seem we need to trigger an event on the HTTPRoute (status update / annotation) or we need to track the parent policies. Do we have an issue for looking into a better option? I know it was discussed. IE have an in mem structure that we maintain of the policy hierarchy. Trigger a reconcile on parents on child change (or something along those lines) so I guess if we have an issue for that we could settle with this for now?

guicassolato: I'll make sure to open one and add it here for ref.

guicassolato: https://github.com/Kuadrant/architecture/issues/29

Issue #259, opened 2023-09-25
RateLimitPolicy Gateway needs to be created before Kuadrant resource


azgabur:
While testing new images I encountered following bug:
Error message in RateLimitPolicy status: `'Internal error occurred: object *v1beta1.Gateway is not Kuadrant managed'`

I created the gateway and HttpRoute after Kuadrant deployment, it is possible that is the problem.

Tested image/catalog-source: https://github.com/Kuadrant/kuadrant-operator/actions/runs/6286740056/job/17070796884
Istio version: 1.16.1
Openshift version: 4.12.32
I attach used gateway, httproute and ratelimitpolicy yaml:
[gateway.yaml.txt](https://github.com/Kuadrant/kuadrant-operator/files/12716641/gateway.yaml.txt)
[httproute.yaml.txt](https://github.com/Kuadrant/kuadrant-operator/files/12716642/httproute.yaml.txt)
[ratelimit.yaml.txt](https://github.com/Kuadrant/kuadrant-operator/files/12716643/ratelimit.yaml.txt)

Kuadrant operator log:
```
{"level":"error","ts":"2023-09-25T14:32:55Z","logger":"kuadrant-operator","msg":"Reconciler error","controller":"ratelimitpolicy","controllerGroup":"kuadrant.io","controllerKind":"RateLimitPolicy","RateLimitPolicy":{"name":"limit-azgabur-testste-ibqm","namespace":"authorino"},"namespace":"authorino","name":"limit-azgabur-testste-ibqm","reconcileID":"4ce1ba01-ac4b-4878-a0fa-9f94cffa40d1","error":"Internal error occurred: object *v1beta1.Gateway is not Kuadrant managed","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:329\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:274\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.14.6/pkg/internal/controller/controller.go:235"}
[...]

PR #260, opened 2023-09-27
kuadrant gateway controller to annotate gateways

eguzki:
### What

Fixes https://github.com/Kuadrant/kuadrant-operator/issues/259

Custom controller to add the kuadrant annotation to every new gateway deployed after Kuadrant CR has been created.

It does not reconcile the annotation value. If there is an annotation with the expected key `kuadrant.io/namespace`

Fixed [codecov](https://app.codecov.io) configuration.

```
❯ cat .github/codecov.yaml | curl --data-binary @- https://codecov.io/validate
Valid!

```

### Verification Steps

* Setup the environment:   
                         
```sh                    
make local-setup         
```                

Checking the annotations of the  `istio-ingressgateway`  gateway
```sh
❯ k get gateways istio-ingressgateway -n istio-system -o jsonpath='{.metadata.annotations}' | yq e 'keys' -P
```
It does not have any kuadrant namespace annotation

```yaml
- kubectl.kubernetes.io/last-applied-configuration
```

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```                              

The istio  `istio-ingressgateway`  gateway, should be annotated with the kuadrant annotation
```sh
❯ k get gateways istio-ingressgateway -n istio-system -o jsonpath='{.metadata.annotations}' | yq e 'keys' -P
```
```yaml
- kuadrant.io/namespace
- kubectl.kubernetes.io/last-applied-configuration
```

Create a new gateway

```
kind: Gateway
metadata:
  name: second
spec:
  gatewayClassName: istio
  listeners:
    - allowedRoutes:
        namespaces:
          from: All
      hostname: '*.example.com'
      name: api
      port: 80
      protocol: HTTP
EOF
```
Note that the gateway does not have kuadrant annotation

The new gateway  `second`  gateway, should be annotated with the kuadrant annotation
```sh
❯ k get gateways second -n kuadrant-system -o jsonpath='{.metadata.annotations}' | yq e 'keys' -P
```

```yaml
- kuadrant.io/namespace
- kubectl.kubernetes.io/last-applied-configuration
```

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -2,10 +2,17 @@
 codecov:
   bot: "Codecov Bot"
   max_report_age: 12
-  require_ci_to_pass: yes
+  require_ci_to_pass: "yes"
   notify:
     after_n_builds: 1
-    wait_for_ci: yes
+    wait_for_ci: "yes"
+  status:
+    project:
+      default:
+        threshold: 1%
+    patch:
+      default:
+        threshold: 1%
 
 # Layout of the PR comment produced by Codecov bot
 comment:
@@ -13,8 +20,8 @@ comment:
 
 # Find more at https://docs.codecov.com/docs/ignoring-paths
 ignore:
-  - api/external/** #  ignoring external vendor code
-  - "**/*.deepcopy.go" # ignore controller-gen generated code
+  - api/external/**  #  ignoring external vendor code
+  - "**/*.deepcopy.go"  # ignore controller-gen generated code
 
 flag_management:
   individual_flags:
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -2,10 +2,12 @@
 codecov:
   bot: "Codecov Bot"
   max_report_age: 12
-  require_ci_to_pass: "yes"
+  require_ci_to_pass: true
   notify:
     after_n_builds: 1
-    wait_for_ci: "yes"
+    wait_for_ci: true
+
+coverage:
   status:
     project:
       default:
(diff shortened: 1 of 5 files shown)


guicassolato (approved):

.github/codecov.yaml
```diff
@@ -2,19 +2,28 @@
...
   notify:
     after_n_builds: 1
-    wait_for_ci: yes
+    wait_for_ci: true
+
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1%
```
eguzki, line 14: cc @andrewdavidmackenzie 

Allow the coverage to drop by X%, and posting a success status.

eguzki: BTW new patch coverage raised up to [79.71%](https://app.codecov.io/gh/Kuadrant/kuadrant-operator/pull/260)

andrewdavidmackenzie: Sounds good. Variability in test sampling can cause the % to vary a little (you can also specify the number of digits of rounding to do before comparing BTW), causing it to drop a little (with same code and tests) and fail - so this guards against that.

andrewdavidmackenzie: e.g.
```
coverage:
  round: up
  precision: 0
```

eguzki: TIL: codecov [target](https://docs.codecov.com/docs/commit-status#target)
```
target

auto | <number>
Choose a minimum coverage ratio that the commit must meet to be considered a success.

    auto will use the coverage from the base commit (pull request base or parent commit) coverage to compare against.
    <number> you can specify a target of an exact coverage number such as 75% or 100% (string, int, or float accepted).
```

PR #275, opened 2023-10-16
[authpolicy-v2] docs

guicassolato:


doc/auth.md
```diff
@@ -0,0 +1,320 @@
+# Kuadrant Auth
+
+A Kuadrant AuthPolicy custom resource:
+
+1. Allows it to target Gateway API networking resources such as [HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute) and [Gateways](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway), using these resources to obtain additional context, i.e., which traffic workload (HTTP attributes, hostnames, user attributes, etc) to enforce auth.
+2. Allows to specify which specific subsets of the targeted network resource to apply the auth rules to.
+3. Abstracts the details of the underlying external authorization protocol and configuration resources, that have a much broader remit and surface area.
+4. Supports cluster operators to set overrides (soon) and defaults that govern what can be done at the lower levels.
```
maleck13, line 8: perhaps this should be reflected somewhere else? Wondering if we could have an issue label for AuthPolicy for example and link to the GH issues instead?

guicassolato: We have the same line in the RL docs: https://github.com/Kuadrant/kuadrant-operator/blob/ee2a4b36132ed597aeb47ff5eef354267a867451/doc/rate-limiting.md?plain=1#L8. This is basically our promise to support inherited policy attachment eventually.

doc/auth.md
```diff
@@ -0,0 +1,320 @@
+# Kuadrant Auth
+
+A Kuadrant AuthPolicy custom resource:
+
+1. Allows it to target Gateway API networking resources such as [HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute) and [Gateways](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway), using these resources to obtain additional context, i.e., which traffic workload (HTTP attributes, hostnames, user attributes, etc) to enforce auth.
+2. Allows to specify which specific subsets of the targeted network resource to apply the auth rules to.
```
Boomatang, line 6: The start of these points reads funny to me. I wonder if the suggested is better. 
```suggestion
1. Targets Gateway API networking resources such as [HTTPRoutes](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute) and [Gateways](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway), using these resources to obtain additional context, i.e., which traffic workload (HTTP attributes, hostnames, user attributes, etc) to enforce auth.
2. Specify which specific subsets of the targeted network resource to apply the auth rules to.
```

doc/auth.md
```diff
@@ -0,0 +1,320 @@
...
+
+#### Overview
+
+The `AuthPolicy` spec includes the following parts:
+
+* A reference to an existing Gateway API resource (`spec.targetRef`)
+* Authentication/authorization scheme (`spec.rules`)
+* Top-level route selectors (`spec.routeSelectors`)
+* Top-level additional conditions (`spec.when`)
+* List of stored named patterns (`spec.patterns`)
```
Boomatang, line 31: Why stored named patterns? Is there a way to list non stored patterns? 
```suggestion
* List of named patterns (`spec.patterns`)
```

doc/auth.md
maleck13: So I think it would still be good to call out that this supports cluster operators setting a default policy based on the Gateway hierarchy? Yes its not defaults overrides but is is a form of default WDYT?

doc/auth.md
```diff
@@ -0,0 +1,320 @@
...
+
+The auth scheme specify rules for:
+* Authentication (`spec.rules.authentication`)
+* External auth metadata fetching (`spec.rules.metadata`)
+* Authorization (`spec.rules.authorization`)
+* Custom response items (`spec.rules.response`)
+* Callbacks (`spec.rules.callbacks`)
+
+where each auth rule can declare specific `routeSelectors` and `when` conditions for the rule to apply.
```
Boomatang, line 40: Very surprised I spotted this.
```suggestion
Where each auth rule can declare specific `routeSelectors` and `when` conditions for the rule to apply.
```

doc/auth.md
```diff
@@ -0,0 +1,320 @@
...
+    callbacks:
+      "my-webhook":
+        http: {…}
+```
+
+Check out the [API reference](reference/authpolicy.md) for a full specification of the AuthPolicy CRD.
+
+## Using the AuthPolicy
+
+### Targeting a HTTPRoute networking resource
```
maleck13, line 147: Interesting I think I would have inverted these to follow the hierarchy. IE cover gateway first and HTTPRoute second?

doc/rate-limiting.md
```diff
@@ -226,36 +226,19 @@ spec:
 | `admin.toystore.com` | 250rps                                                       |
 | `other.toystore.com` | 5000rps                                                      |
 
-#### Route selectors
+### Route selectors
 
-The `routeSelectors` field of the limit definition allows to specify **selectors of routes** (or parts of a route), that _transitively induce a set of conditions for a limit to be enforced_. It is defined as a set of route matching rules, where these rules must exist, partially or identically stated within the HTTPRouteRules of the HTTPRoute that is targeted by the RLP.
+Route selectors allow to target sections of a HTTPRoute, by specifying sets of HTTPRouteMatches and/or hostnames that make the policy controller look up within the HTTPRoute spec for compatible declarations, and select the corresponding HTTPRouteRules and hostnames, to then build conditions that activate the policy or policy rule.
```
Boomatang, line 231: To me this reads funny at the start. I believe the suggestion reads better.
```suggestion
Route selectors allow targeting sections of a HTTPRoute, by specifying sets of HTTPRouteMatches and/or hostnames that make the policy controller look up within the HTTPRoute spec for compatible declarations, and select the corresponding HTTPRouteRules and hostnames, to then build conditions that activate the policy or policy rule.
```

Boomatang (changes requested):
I have added a couple of points. Most are minor and some are subjective.

maleck13 (commented):
Will leave it to others to approve. But these docs look pretty good and complete to me and are a great enhancement to what we on the website

doc/auth.md
guicassolato: I see targeting the HTTPRoute as the simplest case and therefore the one to explain first, but maybe that's because of how later we explain, when targeting a gateway, what happens in case there are also other policies targeting HTTPRoutes under that gateway. If we keep that example as-is, then the reader needs to know what targeting a HTTPRoute means beforehand.

Should we go then?

**OPTION 1 (current)**
- _Targeting a HTTPRoute_ – focus on the direct effect of the policy over that route, no mention about Gateway policies
- _Targeting a Gateway_ – explain how the policy affects all routes under the gateway EXCEPT the ones that have a more specific policy of its own

**OPTION 2**
- _Targeting a Gateway_ – explain how it affects all HTTPRoutes under the gateway, no mention about HTTPRoute policies
- _Targeting a HTTPRoute_ – explain how the policy directly affects that route AND replaces any other policy targeting gateways the route is under

doc/auth.md
guicassolato: The lower case was intentional.

I know this is not scientific writing, but it's a common pattern in texts where you have formalisation of concepts with blocks in between sentences.

The sentence that starts with "where" should be read in continuity to "The auth scheme specify rules for..."; otherwise it loses context.

But perhaps the following can please us both 🙂:

```suggestion
Each auth rule can declare specific `routeSelectors` and `when` conditions for the rule to apply.
```

doc/auth.md
guicassolato: I can cope with the suggestion for bullet point 1 (to be ported to https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/rate-limiting.md as well), but 2 needs to be clear that specific subsets is optional.

doc/reference/authpolicy.md
```diff
@@ -0,0 +1,151 @@
...
+| `unauthenticated` | [Custom denial status spec](https://docs.kuadrant.io/authorino/docs/features/#custom-denial-status-responseunauthenticated-and-responseunauthorized)               | No           | Customizations on the denial status and other HTTP attributes when the request is unauthenticated. (Default: `401 Unauthorized`)   |
+| `unauthorized`    | [Custom denial status spec](https://docs.kuadrant.io/authorino/docs/features/#custom-denial-status-responseunauthenticated-and-responseunauthorized)               | No           | Customizations on the denial status and other HTTP attributes when the request is unauthorized. (Default: `403 Forbidden`)         |
+| `success`         | [SuccessResponseSpec](#successresponsespec)                                                                                                                        | No           | Response items to be included in the auth response when the request is authenticated and authorized.                               |
+
+##### SuccessResponseSpec
```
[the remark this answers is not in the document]
guicassolato: `SuccessResponseSpec` (h4) is used within `ResponseSpec` (h3). So it's correct.

It's actually `SuccessResponseItem` that should be h5, because it's used only within `SuccessResponseSpec` (h4).

doc/auth.md
guicassolato: Changed a bit. Please let me know what you think.

doc/auth.md
maleck13: So I don't see it as a big deal and not a blocker to merge. But my preference is prob Option 2 as it (perhaps I am biased) fits how I think about gateway API resources

doc/auth.md
maleck13: +1

doc/reference/authpolicy.md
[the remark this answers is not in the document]
Boomatang: I will leave this up to you, if you want to change it. It did seem odd but that could also be how it was rendered in the rich diff.

doc/auth.md
Boomatang: If point 2 needs to be clear that specific subsets are optional I think both version are not clear. Possible might need to use the word optional in there.

Boomatang (approved):

PR #276, opened 2023-10-17
[authpolicy-v2] Move AuthPolicy top-level 'patterns' and 'when' fields one level up

guicassolato:
This will pair the level of these policy-wide options to the top-level 'routeSelectors', rather than having two things that have semantics over the same scope defined at different levels in the API.

This change also separates the auth scheme, making it now exclusively about auth rules.


pehala (approved):
I like this change, it definitely makes more sense at the top level.

PR #280, opened 2023-10-18
We need to update the GWs finalizers

alexsnaps:
I actually don't know that this change suffice, but I think it should... Unsure how to best test that _before_ merging.

See [here](https://kubernetes.slack.com/archives/C05J0D0V525/p1697646899353679), but we can discuss that tomorrow...

### DIFF
--- a/controllers/limitador_cluster_envoyfilter_controller.go
+++ b/controllers/limitador_cluster_envoyfilter_controller.go
@@ -46,6 +46,7 @@ type LimitadorClusterEnvoyFilterReconciler struct {
 
 //+kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
 
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -92,7 +92,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2023-10-18T09:13:46Z"
+    createdAt: "2023-10-19T10:15:14Z"
     operators.operatorframework.io/builder: operator-sdk-v1.28.1
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/Kuadrant/kuadrant-operator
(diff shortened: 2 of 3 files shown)


bundle/manifests/kuadrant.io_kuadrants.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.13.0
```
alexsnaps, line 5: Ugh

bundle/manifests/kuadrant.io_authpolicies.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.13.0
+    controller-gen.kubebuilder.io/version: v0.10.0
```
guicassolato, line 5: You may need to `rm -rf ./bin/controller-gen`, then `make generate manifests bundle` again.

Ref.: https://github.com/Kuadrant/kuadrant-operator/pull/277

controllers/limitador_cluster_envoyfilter_controller.go
```diff
@@ -46,6 +46,7 @@ type LimitadorClusterEnvoyFilterReconciler struct {
 
 //+kubebuilder:rbac:groups=networking.istio.io,resources=envoyfilters,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways,verbs=get
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gateways/finalizers,verbs=update
```
guicassolato, line 49: I don't get it. Is that because we create the EnvoyFilter CR? Our controller does not touch the gateway finalizer ever. Is there any path (perhaps through Istio?) that requires this permission? Effectively what it does is grating permission to _our_ `controller-manager` Service Account to update this kind of resource.

guicassolato (changes requested):

controllers/limitador_cluster_envoyfilter_controller.go
guicassolato: Hmmm. Nothing to do with Istio.

It looks like the issue is at this line: https://github.com/Kuadrant/kuadrant-operator/blob/982c978619675d912dd55b32fdda09ab823609ae/controllers/limitador_cluster_envoyfilter_controller.go#L155

And more concretely: https://github.com/Kuadrant/kuadrant-operator/blob/982c978619675d912dd55b32fdda09ab823609ae/pkg/reconcilers/base_reconciler.go#L189

alexsnaps: Yes, sorry, I should have pointed it out explicitly... That's indeed what I think is the issue in [OpenShift](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement)

guicassolato (approved):

alexsnaps:
Other than @eguzki telling me this is nonsense, I'll merge that in my afternoon 🙏

PR #282, opened 2023-10-25
docs: fix user guide authenticated rl for app devs based on authpolicy/v1beta2

guicassolato:


Boomatang (approved):
Changes work as expected

PR #287, opened 2023-11-02
Fix bug in response validation rules

adam-cattermole:
There is a bug where if the response field is present, but no response.success field in a submitted `AuthPolicy` it cannot be reconciled by the controller. The `Success` field is not a pointer so the validation evaluates to false when trying to add the finalizer.

The `exists` check for the routeselectors in `dynamicMetadata` and `headers` is also inverted.

The (second) `AuthPolicy` [in the user guide](https://docs.kuadrant.io/kuadrant-operator/doc/user-guides/auth-for-app-devs-and-platform-engineers/#4-create-a-default-deny-all-policy-at-the-level-of-the-gateway-persona-platform-engineer) can be used as validation - it fails to reconcile from main but works in this PR:

```yaml
kind: AuthPolicy
metadata:
  name: gw-auth
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    authorization:
      deny-all:
        opa:
          rego: "allow = false"
    response:
      unauthorized:
        headers:
          "content-type":
            value: application/json
        body:
          value: |
            {
              "error": "Forbidden",
            }
EOF

```


guicassolato (approved):
This is my bad and I'm sorry about it. We're gonna need to patch the latest release of the Kuadrant Operator with this fix.

Thanks, @adam-cattermole. Nice catch!

guicassolato:
Let's bring this to https://github.com/Kuadrant/kuadrant-operator/tree/release-v0.4.0 as well.

PR #292, opened 2023-11-07
Adding craft doc for RLP

Boomatang:
Add a more detail guide to for the sections of a RLP.

## Verification
As the change is to the docs and the main mode of render will be the docs.kuadrant.io there is a draft PR opened to the docs site which includes these changes. That PR can be found here https://github.com/Kuadrant/docs.kuadrant.io/pull/39

Expected:
* Guide explains all components in the RLP.
* Link in the guides are crafted for use on the site and may not work as expected if the Markdown file is view in isolation.

### DIFF
--- a/doc/craft/ratelimitpolicy.md
+++ b/doc/craft/ratelimitpolicy.md
@@ -1,0 +1,255 @@
+# Crafting A RateLimitPolicy
+A detail explanation of the fields in the RateLimitPolicy is outlined here.
+For a quick reference refer to [Reference](/kuadrant-operator/doc/reference/ratelimitpolicy/) section.
+
+## apiVersion and kind
+The latest version of the RateLimitPolicy is `kuadrant.io/v1beta2`.
+With the kind being `RateLimitPolicy`.
+```yaml
+apiVersion: kuadrant.io/v1beta2
+kind: RatelimitPplicy
+```
+
+## Metadata
+When creating the RateLimitPolicy the policy must be crated in the same namespace as the resource being targeted.
+```yaml
+metadata:
+    name: toystore
+    namespace: default
+```
+Once created Kuadrant may add and manage additional annotations and labels.
+For example:
+```yaml
+metadata:
+    annotations:
+        kuadrant.io/namespace: default
+```
+
+## Spec
+
+### limits
+`spec.limits` is a list of named limits, which describe what limits should be placed on a resource.
+Many named limits are allowed.
+Limit names do require to be valid yaml keys.
+```yaml
+spec:
+    limits:
+        limit_1:
+            ...
+        limit_2:
+            ...
+```
+
+#### counters
+`counters` defines additional rate limit counters based on context qualifiers and well known selectors.
+The full list of well known attributes can be found [here](/architecture/rfcs/0002-well-known-attributes/#request-attributes).
+All defined counters are ANDed together.
+```yaml
+spec: 
+    limits:
+        limit_1:
+            counters:
+                - request.url_path
+```
+The above example creates a new counter for every url_path that is requested.
+
+A more concrete example of why creating new counters base off the well known attributes may relate to comments on social network.
+It might be business policy to only allow users to update comments once pre day. 
+The example limit is defined below.
+Comments are updated at a URI like "/comment/<UUID\>".
+```yaml
+spec: 
+    limits:
+        comments_update:
+            counters:
+                - request.url_path
+            rates:
+                - duration: 1
+                  limit: 1
+                  unit: day
+            routeSelectors:
+                - hostnames:
+                    - 'api.toystore.com'
+                  matches:
+                    - method: POST
+                      path:
+                        type: PathPrefix
+                        value: /comment
+```
+
+#### rates
+`rates` defines the limit of requests allowed in a time period.
+This allows protecting resources over different time periods without the need for duplicating limits.
+A `rate` requires three fields: `duration`, `limit` and `unit`. 
+
+- `duration`: an integer value of how long the time period is, e.g. 10
+- `limit`: an integer value of how request are allowed in the time period, e.g. 5
+- `unit`: unit of time that the `duration` is over.
+Possible values are second, minute, hour and day.
+
+```yaml
+spec:
+    limits:
+        limit_1:
+            rates:
+                - duration: 10
+                  limit: 5
+                  unit: second
+                - duration: 1
+                  limit: 1500
+                  unit: hour
+```
+The above example restricts "limit_1" to 5 requests every 10 seconds or 1500 requests every hour.
+1500 requests an hour is more restrictive than 5 requests every 10 seconds.
+
+#### routeSelectors
+`routeSelectors` defines the semantics for matching an HTTP request based on conditions.
+Each `routeSelector` can contain a list of `hostnames` and `matches`.
+
+- `hostnames`: defined limits apply to the hostnames.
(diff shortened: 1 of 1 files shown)


doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
+# Crafting A RateLimitPolicy
+A detail explanation of the fields in the RateLimitPolicy is outlined here.
+For a quick reference refer to [Reference](/kuadrant-operator/doc/reference/ratelimitpolicy/) section.
```
guicassolato, line 3: Link is broken in GitHub.

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+apiVersion: kuadrant.io/v1beta2
+kind: RatelimitPplicy
+```
+
+## Metadata
+When creating the RateLimitPolicy the policy must be crated in the same namespace as the resource being targeted.
+```yaml
+metadata:
+    name: toystore
```
guicassolato, line 17: Any special reason to be using 4-space indentation? I think most (if not all) our examples use 2 spaces.

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+## apiVersion and kind
+The latest version of the RateLimitPolicy is `kuadrant.io/v1beta2`.
+With the kind being `RateLimitPolicy`.
+```yaml
+apiVersion: kuadrant.io/v1beta2
+kind: RatelimitPplicy
+```
+
+## Metadata
+When creating the RateLimitPolicy the policy must be crated in the same namespace as the resource being targeted.
```
guicassolato, line 14:
```suggestion
The RateLimitPolicy must be created in the same namespace as the resource being targeted.
```

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+metadata:
+    name: toystore
+    namespace: default
+```
+Once created Kuadrant may add and manage additional annotations and labels.
```
guicassolato, line 20:
```suggestion
Once created, Kuadrant may add additional annotations and labels to the policy.
```

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+metadata:
+    annotations:
+        kuadrant.io/namespace: default
+```
+
+## Spec
+
+### limits
+`spec.limits` is a list of named limits, which describe what limits should be placed on a resource.
```
guicassolato, line 31: What does "resource" mean here?

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+spec:
+    limits:
+        limit_1:
+            ...
+        limit_2:
+            ...
+```
+
+#### counters
+`counters` defines additional rate limit counters based on context qualifiers and well known selectors.
```
guicassolato, line 44: I wouldn't say the counters are "additional." These are the qualifiers that define _the_ counters that are created for each possible value assumed for the dynamic attributes specified here.

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+The full list of well known attributes can be found [here](/architecture/rfcs/0002-well-known-attributes/#request-attributes).
```
guicassolato, line 45: Broken link.

Also, pointing to one group of well-known attributes only (i.e. request attributes.) Attributes from the other groups also apply.

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+```yaml
+spec: 
+    limits:
+        limit_1:
+            counters:
+                - request.url_path
+```
+The above example creates a new counter for every url_path that is requested.
+
+A more concrete example of why creating new counters base off the well known attributes may relate to comments on social network.
```
guicassolato, line 56: Sorry. I don't understand this sentence.

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+                    - 'api.toystore.com'
+                    - 'admin.toystore.com'
+                    - '*.example.com'
+                  matches:
+                    - method: GET
+                      path:
+                        type: PathPrefix
+                        value: /
+```
+The above example selects an number of hostnames to protect, and the `matches` refines the selection to GET requests on any URI that is prefix with `/`.
```
guicassolato, line 130:
```suggestion
The above example selects a number of hostnames to protect, and the `matches` refines the selection to GET requests on any URI that is prefix with `/`.
```

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+    limits:
+        limit_1:
+            when:
+                - operator: "eq"
+                  selector: "request.useragent"
+                  value: "curl/8.0.1"
+```
+The above example will match HTTP requests that are sent with the header HTTP_USER_AGENT set with a value of "curl/8.0.1". 
+
+### targetRef
```
guicassolato, line 193: Maybe explain this before explaining `routeSelectors`?

doc/craft/ratelimitpolicy.md
```diff
@@ -0,0 +1,255 @@
...
+metadata:
+    name: toystore
+    namespace: default
+spec: 
+    limits:
+        toystore:
+            rates:
+                - duration: 10
+                  limit: 5
+                  unit: secound
```
guicassolato, line 223:
```suggestion
                  unit: second
```

guicassolato (changes requested):
I understand this is intended to be some RLP 101. However, IMHO, it overloads a bit with https://docs.kuadrant.io/kuadrant-operator/doc/rate-limiting; it adds new wording for fields and concepts more carefully defined in the [Reference](https://docs.kuadrant.io/kuadrant-operator/doc/reference/ratelimitpolicy/), as well as in https://docs.kuadrant.io/kuadrant-operator/doc/reference/route-selectors/.

I think bits of this doc could perhaps be extracted to enhance other existing docs – either the ones I linked here or some of the more basic user guides. But, honestly, I don't think we need it as-is as part of our documentation.

Not that the approach is not desirable. On the contrary! I'm sure first-time users could really use such a detailed explanation that slices down each block and level of the RLP. Maybe it could be turned into a blog post instead, with a context use case to drive the explanations?

doc/craft/ratelimitpolicy.md
Boomatang: No special reason, my editor has a default of 4 spaces. I always found it as a good guide to highlight deeply nested code block and the cyclomatic complexity that brings.

doc/craft/ratelimitpolicy.md
Boomatang: I explained the different sections in an alphabetically order as this is how a user would see items listed in a CR.  If this was to be ordered by the perceived importances should `routeSelectors` be listed before `counters`?

Boomatang:
@guicassolato and I had a good conversation about this change. We are in agreement that the information contained in the proposed doc is valuable but adding this information by this doc might not be the best way. The next step is to review this content with the aim of refactoring it into the existing documents (overview, reference and how-to guides).

Marking this PR as a draft for now.

PR #295, opened 2023-11-13
Add mandatory Gateway API label to the policy CRDs

### LINKED ISSUE #245: Add `gateway.networking.k8s.io/policy` label to the policy CRDs
As stated in the Gateway API [spec](https://gateway-api.sigs.k8s.io/geps/gep-713/#standard-label-on-crd-objects), each CRD that defines a policy object MUST include the standard label `gateway.networking.k8s.io/policy: inherited|direct`.

guicassolato:
- [x] Add mandatory Gateway API `gateway.networking.k8s.io/policy` label to the RateLimitPolicy CRD
- [x] Fixes mandatory Gateway API `gateway.networking.k8s.io/policy` label in the AuthPolicy CRD

Ref.: https://github.com/youngnick/gateway-api/blob/main/geps/gep-713.md#standard-label-on-crd-objects

Closes #245.

### DIFF
--- a/bundle/manifests/kuadrant.io_ratelimitpolicies.yaml
+++ b/bundle/manifests/kuadrant.io_ratelimitpolicies.yaml
@@ -6,7 +6,7 @@ metadata:
   creationTimestamp: null
   labels:
     app: kuadrant
-    gateway.networking.k8s.io/policy: inherited
+    gateway.networking.k8s.io/policy: direct
   name: ratelimitpolicies.kuadrant.io
 spec:
   group: kuadrant.io
--- a/bundle/manifests/kuadrant.io_ratelimitpolicies.yaml
+++ b/bundle/manifests/kuadrant.io_ratelimitpolicies.yaml
@@ -6,6 +6,7 @@ metadata:
   creationTimestamp: null
   labels:
     app: kuadrant
+    gateway.networking.k8s.io/policy: inherited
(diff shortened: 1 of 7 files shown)


bundle/manifests/kuadrant.io_ratelimitpolicies.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
   creationTimestamp: null
   labels:
     app: kuadrant
+    gateway.networking.k8s.io/policy: inherited
```
eguzki, line 9: `inherited`? Shouldn't be `direct` ?

guicassolato: Of course! My distraction! Fixing it now.

guicassolato: In fact, I've just noticed that the one in the AuthPolicy is also wrong!!! https://github.com/Kuadrant/kuadrant-operator/blob/3136ed961c704275338aa83ee6c9490efbd5590f/api/v1beta2/authpolicy_types.go#L202

guicassolato: Fixed both, and rebased.

eguzki (approved):

PR #303, opened 2023-11-16
Update istio to 1.20

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/297

This PR bumps the istio version in our Makefile targets in order to setup Kuadrant for local development. It's also an effort to check if our user guides and general usage is not broken due their new version. As a bonus, it fixes some user guides.

## User guides tried

- [x] [Simple Rate Limiting for Application Developers](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/simple-rl-for-app-developers.md) 👍🏼 
- [x] [Authenticated Rate Limiting for Application Developers](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-for-app-developers.md) 👍🏼 
- [x] [Gateway Rate Limiting for Cluster Operators](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/gateway-rl-for-cluster-operators.md) 👍🏼 
- [x] [Authenticated Rate Limiting with JWTs and Kubernetes RBAC](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-with-jwt-and-k8s-authnz.md)  ~The ACCESS_TOKEN retrieved from Keycloak is null~ Fixed by https://github.com/Kuadrant/kuadrant-operator/pull/311 👍🏼

PR #311, opened 2023-11-16
Update Keycloak examples

guicassolato:
Update AuthPolicy examples that use the latest version of Keycloak from https://github.com/Kuadrant/authorino-examples.

Follow-up on https://github.com/Kuadrant/authorino-examples/pull/35


alexsnaps (approved):
How is this even breaking the build?! 
I'm all for ignoring this, and merging!

guicassolato:
> How is this even breaking the build?!

It's not. It only fixes the failing 4th "user guide tried" of #303.

PR #313, opened 2023-11-17
Update bundle (policy-controller)

mikenairn:
Should [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/policy-controller/kustomization.yaml#L2 ) pull in a release version to avoid the `verify-bundle` failing randomly when we update anything in a different repo.

@maleck13 fyi


maleck13:
Yes agree. I have created a release-0.2 branch now and will update the kustomize to point at that

eguzki (approved):

PR #314, opened 2023-11-17
set the policy controller to the latest release branch

maleck13:
moves the kustomize to ref the release branch instead of main

### DIFF
--- a/config/policy-controller/kustomization.yaml
+++ b/config/policy-controller/kustomization.yaml
@@ -1,5 +1,5 @@
 resources:
-- github.com/Kuadrant/multicluster-gateway-controller/config/policy-controller/default?ref=main
+- github.com/Kuadrant/multicluster-gateway-controller/config/policy-controller/default?ref=release-0.2
 
 patchesStrategicMerge:
   - delete-ns.yaml
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -41,7 +41,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2023-11-17T10:07:35Z"
+    createdAt: "2023-11-17T10:46:25Z"
     operators.operatorframework.io/builder: operator-sdk-v1.28.1
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
     repository: https://github.com/Kuadrant/kuadrant-operator


config/policy-controller/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 resources:
-- github.com/Kuadrant/multicluster-gateway-controller/config/policy-controller/default?ref=main
+- github.com/Kuadrant/multicluster-gateway-controller/config/policy-controller/default?ref=release-0.2
```
didierofrivia, line 2: I'd say that these kind of version pointing in our manifests in main branch, should point to floating `latest` tag or `main` branch, and use the specific version for release branches. Does it make sense?

didierofrivia: on a second thought, if we want to treat the policy controller as an external dependencies, say, like istio, it's a different story

maleck13: issue with that is, any time something changes in the policy controller repo yamls, it will mean the verify-bundle will fail in this repo

guicassolato: Authorino is listed as a dependency and it points it to main: https://github.com/Kuadrant/kuadrant-operator/blob/61711f31372d2e72c28d5c0dceaa92ded8a00983/config/dependencies/authorino/kustomization.yaml#L2

guicassolato: In fact, I think the policy controller could be treated the same way. Why not having it under `config/dependencies`?

maleck13: so if something in authorino CRD changes then the bundle here also has to be regenerated? If that is the behaviour we would prefer I am ok with it, it had just seemed a bit of a pain rather than pinning to latest release

mikenairn: This really depends what you are trying to achieve. If you want the TLSPolicy and DNSPolicy CRDs and policy controller deployment included in the kuadrant-operator bundle, then adding them to config/dependencies probably won't work. AFAIK, none of those dependencies are pulled into the bundle and the likes of `kuadrant-operator/config/dependencies/authorino/kustomization.yaml` is only used where you want to run everything without OLM. In a normal OLM installation authorinio-operator is installed as an OLM dependency since it has it's own bundle, policy-controller doesn't.

guicassolato: Makes sense. Indeed the APIs defined by those dependencies are not part of Kuadrant Operator's bundle.

maleck13: Ok so only question is whether it should ref main or not. I am thinking not at least initially

guicassolato: About fixing a version or using latest, I think the logic behind it was that:
- components we control we default to latest – and therefore we pick up changes in those that affect Kuadrant Operator as soon as possible, whereas
- components we do not control we fix a version – and therefore ensure stability of Kuadrant Operator before externalities going on on those projects.

maleck13: Ok so then we don't change this one and we close this PR?

alexsnaps:
Closing then?

PR #317, opened 2023-11-20
Wasmplugin controller: new approach using Kuadrant Topology (DAG)

eguzki:
### What

Dedicated WASMPlugin controller

Description about the new approach using Kuadrant Topology.

#### Main object type
The type of object being *reconciled* is the `Gateway`. There is one-to-one mapping of gateway and wasmplugin. Another approach would be the type of object being reconciled being the WASMPlugin. Then it would be needed one way to get the gateway name from `{name, namespace}` of the wasmplugin. The event mappers would be more complex. Worth exploring anyway.

#### Watched resources and event mappers
The controller watches all resources that affect the configuration (spec) of the wasmplugin. I.e. HTTPRoutes, Gateways and RateLimitPolicies. On events of those resources, the event mapper implemented walks the Gateway API topology, via targetRef and parentRefs up to the peak of the topology, as far as kuadrant is concerned today, i.e. the Gateway. In other words, if you change a rate limit policy, it will generate events on all gateways. If you change a httproute, it will generate events on all gateways involved. Interesting to note that if you change the parentRef of a HTTProute, it will generate events on the gateways that **were** targeted and also on the new gateways currently being targeted. That allows doing cleaning for those gateways where a RLP is no longer affecting.

#### The Kuadrant Topology
On gateway reconciliation event,  the controller builds what it is named "The Kuadrant Topology". It reads all HTTPRoutes targeting the given gateway as well as **ALL** rate limit policies in the cluster. With all that info, it builds a set of indexes that can be queried to provide information to build the wasmplugin. Interesting to note here that a field indexer is being added to the manager to index all httproute parenting a gateway. Basically you can use this manager index (not the kuadrant topology index), to query "give me all the HTTPRoutes parenting Gateway A". 

The implementation assumes "cheap" operations of reading **ALL** the ratelimitpolicies and all routes parenting a given gateway. That assumption comes from the controller runtime k8s client that holds a cache. At least for `list` and `get` operations. That cache is being kept up to date  in the background when there are updates/deletes in the objects triggered by kuadrant controllers or any other third party controllers or any other k8s API server user.

TODO: Integration tests:
- [x] Bring some tests from `ratelimitpolicy_controller_test.go` to `rate_limiting_wasmplugin_controller_test.go`
- [x] No routes targeting the gateway
- [x] No routes selected from the HTTPRoute
- [x] HTTPRoute moves from one gateway to another.
- [x] RLP changes target from one route to another
- [x] RLP A targeting Gateway, 1 free route.  New RLP B targeting existing Route. The Wasmplugin should change. The RLP A should not add configuration to the WASMPlugin
- [x] New httproute without RLP: wasmplugin should change when there is RLP targeting the gateway

:left_right_arrow: Next :train: station: Limitador controller using Kuadrant Topology


guicassolato:
Relates to https://github.com/Kuadrant/architecture/issues/29

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+func (r *RateLimitingWASMPluginReconciler) gatewayAPITopologyFromGateway(ctx context.Context, gw *gatewayapiv1.Gateway) (*common.KuadrantTopology, error) {
+	logger, err := logr.FromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	routeList := &gatewayapiv1.HTTPRouteList{}
+	// Get all the routes having the gateway as parent
+	err = r.Client().List(ctx, routeList, client.MatchingFields{HTTPRouteParents: client.ObjectKeyFromObject(gw).String()})
```
guicassolato, line 191: This is very nice! I imagine the object kind is also taken into account, right?

guicassolato: Answering my own question... Yes, only `Gateway` parents: https://github.com/Kuadrant/kuadrant-operator/blob/2f943d0118d56bef6aa640f2802a2c2ec20ffd6d/controllers/rate_limiting_wasmplugin_controller.go#L309-L312

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+// addHTTPRouteByGatewayIndexer declares an index key that we can later use with the client as a pseudo-field name,
+// allowing to query all the routes parenting a given gateway
+// to prevent creating the same index field multiple times, the function is declared private to be
+// called ontly by this controller
+func addHTTPRouteByGatewayIndexer(mgr ctrl.Manager, logger logr.Logger) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &gatewayapiv1.HTTPRoute{}, HTTPRouteParents, func(rawObj client.Object) []string {
+		// grab the route object, extract the parents
+		route, assertionOk := rawObj.(*gatewayapiv1.HTTPRoute)
+		logger.Info("assertionOK", "ok", assertionOk)
```
guicassolato, line 299: For debugging?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
+	"github.com/kuadrant/kuadrant-operator/pkg/reconcilers"
+	"github.com/kuadrant/kuadrant-operator/pkg/rlptools"
+	"github.com/kuadrant/kuadrant-operator/pkg/rlptools/wasm"
+)
+
+const (
+	HTTPRouteParents = ".metadata.parent"
```
guicassolato, line 43: Maybe using a key name that makes it clear we only care about gateway parents?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+			continue
+		}
+
+		wasmPlugin.RateLimitPolicies = append(wasmPlugin.RateLimitPolicies, *wasmRLP)
+	}
+
+	return wasmPlugin, nil
+}
+
+func (r *RateLimitingWASMPluginReconciler) gatewayAPITopologyFromGateway(ctx context.Context, gw *gatewayapiv1.Gateway) (*common.KuadrantTopology, error) {
```
guicassolato, line 183: Wondering if this func could be enhanced for more kinds of policies, maybe using generics. The fact that it fetches a `RateLimitPolicyList`, which then it iterates through to cast each `RateLimitPolicy` to `KuadrantPolicy`, seems to be a minor detail. WDYT?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+	return &wasm.RateLimitPolicy{
+		Name:      client.ObjectKeyFromObject(rlp).String(),
+		Domain:    rlptools.LimitsNamespaceFromRLP(rlp),
+		Hostnames: common.HostnamesToStrings(hostnames), // we might be listing more hostnames than needed due to route selectors hostnames possibly being more restrictive
+		Service:   common.KuadrantRateLimitClusterName,
+		Rules:     rules,
+	}, nil
+}
+
+func (r *RateLimitingWASMPluginReconciler) RouteFromRLP(ctx context.Context, t *common.KuadrantTopology, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*gatewayapiv1.HTTPRoute, error) {
```
guicassolato, line 250: Could this func belong to the `KuadrantTopology` type? I imagine that extracting the scope HTTPRoute for an AuthPolicy would look like pretty much the same, but I could be missing something.

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+	return common.NewKuadrantTopology(
+		[]*gatewayapiv1.Gateway{gw},
+		common.Map(routeList.Items, func(r gatewayapiv1.HTTPRoute) *gatewayapiv1.HTTPRoute { return &r }),
+		common.Map(rlpList.Items, func(p kuadrantv1beta2.RateLimitPolicy) common.KuadrantPolicy { return &p }),
+	), nil
+}
+
+func (r *RateLimitingWASMPluginReconciler) WASMRateLimitPolicy(ctx context.Context, t *common.KuadrantTopology, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*wasm.RateLimitPolicy, error) {
+	gwHostnamesTmp := common.TargetHostnames(gw)
+	gwHostnames := common.Map(gwHostnamesTmp, func(str string) gatewayapiv1.Hostname { return gatewayapiv1.Hostname(str) })
```
guicassolato, line 215: This seems to be the reverse of https://github.com/Kuadrant/kuadrant-operator/blob/d4758585cfe4f17efd74207042991722ce06bc09/pkg/common/common.go#L187

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+		// no need to add the policy if there are no rules; a rlp can return no rules if all its limits fail to match any route rule
+		return nil, nil
+	}
+
+	// narrow the list of hostnames specified in the route so we don't generate wasm rules that only apply to other gateways
+	// this is a no-op for the gateway rlp
+	hostnames := common.FilterValidSubdomains(gwHostnames, route.Spec.Hostnames)
+	if len(hostnames) == 0 { // it should only happen when the route specifies no hostnames
+		hostnames = gwHostnames
+	}
```
guicassolato, line 239: Shouldn't the wasm rules be computed _after_ fixing the hostnames for rules that don't declare any?

guicassolato:
Sharing here a few thoughts to continue the conversation online:

1. How to decouple the DAG (aka “Kuadrant Topology”, and more internally “GWAPI Topology” there) from our specific kinds of policies, and make it generic enough for any kind of GWAPI policy – i.e. potentially targeting other kinds of network resource.

2. How do we share the cached topology across reconciliation events? Alternatively, how do we test controller-runtime is not rebuilding it at every reconciliation event?

3. Related to above, how do we respond to events that only update the topology without necessarily requiring any recomputing of the effective policy? Only to update the status of routes protected vs out of scope, for example.

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+	pluginConfig, err := r.wasmPluginConfig(ctx, gw)
+	if err != nil {
+		return nil, err
+	}
+
+	if pluginConfig == nil || len(pluginConfig.RateLimitPolicies) == 0 {
+		common.TagObjectToDelete(wasmPlugin)
+		return wasmPlugin, nil
+	}
```
KevFan, line 122: nit: `nil` wasmplugin seems to be only returned when `len(wasmPlugin.RateLimitPolicies) == 0` already at 

https://github.com/Kuadrant/kuadrant-operator/blob/e76cd28f7ee1633714a4e853f260e976f8e018f9/controllers/ratelimitpolicy_istio_wasmplugin.go#L231-L234

so I think there's no need to check this again here 🤔 

```suggestion
	if pluginConfig == nil {
		common.TagObjectToDelete(wasmPlugin)
		return wasmPlugin, nil
	}
```

controllers/rate_limiting_wasmplugin_controller.go
didierofrivia: I agree with Gui's comment, we could benefit from using generics and retrieving a `PolicyList`. I understand we will be requiring it heavily

pkg/common/kuadrant_topology.go
```diff
@@ -0,0 +1,385 @@
...
+		return client.ObjectKey{}
+	}
+
+	return client.ObjectKeyFromObject(r.Route)
+}
+
+type policyNode struct {
+	Policy          KuadrantPolicy
+	TargetedGateway *gatewayNode
+	TargetedRoute   *routeNode
```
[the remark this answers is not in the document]
didierofrivia: possibly using generics too

pkg/common/kuadrant_topology.go
```diff
@@ -0,0 +1,385 @@
...
+
+func (g *gatewayNode) ObjectKey() client.ObjectKey {
+	if g.Gateway == nil {
+		return client.ObjectKey{}
+	}
+
+	return client.ObjectKeyFromObject(g.Gateway)
+}
+
+type routeNode struct {
```
didierofrivia, line 263: Not sure if worth it or not, leaving it to you to decide or a future improvement. Maybe we could have an interface _networkNode_ and implement the different GW, routes, etc.. structs so the topology object could be "leaner" and not change much when we have to implement possible future nodes (grpcRoute, Service, etc)

controllers/rate_limiting_wasmplugin_controller.go
eguzki: This method logic is simple: when the policy is targeting a (valid) route, return that route. If the policy is targeting a (valid) gateway, compute an "imaginary" route made up from all the "free" routes attached to the gateway. 

I do not think that this logic belongs to the kuadrant topology type which is only, and nothing else than a representation or view of the cluster state. The logic above is the implementation of the current kuadrant's design of policies targeting a gateway and its semantics. 

I agree, though, that this `RouteFromRLP` method can be generalized to accomodate any kuadrant policy as long as kuadrant implements the same approach for any policy targeting a gateway. 

Let me stick with this implementation for now and let's refactor and generalize when the authpolicy controller leverages the kuadrant policy here presented.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: Yeah, I am aware there is a little mess around the type of the hostnames values. Some functions/APIs require `string` while other require `gatewayapiv1.Hostname`. 

In this particular case, The gateway holds `gatewayapiv1.Hostname`, but the method `TargetHostnames` converts them to `string`. The wasm plugin wants `string` but before that we need to process them and filter valid subdomains and the method ` FilterValidSubdomains` requires `gatewayapiv1.Hostname`. 

Too many back and forth conversions and copies. Unacceptable for a processing in the data plane. I guess not a big issue in the control plane.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: good catch. When any route selector had not empty `hostnames` fields, the routes without hostname were skipped. Added e2e test to cover that use case as well.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: I think this comment no longer applies to the current codebase in this PR. Let me know if it still does.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: changed

controllers/rate_limiting_wasmplugin_controller.go
eguzki: The current implementation of the "Kuadrant Topology" does not use a k8s client to query Gateway API and kuadrant policies. Instead, fetching resources must be done externally and then the kuadrant topology accepts list of resources. It could be done otherwise and have generic function with a k8s client to build topology for a given policy of one kind (for example RLP or KAP). The main reason for the current implementation is to decouple kuadrant topology and the cluster. Meaning that I can build kuadrant topology from a set of resources filtered with some criteria. The usecase used here is that I want to build a kuadrant topology only for one gateway and all policies and routes involved. But only for one gateway. 

Interesting. Worth exploring in future PRs.

controllers/rate_limiting_wasmplugin_controller.go
guicassolato: I was thinking something like `.metadata.gateway`, because the filter function added by `controllers.addHTTPRouteByGatewayIndexer` only cares about route `parentRefs` that are of kind `Gateway`. If one tries to use the filter with a `Service` route parent, even if the route declares such parent, it won't return the route.

eguzki:
# WIP (not finished yet)

> How to decouple the DAG (aka “Kuadrant Topology”, and more internally “GWAPI Topology” there) from our specific kinds of policies, and make it generic enough for any kind of GWAPI policy – i.e. potentially targeting other kinds of network resource.

Sure. That is the next logical step forward. The "Gateway API policy Topology". The topology only needs to know about the `targetRef` field. 

> How do we share the cached topology across reconciliation events? Alternatively, how do we test controller-runtime is not rebuilding it at every reconciliation event?

I have been working on this. TL;DR is fortunately *yes*, the resources are cached across reconciliation events and, even more, across controllers using the same manager. But we must be careful, as this could not be true if we change things here. 

1) The topology is built from resources read using the `Client` k8s client instead of `APIReader` client
2) The client is shared across controllers
3) The `Client` client is built using default options. 

To be clear, the topology is *not* cached, the resources used to build the topology (a set of indexes) are cached. The topology needs to be rebuild (the indexes are re-created) on each reconciliation event. I consider this procedure lightweight as long as the resources are cached and no I/O is involved. 

It's very clear from [controller-runtime doc](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client) that for "get" and "list" methods the interface implementation uses a cache. This, however, for the default options when building the client. The kuadrant operator uses (so far) the default options when building the client, via the manager:

https://github.com/Kuadrant/kuadrant-operator/blob/main/main.go#L127-L140

```go
	options := ctrl.Options{
		Scheme:                 scheme,
		Metrics:                metricsserver.Options{BindAddress: metricsAddr},
		WebhookServer:          webhook.NewServer(webhook.Options{Port: 9443}),
		HealthProbeBindAddress: probeAddr,
		LeaderElection:         enableLeaderElection,
		LeaderElectionID:       "f139389e.kuadrant.io",
	}

	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), options)
	if err != nil {
		setupLog.Error(err, "unable to start manager")
		os.Exit(1)
	}
```

I have tested the cache using the k8s auditing policy together with `kind` tool. https://kind.sigs.k8s.io/docs/user/auditing/

I have updated the `kind` configuration file to add APIServer logging based on audit policy

```diff
diff --git a/utils/kind-cluster.yaml b/utils/kind-cluster.yaml
index 4697238..4caa8f1 100644
--- a/utils/kind-cluster.yaml
+++ b/utils/kind-cluster.yaml
@@ -4,3 +4,28 @@ apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
   image: kindest/node:v1.27.3
+  kubeadmConfigPatches:
+  - |
+    kind: ClusterConfiguration
+    apiServer:
+      # enable auditing flags on the API server
+      extraArgs:
+        audit-log-path: /var/log/kubernetes/kube-apiserver-audit.log
+        audit-policy-file: /etc/kubernetes/policies/audit-policy.yaml
+      # mount new files / directories on the control plane
+      extraVolumes:
+        - name: audit-policies
+          hostPath: /etc/kubernetes/policies
+          mountPath: /etc/kubernetes/policies
+          readOnly: true
+          pathType: "DirectoryOrCreate"
+        - name: "audit-logs"
+          hostPath: "/var/log/kubernetes"
+          mountPath: "/var/log/kubernetes"
+          readOnly: false
+          pathType: DirectoryOrCreate
+  # mount the local file on the control plane
+  extraMounts:
+  - hostPath: utils/audit-policy.yaml
+    containerPath: /etc/kubernetes/policies/audit-policy.yaml
+    readOnly: true
```

And then created the audit policy to log only metadata about request for HTTPRoutes, Gateways and kuadrant policies. I also added one more filter to be the user being the kuadrant operator service account. This way the kubectl and k9s requests are not being logged. 

```yaml
cat <<EOF > utils/audit-policy.yaml 
---
apiVersion: audit.k8s.io/v1
kind: Policy
rules:
- level: Metadata
  users:
    - "system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager"
  resources:
    - group: "gateway.networking.k8s.io"
      resources:
        - httproutes
        - gateways
    - group: "kuadrant.io"
      resources:
        - ratelimitpolicies
        - authpolicies
EOF
```
Then, follow the guide for [Simple Rate Limiting for Application Developers](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/simple-rl-for-app-developers.md)

controllers/rate_limiting_wasmplugin_controller.go
guicassolato: I guess I'm all right with the `KuadrantTopology` being just a data structure, decoupled from the mechanisms to communicate with the cluster (e.g. API client.)

I still don't see this function as belonging to the WasmPlugin reconciler nevertheless. Despite the RLP kind hard-coded in the core of the function's (current) implementation, it's surrounded by generic code and, in the end, returns something that is not RLP-specific. It feels like the specific policy kind should be a parameter of the function.

controllers/rate_limiting_wasmplugin_controller.go
guicassolato: I'd argue the function _already_ is generic enough for all kinds of Kuadrant policies. I think it's literally only the signature that makes it RLP-specific now, no?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,386 @@
...
+
+	rateLimitPolicies := gatewayAPITopology.PoliciesFromGateway(gw)
+
+	logger.V(1).Info("wasmPluginConfig", "#RLPS", len(rateLimitPolicies))
+
+	// Shallow copy by assignment
+	rateLimitPoliciesSorted := rateLimitPolicies
+
+	// Sort RLPs for consistent comparison with existing objects
+	sort.Sort(common.PolicyByKey(rateLimitPoliciesSorted))
```
guicassolato, line 171: This is interesting. It could be a breaking change (depending on the behaviour implemented by the wasm shim.)

We definitely need to order the policies in the config. That's for sure! However, in the previous implementation, the order was dictated by the how the policy refs were stored in the gateway annotation. This was closer to the order of creation of the policies than alphabetical order on the qualified name of the policy.

If, in the wasm shim, the order of the policies in the config affects (today or in the future) the behaviour in the data plane, then depending what name one chooses for one's RLP/RLP namespace, users may observe different behaviours. To avoid that, I would stick with the creation timestamp for sorting the policies instead.

pkg/common/gatewayapi_event_mappers.go
```diff
@@ -0,0 +1,49 @@
...
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+// HTTPRouteToParentGatewaysEventMapper is an EventHandler that maps HTTPRoute events to gateway events,
+// by going through the parentRefs of the route
+type HTTPRouteToParentGatewaysEventMapper struct {
+	Logger logr.Logger
+}
+
+func (m *HTTPRouteToParentGatewaysEventMapper) Map(_ context.Context, obj client.Object) []reconcile.Request {
```
guicassolato, line 20: Except for the type it returns (`[]reconcile.Request` vs `[]string`), isn't this function practically the same as the one defined by `controllers.addHTTPRouteByGatewayIndexer` for the list filtering?

pkg/common/gatewayapi_utils.go
```diff
@@ -17,7 +17,9 @@ import (
 	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 )
 
-const GatewayProgrammedConditionType = "Programmed"
+const (
+	GatewayProgrammedConditionType = "Programmed"
+)
```
guicassolato, line 22: Linter? I actually prefer the one-line version. Am I the only the one?

pkg/common/kuadrant_policy_to_gateway_eventmapper.go
```diff
@@ -0,0 +1,77 @@
...
+				logger.V(1).Info("no route found", "route", routeKey)
+				return []reconcile.Request{}
+			}
+			logger.Error(err, "failed to get target", "route", routeKey)
+			return []reconcile.Request{}
+		}
+
+		requests := make([]reconcile.Request, 0)
+
+		for _, parentRef := range route.Spec.ParentRefs {
```
guicassolato, line 56: Same filter again? https://github.com/Kuadrant/kuadrant-operator/pull/317/files#r1502498973

pkg/common/kuadrant_topology.go
```diff
@@ -0,0 +1,385 @@
...
+	gatewayPolicies map[client.ObjectKey][]KuadrantPolicy
+
+	// policyRoute is an index of policies mapping to HTTPRoutes
+	// The index only includes policies targeting only existing and accepted (by parent gateways) HTTPRoutes
+	// Type: Policy -> HTTPRoute
+	policyRoute map[client.ObjectKey]*gatewayapiv1.HTTPRoute
+
+	// freeRoutes is an index of gateways mapping to HTTPRoutes not targeted by a kuadrant policy
+	// Gateway -> []HTTPRoute
+	freeRoutes map[client.ObjectKey][]*gatewayapiv1.HTTPRoute
```
guicassolato, line 27:
```suggestion
	untargetedRoutes map[client.ObjectKey][]*gatewayapiv1.HTTPRoute
```

?

pkg/common/kuadrant_topology.go
```diff
@@ -0,0 +1,385 @@
...
+	if g.Gateway == nil {
+		return client.ObjectKey{}
+	}
+
+	return client.ObjectKeyFromObject(g.Gateway)
+}
+
+type routeNode struct {
+	Parents      map[client.ObjectKey]*gatewayNode
+	DirectPolicy *policyNode
```
guicassolato, line 265: Not that I have a better suggestion for it, but I can see the use of "Direct Policy" becoming confusing, especially in light of the new defs proposed by https://github.com/kubernetes-sigs/gateway-api/pull/2813.

Technically, Kuadrant RLPs and APs are always "Inherited Policies" now, I reckon.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: The wasm shim applies the policy with the longest match of the hostname. 

Ok, After reading the [Gateway API spec](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.HTTPRouteSpec), it seems that in the event of multiple policies with exact hostnames, 

```
* The oldest Route based on creation timestamp.
* The Route appearing first in alphabetical order by “{namespace}/{name}”.
```

I will follow the same approach.

pkg/common/gatewayapi_utils.go
eguzki: Just a matter of personal taste. 

I would rather go for grouping several constant declarations, even if it's only one. If another constant needs to be added, the git diff is cleaner.

pkg/common/kuadrant_policy_to_gateway_eventmapper.go
eguzki: will be refactored to avoid DRY

pkg/common/kuadrant_topology.go
eguzki: After reviewing the implementation of the topolocy and trying to run away from "kuadrant design decisions" to Gateway API topology with policies included, I think we do not need that "direct policy" field in either `routeNode` and `gatewayNode`. 

I will give it a try and come back to this later.

pkg/common/kuadrant_topology.go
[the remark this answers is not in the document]
eguzki: this is internal implementation is going to change.

controllers/rate_limiting_wasmplugin_controller.go
eguzki: done, ready for review

eguzki:
superseded by #447

PR #301, opened 2023-11-15
Enhanced observability for the limitador instance

eguzki:
### What

Fixes #296 

- [x] Upgrade Limitador type to v0.6.0
- [x] Upgrade Limitador type to v0.7.0
- [x] Limitador deployment with rateLimitHeaders set to `DRAFT_VERSION_03`
- [x] Limitador: enable the exhaustive metrics

PR #318, opened 2023-11-20
Limitador-oparetor version update

Boomatang:
Bumping the api version of limitador-operator from v0.4.0 to v0.6.0. This allows use of the new CR types if need be.

# Verification
* Test should pass.


eguzki:
Overlap here https://github.com/Kuadrant/kuadrant-operator/pull/301

alexsnaps:
Dup

PR #322, opened 2023-11-21
docs: TLS and DNS Policy user guides

mikenairn:
Add user guide documentation for TLS and DNS Policy focused on the single cluster context and using the simple routing strategy for DNS.

To support the DNS and TLS policy guides, updates were made to install and configure, metallb and the policy controller:

* Add script to generate MetalLB IPAddressPool for docker network nad execute as part of `make install-metallb` 
```
./utils/docker-network-ipaddresspool.sh <docker network name> | kubectl apply -n metallb-system -f -
```
* Expose port 80 and port 443 of the kind cluster so you can access it locally

### DIFF
--- a/doc/user-guides/gateway-dns.md
+++ b/doc/user-guides/gateway-dns.md
@@ -1,0 +1,224 @@
+# Gateway DNS for Cluster Operators
+
+This user guide walks you through an example of how to configure DNS for all routes attached to an ingress gateway.
+
+<br/>
+
+## Requisites
+
+- [Docker](https://docker.io)
+- [Rout53 Hosted Zone](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/CreatingHostedZone.html)
+
+### Setup
+
+This step uses tooling from the Kuadrant Operator component to create a containerized Kubernetes server locally using [Kind](https://kind.sigs.k8s.io),
+where it installs Istio, Kubernetes Gateway API and Kuadrant itself.
+
+Clone the project:
+
+```shell
+git clone https://github.com/Kuadrant/kuadrant-operator && cd kuadrant-operator
+```
+
+Setup the environment:
+
+```shell
+make local-setup
+```
+
+Install metallb:
+```shell
+make install-metallb
+```
+
+Fetch the current kind networks subnet:
+```shell
+docker network inspect kind -f '{{ (index .IPAM.Config 0).Subnet }}'
+```
+Response:
+```shell
+"172.18.0.0/16"
+```
+
+Create IPAddressPool within kind network(Fetched by the command above) e.g. 172.18.200
+```shell
+kubectl -n metallb-system apply -f -<<EOF
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: kuadrant-local
+spec:
+  addresses:
+  - 172.18.200.0/24
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: empty
+EOF
+```
+
+Create a namespace:
+```shell
+kubectl create namespace my-gateways
+```
+
+Export a root domain and hosted zone id:
+```shell
+export ROOT_DOMAIN=<ROOT_DOMAIN>
+export AWS_HOSTED_ZONE_ID=<AWS_HOSTED_ZONE_ID>
+```
+
+> **Note:** ROOT_DOMAIN and AWS_HOSTED_ZONE_ID should be set to your AWS hosted zone *name* and *id* respectively.
+
+### Create a ManagedZone
+
+Create AWS credentials secret
+```shell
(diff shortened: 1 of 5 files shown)


doc/user-guides/gateway-dns.md
```diff
@@ -0,0 +1,224 @@
...
+spec:
+  addresses:
+  - 172.18.200.0/24
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: empty
+EOF
+```
```
mikenairn, line 59: I think this would be better scripted and part of the default local setup (`make local-setup`) I'd say most user guides, docs, demos etc.. would be better for it since you won't need to do stuff like this https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/gateway-rl-for-cluster-operators.md?plain=1#L50 and forward ports to your host. 

@guicassolato @eguzki Any thoughts on that? Don't mind doing it, just don't want to waste my time if there is a reason not to.

utils/kind-cluster.yaml
```diff
@@ -4,9 +4,14 @@ apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
   image: kindest/node:v1.27.3
-  # port forward 80 on the host to 80 on this node
   extraPortMappings:
     - containerPort: 30950
       hostPort: 9080
     - containerPort: 30951
       hostPort: 9443
```
mikenairn, line 11: If we decide to use metalb in the default setup we can probably drop these ports and use 9080 -> 80, 9443 -> 443

doc/user-guides/gateway-dns.md
maleck13: do we do this in MGC also? If not seems it would be useful there too?

doc/user-guides/gateway-dns.md
```diff
@@ -0,0 +1,224 @@
...
+```shell
+kubectl get managedzones -n my-gateways
+```
+
+### Create an ingress gateway
+
+Create a gateway using your ROOT_DOMAIN as part of a listener hostname:
+```sh
+kubectl -n my-gateways apply -f - <<EOF
+apiVersion: gateway.networking.k8s.io/v1beta1
```
maleck13, line 113:
```suggestion
apiVersion: gateway.networking.k8s.io/v1
```

doc/user-guides/gateway-tls.md
```diff
@@ -0,0 +1,209 @@
...
+spec:
+  addresses:
+  - 172.18.200.0/24
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: empty
+EOF
+```
```
maleck13, line 58: should we move this into a different file and set it as a pre-req? That way it wont be repeated and also when we have a quick start we can just update these docs to remove this setup piece

maleck13 (commented):
docs look ok, I tihnk it would be good to pull out the common setup stuff and keep each doc just focused on the task, that way if you move from one to the other you can just get on with it. 
Also we need to change the ref version in Gateways to be v1

doc/user-guides/gateway-tls.md
mikenairn: Meh, i think these docs just listing all steps is better than docs linking to other docs expecting you to run all those steps before continuing steps in this doc. I'd rather just script this tbh, can just be done as part of `make install-metallb` initially and decide if you want to make it part of `make local-setup` later, wdyt?

doc/user-guides/gateway-dns.md
mikenairn: We, for reasons i don't know or can't remember, create an entirely different docker network in MGC and so know what the subnet is and create the IPAddressPool with that. This is more like what the kind docs tell you to do with the existing kind docker network https://kind.sigs.k8s.io/docs/user/loadbalancer/

@sergioifg94 Do you know if there was a reason we created a different docker network rather than just using the one that was already being created?

doc/user-guides/gateway-tls.md
maleck13: Yeah I think adding it to make install-metallb is a good call

doc/user-guides/gateway-dns.md
sergioifg94: Yes, the reason for it is that if you don't create one and let it automatically create, it it will assign the next free address, so the addresses could differ depending on what local networks people might already have. Creating one beforehand ensures that the addresses are consistent for everyone

maleck13:
@mikenairn the issue is with the 8080 port in the tests

mikenairn:
> @mikenairn the issue is with the 8080 port in the tests

Yeah sure, 8080 was probably a bad choice

maleck13 (commented):
went through all the steps. Worked as expected. Once we have a single cluster quick start then we prob want to drop some of these steps but this is good for now.

maleck13 (approved):

Issue #299, opened 2023-11-14
Look into OSSM 3.x


alexsnaps:
https://github.com/maistra/istio-operator/tree/maistra-3.0

This should (eventually?) install Istio 1.20 (and GwAPI 1.0)... would kuadrant work? Are we missing something?

adam-cattermole:
Yes it did

PR #323, opened 2023-11-21
Add support for meshConfig from Istio CR

adam-cattermole:
#### Changes
- There's an `Istio` CR that provides the `meshConfig` instead of an `IstioOperator` CR for the sail operator
	- Added new `IstioWrapper` for updating the `meshConfig` stored in the `Istio` CR
	- We check for `IstioOperator` presence first, then `Istio`, then fall back to OSSM 2.x

#### TODO:
* To use the `Istio` type I've included the maistra operator as a go dep but I think I'm leaning toward duplicating the types as done prior in our https://github.com/Kuadrant/kuadrant-operator/tree/main/api/external/maistra folder rather than including all of these dep changes in the go.mod
* I might investigate the option of using sail to install istio for local development rather than using `istioctl`


alexsnaps:
@adam-cattermole is that an attempt to fix #299 ?

adam-cattermole:
> @adam-cattermole is that an attempt to fix #299 ?

Yes it is, still a WIP though

adam-cattermole:
I'm still looking at the other points in the to do section but given they could be follow ups I've marked ready for review

adam-cattermole:
> @adam-cattermole I am a fan of using project sail in regular development but I think we need to be careful to ensure we are still staying compatible with "regular" istio installs

I've left the original install options in the Makefile but added a commit moving to installing project sail locally. Will need an additional change to the docs as there is a minor difference in how we access the deployed application.

I'll make a change to have the integration tests run on each istio install type - that might be enough to make sure we aren't breaking anything on each PR.

config/dependencies/gateway-api/gateway/gateway.yaml
```diff
@@ -8,12 +8,9 @@ metadata:
 spec:
   gatewayClassName: istio
   listeners:
-  - name: default
```
maleck13, line 11: I don't mind the name change, but any specific reason to change it?

config/dependencies/gateway-api/gateway/gateway.yaml
```diff
@@ -8,12 +8,9 @@ metadata:
...
+  - name: http
     port: 80
     protocol: HTTP
     allowedRoutes:
       namespaces:
         from: All
-  addresses:
```
maleck13, line 17: glad we can remove this,but does it cause any issues with other user-guides?

config/dependencies/istio/sail/istio.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: istio-system
+spec:
+  version: v1.20.0
+  values:
+    pilot:
+      autoscaleEnabled: false
+  rawValues:
+    gateways:
+      istio-ingressgateway:
+        autoscaleEnabled: false
```
maleck13, line 14: be good to get a couple of comments as to what it does or why its needed?

config/dependencies/gateway-api/gateway/gateway.yaml
adam-cattermole: This was just to match the naming when retrieving the port from the gateway as used in the [istio docs](https://istio.io/latest/docs/examples/bookinfo/#determine-the-ingress-ip-and-port) - no real reason though we can leave it as default and I can update the exports in each of the guides

config/dependencies/gateway-api/gateway/gateway.yaml
adam-cattermole: On a sail install this part of the spec gets populated when it creates the loadbalancer service so it should be present anyway. I'll double check that I didn't miss any user guides though.

pkg/istio/mesh_config_test.go
```diff
@@ -199,3 +201,42 @@ func TestOSSMControlPlaneWrapper_SetMeshConfig(t *testing.T) {
...
 }
+
+func TestSailWrapper_GetConfigObject(t *testing.T) {
+	ist := &istiov1alpha1.Istio{}
+	wrapper := NewSailWrapper(ist)
+
+	assert.Equal(t, wrapper.GetConfigObject(), ist)
+}
+
+func TestSailWrapper_GetMeshConfig(t *testing.T) {
```
maleck13, line 212: do we want to cover the error cases?

maleck13 (commented):
Couple of small comments but the changes look reasonable to me. I don't want to approve this as I don't work in the kuadrant-operator enough. Will leave that to @eguzki @guicassolato or @alexsnaps 

I will try out the verification steps next

config/dependencies/istio/sail/istio.yaml
adam-cattermole: This was just to match our original local deploy configuration but isn't necessary and could be removed https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/istio/istio-operator.yaml#L48-L54

pkg/istio/mesh_config_test.go
```diff
@@ -199,3 +209,59 @@ func TestOSSMControlPlaneWrapper_SetMeshConfig(t *testing.T) {
...
+		assert.NilError(t, err)
+	}
+	config := &istiov1alpha1.Istio{}
+	if err := config.Spec.SetValues(values); err != nil {
+		assert.NilError(t, err)
+	}
+	wrapper := NewSailWrapper(config)
+
+	meshConfig, err := wrapper.GetMeshConfig()
+	assert.Error(t, err, "proto: (line 1:2): unknown field \"invalid-field\"")
```
adam-cattermole, line 251: I'll change this test.. https://github.com/protocolbuffers/protobuf-go/blob/master/internal/errors/errors.go#L27-L28

make/istio.mk
```diff
@@ -19,15 +26,32 @@ $(ISTIOCTL):
...
-.PHONY: istio-verify-install
-istio-verify-install: istioctl ## Verify istio installation.
+.PHONY: istioctl-verify-install
+istioctl-verify-install: istioctl ## Verify istio installation.
 	$(ISTIOCTL) verify-install -i $(ISTIO_NAMESPACE)
+
+.PHONY: sail-install
+sail-install: kustomize
+	$(MAKE) install-metallb
+	$(KUSTOMIZE) build $(ISTIO_INSTALL_DIR)/sail | kubectl apply -f -
```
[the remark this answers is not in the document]
adam-cattermole: I see your suggestion, but given we use this for all of our kustomize uses elsewhere I think it would be strange to have this single occurrence use `kubectl apply -k`. Perhaps we should have a follow up issue to replace all `$(KUSTOMIZE) build .. | kubectl apply -f` if we think that makes sense?

maleck13:
I was able to work with this in kubernetes. https://gist.github.com/maleck13/eb5860181404f055ac5ec64d81945a4f @alexsnaps @adam-cattermole this ready to merge?

adam-cattermole:
@maleck13 I'm happy with the changes but I think we could use an approving review from service protection

controllers/kuadrant_controller.go
```diff
@@ -320,23 +322,32 @@ func (r *KuadrantReconciler) getIstioConfigObjects(ctx context.Context, logger l
...
+	istKey := client.ObjectKey{Name: controlPlaneProviderName(), Namespace: controlPlaneProviderNamespace()}
+	err := r.GetResource(ctx, istKey, iop)
+	if err == nil || apierrors.IsNotFound(err) {
 		configsToUpdate = append(configsToUpdate, istio.NewOperatorWrapper(iop))
+	} else if !apimeta.IsNoMatchError(err) {
+		logger.V(1).Info("failed to get istiooperator object", "key", istKey, "err", err)
+		return nil, err
+	} else {
+		// Error is NoMatchError so check for Istio CR instead
+		ist := &istiov1alpha1.Istio{}
```
didierofrivia, line 334: Is this CR watched by the istio operator too?

adam-cattermole: Yep it is

didierofrivia (approved):
The way the sail project is added through a wrapper follows the same pattern as the alternatives, however this clearly exposes that we need to find a more declarative way of choosing what mesh provider we are installing. I'd say it's OK now, but we need to consider kickstarting a RFC, say part of the Kuadrant CR spec definitions efforts going like https://github.com/Kuadrant/architecture/pull/5/, https://github.com/Kuadrant/architecture/pull/18 and https://github.com/Kuadrant/architecture/pull/6

PR #324, opened 2023-11-21
Using Limitador CR condition ready

didierofrivia:


controllers/kuadrant_status.go
```diff
@@ -119,29 +119,23 @@ func (r *KuadrantReconciler) readyCondition(ctx context.Context, kObj *kuadrantv
...
+	if errors.IsNotFound(err) {
+		reason := "Limitador not found"
+		return &reason, nil
 	}
 
-	if availableCondition.Status != corev1.ConditionTrue {
-		return &availableCondition.Message, nil
+	statusConditionReady := meta.FindStatusCondition(limitadorObj.Status.Conditions, "Ready")
+	if statusConditionReady.Status != metav1.ConditionTrue {
+		return &statusConditionReady.Message, fmt.Errorf(statusConditionReady.Message)
```
eguzki, line 138: it should not return error. It is not an error to have the limitador not ready. It is a condition or state.

eguzki (approved):

controllers/kuadrant_status.go
```diff
@@ -119,29 +119,23 @@ func (r *KuadrantReconciler) readyCondition(ctx context.Context, kObj *kuadrantv
...
-		tmp := "Available condition not found"
-		return &tmp, nil
+	if errors.IsNotFound(err) {
+		reason := "Limitador not found"
+		return &reason, nil
 	}
 
-	if availableCondition.Status != corev1.ConditionTrue {
-		return &availableCondition.Message, nil
+	statusConditionReady := meta.FindStatusCondition(limitadorObj.Status.Conditions, "Ready")
```
eguzki, line 136: `statusConditionReady ` can be `nil`

eguzki (approved):

PR #326, opened 2023-11-22
bump dependencies to fix vulnerabilities

eguzki:
bump istio.io/istio 1.20.0
bump google.golang.org/grpc v1.58.3

PR #327, opened 2023-11-22
Update all the deps

alexsnaps:
Lame attempt at upgrading all the things, but this has resulted in [breaking an Integration test in my previous attempt](https://github.com/Kuadrant/kuadrant-operator/pull/326#issuecomment-1822679426), see how this goes this time around. We might also not want to update _all_ the things…


alexsnaps:
Closing in favor of #326 - all deps don't solve our issue.

PR #338, opened 2023-11-24
remove deployment of policy-controller for now

maleck13:
When we deploy a hub and spoke in the same cluster, we end up with two policy controllers running. So rather than rush into this deployment model, going to remove the policy controller from the kuadrant-operator CSV for now. We will update a the docs and explain that in a single cluster you can use the policy-controller in a single cluster and provide a command to install via kustomize

https://github.com/Kuadrant/multicluster-gateway-controller/issues/693


mikenairn:
You'll need to manually remove the CRDs and Roles added by the policy-controller from here https://github.com/Kuadrant/kuadrant-operator/tree/main/bundle/manifests. The bundle command doesn't remove these unfortunately.

mikenairn:
> I reckon we can leave the targets for metallb even if atm there's no practical use for those.

fyi, we plan to still add these docs but with an extra step to install the policy-controller https://github.com/Kuadrant/kuadrant-operator/pull/322, this will still require metallb.

maleck13:
@didierofrivia I see the integration tests failed. I don't think they are to do with this change. Still ok to merge?

didierofrivia:
@maleck13 rebasing with main should fix the test failing

maleck13:
@didierofrivia still looks to be failing any help would be appreciated . Seems to also be failing on just a docs update https://github.com/Kuadrant/kuadrant-operator/pull/322

PR #343, opened 2023-11-28
Release v0.5.0

alexsnaps:
Not a PR to merge, just checking this is correct and we can run the job: 

```shell
❯ make bundle IMAGE_TAG=v0.5.0 REGISTRY=quay.io VERSION=0.5.0 REPLACES_VERSION=0.4.1 CHANNELS=stable LIMITADOR_OPERATOR_VERSION=0.7.0 AUTHORINO_OPERATOR_VERSION=0.10.0 WASM_SHIM_VERSION=0.3.0
```

see how [this](https://github.com/Kuadrant/kuadrant-operator/actions/runs/7021911340) goes...


didierofrivia:
something to comment that I'll also add it to the RFC:

The _push_ event that triggers the build of the images are not passing all the params needed. I usually cancel it, then trigger manually the _workflow_dispatch_ "Build Images" with the same params you ran the `make bundle`

didierofrivia:
I reckon the manifest failed because of your local version of `controller-gen`

didierofrivia:
Oops, @guicassolato we (royal) forgot to do it the last time, but yes, we should do it

alexsnaps:
So, I will redo this yet again, but I challenge the usefulness of the changelog in this case... 
where/how do we plan on distributing that changelog? If it has no use, I think we should just _not_ have it at all. 
I don't know (if at all) where that changelog is currently made available to end users... can someone shed some light?

PR #344, opened 2023-11-28
Include missing unit test

Boomatang:
The controllers directory was not included in the `make test-unit`, this PR fixes that and the broken unit tests.

# Verification
* all tests should pass.
* the fixed tests should be included in the the test run. Check this by running `make test-unit | grep QueryParamMatchRegularExpression`


Boomatang:
@guicassolato It would be good to get you to confirm the fixes to the `PatternExpression` in the unit tests is correct. Some `.` have being replaced by `|` and I am not sure if this is the correct behaviour.

PR #346, opened 2023-11-29
Pin version, dep versions, channel and default channel

eguzki:


alexsnaps:
@didierofrivia I am not per se surprised by these changes, but I also think they make no sense... 
What this PR does, in order to get test to be 🟢 is to explicitly set all the variable that are then asked to be overwritten in the action... 🤔 
So I guess I don't get it. What is _not_ desirable in the release workflow? The testing of the bundle? The explicit setting of the proper operand versions in our build action? None of the above? All of it?
I guess that was already my point in a previous chat: we keep on repeating ourselves... Whether in the `make bundle ...` step locally or, as in this PR, changes to the `Makefile` to "hardcode" these versions; and then again when triggering the job. 

Right now, I think I like this approach better: hardcode the values in that changeset, and then... not set them at the build step? Or "just" the operator version/tag ones? Or...?

Anyways, I hope we get to discuss this today and make a call, one that possibly impacts open RFC PRs and/or changes in all our repos wrt this release process. I _think_ we can make it better, tho I also know I don't necessarily understand all requirement downstream when pushing to OperatorHub neither... anyways. I've been trying to wrap my head around this since this PR was open, and... I guess I'm just _very_ perplex about this all.

PR #293, opened 2023-11-08
Gh 639 policy controller

maleck13:
Draft PR adding the new policy controller to the deployment and CSV for kuadrant operator

This PR adds in the policy controller via kustomize it also add metal lb as an option

**Verification**
`make local-cluster-setup`
Build a new bundle and catalog and install it in to the cluster

expect to see a new `kuadrant-operator-policy-controller` in the `kuadrant-system` namespace
validate that the DNSPolicy and TLSPolicy CRDS are present in the cluster

**Extra Validation**

Before doing this you will need an AWS credential that gives access to route53 a domain (talk to craig as he can set you up with this)

deploy metallb to a local cluster
`make install-metallb`
`k create -f config/metallb/metal-lb.yaml`

create a new gateway isito gateway

```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: prod-web
spec:
  gatewayClassName: istio
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: specific
    hostname: 'REPLACEME'
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - kind: Secret
        name: REPLACEME
  ```

Setup TLS

create an issuer

```
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: glbc-ca
spec:
  selfSigned: {}

```

create a TLSPolicy

```
apiVersion: kuadrant.io/v1alpha1 
kind: TLSPolicy 
metadata: 
  name: prod-web
spec: 
  targetRef: 
    name: prod-web 
    group: gateway.networking.k8s.io 
    kind: Gateway    
  issuerRef: 
    group: cert-manager.io
    kind: ClusterIssuer
    name: glbc-ca
```

Observe the gateway is now configured and has an address and tls secrets

```
get gateway
NAME       CLASS   ADDRESS        PROGRAMMED   AGE
prod-web   istio   172.31.200.0   True         7m56s
```

Setup DNS
setup a dnsprovider 
https://docs.kuadrant.io/multicluster-gateway-controller/docs/dnspolicy/dns-provider/
setup a managedzone
https://docs.kuadrant.io/multicluster-gateway-controller/docs/dnspolicy/managed-zone/

create a dnspolicy

```
apiVersion: kuadrant.io/v1alpha1
kind: DNSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    group: gateway.networking.k8s.io
[...]

PR #348, opened 2023-11-30
re-add policy-controller

maleck13:
This is re-adding the policy-controller. 

This was verified here 
https://github.com/Kuadrant/kuadrant-operator/pull/293


adam-cattermole (approved):
Changes look good to me. Quickly validated the basic install as in original PR and see both containers running.

PR #360, opened 2023-12-06
Small enhancements to building actions

pehala:
* Do daily scheduled builds instead of weekly
   * Scheduled it at 1am so the gather job can finish in time
* Use SHAs only for scheduled build
  * It does not make sense as the rest of the components is only fixed in scheduled builds.
  * Should help with "mess" in Quay


.github/workflows/build-images-base.yaml
```diff
@@ -63,7 +67,7 @@ on:
         type: string
 
 env:
-  IMG_TAGS: ${{ github.sha }} ${{ inputs.kuadrantOperatorTag }}
```
didierofrivia, line 66: so no more sha tags in our images by default... I reckon we are not using them for anything else than the QE builds, no?

.github/workflows/build-images-base.yaml
```diff
@@ -63,7 +67,7 @@ on:
...
+  IMG_TAGS: ${{ inputs.kuadrantOperatorTag }} ${{ inputs.additionalTags == null && '' || inputs.additionalTags }}
```
didierofrivia, line 70: I'd leave it as...

```suggestion
  IMG_TAGS: ${{ inputs.kuadrantOperatorTag }} ${{ inputs.additionalTags }}
```

since the default is an empty string, no damage done for the rest of the consumers of this workflow base

.github/workflows/build-images-scheduled.yaml
```diff
@@ -15,3 +15,4 @@ jobs:
       authorinoOperatorVersion: ${{ vars.AUTHORINO_OPERATOR_SHA }}
       limitadorOperatorVersion: ${{ vars.LIMITADOR_OPERATOR_SHA }}
       wasmShimVersion: ${{ vars.WASM_SHIM_SHA }}
+      additionalTags: ${{ github.sha }}
```
didierofrivia, line 18: Actually, this might not even needed, since we are passing the `kuadrantOperatorTag` with the value of `github.sha` in line 13 and 14 of this very same file... so the `build-images-base` already has that tag... does it make sense?

pehala: Thanks for the suggestion! I totally missed that

didierofrivia (approved):
🍺

PR #372, opened 2023-12-11
Add descriptions and samples to CSV

### LINKED ISSUE #369: New CRDs names and descriptions missing in Operator Hub
#348 introduced new CRDs which have the names, descriptions and examples missing in CSV, which then renders in OperatorHub:

image

These need to be added to the [CSV template](https://github.com/Kuadrant/kuadrant-operator/blob/13c75de18c06a459f82c0cd682988b3e11926b92/config/manifests/bases/kuadrant-operator.clusterserviceversion.yaml#L18) to propagate correctly to the bundle manifest

grzpiotrowski:
Fixes #369 

Added descriptions to CSV template for:
- DNSHealthCheckProbe
- DNSPolicy
- DNSRecord
- ManagedZone
- TLSPolicy

Added a kustomize remote base for the yaml samples from MGC repo:
- DNSPolicy
- ManagedZone
- TLSPolicy

DNSRecord and DNSHealthCheckProbe sample not included since these are created by the controller.

Changes/suggestions very welcome.

The final looks in the OperatorHub can be checked by pasting the whole content of the generated [kuadrant-operator.clusterserviceversion.yaml](https://github.com/Kuadrant/kuadrant-operator/blob/csv-crd-desc/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml) to the [OperatorHub Preview](https://operatorhub.io/preview) tool.

### DIFF
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -5,6 +5,68 @@ metadata:
     alm-examples: |-
       [
         {
+          "apiVersion": "kuadrant.io/v1alpha1",
+          "kind": "DNSPolicy",
+          "metadata": {
+            "name": "example-dnspolicy"
+          },
+          "spec": {
+            "healthCheck": {
+              "endpoint": "/",
+              "protocol": "HTTP"
+            },
+            "targetRef": {
+              "group": "",
+              "kind": "Gateway",
+              "name": "example-gateway"
+            }
+          }
+        },
+        {
+          "apiVersion": "kuadrant.io/v1alpha1",
+          "kind": "ManagedZone",
+          "metadata": {
+            "labels": {
+              "app.kubernetes.io/created-by": "multicluster-gateway-controller",
+              "app.kubernetes.io/instance": "managedzone-sample",
+              "app.kubernetes.io/managed-by": "kustomize",
+              "app.kubernetes.io/name": "managedzone",
+              "app.kubernetes.io/part-of": "multicluster-gateway-controller"
+            },
+            "name": "managedzone-sample"
+          },
+          "spec": {
+            "description": "My managed domain",
+            "domainName": "testmz.hcapps.net"
+          }
+        },
+        {
+          "apiVersion": "kuadrant.io/v1alpha1",
+          "kind": "TLSPolicy",
+          "metadata": {
+            "labels": {
+              "app.kubernetes.io/created-by": "tmp",
+              "app.kubernetes.io/instance": "tlspolicy-sample",
+              "app.kubernetes.io/managed-by": "kustomize",
+              "app.kubernetes.io/name": "tlspolicy",
+              "app.kubernetes.io/part-of": "tmp"
+            },
+            "name": "tlspolicy-sample"
+          },
+          "spec": {
+            "issuerRef": {
+              "group": "cert-manager.io",
+              "kind": "ClusterIssuer",
+              "name": "glbc-ca"
+            },
+            "targetRef": {
+              "group": "gateway.networking.k8s.io",
+              "kind": "Gateway",
+              "name": "prod-web"
+            }
+          }
+        },
+        {
           "apiVersion": "kuadrant.io/v1beta1",
           "kind": "Kuadrant",
           "metadata": {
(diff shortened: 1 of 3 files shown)


bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -4,6 +4,78 @@ metadata:
   annotations:
     alm-examples: |-
       [
+        {
```
maleck13, line 7: I think these might need to be added in the MGC repo? I think this will get nuked when we do a new bundle as it is pulled in from MGC

config/manifests/bases/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -31,6 +31,31 @@ spec:
         kind: RateLimitPolicy
         name: ratelimitpolicies.kuadrant.io
         version: v1beta2
+      - description: DNSHealthCheckProbe enables performing health checks against a DNS endpoint (A or CNAME record)
```
maleck13, line 34: same with these

config/samples/kuadrant.io_v1alpha1_dnspolicy.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: kuadrant.io/v1alpha1
```
maleck13, line 1: looks good

config/samples/kuadrant.io_v1alpha1_dnsrecord.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kuadrant.io/v1alpha1
```
maleck13, line 1: probably less valuable as really only the dns conrtroller creates these but it looks accurate

maleck13 (commented):
If the changes have been made directly in the CSV I think they will get wiped out via a future generated CSV

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
grzpiotrowski: Right, here I just added the sample files manually to the [kuadrant-operator/config/samples/kustomization.yaml ](https://github.com/Kuadrant/kuadrant-operator/blob/csv-crd-desc/config/samples/kustomization.yaml )from where they propagate to the examples in CSV. 

Fetching these samples automatically from the MGC repo would make more sense I guess. I need to check how to do that in Kustomize. If not possible than I think the samples would need to be fetched in the `make bundle` script.

maleck13: ah I see. Yeah I mean I guess this is ok, but would be better to have them from the source

maleck13 (approved):
ok so generally looks fine to me

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
grzpiotrowski: hm. Ok, did some experimentation and looks like we could put a kustomization.yaml to have a remote base with the samples in the MGC repo. The samples for TLSPolicy, ManagedZone and DNSPolicy are already there.
And since the DNSHealthCheckProbe and DNSRecord are created by the controller I suppose there is no real need to expose examples for these two in the Operator Hub through the CSV?

The snipped below would go to: `multicluster-gateway-controller/config/samples/kustomization.yaml`:
```yaml
resources:
- kuadrant.io_v1alpha1_tlspolicy.yaml
- kuadrant.io_v1alpha1_managedzone.yaml
- dnspolicy.yaml
```

Then we could have a reference to it as a remote resource for kustomize in kuadrant-operator:
```yaml
resources:
- github.com/Kuadrant/multicluster-gateway-controller/config/samples?ref=main
```

This would let us keep the samples in one place and not worry about updating them manually in kuadrant-operator.

maleck13: sounds like a good option. If you open the PR I will take a look at it

grzpiotrowski: PR opened. I'll leave the link here as a cross reference: [Add config/samples/kustomization.yaml #720](https://github.com/Kuadrant/multicluster-gateway-controller/pull/720)

grzpiotrowski:
Ok, so I made the changes to have the DNSPolicy, TLSPolicy and ManagedZone samples being pulled from MGC repo via kustomize remote base, instead of having the samples duplicated into kuadrant-operator repo.

This is a screenshot of how it will appear in OperatorHub now:
image

grzpiotrowski:
Bundle test failing now as [multicluster-gateway-controller/pull/720](https://github.com/Kuadrant/multicluster-gateway-controller/pull/720) will need to be merged first.

didierofrivia (approved):
🍔

Issue #370, opened 2023-12-11
New single cluster quick start script in place that can be executed remotely


adam-cattermole:
New single cluster quick start script in place that can be executed remotely `curl "https://raw.githubusercontent.com/kuadrant/kuadrant-operator/${BRANCH}/hack/quickstart-setup.sh" | bash` Note the quick start should optionally be able to setup a managed zone and dns provider for use with `DNSPolicy`

PR #373, opened 2023-12-11
Add new quickstart script

adam-cattermole:
#370


hack/quickstart-setup.sh
```diff
@@ -0,0 +1,110 @@
...
+if [ -z $KUADRANT_ORG ]; then
+  KUADRANT_ORG=${KUADRANT_ORG:="kuadrant"}
+fi
+if [ -z $MGC_BRANCH ]; then
+  MGC_BRANCH=${MGC_BRANCH:="main"}
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/adam-cattermole/multicluster-gateway-controller/update-quickstart/hack/.quickstartEnv)"
+source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/adam-cattermole/multicluster-gateway-controller/update-quickstart/hack/.deployUtils)"
```
adam-cattermole, line 31: Needs updating once this is merged https://github.com/Kuadrant/multicluster-gateway-controller/pull/731

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,110 @@
...
+# limitations under the License.
+#
+
+if [ -z $KUADRANT_ORG ]; then
+  KUADRANT_ORG=${KUADRANT_ORG:="kuadrant"}
+fi
+if [ -z $KUADRANT_BRANCH ]; then
+  KUADRANT_BRANCH=${KUADRANT_BRANCH:="main"}
+fi
+if [ -z $MGC_BRANCH ]; then
```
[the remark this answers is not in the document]
adam-cattermole: Do we want the branch you install from to be configurable? or do you mean it'd already be set at this stage before the curl request so no need to set a default?

I followed what was done here in MGC quickstart https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/hack/quickstart-setup.sh#L19-L24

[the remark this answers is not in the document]
maleck13: right but I don't see `MGC_BRANCH` being used in the script. Perhaps I am missing something

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,110 @@
...
+${KUSTOMIZE_BIN} build ${KUADRANT_DEPLOY_KUSTOMIZATION} | kubectl apply -f -
+kubectl -n ${KUADRANT_NAMESPACE} patch deployment kuadrant-operator-controller-manager --type='merge' -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","image":"'"${KUADRANT_IMAGE}"'"}]}}}}'
+
+# Configure managedzone
+if [ ! -z "$DNS_PROVIDER" ]; then
+  configureController ${KUADRANT_CLUSTER_NAME} ${KUADRANT_NAMESPACE}
+fi
+
+# Deploy kuadrant
+kubectl -n ${KUADRANT_NAMESPACE} apply -f ${KUADRANT_REPO_RAW}/config/samples/kuadrant_v1beta1_kuadrant.yaml
```
maleck13, line 110: I think it might be good to echo out a link to the follow on walkthrough WDYT?

adam-cattermole: 100% agree

hack/quickstart-setup.sh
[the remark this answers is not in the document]
adam-cattermole: Ah I see, it will be in the MGC links once I've merged the other PR to main, I'll update the commands:
```bash
source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${KUADRANT_ORG}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.quickstartEnv)"
source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/${KUADRANT_ORG}/multicluster-gateway-controller/${MGC_BRANCH}/hack/.deployUtils)"
```

hack/quickstart-setup.sh
adam-cattermole: I guess this should be the link to the location on kuadrant.io rather than github once we have the doc PR merged?

adam-cattermole: Ah nevermind, I see the link is already there - updated now

hack/quickstart-setup.sh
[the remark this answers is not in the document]
maleck13: right ok. Hmm. I am wondering if perhaps we should split these pieces of work. I fear here we could end up with a situation where you need to know the MGC_BRANCH to use and the KUADRANT_BRANCH to use. As we could change something in MGC deploy utils and we might accidentally break this quick start. So I am thinking that it may keep us more sane and be in our best interest right now to copy over the scripts you need and add an issue to discuss how the projects can share these functions??

maleck13:
related https://github.com/Kuadrant/multicluster-gateway-controller/pull/731

hack/quickstart-setup.sh
```diff
@@ -0,0 +1,111 @@
...
+KUADRANT_REPO_RAW="https://raw.githubusercontent.com/${KUADRANT_ORG}/kuadrant-operator/${KUADRANT_BRANCH}"
+KUADRANT_DEPLOY_KUSTOMIZATION="${KUADRANT_REPO}/config/deploy"
+KUADRANT_GATEWAY_API_KUSTOMIZATION="${KUADRANT_REPO}/config/dependencies/gateway-api"
+KUADRANT_ISTIO_KUSTOMIZATION="${KUADRANT_REPO}/config/dependencies/istio/sail"
+KUADRANT_CERT_MANAGER_KUSTOMIZATION="${KUADRANT_REPO}/config/dependencies/cert-manager"
+KUADRANT_METALLB_KUSTOMIZATION="${KUADRANT_REPO}/config/metallb"
+
+set -e pipefail
+
+if [[ "${KUADRANT_BRANCH}" != "main" ]]; then
```
maleck13, line 44: I wonder if this would be better named ad `KUADRANT_RELEASE` that way the doc could say set ```export KUADRANT_RELEASE=v0.6.0```

hack/quickstart-setup.sh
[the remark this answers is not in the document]
maleck13: So based on discussion we will do the following:
rely on mgc `main` scripts for 0.6.0 of kuadrant and then move to pin the scripts in 0.7.0

david-martin:
👀

maleck13:
This seems to fail quite early at the moment with the following message

```
The Istio "istiocontrolplane" is invalid: spec.namespace: Required value
```

adam-cattermole:
Work in progress fix here https://github.com/Kuadrant/kuadrant-operator/pull/380

maleck13:
Is there a previous tag we could use that would keep this working and we can work towards supporting the changes then in a separate release?

maleck13:
Once this merges we need to merge https://github.com/Kuadrant/docs.kuadrant.io/pull/44/files I am updating the quick start doc with the new env vars now

hack/quickstart-setup.sh
adam-cattermole: I've updated this now, so for the time being can be run with:

```bash
export KUADRANT_ORG=adam-cattermole
export MGC_REF=update-quickstart
curl "https://raw.githubusercontent.com/adam-cattermole/kuadrant-operator/quickstart-script/hack/quickstart-setup.sh" | bash
```

maleck13:
relies on https://github.com/Kuadrant/multicluster-gateway-controller/pull/731

david-martin:
I got through the script and guide to the end.
1 thing I noticed, that might be a misconfiguration on the RLP, is that both alice & bob are affected by the same limit i.e. 1 every 3 seconds.
From reading the RLP in the guide, should it be 2 every 3 seconds for bob?

```
$ while :; do curl -k --resolve api.${KUADRANT_ZONE_ROOT_DOMAIN}:443:${INGRESS_HOST} --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' "https://api.$KUADRANT_ZONE_ROOT_DOMAIN/cars" | grep -E --color "\b(429)\b|$"; sleep 1; done
200
429
429
200
429
429
200
429
429
200
^C

$ while :; do curl -k --resolve api.${KUADRANT_ZONE_ROOT_DOMAIN}:443:${INGRESS_HOST} --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMBOB' "https://api.$KUADRANT_ZONE_ROOT_DOMAIN/cars" | grep -E --color "\b(429)\b|$"; sleep 1; done
429
429
200
429
429
200
429
429
200
^C
```

maleck13:
@david-martin good spot. The RLP looks incorrect in the doc, I will update

philbrookes:
This is all working for me.

maleck13:
going to merge this now we can update the doc if needed afterwards

maleck13 (approved):

PR #380, opened 2023-12-15
Fixes to local setup for sail

adam-cattermole:
New changes to https://github.com/maistra/istio-operator/tree/maistra-3.0 broke our `make local-setup`.

The operator became cluster scoped and relies on `spec.namespace` to be set in Istio CR.

TODO:
- [x] Currently investigating why service is not being created when we create the gateway - gateway does not become programmed


adam-cattermole:
The image for `latest-3.0` was not tagged to the nightly build that included cluster-scoped changes. This resulted in us installing the newer manifests with the cluster-scoped changes but with an older operator image that wasn't handling it which resulted in the gateway not being reconciled.

Image has been re-tagged so the changes here work now.

guicassolato (approved):

PR #381, opened 2023-12-15
deps: update maistra.io/istio-operator@maistra-3.0

guicassolato:


adam-cattermole:
I think this is dup of #380. There's still an extra missing change needed to get this working

guicassolato:
> I think this is dup of #380. There's still an extra missing change needed to get this working

100%. Sorry about that. Closing this one.

PR #391, opened 2023-12-19
fix cluster scoped istio CR

eguzki:
:shrug:

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -50,6 +50,8 @@ import (
 
 const (
 	kuadrantFinalizer = "kuadrant.io/finalizer"
+	// (Sail) The istio CR must be named default to process GW API resources
+	istioCRName = "default"
 )
 
 // KuadrantReconciler reconciles a Kuadrant object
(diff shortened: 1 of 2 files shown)


controllers/kuadrant_controller.go
```diff
@@ -380,6 +381,10 @@ func controlPlaneProviderName() string {
 	return env.GetString("ISTIOOPERATOR_NAME", "istiocontrolplane")
 }
 
+func istioCRName() string {
+	return env.GetString("ISTIO_NAME", "default")
```
guicassolato, line 385: So we need two environment variables, one for specifying the name of the `IstioOperator` CR and another one for specifying the name of the `Istio` CR in case the IstioOperator API is not installed?

adam-cattermole: I'm investigating this as well as part of #392. It appears that currently the gateway is only reconciled if the Istio CR is named `default`, in which case the associated configmap ends up being named `istio`. If the Istio CR is named anything else the configmap ends up being called `istio-<istio-cr-name>` and (currently in this case) the gateway is not being reconciled correctly.

adam-cattermole: I've reached out to the sail team and it appears that the `istio` CR **must** be named `default` to process GW API resources (at least for now). This is due to the way upstream istio handles revisions (they are not fully supported yet) and it may change in the future.

Perhaps to avoid confusion for now we could remove the configuration through an environment variable and instead have it always search for `default` - I'll push a commit for this - I've also checked the docs and there's no mention of the previous name as they rely on the quickstart script or the Makefile which both use this yaml.

guicassolato (approved):

PR #392, opened 2024-01-03
wip: Rename configmap for sail

adam-cattermole:
Test to see if this fixes sail integration tests.

I don't like the solution will look at refactoring.

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/env"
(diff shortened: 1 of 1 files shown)


controllers/kuadrant_controller.go
```diff
@@ -380,7 +383,10 @@ func controlPlaneProviderName() string {
 	return env.GetString("ISTIOOPERATOR_NAME", "istiocontrolplane")
 }
 
-func controlPlaneConfigMapName() string {
+func controlPlaneConfigMapName(isSail bool) string {
```
guicassolato, line 386: I wonder if this function doesn't belong to `pkg/istio/mesh_config.go`. It looks like the name of the ConfigMap is an implementation detail of each kind of Istio operator.

If this is indeed the direction we want to go, WDYT about something like the following?

```go
func (r *KuadrantReconciler) getIstioConfigObjects(ctx context.Context, logger logr.Logger) ([]common.ConfigWrapper, error) {
	// ...
	if err == nil || apierrors.IsNotFound(err) {
		opw := istio.NewOperatorWrapper(iop)
		configMapName = opw.ConfigMapName()
		configsToUpdate = append(configsToUpdate, opw)
	} else if !apimeta.IsNoMatchError(err) {
		// ...
	} else {
		// ...
		opw := istio.NewSailWrapper(ist)
		configMapName = opw.ConfigMapName()
		configsToUpdate = append(configsToUpdate, opw)
	}
	// ...
}
```

Of course, `OperatorWrapper` and `SailWrapper` structs would implement `ConfigMapName() string` as well (apart from `ConfigWrapper`).

adam-cattermole: Yep that's along the lines of what I was thinking. Unfortunately these changes aren't enough though to solve the install issue, still investigating.

adam-cattermole:
I'm going to close this in favour of #391

Issue #307, opened 2023-11-16
Reconcile Spec to sub components


Boomatang:
Add the new Spec into the kuadrant CR. Check the [Fields to reconcile](https://github.com/Kuadrant/architecture/blob/main/rfcs/0006-kuadrant_sub_components_configurations.md#fields-to-reconcile) in the RFC for a list of fields. Each field should be double-checked with the component owners to ensure there is a valid use case for having that field in the Kuadrant CR.

## Expected Behaviour 

- Spec configured in the Kuadrant CR should be reconcile to the relevant sub component CR
- If the spec is defined in the kuadrant CR only that spec is allowed.
Meaning any changes to that spec in the sub component CRs while be reverted by the kuadrant operator back to the spec in the kuadrant CR.
- When a definable field is not configured in the Kuadrant CR the operator should not try to reconcile that field.
This allows for upgrades by not overriding configuration users many have already configured in the sub component before this release.
- No labels should be reconciled as part of this work.

## PRs
- [ ] https://github.com/Kuadrant/kuadrant-operator/pull/393
- [x] https://github.com/Kuadrant/kuadrant-operator/pull/350

Boomatang:
Closing: a number of issue with this, work even got reverted.

PR #393, opened 2024-01-11
Reconcile Sub Component Authorino CR

Boomatang:
Part of: #307 
RFC: https://github.com/Kuadrant/architecture/blob/main/rfcs/0006-kuadrant_sub_components_configurations.md

These changes allow a user to configure settings in the Authorino CR from the Kuadrant CR.

# Verification 

* Deploy this branch using `make local-setup`.
* Wait for deployment to complete and all resources to be a ready state.
* Deploy the custom resources.
  * `kubectl apply -f https://raw.githubusercontent.com/Boomatang/k8s_configs/kuadrant-operator-PR393/manifest/kuadrant_spec_authorino.yaml`
  * The custom manifest will deploy all the resources required for testing including an instance of Jaeger for OpenTelemetry. Jaeger needs a number of resources set up so the deployment will fail on the first run but rerun the command again to get all resources installed.
  * The Kuadrant CR has every [spec](https://github.com/Boomatang/k8s_configs/blob/kuadrant-operator-PR393/manifest/kuadrant_spec_authorino.yaml#L15681-L15687) that this PR allows configuring set. 
  * The manifest is linked to a tag [kuadrant-operator-PR393](https://github.com/Boomatang/k8s_configs/tree/kuadrant-operator-PR393) to ensure state does not change.
* Wait for the deployment to complete
* The spec that is being reconciled are listed in the RFC: [Authorino Spec](https://github.com/Kuadrant/architecture/blob/main/rfcs/0006-kuadrant_sub_components_configurations.md#authorino-spec)
* Editing the spec directly in the Authorino CR should be reverted to the spec in the Kuadrant CR
* Editing the spec in the Kuadrant CR should update the spec in the Authorino CR
  * NOTE: fully removing a spec from the kuadrant CR will not remove it from the Authorino CR. This is done to allow upgrades. A user may have custom spec already in the Limitador CR. This change only overrides the Authorino CR spec if that spec is defined in Kuadrant CR.


controllers/kuadrant_controller.go
```diff
@@ -493,38 +494,71 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
-		},
-		Spec: authorinov1beta1.AuthorinoSpec{
-			ClusterWide:            true,
-			SupersedingHostSubsets: true,
-			Listener: authorinov1beta1.Listener{
-				Tls: authorinov1beta1.Tls{
-					Enabled: &tmpFalse,
+	authorinoKey := client.ObjectKey{Name: common.AuthorinoName, Namespace: kObj.Namespace}
+	authorino := &authorinov1beta1.Authorino{}
+	err := r.Client().Get(ctx, authorinoKey, authorino)
```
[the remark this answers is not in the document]
Boomatang: Good question. This has to do with not overriding exist user settings if they are not defined in the kuadrant CR. An example is, the user has set the number of replicas in the Authorino CR before this feature was released. We do not want to over ride these existing setting.  If we don't do the Get your questioning, the mutator function will always say there is a difference between the desired and existing. It may be possible to do all this work inside the mutator function but that would meant the desired configuration being past in would not be the whole truth as the mutator function would be able to change the desired configuration on the fly.  

I believe it is better to pass the correct desired configuration to the mutator function, over allowing the mutator change the desired configuration on the fly.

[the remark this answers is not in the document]
adam-cattermole: Aha I see makes sense thanks, precedence is kuadrant CR > existing settings > defaults.

pkg/kuadranttools/authorino_tools.go
```diff
@@ -0,0 +1,71 @@
...
+
+func AuthorinoMutator(existingObj, desiredObj client.Object) (bool, error) {
+	update := false
+	existing, ok := existingObj.(*authorinov1beta1.Authorino)
+	if !ok {
+		return false, fmt.Errorf("existingObj %T is not a *authorinoauthorinov1beta1.Authorino", existingObj)
+	}
+	desired, ok := desiredObj.(*authorinov1beta1.Authorino)
+	if !ok {
+		return false, fmt.Errorf("desiredObj %T is not a *authorinoauthorinov1beta1.Authorino", desiredObj)
```
adam-cattermole, line 20: nit typo in log:
```suggestion
		return false, fmt.Errorf("existingObj %T is not a *authorinov1beta1.Authorino", existingObj)
	}
	desired, ok := desiredObj.(*authorinov1beta1.Authorino)
	if !ok {
		return false, fmt.Errorf("desiredObj %T is not a *authorinov1beta1.Authorino", desiredObj)
```

adam-cattermole (commented):
Overall the changes here look good - just a couple things:

- Do you think we should add an integration test where kuadrant CR has some authorino spec values applied?
- I'm wondering if we should document the behaviour where items removed from the spec do not result in defaults in the authorino CR. I understand this behaviour is to avoid overriding existing settings but if I'm a new user of Kuadrant and all I want is to set my authorino settings through the Kuadrant CR I would find this confusing - I deploy with defaults and then override them and then remove the setting and I don't end up with defaults again.

controllers/kuadrant_controller.go
```diff
@@ -493,38 +494,71 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
```
guicassolato, line 501: Maybe get rid of the else clause.

```suggestion
		if !apierrors.IsNotFound(err) {
			return err
		}
```

controllers/kuadrant_controller.go
```diff
@@ -493,38 +494,71 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
-				Tls: authorinov1beta1.Tls{
-					Enabled: &tmpFalse,
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      common.AuthorinoName,
+					Namespace: kObj.Namespace,
 				},
-			},
-		},
+				Spec: authorinov1beta1.AuthorinoSpec{
+					ClusterWide:            true,
```
guicassolato, line 513: There aren't many use cases for allowing multiple cluster-wide instances and yet do not to expose `authConfigLabelSelectors`. I can think of some, but in general it's not recommended.

controllers/kuadrant_controller.go
```diff
@@ -493,38 +494,71 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
+					SupersedingHostSubsets: true,
+					Listener: authorinov1beta1.Listener{
+						Tls: authorinov1beta1.Tls{
+							Enabled: &tmpFalse,
```
guicassolato, line 517: Can use [`ptr.To`](https://pkg.go.dev/k8s.io/utils/ptr#To) here.

controllers/kuadrant_controller.go
Boomatang: This was part of the existing configuration. I made the assumption that the existing configuration was correct.

controllers/kuadrant_controller.go
Boomatang: This was part of the existing code. Happy to make the change as there are a few other changes needed.

controllers/kuadrant_controller.go
Boomatang: Ya I can make this change. Personal I don't like negative conditionals and my natural instinct is not to use them.

pkg/kuadranttools/authorino_tools.go
Boomatang: Going to blame copy and paste. Will fix.

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
Boomatang: Sorry @guicassolato, I have to strongly disagree with you on this. If the user has configuration in the Authorino or Limitador CR, blank values in the Kuadrant CR should not override those values. Because if it does how does the user upgrade without bring down their service? That would be a breaking change to the Users API without any deprecation notice. I know currently today we do not have good support for upgrades but that does not mean we should break customers stuff if it can be avoided. 

This behaviour follows what was outlined in the RFC. If this is unwanted behaviour a deprecation notice should be give and the behaviour removed in a later version allow user time to migrate their custom configurations from the sub components CRs to the Kuadrant CR. 

For clarity, this only applies to fields the user is currently allow configure. If the Kuadrant operator requires a field to be Y after this change that field will still be Y.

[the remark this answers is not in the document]
guicassolato: It feels like what you're proposing inverts the relation between controller and controlled resources, IMO – in contrast to API Server >>> Controller >>> Controlled resources.

E.g., should we apply the same policy to Authorino AuthConfigs? One manually sets a field in a AuthConfig resource that is omitted in the AuthPolicy. Should the controller override what's in the AuthConfig or should it honour it? IMO, it should override because, in this case, the AuthPolicy is the API, not the AuthConfig. Similarly, the Kuadrant CRD is the API, not the Authorino one, nor Limitador.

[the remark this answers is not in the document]
Boomatang: I know see what you are getting at. Would you class this as a breaking change then? I would, if the user upgrades to the version here any setting that can now be configured in the Kuadrant CR would override any of the configuration in the sub component CRs. That would be introducing a breaking change without any deprecation notice and no way for the user to upgrade without bring down their service. 

I would be happy if this change was added in and a deprecation notice give to say that in the future configuration of in sub components would be removed.

api/v1beta1/kuadrant_types.go
```diff
@@ -19,6 +19,7 @@ package v1beta1
 import (
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
+	authorinov1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
```
guicassolato, line 22: Let's make sure we use the latest version of the `Authorino` API here: https://github.com/Kuadrant/kuadrant-operator/pull/561

doc/reference/kuadrant.md
```diff
@@ -16,6 +16,80 @@ The Kuadrant operator creates a Limitador CR named `limitador` in the same names
...
+|------------|-------------|:------------:|-------------------------------------------------------------------------------|
+| port       | Integer     |      No      | Port number of OIDC Discovery server for Festival Wristband tokens.           |
+| tls        | [TLS](#tls) |     Yes      | TLS configuration of the ODIC Discovery server for Festival Wristband tokens. |
+
+#### Tracing
+
+| **Field** | **Type** | **Required** | **Description**                                                                                     |
+|-----------|----------|:------------:|-----------------------------------------------------------------------------------------------------|
+| endpoint  | String   |     Yes      | Full endpoint of the OpenTelemetry tracing collector service (e.g. http://jaegar:14268/api/traces). |
+| tags      | Map      |      No      | Key-value map of fixed tags to add to all OpenTelemetry traces emitted by Authorino.                |
```
guicassolato, line 68: Ref. https://github.com/Kuadrant/kuadrant-operator/pull/393#discussion_r1591051701

The latest version of the `Authorino` API shall include another field here, `insecure`.

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
guicassolato: https://github.com/Kuadrant/kuadrant-operator/issues/617

api/v1beta1/kuadrant_types.go
Boomatang: Once that PR is merged I will make sure to rebase to include it.

guicassolato: Merged

maleck13:
@Boomatang @guicassolato @adam-cattermole what is the story with this PR is it still relevant or should we close it?

Boomatang:
Yes @maleck13,  I do feel that this work is still relevant, it is part of the [RFC006](https://github.com/Kuadrant/architecture/blob/main/rfcs/0006-kuadrant_sub_components_configurations.md). After me being out sick for two months priorities shifted and this work did get left behind a bit. It is now harder to get reviews on the work. 

If it is decide that this work is not relevant any more then it would make sense for the RFC to be removed and the existing work done around the limitador CR to be reverted.

Boomatang:
Changes in RFC design means that will not be the final approach

Issue #382, opened 2023-12-15
Pin version of Sail used


adam-cattermole:
We currently install from the https://github.com/maistra/istio-operator/tree/maistra-3.0 branch directly which results in any downstream changes potentially breaking our install (see #380). We should look to pin the version that we install alongside the image that we use.

That being said, it's worth considering that when installing from the community operators on OpenShift currently the only available build is the most recent nightly build. If we pin to a version and don't test frequently on OpenShift, changes could be made that would break our compatibility without our knowledge.

adam-cattermole:
The release of the sail without a floating tag should be completed as part of [OSSM-5547](https://issues.redhat.com/browse/OSSM-5547)

adam-cattermole:
Until we find a way forward with #401 I'll pin to the latest nightly that still had the kustomize manifests.

PR #395, opened 2024-01-16
Use istioctl for local install and skip sail tests on PR

adam-cattermole:
- Reverted the default install type for `make local-setup` to use `istioctl` instead of project sail
- Disabled the sail integration tests from PRs but still runs nightly
- Updated quickstart to depend on `ISTIO_INSTALL_SAIL` where default is to install without sail (false)

You can test the quickstart changes in this PR like this:

```bash
curl "https://raw.githubusercontent.com/adam-cattermole/kuadrant-operator/make-istioctl-default/hack/quickstart-setup.sh" | bash
```

TODO:
- [x] Check docs + quickstart

Related #382

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -48,6 +48,11 @@ jobs:
     strategy:
       matrix:
         istio-type: [ sail, istioctl ]
+        pr-event:
+          - ${{ github.event_name == 'pull_request' }}
+        exclude:
+          - istio-type: sail
+            pr-event: true
     runs-on: ubuntu-latest
     env:
       KIND_CLUSTER_NAME: kuadrant-test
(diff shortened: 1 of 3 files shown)


didierofrivia (approved):
Thanks for the PR!, this will solve the failing scripts in other branches.

```
error: no objects passed to apply
make[4]: *** [sail-install] Error 1
make[3]: *** [istio-install] Error 2
make[2]: *** [local-cluster-setup] Error 2
make[1]: *** [local-env-setup] Error 2
make: *** [local-setup] Error 2
```

We might want to review in the future the scripts that are related to user guides and place them closer to those instead of the `hack` directory. Maybe the ones we have in the `utils` directory are good candidates for merging into `hack` following https://github.com/kubernetes/kubernetes/blob/master/hack/README.md

PR #396, opened 2024-01-17
[WIP] refactor: api machainary 

KevFan:


api/v1beta2/ratelimitpolicy_types_test.go
```diff
@@ -6,11 +6,10 @@ import (
 	"strings"
 	"testing"
 
+	policy2 "github.com/kuadrant/kuadrant-operator/pkg/library/utils"
```
adam-cattermole, line 9: nit
```suggestion
	utils "github.com/kuadrant/kuadrant-operator/pkg/library/utils"
```

pkg/library/reconcilers/gateway_diffs.go
```diff
@@ -0,0 +1,122 @@
+package reconcilers
+
+import (
+	"context"
+	"fmt"
+	"slices"
+
+	"github.com/go-logr/logr"
+	policy2 "github.com/kuadrant/kuadrant-operator/pkg/library/utils"
```
adam-cattermole, line 9: nit
```suggestion
	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
```

controllers/ratelimitpolicy_controller.go
```diff
@@ -22,6 +22,9 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/kuadrant/kuadrant-operator/pkg/library/mappers"
+	reconcilers2 "github.com/kuadrant/kuadrant-operator/pkg/library/reconcilers"
```
didierofrivia, line 25: maybe rename to something a bit more descriptive, like.. `reconciler_utils`, `reconcilers_tools` wdyt?

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -252,6 +252,10 @@ func (r *RateLimitPolicy) BackReferenceAnnotationName() string {
 	return "kuadrant.io/ratelimitpolicies"
 }
 
+func (r *RateLimitPolicy) DirectReferenceAnnotationName() string {
+	return "kuadrant.io/ratelimitpolicy"
```
didierofrivia, line 256: maybe we could move these functions returning static strings to constants in the future

controllers/ratelimitpolicy_controller.go
KevFan: Agreed, will update 👍

adam-cattermole (approved):
There's a lot of changes so would appreciate another review but I think this looks pretty good now, nice work!

api/v1beta2/authpolicy_types.go
```diff
@@ -197,7 +202,8 @@ func (s *AuthPolicyStatus) Equals(other *AuthPolicyStatus, logger logr.Logger) b
 	return true
 }
 
-var _ common.KuadrantPolicy = &AuthPolicy{}
+var _ utils.KuadrantPolicy = &AuthPolicy{}
+var _ utils.Referrer = &AuthPolicy{}
```
mikenairn, line 206: Is "utils" the best package for these?

eguzki (approved):
I say we go with this :100: 

> Currently, this pull request may not be the ideal refactoring, as several functions were refactored into this library package to address issues such as cyclic imports. However, this PR is likely to be the first of many refactors as we determine what should truly exist in the gateway-api-machinery library versus the Kuadrant Operator."

I have the same feeling. And I do not think this PR adds more mess, on the contrary, I think that the packages have been organized in a neat way.

eguzki:
I also have to anticipate that with the WIP topology implementation approaches (at least 2 I am aware of), some references (back refs and direct back refs) between resources will vanish, making some code refactored here useless.

api/v1beta2/authpolicy_types.go
KevFan: Possibly not, I've added a kuadrant package to break up the `utils` package a bit, which might be a better package for these (some are required to prevent a cyclic import)

KevFan:
Merging this as this has multiple approvals now and have been rebased from main with changes for DNSPolicy and TLSPolicy. All comments are addressed and tests are passing so I think this is safe to merge as is

PR #403, opened 2024-01-23
Remove --ocm-hub flag and update bundle

adam-cattermole:
Policy controller pod was failing to start due to additional `--ocm-hub` flag which was removed (see https://github.com/Kuadrant/multicluster-gateway-controller/pull/711).

Manifests also needed updating from `make bundle`


didierofrivia (approved):
👷🏼

Issue #305, opened 2023-11-16
Sub component testing


Boomatang:
As now this allows configuration of the sub components from a higher level we need to create automated tests that will configure the sub components. The aim is to ensure as further development of the sub components continues these updates don't break the configuration from a higher level.

Boomatang:
Closing: a number of issue with this, work even got reverted.

PR #404, opened 2024-01-23
Integration tests for configuration reconcile

Boomatang:
Part of: https://github.com/Kuadrant/kuadrant-operator/issues/305

Add integration test for testing the reconcile of limitador configuration in the kuadrant CR into the limitador CR.

### DIFF
--- a/controllers/kuadrant_controller_test.go
+++ b/controllers/kuadrant_controller_test.go
@@ -6,13 +6,14 @@ import (
 	"reflect"
 	"time"
 
-	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
-	"github.com/kuadrant/kuadrant-operator/pkg/common"
 	"github.com/kuadrant/limitador-operator/api/v1alpha1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
 )
 
 var _ = Describe("Kuadrant controller", func() {
--- a/controllers/kuadrant_controller_test.go
+++ b/controllers/kuadrant_controller_test.go
@@ -4,18 +4,19 @@ package controllers
 
 import (
 	"context"
+	"reflect"
+	"time"
+
 	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
 	"github.com/kuadrant/kuadrant-operator/pkg/common"
 	"github.com/kuadrant/limitador-operator/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


controllers/kuadrant_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+		AfterEach(func(ctx SpecContext) {
+			DeleteNamespaceCallbackWithContext(ctx, testNamespace)
+		}, afterEachTimeOut)
+		It("Copy configuration from Kuadrant CR to Limitador CR", func(ctx SpecContext) {
+			kObj := &kuadrantv1beta1.Kuadrant{}
+			lObj := &v1alpha1.Limitador{}
+
+			Eventually(func() bool {
+				err := k8sClient.Get(context.Background(), client.ObjectKey{Name: kuadrant, Namespace: testNamespace}, kObj)
```
KevFan, line 41:
```suggestion
				err := k8sClient.Get(ctx, client.ObjectKey{Name: kuadrant, Namespace: testNamespace}, kObj)
```

nit: perhaps use ctx from `SpecContext` thoughout instead of creating new context ? Also do we want define a `SpecTimeout` for the `SpecContext` so that `Eventually` using this context will timeout after a certain point if it gets stuck ?

Boomatang: It's not a nit. I missed that one when updating the PR to use `Spec'Context`. Good catch.

KevFan: Cool, any thoughts on the `SpecTimeout`'s ?

Boomatang: I like the idea of the `SpecTimeout`. It is good to have. The only problem that I see is we don't record the length of each test running. So our timeouts are a best guess value. In the future if we start hitting the `SpecTimeout` do we just up the value or track down the reason for then hitting it. This is a cause of missing metrics about our tests.

controllers/kuadrant_controller_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"reflect"
+	"time"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
+	"github.com/kuadrant/limitador-operator/api/v1alpha1"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
KevFan, line 15:
```suggestion
	"github.com/kuadrant/limitador-operator/api/v1alpha1"
	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
	"k8s.io/utils/ptr"
	"sigs.k8s.io/controller-runtime/pkg/client"
	
	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
	"github.com/kuadrant/kuadrant-operator/pkg/common"
```
nit: separate kuadrant imports

KevFan (approved):
Looks good overall, left some minor nits / comments

PR #405, opened 2024-01-24
Disable fail-fast for nightly integration tests

adam-cattermole:
When the nightly integration tests run if one of the sail operator / istioctl fails the other test in the matrix is cancelled due to `fail-fast: true`. Setting to false so we run both integration tests no matter the outcome of the first to complete.

See latest nightly here where the integration test for sail cancelled the integration test for istioctl:
https://github.com/Kuadrant/kuadrant-operator/actions/runs/7634245121

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -53,6 +53,7 @@ jobs:
         exclude:
           - istio-type: sail
             pr-event: true
+      fail-fast: false
     runs-on: ubuntu-latest
     env:
       KIND_CLUSTER_NAME: kuadrant-test


Boomatang (commented):
For nightly tests this makes sense. Does this affect standard PR checks? On PRs it might be better to fail fast and not using the extra resources.

adam-cattermole:
At the moment it doesn't affect PR's as only one of the matrix options is run (`istioctl`). In my opinion even if we reintroduce the integration tests on sail operator for PR's I think it makes more sense for both to run in case a change breaks integration in one and not the other?

KevFan (approved):
LGTM !

PR #407, opened 2024-01-29
Change scheduled build naming

pehala:
# Summary

This PR aims to solve these two issues:

- Since the scheduled build actions effectively became `nightly` builds, update the naming to make it more apparent in both tags and Action name
- If the kuadrant-operator did not change between builds, but components did, the tag would stay the same but the contents would change
   - This will cause more tags in the quay (literally for each day) 

I still need to do a bit of testing to make sure the change is correct but the idea stands and I would like to know your opinion

### DIFF
--- a/.github/workflows/build-images-nightly.yaml
+++ b/.github/workflows/build-images-nightly.yaml
@@ -1,4 +1,4 @@
-name: Schedule build with latest image SHA versions
+name: Schedule nightly build with latest image SHA versions
 
 on:
   schedule:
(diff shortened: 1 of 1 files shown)


adam-cattermole:
Just a thought based off your comment (not within scope for this PR):

> This will cause more tags in the quay (literally for each day)

I wonder whether we want to look into using the `LABEL quay.expires.after=` again, in particular for Nightly builds to only keep the tags for a certain period of time.

pehala:
> I wonder whether we want to look into using the `LABEL quay.expires.after=` again, in particular for Nightly builds to only keep the tags for a certain period of time.

I think that is a great idea

didierofrivia (approved):
Let's keep in mind we could have several different named builds but with the exact same dependencies... maybe we could optimize in the future. 

🥟

Issue #349, opened 2023-11-30
Authpolicy status should not report error when authconfig resource is not required.


eguzki:
Given the following topology:

```
Gateway prod-web

HTTPRoute petstore-route   ->  gateway prod-web

AuthPolicy gw-auth -> Gateway prod-web

Authpolicy petstore-route -> HTTPRoute petstore-route
```

The status field of the `gw-auth` policy reports as "Not Available"

```yaml
 status:                                           
  conditions:                                     
  - lastTransitionTime: "2023-11-30T15:40:34Z"    
    message: AuthScheme is not ready yet          
    reason: AuthSchemeNotReady                    
    status: "False"                               
    type: Available                               
  observedGeneration: 2                           
```

This is wrong because the `gw-auth` policy does not need an AuthConfig resource as there is no "free" HTTPRoutes.  It should report as "Available" instead.

KevFan:
@eguzki @guicassolato Given the RFC on https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md, I'm interested on how this scenario would work for the `Enforced` condition for the Gateway AuthPolicy `gw-auth` in the example.

To me, there is 2 options from https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md#conditions:
1. This could be `Enforced: True: Enforced` - that this policy has been successfully enforced even though all httproutes from this Gateway has their own AuthPolicy

or alternatively:

2. This could be `Enforced: False: Overridden` - This policy has been overridden as all httproutes from this Gateway has their own AuthPolicy

I think currently in my head no 2) makes more sense, the gateway AuthPolicy is not enforced on any underlying HTTPRoute and has been overridden by the AuthPolicy on the HTTPRoute attached to the Gateway. If a free httproute using this gateway becomes available with no authpolicy, it should then go back to no 1) as its successfully enforcing it's AuthPolicy on a httproute

Interested in your thoughts on this on if you agree or not 🤔

didierofrivia:
@KevFan @guicassolato I share your view point on this, following the states conditions listed in https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md#conditions, the use case should reflect `Enforced: False: Overridden`

eguzki:
Sorry for being late to the party. I just wanted to tell that I agree with the decision taken here. "Enforced" and "Overriden" at the same time represents better the actual configuration applied.

Issue #290, opened 2023-11-06
AuthPolicy status


guicassolato:
- [x] Implementation of [RFC 0004](https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md) for the AuthPolicy
- [ ] #690
  - Maybe a list of auth rules by name that will apply to which HTTPRouteRule (?)

KevFan:
@guicassolato for task 2, it seems like there is some description missing "Update the status stanza of the AuthPolicy about route selectors that failed to" 🤔

guicassolato:
> @guicassolato for task 2, it seems like there is some description missing "Update the status stanza of the AuthPolicy about route selectors that failed to" 🤔

Thanks @KevFan. Updated the description.

guicassolato:
Extracted the detail into a separate issue so this one can be closed as done

PR #411, opened 2024-01-31
feat: auth policy enforced condition

KevFan:
# Description
Part of https://github.com/Kuadrant/kuadrant-operator/issues/290 https://github.com/Kuadrant/architecture/issues/38
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/349

Add `Enforced` condition type to AuthPolicy. 

# Verification
Functionality is generally already tested with the integration tests added.

To verify manually instead:

## Setup
* Checkout this branch
* Deploy using `make local-setup`
* Create HTTPRoute
```sh
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
EOF
```
* Create AuthPolicy
```sh
kubectl apply -f - <<EOF
---
kind: AuthPolicy
metadata:
  name: auth0
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authorization:
      deny-all:
        opa:
          rego: "allow = false"
    response:
      unauthorized:
        headers:
          "content-type":
            value: application/json
        body:
          value: |
            {
              "error": "Forbidden",
            }
EOF
```
* Verify status:
  * `Accepted`  - `True`
  * `Enforced` - `False`
```sh
kubectl get authpolicy -A -o wide
kubectl get authpolicy auth0 -o yaml | yq '.status'
``` 
image
* Deploy Kuadrant CR
```
kubectl apply -f - <<EOF
---                      
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```
* Verify status:
  * `Accepted`  - `True`
  * `Enforced` - `True`
```sh
kubectl get authpolicy -A -o wide
kubectl get authpolicy auth0 -o yaml | yq '.status'
``` 
image

### DIFF
--- a/api/v1beta2/ratelimitpolicy_types.go
+++ b/api/v1beta2/ratelimitpolicy_types.go
@@ -168,7 +168,8 @@ var _ common.KuadrantPolicy = &RateLimitPolicy{}
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:metadata:labels="gateway.networking.k8s.io/policy=direct"
-// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[0].reason`,description="RateLimitPolicy Status",priority=2
+// +kubebuilder:printcolumn:name="Accepted",type=string,JSONPath=`.status.conditions[?(@.type=="Accepted")].status`,description="RateLimitPolicy Accepted",priority=2
+// +kubebuilder:printcolumn:name="Enforced",type=string,JSONPath=`.status.conditions[?(@.type=="Enforced")].status`,description="RateLimitPolicy Enforced",priority=2
 // +kubebuilder:printcolumn:name="TargetRefKind",type="string",JSONPath=".spec.targetRef.kind",description="Type of the referenced Gateway API resource",priority=2
 // +kubebuilder:printcolumn:name="TargetRefName",type="string",JSONPath=".spec.targetRef.name",description="Name of the referenced Gateway API resource",priority=2
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
--- a/controllers/authpolicy_authconfig.go
+++ b/controllers/authpolicy_authconfig.go
@@ -92,6 +92,7 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 		if len(rules) == 0 {
 			logger.V(1).Info("no httproutes attached to the targeted gateway, skipping authorino authconfig for the gateway authpolicy")
 			common.TagObjectToDelete(authConfig)
+			r.setOverriddenPolicy(ap)
 			return authConfig, nil
 		}
 		route = &gatewayapiv1.HTTPRoute{
@@ -101,6 +102,9 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 			},
 		}
 	}
+
+	// AuthPolicy is not overridden if code execution flow reaches here
+	r.removeOverriddenPolicy(ap)
 
 	// hosts
 	authConfig.Spec.Hosts = hosts
--- a/controllers/authpolicy_authconfig.go
+++ b/controllers/authpolicy_authconfig.go
@@ -92,7 +92,7 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 		if len(rules) == 0 {
 			logger.V(1).Info("no httproutes attached to the targeted gateway, skipping authorino authconfig for the gateway authpolicy")
 			common.TagObjectToDelete(authConfig)
-			r.setOverriddenPolicy(ap)
+			r.OverriddenPolicyMap.SetOverriddenPolicy(ap)
 			return authConfig, nil
 		}
 		route = &gatewayapiv1.HTTPRoute{
@@ -103,8 +103,8 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 		}
 	}
 
-	// AuthPolicy is not overridden if code execution flow reaches here
-	r.removeOverriddenPolicy(ap)
+	// AuthPolicy is not overridden if we still need to create an AuthConfig for it
+	r.OverriddenPolicyMap.RemoveOverriddenPolicy(ap)
 
 	// hosts
 	authConfig.Spec.Hosts = hosts
--- a/controllers/authpolicy_status.go
+++ b/controllers/authpolicy_status.go
@@ -11,6 +11,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	v1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -64,8 +65,8 @@ func (r *AuthPolicyReconciler) calculateStatus(ctx context.Context, ap *api.Auth
 		ObservedGeneration: ap.Status.ObservedGeneration,
 	}
 
-	availableCond := r.acceptedCondition(ap, specErr)
-	meta.SetStatusCondition(&newStatus.Conditions, *availableCond)
+	acceptedCond := r.acceptedCondition(ap, specErr)
+	meta.SetStatusCondition(&newStatus.Conditions, *acceptedCond)
 
 	// Do not set enforced condition if Accepted condition is false
 	if meta.IsStatusConditionFalse(newStatus.Conditions, string(gatewayapiv1alpha2.PolicyReasonAccepted)) {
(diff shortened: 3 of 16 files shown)


controllers/authpolicy_status.go
```diff
@@ -71,27 +58,83 @@ func (r *AuthPolicyReconciler) reconcileStatus(ctx context.Context, ap *api.Auth
...
+		Namespace: policy.Namespace,
+		Name:      authConfigName(apKey),
+	}
+	authConfig := &authorinoapi.AuthConfig{}
+	err := r.GetResource(ctx, authConfigKey, authConfig)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return false, false, fmt.Errorf("failed to get AuthConfig: %w", err)
+		}
+		return true, common.IsTargetRefGateway(policy.GetTargetRef()), nil
```
guicassolato, line 121: If we cannot find the AuthConfig, then the AuthConfig is ready? Twisted as it sounds, I understand this is what we want. Maybe add a comment?

didierofrivia: This would happen when the AuthPolicy hasn't reconciled yet the AuthConfig, right?

guicassolato:
Verification steps work as expected 🎉

After the last step, I've added:

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: gw-auth
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    authorization:
      allow-all:
        opa:
          rego: "allow = true"
EOF
```

Then:

```sh
kubectl get authpolicy -A -o wide
```
```
NAMESPACE      NAME      ACCEPTED   ENFORCED   TARGETREFKIND   TARGETREFNAME          AGE
default        auth0     True       True       HTTPRoute       toystore               19m
istio-system   gw-auth   True       False      Gateway         istio-ingressgateway   7s
```

<br/>

```sh
kubectl get authpolicy gw-auth -n istio-system -o yaml | yq '.status'
```
```yaml
conditions:
  - lastTransitionTime: "2024-02-13T13:23:05Z"
    message: AuthPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-02-13T13:23:05Z"
    message: AuthPolicy is overridden by [{"Namespace":"default","Name":"auth0"}]
    reason: Overridden
    status: "False"
    type: Enforced
observedGeneration: 1
```

<br/>

```sh
kubectl delete authpolicy auth0
```

<br/>

```sh
kubectl get authpolicy -A -o wide
```
```
NAMESPACE      NAME      ACCEPTED   ENFORCED   TARGETREFKIND   TARGETREFNAME          AGE
istio-system   gw-auth   True       True       Gateway         istio-ingressgateway   3m37s
```

<br/>

```sh
kubectl get authpolicy gw-auth -n istio-system -o yaml | yq '.status'
```
```yaml
conditions:
  - lastTransitionTime: "2024-02-13T13:23:05Z"
    message: AuthPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2024-02-13T13:26:34Z"
    message: AuthPolicy has been successfully enforced
    reason: Enforced
    status: "True"
    type: Enforced
observedGeneration: 1
```

controllers/authpolicy_status.go
KevFan: Yeah, not too happy with this logic either to detect that an AuthConfig is not required from an AuthPolicy in this overridden scenario 🤔 

Re-thinking this, ideally we should be able to infer this from the AuthPolicy struct from a value set which denotes this scenario here:
https://github.com/Kuadrant/kuadrant-operator/blob/133645e9080f55022eff7c67d22da68424bce04a/controllers/authpolicy_authconfig.go#L92-L95

Instead I'm thinking we can add an annotation to denote that a gateway AuthConfig is required or not `gwAuthConfigRequired: true|false` instead of this twisted logic and also to prevent getting this is code logic if we are actually waiting for Authconfig to be created 🤔 

What do you guys think ?

guicassolato: > ideally we should be able to infer this from the AuthPolicy struct from a value set which denotes this scenario here:

Thing is that it's not the structure of the Gateway AuthPolicy itself that says that, but the presence of other AuthPolicies targeting the HTTPRoutes, as you know.

The LOC you pointed is indeed where this decision happens. Signaling (possibly by annotating, but not necessarily) from there, all the way to the step where we set the status, sounds a good idea to me.

> I'm thinking we can add an annotation to denote that a gateway AuthConfig is required or not 

Do you mean an annotation in the AuthPolicy CR? Would that be actually pushed to etcd or just to carry in-memory the information to the set the status later on?

In Authorino, we implemented an approach of writing to a shared memory space that the status controller reads from: https://github.com/Kuadrant/authorino/blob/744220a503dbf231f104652f16084bb51feef9a3/controllers/auth_config_controller.go#L83

> `gwAuthConfigRequired: true|false`

I'm not yet 100% convinced about the annotation plan, but if we do that, please let's use a name such as `kuadrant.io/authconfig-required` instead please.

KevFan: > Do you mean an annotation in the AuthPolicy CR? Would that be actually pushed to etcd or just to carry in-memory the information to the set the status later on?
> 

Yes, annotation in the AuthPolicy CR that is kept in memory since this information is implied in the AuthPolicy status once the `Overridden` reason is set. I don't think there is a need to hold this information again in the AuthPolicy CR via annotations unless it proves useful for end users / consumers of AuthPolicy CR to know that there is no AuthConfig created for an AuthPolicy. Although I feel there is no major harm in pushing this annotation to etcd either 🤔 Using annotations was just an easy way of passing this information down to the Enforced status logic

> In Authorino, we implemented an approach of writing to a shared memory space that the status controller reads from: https://github.com/Kuadrant/authorino/blob/744220a503dbf231f104652f16084bb51feef9a3/controllers/auth_config_controller.go#L83
> 
> > `gwAuthConfigRequired: true|false`
> 
> I'm not yet 100% convinced about the annotation plan, but if we do that, please let's use a name such as `kuadrant.io/authconfig-required` instead please.

Thanks ! I'll have a look at Authorino to see if something similar is the preferred approach 👍

controllers/authpolicy_authconfig.go
```diff
@@ -102,6 +103,9 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 		}
 	}
 
+	// AuthPolicy is not overridden if code execution flow reaches here
```
guicassolato, line 106:
```suggestion
	// AuthPolicy is not overridden if we still need to create an AuthConfig for it
```

controllers/authpolicy_status.go
```diff
@@ -82,10 +83,19 @@ func (r *AuthPolicyReconciler) acceptedCondition(policy common.KuadrantPolicy, s
...
 
-// enforcedCondition checks if the provided AuthPolicy is enforced based on the status of the associated AuthConfig and Gateway.
+// enforcedCondition checks if the provided AuthPolicy is enforced, ensuring it is properly configured and applied based
+// on the status of the associated AuthConfig and Gateway.
 func (r *AuthPolicyReconciler) enforcedCondition(ctx context.Context, policy *api.AuthPolicy, targetNetworkObject client.Object) *metav1.Condition {
 	logger, _ := logr.FromContext(ctx)
-	authConfigReady, gwPolicyOverridden, err := r.checkAuthConfigAndGateway(ctx, policy)
+
+	// Check if the policy is overridden
+	if r.isPolicyOverridden(policy) {
```
guicassolato, line 92: This assumes that computing the desired AuthConfig (occasionally none), then marking the AuthPolicy as overridden and finally calculating the Enforced condition happen in a synchronous way, which is a fair assumption because it's the logical path. However, a simple goroutine added somewhere along this flow could break this logic, which spreads across multiple functions. So maybe add a comment here explaining how critical it is to keep this synchronous? WDYT?

api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -168,7 +168,8 @@ var _ common.KuadrantPolicy = &RateLimitPolicy{}
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:metadata:labels="gateway.networking.k8s.io/policy=direct"
-// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.conditions[0].reason`,description="RateLimitPolicy Status",priority=2
+// +kubebuilder:printcolumn:name="Accepted",type=string,JSONPath=`.status.conditions[?(@.type=="Accepted")].status`,description="RateLimitPolicy Accepted",priority=2
+// +kubebuilder:printcolumn:name="Enforced",type=string,JSONPath=`.status.conditions[?(@.type=="Enforced")].status`,description="RateLimitPolicy Enforced",priority=2
```
guicassolato, line 172: Though the title says "auth policy", it's good to have this here as well, for completeness 👍

main.go
```diff
@@ -22,6 +22,7 @@ import (
 	"os"
 	"runtime"
 
+	"github.com/kuadrant/kuadrant-operator/pkg/common"
```
guicassolato, line 25: nitpick: I fail to see the logic behind how we group these dependencies. Maybe in another PR we can work on this and give them some order people can reason about. E.g. Kuadrant-controlled packages always grouped together, or layered based on proximity to the Golang std lib.

guicassolato (approved):
Nice job!

I've tested it again after last refactoring and still working perfectly fine.

main.go
KevFan: Yeah, grouping is inconsistent in a few places. I think the main reason, at least for me, for this inconsistency is that these are frequently auto added by an IDE but ignores the groups that have been already been logically broken up previously. Goland IDE seems to always auto add a dependency to the first group list after the Golang std libs 🤔

mikenairn: fyi, i had opened a PR to add the openshift goimports to keep module grouping consistent https://github.com/Kuadrant/kuadrant-operator/pull/415. Backed it out as we seem to already use `goimports` in CI tests, but maybe we could look into it more.

KevFan: +1

PR #415, opened 2024-02-07
Add goimports

mikenairn:
Something we added to MGC a while back which i think would be good here https://github.com/Kuadrant/multicluster-gateway-controller/pull/365

Adds new Make targets to sort and group project imports `make imports` and verify imports are sorted and grouped `make verify-imports`. 

```
make help | grep imports
  imports          Run openshift goimports against code.
  verify-imports   Verify go imports are sorted and grouped correctly.
  openshift-goimports  Download openshift-goimports locally if necessary
```

Sorting and grouping based on [openshift-goimports]( https://github.com/openshift-eng/openshift-goimports) defaults, with all kuadrant* modules added as an intermediate so they get grouped nearer the modules imports. Script inspired by KCP and other openshift projects.

From openshift-goimports docs:
```
Organizes Go imports into the following groups:
  
  standard - Any of the Go standard library packages
  other - Anything not specifically called out in this list
  kubernetes - Anything that starts with k8s.io
  openshift - Anything that starts with github.com/openshift
  intermediates - Optional list of groups
  module - Anything that is part of the current module
```

Also adds job to CI GH action to execute the `verify-imports` command on all PRs.


mikenairn:
CI already has a check using `goimports` so closing this for now

PR #416, opened 2024-02-08
Unified kuadrant - Add TLSPolicy and DNSPolicy

mikenairn:
* Adds the DNSPolicy [api](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/apis/v1alpha1/dnspolicy_types.go) and [controllers](https://github.com/Kuadrant/multicluster-gateway-controller/tree/main/pkg/controllers/dnspolicy) from MGC 
* Adds the TLSPolicy [api](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/pkg/apis/v1alpha1/tlspolicy_types.go) and [controllers](https://github.com/Kuadrant/multicluster-gateway-controller/tree/main/pkg/controllers/tlspolicy) from MGC 
* Removes the policy-controller
* Adds a new dependency for DNS support [dns-operator](https://github.com/Kuadrant/dns-operator)

As much as possible the DNS and TLS policy code has not changed from what is currently in MGC, but has been modified to fit in with the kuadrant operator code base. The only noticeable difference is the status of the polices is now different, as it now shows a  "Status" instead of "Ready":

```
kubectl get dnspolicy -A -o wide                                                                                                                                                                                                                           
NAMESPACE                NAME       STATUS     TARGETREFKIND   TARGETREFNAME   AGE                                                                             
multi-cluster-gateways   prod-web   Accepted   Gateway         prod-web        27m
```

**Verification**


```
make local-setup
kubectl get deployments -n kuadrant-system
NAME                                    READY   UP-TO-DATE   AVAILABLE   AGE
authorino-operator                      1/1     1            1           7m3s
authorino-webhooks                      1/1     1            1           7m3s
dns-operator-controller-manager         1/1     1            1           7m3s
kuadrant-operator-controller-manager    1/1     1            1           3m48s
limitador-operator-controller-manager   1/1     1            1           7m3s
```

Run though DNS and TLS user guides:

* https://github.com/Kuadrant/kuadrant-operator/blob/6f0668ba8402da9c5d0612aa92febf4b3f035a78/doc/user-guides/gateway-dns.md
* https://github.com/Kuadrant/kuadrant-operator/blob/6f0668ba8402da9c5d0612aa92febf4b3f035a78/doc/user-guides/gateway-tls.md


Makefile
```diff
@@ -137,17 +139,26 @@ LIMITADOR_OPERATOR_GITREF = $(LIMITADOR_OPERATOR_BUNDLE_VERSION)
...
-POLICY_CONTROLLER_VERSION = 0.0.0
-POLICY_CONTROLLER_GITREF = main
-else ifeq (true,$(policy_controller_is_semantic))
-POLICY_CONTROLLER_GITREF = v$(POLICY_CONTROLLER_VERSION)
+## kuadrant dns
+#ToDo Pin this version once we have an initial release of kuadrant dns
+KUADRANT_DNS_OPERATOR_VERSION ?= latest
+KUADRANT_DNS_OPERATOR_VERSION = unified_kuadrant #ToDo(mnairn) Remove this before merge
+
+limitador_bundle_is_semantic := $(call is_semantic_version,$(KUADRANT_DNS_OPERATOR_VERSION))
```
maleck13, line 147: should this be `dns_bundle_is_semantic` ?

Makefile
```diff
@@ -137,17 +139,26 @@ LIMITADOR_OPERATOR_GITREF = $(LIMITADOR_OPERATOR_BUNDLE_VERSION)
...
+ifeq (latest,$(KUADRANT_DNS_OPERATOR_VERSION))
+KUADRANT_DNS_OPERATOR_BUNDLE_VERSION = 0.0.0
+KUADRANT_DNS_OPERATOR_BUNDLE_IMG_TAG = latest
+KUADRANT_DNS_OPERATOR_GITREF = main
+else ifeq (true,$(limitador_bundle_is_semantic))
```
maleck13, line 152: same as above

api/v1alpha1/dnspolicy_types.go
```diff
@@ -0,0 +1,346 @@
...
+	Status DNSPolicyStatus `json:"status,omitempty"`
+}
+
+func (p *DNSPolicy) GetWrappedNamespace() gatewayapiv1.Namespace {
+	return gatewayapiv1.Namespace(p.Namespace)
+}
+
+func (p *DNSPolicy) GetRulesHostnames() []string {
+	//TODO implement me
+	panic("implement me")
```
maleck13, line 156: guessing this was already here. Would it be better to return an empty slice or do we want to keep the panic

bundle.Dockerfile
```diff
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=kuadrant-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.32.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.28.1
```
maleck13, line 9: any reason for the step down in versions?

Makefile
mikenairn: yep, fixed now

api/v1alpha1/dnspolicy_types.go
mikenairn: fixed

Makefile
mikenairn: fixed

bundle.Dockerfile
mikenairn: fixed now, i had an old operator-sdk version in my ./bin directory

api/v1alpha1/dnspolicy_types.go
mikenairn: yeah probably not, we can have return an empty array

maleck13:
So I have done a scan of the code and other files. All looks as expected. I will try run it locally to verify. But I don't think we need to spend too much time on this PR as it is basically just code moving over.

@alexsnaps @eguzki what are your thoughts here?

maleck13:
@eguzki would like thoughts on the code coverage. We have focused more on the integration tests than unit tests. So although the code bot says its not covered, it is likely covered by an integratation rather than unit. What is the expectation here for kuadrant operator. Do we need to up the unit coverage before merge, or are we happy to merge knowing the coverage is not reflecting the integration tests?

mikenairn:
> @eguzki would like thoughts on the code coverage. We have focused more on the integration tests than unit tests. So although the code bot says its not covered, it is likely covered by an integratation rather than unit. What is the expectation here for kuadrant operator. Do we need to up the unit coverage before merge, or are we happy to merge knowing the coverage is not reflecting the integration tests?

I'm looking into improving the coverage, there are still tests that have not been added. Also looks like it is including integration tests in the coverage report, lack of TLS tests are probably bringing it down a good bit.

alexsnaps:
I wouldn't over sweat the Codecov stuff… If you know it's covered by integration tests, that's good enough to me certainly

.github/workflows/update-stored-dependencies-version.yaml
```diff
@@ -24,7 +24,6 @@ jobs:
           - limitador
           - limitador-operator
           - wasm-shim
-          - policy-controller
```
adam-cattermole, line 27: I think we need to add the `dns-operator` here so the github actions variables are updated to be used in the scheduled builds

didierofrivia: Right, I reckon this will be also needed by QA, keeping the latest hash stored or in the end we will be moving to date based? Maybe @pehala could clarify. In the case we still need to keep track of the latest SHAs, there's also some manual work to do, which is adding the action variable in our organization account. I could show you how to or do it myself @mikenairn

api/v1alpha1/dnspolicy_types.go
```diff
@@ -0,0 +1,345 @@
...
+func (p *DNSPolicy) GetTargetRef() gatewayapiv1alpha2.PolicyTargetReference {
+	return p.Spec.TargetRef
+}
+
+func (p *DNSPolicy) Kind() string { return p.TypeMeta.Kind }
+
+// Validate ensures the resource is valid. Compatible with the validating interface
+// used by webhooks
+func (p *DNSPolicy) Validate() error {
+	if p.Spec.TargetRef.Group != "gateway.networking.k8s.io" {
```
adam-cattermole, line 167: nit:
```suggestion
	if p.Spec.TargetRef.Group != gatewayapiv1.GroupName {
```

config/samples/kuadrant_v1alpha1_dnspolicy.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: kuadrant.io/v1alpha1
+kind: DNSPolicy
+metadata:
+  name: dnspolicy-sample
+spec:
+  targetRef:
+    name: example-gateway
+    group: ""
```
adam-cattermole, line 9: I see the [original sample](https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/config/samples/dnspolicy.yaml) didn't have a group set, is there a reason why this isn't the same as the tlspolicy sample:
```suggestion
    group: gateway.networking.k8s.io
```

.github/workflows/update-stored-dependencies-version.yaml
mikenairn: image
Added. Just used the latest commit from main.

config/samples/kuadrant_v1alpha1_dnspolicy.yaml
mikenairn: No reason, added it now.

api/v1alpha1/dnspolicy_types.go
```diff
@@ -0,0 +1,343 @@
...
+// Validate ensures the resource is valid. Compatible with the validating interface
+// used by webhooks
+func (p *DNSPolicy) Validate() error {
+	if p.Spec.TargetRef.Group != gatewayapiv1.GroupName {
+		return fmt.Errorf("invalid targetRef.Group %s. The only supported group is gateway.networking.k8s.io", p.Spec.TargetRef.Group)
+	}
+
+	if p.Spec.TargetRef.Kind != ("Gateway") {
+		return fmt.Errorf("invalid targetRef.Kind %s. The only supported kind is Gateway", p.Spec.TargetRef.Kind)
+	}
```
KevFan, line 171: AuthPolicy and RateLimitPolicy switched to use CEL for this part of the validation if you want to follow the same pattern for DNSPolicy

https://github.com/Kuadrant/kuadrant-operator/blob/c831a7005b4bc5e772f6d0f790fb4e2796d79973/api/v1beta2/authpolicy_types.go#L134-L135

api/v1alpha1/tlspolicy_types.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (p *TLSPolicy) Validate() error {
+	if p.Spec.TargetRef.Group != ("gateway.networking.k8s.io") {
+		return fmt.Errorf("invalid targetRef.Group %s. The only supported group is gateway.networking.k8s.io", p.Spec.TargetRef.Group)
+	}
+
+	if p.Spec.TargetRef.Kind != ("Gateway") {
+		return fmt.Errorf("invalid targetRef.Kind %s. The only supported kind is Gateway", p.Spec.TargetRef.Kind)
+	}
```
KevFan, line 154: Same here

KevFan (commented):
Minor suggestion to use CEL validation for part of the existing validation for DNSPolicy and TLSPolicy 🧑‍💻

mikenairn:
> Minor suggestion to use CEL validation for part of the existing validation for DNSPolicy and TLSPolicy 🧑‍💻

Thanks, sounds like a good idea. I'm probably not going to make that change in this PR though. I am however gathering a list of follow on tasks to better align these polices with auth and rlp so will create one for that.

eguzki:
> @eguzki would like thoughts on the code coverage. We have focused more on the integration tests than unit tests. So although the code bot says its not covered, it is likely covered by an integratation rather than unit. What is the expectation here for kuadrant operator. Do we need to up the unit coverage before merge, or are we happy to merge knowing the coverage is not reflecting the integration tests?

Integration tests generate code coverage report. But e2e tests only report about `controllers` golang package. I will look for a fix to get code coverage report for all packages when running e2e tests

pkg/common/k8s_utils.go
```diff
@@ -187,3 +187,19 @@ func FindDeploymentStatusCondition(conditions []appsv1.DeploymentCondition, cond
 
 	return nil
 }
+
+func HasLabel(obj metav1.Object, key string) bool {
+	labels := obj.GetLabels()
+	if labels == nil {
```
eguzki, line 193: nit: no need to check if it is `nil`. Reading from a `nil` map reference is safe. It will always return `false`

eguzki:
Too big and out of context to review. 

If Craig can run verification steps and user guides successfully, then LGTM as well.

maleck13:
@mikenairn I was able to complete the verification steps. I can't curl locally but I beleive this is an issue with my local setup as it stopped working some time ago and so has nothing to do with this change. 
- DNS records created as expected and TLS secret created as expected. 👍

eguzki:
> > @eguzki would like thoughts on the code coverage. We have focused more on the integration tests than unit tests. So although the code bot says its not covered, it is likely covered by an integratation rather than unit. What is the expectation here for kuadrant operator. Do we need to up the unit coverage before merge, or are we happy to merge knowing the coverage is not reflecting the integration tests?
> 
> Integration tests generate code coverage report. But e2e tests only report about `controllers` golang package. I will look for a fix to get code coverage report for all packages when running e2e tests

Done in https://github.com/Kuadrant/kuadrant-operator/pull/425

api/v1alpha1/dnspolicy_types.go
adam-cattermole: There's actually a couple other cases in api/v1alpha1:
- https://github.com/Kuadrant/kuadrant-operator/blob/unified_kuadrant/api/v1alpha1/dnspolicy_types.go#L292
- https://github.com/Kuadrant/kuadrant-operator/blob/unified_kuadrant/api/v1alpha1/tlspolicy_types.go#L148
- https://github.com/Kuadrant/kuadrant-operator/blob/unified_kuadrant/api/v1alpha1/tlspolicy_types.go#L191

There are other cases in the codebase (not introduced by this change) but they are addressed by #396 I think

api/v1alpha1/dnspolicy_types.go
mikenairn: Created issue to track this https://github.com/Kuadrant/kuadrant-operator/issues/428

api/v1alpha1/tlspolicy_types.go
mikenairn: As above https://github.com/Kuadrant/kuadrant-operator/issues/428

didierofrivia:
I've successfully checked that the TLS and DNS records are created correctly. Also, I've checked that the TLS user guide is working on Linux... on macOS one needs to have running [docker-mac-net-connect](https://github.com/chipmk/docker-mac-net-connect) since docker for mac does not expose container networks directly on the macOS host, might want to add a note to the user guides.

mikenairn:
> I've successfully checked that the TLS and DNS records are created correctly. Also, I've checked that the TLS user guide is working on Linux... on macOS one needs to have running [docker-mac-net-connect](https://github.com/chipmk/docker-mac-net-connect) since docker for mac does not expose container networks directly on the macOS host, might want to add a note to the user guides.

I can't get a port forward option to work with the TLS user guide, so i think saying you should use  [docker-mac-net-connect](https://github.com/chipmk/docker-mac-net-connect) on a mac somewhere in the guides will have to do. I don't however have any clue how it works or what is involved with that so can someone give me a sentence to add?

maleck13:
Maybe we add just that for now as a dependency in the docs?
actually it is already there https://docs.kuadrant.io/getting-started-single-cluster/ so we are prob good to go?
https://docs.kuadrant.io/getting-started-multi-cluster/

didierofrivia (approved):
🎖️

eguzki (approved):

PR #421, opened 2024-02-14
remove kind port mappings

eguzki:
### What

Port mappings defined for kind cluster are no longer needed. Hence, removed.

After adding metallb and LoadBalancer service types for the gateways, the gateways get an IP that is not the same as the IP address of the control plane docker container. Thus, port mapping does not work.


didierofrivia (approved):
💣

adam-cattermole (approved):
Nice, lgtm!

PR #447, opened 2024-03-04
Rate limiting wasmplugin controller: using GatewayAPI topology (DAG)

eguzki:
## What

Dedicated WASMPlugin controller for the rate limiting policy. The controller implementation is based on a new approach using GatewayAPI Topology data structure [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph) to represent the GatewayAPI resources.

Supersedes #317  

### Controller main object type
The type of the main object being *reconciled* is the `Gateway`. There is one-to-one mapping of gateway and wasmplugin. Another approach would be having the WASMPlugin as the main object type. In that case, it would be needed one way to get the gateway name from `{name, namespace}` of the wasmplugin. The event mappers would be more complex. Worth exploring anyway.

### Watched resources and event mappers
The controller watches all resources that affect the configuration (spec) of the rate limit Wasmplugin. I.e. HTTPRoutes, Gateways and RateLimitPolicies. On events of those resources, the event mapper walks the Gateway API topology, via targetRef and parentRefs up to the peak of the topology, as far as kuadrant is concerned today, i.e. the Gateway. In other words, if you change a rate limit policy, it will generate events on all gateways affected. If you change an HTTPRroute, it will generate events on all gateways involved. Interesting to note that if you change the parentRef of a HTTProute, it will generate events on the gateways that **were** targeted and also on the new gateways currently being targeted. That allows doing cleaning for those gateways where a RLP is no longer affecting.

#### The GatewayAPI Topology
On gateway reconciliation event,  the controller builds what it is named "The GatewayAPI Topology". It reads **ALL** HTTPRoutes targeting the given gateway as well as **ALL** rate limit policies in the cluster. With all that info, the controller creates a new instance of the Gateway API topology as well as a set of indexes (another "view" of the topology) that can be queried to provide information to build the desired spec of the rate limiting wasmplugin. Interesting to note here that a field indexer is being added to the manager to index all httproute parenting a gateway. Basically you can use this manager index (not a gateway api topology index), to query "give me all the HTTPRoutes parenting Gateway A". 

The current implementation of the "GatewayAPI Topology" does not use a k8s client to query Gateway API and kuadrant policies. Instead, fetching resources must be done externally and then the GatewayAPI topology receives a list of resources. It could be done otherwise and have generic function with a k8s client to build topology for a given policy of one kind (for example RLP or KAP). The main reason for the current implementation is to decouple GatewayAPI topology and the cluster. Meaning that I can build GatewayAPI topology from a set of resources filtered with some criteria. The use case here is that I want to build a GatewayAPI topology, not for the entire cluster, but only focused on one gateway and all policies and routes involved. 

The implementation assumes "cheap" operations of reading **ALL** the ratelimitpolicies and all routes parenting a given gateway. That assumption comes from the controller runtime k8s client that holds a cache. At least for `list` and `get` operations. That cache is being kept up to date  in the background when there are updates/deletes in the objects triggered by kuadrant controllers or any other third party controllers or any other k8s API server user. Check [for this comment](https://github.com/Kuadrant/kuadrant-operator/pull/317#issuecomment-1963852398) with a verification of this "cheap" operation assumption. Indeed, the resource lists are cached across reconciliation events and controllers.

The topology implementation is based on a pure  [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph) data structure implementation that is totally decoupled from the GatewayAPI relations. On top of that, there is an implementation of the GatewayAPI topology which builds a graph from Gateways and HTTPRoutes. In the future, it can be extended for additional GatewayAPI API types. On top of the GatewayAPI topology, a number of indexes are created which are tailored for the needs of the rate limiting wasmplugin controller. Thus:
```
RL WASMPlugin controller 
       |
       |
       V
Custom topology indexes (views)
       |
       |
       V
GatewayAPI topology (focused on one gateway)
       |
       |
       V
     DAG
```

Additionally:

- [X] Added `github.com/kuadrant/kuadrant-operator/pkg/library/gatewayapi` package with all tools and helpers related to GatewayAPI resources and not directly related to any kuadrant controller logic. This package would be a good candidate to be shared and useful for any third party controller dealing with GatewayAPI resources. This package includes the GatewayAPI topology implementation.
- [X] moved few methods from `common` to  `github.com/kuadrant/kuadrant-operator/pkg/library/gatewayapi` package that IMO are generic enough and decoupled from kuadrant controller's implementation needs. This change caused (small) changes on lots of files :disappointed: 
- [X] moved base reconcilers to `pkg/library/reconcilers` package

TODO: Integration tests:
- [x] Bring some tests from `ratelimitpolicy_controller_test.go` to `rate_limiting_wasmplugin_controller_test.go`
- [x] No routes targeting the gateway
- [x] No routes selected from the HTTPRoute
- [x] HTTPRoute moves from one gateway to another.
- [x] RLP changes target from one route to another
- [x] RLP A targeting Gateway, 1 free route.  New RLP B targeting existing Route. The Wasmplugin should change. The RLP A should not add configuration to the WASMPlugin
- [x] New httproute without RLP: wasmplugin should change when there is RLP targeting the gateway

:left_right_arrow: Next :train: station: Limitador controller using the GatewayAPI Topology


pkg/library/gatewayapi/utils.go
```diff
@@ -0,0 +1,142 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+
+	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
+)
+
+const (
+	GatewayProgrammedConditionType = "Programmed"
+)
```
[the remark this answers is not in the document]
eguzki: updated

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,388 @@
...
+
+	logger.V(1).Info("wasmPluginConfig", "#RLPS", len(rateLimitPolicies))
+
+	// Shallow copy by assignment
+	rateLimitPoliciesSorted := rateLimitPolicies
+
+	// Sort RLPs for consistent comparison with existing objects
+	sort.Sort(kuadrantgatewayapi.PolicyByCreationTimestamp(rateLimitPoliciesSorted))
+
+	for _, policy := range rateLimitPolicies {
```
KevFan, line 178:
```suggestion
	// Sort RLPs for consistent comparison with existing objects
	sort.Sort(kuadrantgatewayapi.PolicyByCreationTimestamp(rateLimitPolicies))

	for _, policy := range rateLimitPolicies {
```

Is there a need for shallow copy in order to sort since it's looped using the intial `rateLimitPolicies` anyway ? 🤔

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,388 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	policies := utils.Map(rlpList.Items, func(p kuadrantv1beta2.RateLimitPolicy) kuadrantgatewayapi.Policy { return &p })
+
+	t, err := kuadrantgatewayapi.NewTopology(
+		kuadrantgatewayapi.WithGateways([]*gatewayapiv1.Gateway{gw}),
+		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To)),
+		kuadrantgatewayapi.WithPolicies(policies),
```
KevFan, line 228:
```suggestion
	policies := utils.Map(rlpList.Items, func(p kuadrantv1beta2.RateLimitPolicy) kuadrantgatewayapi.Policy { return &p })
	routes := utils.Map(routeList.Items, func(r gatewayapiv1.HTTPRoute) *gatewayapiv1.HTTPRoute {return &r})

	t, err := kuadrantgatewayapi.NewTopology(
		kuadrantgatewayapi.WithGateways([]*gatewayapiv1.Gateway{gw}),
		kuadrantgatewayapi.WithRoutes(routes),
		kuadrantgatewayapi.WithPolicies(policies),
```
nit: Goland is highlighting this for me but it compiles and runs so not sure on this one .. 🤷 
image

controllers/rate_limiting_wasmplugin_controller.go
eguzki: not really. Updated

controllers/rate_limiting_wasmplugin_controller.go
eguzki: `make run-lint` (golangci-lint) does not complain. It compiles. 

you sure it is valid lint issue? "cannot use generic function without instantiation" does not sound correct for this example.

eguzki: I am happy to change it to `routes := utils.Map(routeList.Items, func(r gatewayapiv1.HTTPRoute) *gatewayapiv1.HTTPRoute {return &r})` if you find this annoying. Let me know.

KevFan (approved):
This looks good to me anyway! Awesome work! 👍

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,385 @@
...
+func (r *RateLimitingWASMPluginReconciler) Reconcile(eventCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Logger().WithValues("Gateway", req.NamespacedName)
+	logger.Info("Reconciling rate limiting WASMPlugin")
+	ctx := logr.NewContext(eventCtx, logger)
+
+	gw := &gatewayapiv1.Gateway{}
+	if err := r.Client().Get(ctx, req.NamespacedName, gw); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("no gateway found")
+			return ctrl.Result{}, nil
```
guicassolato, line 72: Why not returning the error if the Gateway is not found?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,385 @@
...
+		kuadrantgatewayapi.WithLogger(logger),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return kuadrantgatewayapi.NewTopologyIndexes(t), nil
+}
+
+func (r *RateLimitingWASMPluginReconciler) WASMRateLimitPolicy(ctx context.Context, t *kuadrantgatewayapi.TopologyIndexes, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*wasm.RateLimitPolicy, error) {
```
guicassolato, line 235: Only used in this controller, right?

```suggestion
func (r *RateLimitingWASMPluginReconciler) wasmRateLimitPolicy(ctx context.Context, t *kuadrantgatewayapi.TopologyIndexes, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*wasm.RateLimitPolicy, error) {
```

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,385 @@
...
+	return &wasm.RateLimitPolicy{
+		Name:      client.ObjectKeyFromObject(rlp).String(),
+		Domain:    rlptools.LimitsNamespaceFromRLP(rlp),
+		Hostnames: utils.HostnamesToStrings(hostnames), // we might be listing more hostnames than needed due to route selectors hostnames possibly being more restrictive
+		Service:   common.KuadrantRateLimitClusterName,
+		Rules:     rules,
+	}, nil
+}
+
+func (r *RateLimitingWASMPluginReconciler) RouteFromRLP(ctx context.Context, t *kuadrantgatewayapi.TopologyIndexes, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*gatewayapiv1.HTTPRoute, error) {
```
guicassolato, line 283:
```suggestion
func (r *RateLimitingWASMPluginReconciler) routeFromRLP(ctx context.Context, t *kuadrantgatewayapi.TopologyIndexes, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway) (*gatewayapiv1.HTTPRoute, error) {
```

controllers/rate_limiting_wasmplugin_controller.go
guicassolato: This should fix it:

```suggestion
	policies := utils.Map(rlpList.Items, func(p kuadrantv1beta2.RateLimitPolicy) kuadrantgatewayapi.Policy { return &p })

	t, err := kuadrantgatewayapi.NewTopology(
		kuadrantgatewayapi.WithGateways([]*gatewayapiv1.Gateway{gw}),
		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To[gatewayapiv1.HTTPRoute])),
		kuadrantgatewayapi.WithPolicies(policies),
```

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,385 @@
...
+				Rules:     untargetedRules,
+			},
+		}
+	}
+
+	return route, nil
+}
+
+// addHTTPRouteByGatewayIndexer declares an index key that we can later use with the client as a pseudo-field name,
+// allowing to query all the routes parenting a given gateway
```
guicassolato, line 324:
```suggestion
// allowing to query all the routes parented by a given gateway
```

pkg/library/dag/dag.go
```diff
@@ -0,0 +1,283 @@
...
+
+type Node interface {
+	ID() NodeID
+}
+
+type internalNode struct {
+	id       NodeID
+	node     Node
+	parents  map[NodeID]*internalNode
+	children map[NodeID]*internalNode
```
guicassolato, line 39: Should we use [`orderedmap`](https://pkg.go.dev/github.com/elliotchance/orderedmap/v2) here, so we have consistent responses from funcs like `Parents()`, `Children()`?

pkg/library/dag/dag.go
```diff
@@ -0,0 +1,283 @@
...
+		parents:  make(map[string]*internalNode),
+		children: make(map[string]*internalNode),
+	}
+
+	if _, exists := d.nodes[n.id]; exists {
+		return fmt.Errorf("node %s already exists", n.id)
+	}
+
+	d.nodes[n.id] = n
+	d.populateIndexes(node)
```
guicassolato, line 104: We should add a mutex and wrap this as a critical section.

pkg/library/dag/dag.go
```diff
@@ -0,0 +1,283 @@
...
+	if _, exists := parentInternalNode.children[childInternalNode.id]; exists {
+		return fmt.Errorf("parent node %s already has an edge with child %s", parentInternalNode.id, childInternalNode.id)
+	}
+
+	if _, exists := childInternalNode.parents[parentInternalNode.id]; exists {
+		return fmt.Errorf("child node %s already has an edge with parent %s", childInternalNode.id, parentInternalNode.id)
+	}
+
+	parentInternalNode.children[childInternalNode.id] = childInternalNode
+	childInternalNode.parents[parentInternalNode.id] = parentInternalNode
```
guicassolato, line 142: _Idem:_ https://github.com/Kuadrant/kuadrant-operator/pull/447/files#r1528886767

pkg/library/gatewayapi/topology.go
```diff
@@ -0,0 +1,291 @@
...
+				)
+				return RouteNode{}
+			}
+			return RouteNode(rDAGNode)
+		})
+
+		return GatewayNode{
+			Gateway:          gNode.Gateway,
+			attachedPolicies: gNode.attachedPolicies,
+			routes:           routes,
```
guicassolato, line 272: Routes of the gateways added lazily only when reading the gateways of a topology. Maybe add a comment?

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -0,0 +1,385 @@
...
+	kuadrantgatewayapi "github.com/kuadrant/kuadrant-operator/pkg/library/gatewayapi"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/mappers"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/reconcilers"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
+	"github.com/kuadrant/kuadrant-operator/pkg/rlptools"
+	"github.com/kuadrant/kuadrant-operator/pkg/rlptools/wasm"
+)
+
+const (
+	HTTPRouteParents = ".metadata.route.parent"
```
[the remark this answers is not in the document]
eguzki: the winner is `HTTPRouteGatewayParentField = ".metadata.parentRefs.gateway"`

controllers/rate_limiting_wasmplugin_controller.go
eguzki: because that's not an error. And more specifically, it should not trigger another reconciliation event. If the gateway is not found, there is nothing to do and the controller is saying "do not bother me more about this gateway which I cannot see".

pkg/library/dag/dag.go
eguzki: happy to add them, but it is not needed right now, is it?

pkg/library/dag/dag.go
eguzki: I am aware that the dag package is not thread safe. But, as it is used now, it does not need to be thread-safe. Locks can be added when (if) a single instance is accessed by multiple threads.

pkg/library/gatewayapi/topology.go
eguzki: Sorry, I do not follow. When calling `Gateways()`, the returned object has all the routes for each gateway in the topology. It is not loaded lazily. There is no callback from the user perspective.

eguzki:
@guicassolato ready for a new review

pkg/library/dag/dag.go
guicassolato: This is a basic data structure that can be used to build indexes and elaborate logic on top. Because we already had problems with that in the past, I'm personally more inclined every day to ensuring consistent ordering on those lists. But up to you.

pkg/library/dag/dag.go
```diff
@@ -0,0 +1,283 @@
...
+		id       string
+		parents  map[string]interface{}
+		children []*node
+	}
+
+	type graph struct {
+		nodes []*node
+	}
+
+	// build a mutable simple graph representation out of DAG only for validating purposes
```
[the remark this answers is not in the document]
guicassolato: Preventing cycles is one of the built-in features provided by both of those libs, plus they are both thread-safe by design.

Today we may not need all they offer, agreed. For example, we probably don't care about shortest paths between nodes yet. Still, as long as a lib is efficient enough for what we _do_ need, saves us the effort to have to maintain those functionalities ourselves (simple as they might be), and opens up for more applications in the future, I personally think it's worth considering it.

As for "give me all the HTTPRoutes", isn't that the kind of thing we can get from traversing the graph? If you prefer keeping indexed copies of type-specific nodes – say, for efficiency –, I suppose we could also wrap the DAG within another struct that stores additionally all the indexes we need, maybe using the hashes/ids of the vertices to map to value?

Either way, I leave this as a suggestion for the future. In theory, if we want to, i guess replacing this underlying implementation should be possible to do transparently.

guicassolato (commented):
I've tried this PR with the verification steps from https://github.com/Kuadrant/kuadrant-operator/pull/218. Except for the order of the rules in the WasmPlugin resource which seemed completely random (possibly a preexisting issue), it all works fine.

The `kuadrant-operator-controller-manager` keeps crashing but I don't think it's necessarily related to this PR. Rather, due to some `DNSHealthCheckProbe` issue spotted in the logs. The same does NOT happen in `main` though, so maybe rebasing will fix it?

Other than that, I've left a few comments from before – some reinforced in my latest review. The most important one regards thread-safety of the new data structures.

guicassolato (approved):
Thanks for rebasing, @eguzki!

I don't want to block this any further, so approving it and leaving a summary of things I think we could improve in the future:
1. [thread-safety of the DAG data structure](https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528886767)
2. [consistent order for the lists](https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528882189)
3. [rely on a lib to implement and validate the graph](https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528917694)

eguzki:
Fair points, before merging, I will:

> https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528886767

Leave a comment to highlight that the `dag` package is not thread safe

> https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528882189

Check this out as it should not matter the order of the policies being returned by the DAG. The controller [sorts that list](https://github.com/Kuadrant/kuadrant-operator/blob/wasmplugin-controller/controllers/rate_limiting_wasmplugin_controller.go#L172-L173) of policies afterwards based on policy creation timestamp

> https://github.com/Kuadrant/kuadrant-operator/pull/447#discussion_r1528917694

Create an issue for that

eguzki:
Comments addressed @guicassolato 
Asking for one more review :please:

pkg/rlptools/wasm_utils.go
```diff
@@ -30,7 +31,16 @@ func WasmRules(rlp *kuadrantv1beta2.RateLimitPolicy, route *gatewayapiv1.HTTPRou
 		return rules
 	}
 
-	for limitName := range rlp.Spec.Limits {
+	// Sort RLP limits for consistent comparison with existing wasmplugin objects
+	limitNames := make([]string, 0, len(rlp.Spec.Limits))
```
guicassolato, line 35: This mitigates but does not solve 100%, I'm afraid. _Ceteris paribus_ 🤓, a consistent wasmplugin config also depends on the order of the policies which are indexed for a gateway using a regular Golang map.

eguzki: The order of the policies is already being sorted [here](https://github.com/Kuadrant/kuadrant-operator/blob/a46f05f93176c90455d174476d13aa059d0fcde2/controllers/rate_limiting_wasmplugin_controller.go#L173), isn't it?

eguzki: as I mentioned already
> The controller [sorts that list](https://github.com/Kuadrant/kuadrant-operator/blob/wasmplugin-controller/controllers/rate_limiting_wasmplugin_controller.go#L172-L173) of policies afterwards based on policy creation timestamp

PR #449, opened 2024-03-05
GH-641 enforced condition on dnspolicy

### LINKED ISSUE #454: dnspolicy: Aggregate DNSRecord status
When the DNS Operator performs actions based on the state of a DNS Record, it should update the DNS Record status with results of these actions, similar to how it is currently implemented

The kuadrant-operator should expect all of the DNS Records for a particular policy to eventually have an accepted status (see [here](https://github.com/Kuadrant/dns-operator/issues/34) for more information). Once they are all accepted, this DNS Policy should have it's status set to `enforced`.

- [ ] Add logic to aggregate status to DNS Policy
- [ ] Tests created to cover this aggregated status

makslion:
## What 
Following recent changes with the auth policy and introducing DNS Policy to the `enforced` condition. 
Also closes #454 
## How 
The policy is used to create DNS Records and records are left with the controller owner annotation that mentions the policy. Listing DNS records and finding the one controlled by the policy should be sufficient to indicate that the policy was enforced, also including a tinny change to the test suite to reflect new logic. 
## Verification 
Integration suite passing should be sufficient. One could go through the "typical installation" and observe policy getting `enforced` status.

### DIFF
--- a/controllers/authpolicy_status.go
+++ b/controllers/authpolicy_status.go
@@ -101,16 +101,16 @@ func (r *AuthPolicyReconciler) enforcedCondition(ctx context.Context, policy *ap
 	authConfigReady, err := r.isAuthConfigReady(ctx, policy)
 	if err != nil {
 		logger.Error(err, "Failed to check AuthConfig and Gateway")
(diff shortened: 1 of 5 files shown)


maleck13:
Changes look good to me

maleck13 (approved):

PR #457, opened 2024-03-12
Pin dns operator version v0.1.0

mikenairn:
We intend to make non backward compatible changes in the DNS Operator and having the main branch here track the development branch of DNS Operator will result in broken builds/tests here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,8 +136,7 @@ endif
 LIMITADOR_OPERATOR_BUNDLE_IMG ?= quay.io/kuadrant/limitador-operator-bundle:$(LIMITADOR_OPERATOR_BUNDLE_IMG_TAG)
 
 ## dns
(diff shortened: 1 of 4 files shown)


KevFan (approved):
Looks good to me 👍

didierofrivia:
I'd say that nobody should trust any floating tags and people should go for released versions... However, having a failing main branch is not ideal. I have mixed feeling about this, maybe we should define what is expected from the main branch and if we could start pinning versions to it. Would love to hear what others have to say about this, maybe if we decide to go for versioning and `v.x.y.z-dev` tags for main would make more sense, https://github.com/Kuadrant/architecture/pull/41/files?short_path=eeceecb#diff-eeceecbdf5c1a6fc2cc490b55dd4351651d64c8a740458325511493e52f7616a

eguzki (approved):

PR #487, opened 2024-03-13
codecov disabled for forks

eguzki:
### What

On Codecov GitHub Action v4 Tokenless uploading is unsupported. However, PRs made from forks to the upstream public repos will support tokenless (e.g. contributors to OS projects do not need the upstream repo's Codecov token) [doc](https://github.com/codecov/codecov-action?tab=readme-ov-file#v4-release)

However, even if allowed for public repos, it is being rate limited. When rate limited, the report is not being uploaded and therefore the PR's codecov check fails. 

This PR disables codecov reporting for forks.  When the report is not uploaded, the codecov check is not run, hence does not fail. 

The following is a PR for testing purposes from a fork https://github.com/Kuadrant/kuadrant-operator/pull/488, Note that the codecov check is not in the list of checks. Codecov also shows that no report has been uploaded for the #488 PR
Screenshot 2024-03-13 at 18-08-38 Code coverage done right

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -35,7 +35,11 @@ jobs:
       - name: Run make test
         run: |
           make test-unit
-      - name: Upload unit-test coverage reports to CodeCov # more at https://github.com/codecov/codecov-action
+      - name: Upload unit-test coverage reports to CodeCov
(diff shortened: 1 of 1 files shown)


didierofrivia (approved):
🗡️

PR #492, opened 2024-03-18
Import dns operator crds

philbrookes:
import DNS Operator CRDs and update relevant code.


maleck13:
@philbrookes @mikenairn I am a bit confused why the health checks are removed from DNSPolicy when we are adding in support for AWS? I know we talked about possibly moving them under load balanced section but I don't get why we need to remove them totally?

mikenairn:
> @philbrookes @mikenairn I am a bit confused why the health checks are removed from DNSPolicy when we are adding in support for AWS? I know we talked about possibly moving them under load balanced section but I don't get why we need to remove them totally?

I wouldn't have expected it to be getting removed from the DNSPolicy.

philbrookes:
I suppose we could keep that, I was thinking it would come back when it does something. With this being a feature branch I suppose I don't feel strongly either way though.

philbrookes:
I've added the health check bits back to the DNS Policy on this now.

api/v1alpha1/dnspolicy_types.go
```diff
@@ -214,14 +212,14 @@ type DNSPolicyList struct {
...
-	FailureThreshold          *int                                      `json:"failureThreshold,omitempty"`
-	AdditionalHeadersRef      *kuadrantdnsv1alpha1.AdditionalHeadersRef `json:"additionalHeadersRef,omitempty"`
-	ExpectedResponses         []int                                     `json:"expectedResponses,omitempty"`
-	AllowInsecureCertificates bool                                      `json:"allowInsecureCertificates,omitempty"`
-	Interval                  *metav1.Duration                          `json:"interval,omitempty"`
+	Endpoint                  string           `json:"endpoint,omitempty"`
+	Port                      *int             `json:"port,omitempty"`
+	Protocol                  *string          `json:"protocol,omitempty"`
+	FailureThreshold          *int             `json:"failureThreshold,omitempty"`
+	AdditionalHeadersRef      *string          `json:"additionalHeadersRef,omitempty"`
```
mikenairn, line 219: I'm not sure if it matters but we seem to be changing the type here. Would we be better to add the `AdditionalHeadersRef` type to the types here to keep it compatible? tbh this might change considerably as we go, but i believe you can add headers to the route53 health checks so might be something we want to keep.

controllers/dns_helper.go
```diff
@@ -103,12 +104,12 @@ func gatewayDNSRecordLabels(gwKey client.ObjectKey) map[string]string {
...
+	var endpoints []*externaldns.Endpoint
 
 	//Health Checks currently modify endpoints so we have to keep existing ones in order to not lose health check ids
-	currentEndpoints := make(map[string]*kuadrantdnsv1alpha1.Endpoint, len(dnsRecord.Spec.Endpoints))
-	for _, endpoint := range dnsRecord.Spec.Endpoints {
-		currentEndpoints[endpoint.SetID()] = endpoint
+	currentEndpoints := make(map[string]*externaldns.Endpoint, len(dnsRecord.Spec.Endpoints))
+	for _, ep := range dnsRecord.Spec.Endpoints {
+		currentEndpoints[ep.SetIdentifier] = ep
 	}
```
mikenairn, line 113: `SetIdentifier` could be empty which is why we were using the `SetID` method here. You would probably be better to copy that into here and use that instead to avoid any unexpected consequences https://github.com/Kuadrant/dns-operator/blob/v0.1.0/api/v1alpha1/dnsrecord_types.go#L65-L67

api/v1alpha1/dnspolicy_types.go
philbrookes: My intention was to have this be the name of a secret with additional headers specified in it.

controllers/dns_helper.go
philbrookes: ah ok, I'll do that now.

mikenairn (approved):

controllers/dns_helper.go
```diff
@@ -245,50 +250,50 @@ func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTar
 		endpoints = append(endpoints, clusterEndpoints...)
 
 		//Create lbName CNAME (lb-a1b2.shop.example.com -> default.lb-a1b2.shop.example.com)
-		endpoint = createOrUpdateEndpoint(lbName, []string{geoLbName}, kuadrantdnsv1alpha1.CNAMERecordType, string(geoCode), DefaultCnameTTL, currentEndpoints)
+		ep = createOrUpdateEndpoint(lbName, []string{geoLbName}, kuadrantdnsv1alpha1.CNAMERecordType, string(geoCode), DefaultCnameTTL, currentEndpoints)
 
-		//Deal with the default geo endpoint first
+		//Deal with the default geo externaldns first
```
mikenairn, line 255: nit: global replace i guess, but this should just be `endpoint`

philbrookes: 👍

PR #493, opened 2024-03-18
Add Cluster ID

philbrookes:
merge after: https://github.com/Kuadrant/kuadrant-operator/pull/492


bundle/manifests/kuadrant.io_dnspolicies.yaml
```diff
@@ -68,12 +68,7 @@ spec:
                   the listeners assigned to the target gateway
                 properties:
                   additionalHeadersRef:
-                    properties:
```
maleck13, line 71: removing these as we know not supported?

controllers/dns_helper.go
```diff
@@ -232,7 +231,7 @@ func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTar
 			}
 
 			if len(ipValues) > 0 {
-				clusterLbName := strings.ToLower(fmt.Sprintf("%s.%s", cgwTarget.GetShortCode(), lbName))
+				clusterLbName := strings.ToLower(fmt.Sprintf("%s.%s", clusterID, lbName))
```
maleck13, line 234: was chatting with @mikenairn on this. I think we could still use the short code based on the target plus the cluster id? This would allow for two gateways on a single cluster sharing a host for example

philbrookes: do you mean like: `<clusterID>-<shortcode>.<lbName>`?

controllers/dns_helper.go
```diff
@@ -206,7 +205,7 @@ func (dh *dnsHelper) getSimpleEndpoints(mcgTarget *multicluster.GatewayTarget, h
 // ab2.lb-a1b2.shop.example.com A 192.22.2.3
 // ab3.lb-a1b2.shop.example.com A 192.22.2.4
 
-func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTarget, hostname string, currentEndpoints map[string]*externaldns.Endpoint) []*externaldns.Endpoint {
+func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTarget, hostname string, currentEndpoints map[string]*externaldns.Endpoint, clusterID string) []*externaldns.Endpoint {
```
mikenairn, line 208: The idea here was that `mcgTarget` had everything set on it required to generate endpoints. So in here it would either use `GetName` or `GetShortCode` on the mcgTarget (the outer multicluster gateway), or the cgwTarget(the single cluster gateway).

You'd probably be better to add `clusterID` as the name of the cluster on the [ClusterGateway](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/multicluster/gateway_wrapper.go#L153) so that its used when you call `cgwTarget.GetShortCode()` or `cgwTarget.GetName()`

Also see comment from Craig, the `GetName` method of the cluster target probably needs to include the gw name and namespace similar to [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/multicluster/target.go#L37)  so, `<clusterID>`-`<GW Name>`-`<GW Namespace>`.

pkg/common/k8s_utils.go
```diff
@@ -147,3 +156,23 @@ func GetLabel(obj metav1.Object, key string) string {
...
+	ns := &corev1.Namespace{}
+	err := c.Get(ctx, client.ObjectKey{Name: clusterIDNamespace}, ns)
+	if err != nil {
+		return err, ""
+	}
+	clusterID = generateIDFromUID(ns.UID)
+	return nil, clusterID
+}
+
+func generateIDFromUID(uid types.UID) string {
```
maleck13, line 175: do we want to use md5 what about the short code work that was used before?

controllers/dns_helper.go
maleck13: yeah WDYT? would allow more use cases I think without much overhead

maleck13 (commented):
Some small questions and suggestions before approval

pkg/common/k8s_utils.go
philbrookes: I didn't give it much thought, I'll update it to use the existing base36 function.

controllers/dns_helper.go
```diff
@@ -232,7 +231,7 @@ func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTar
 			}
 
 			if len(ipValues) > 0 {
-				clusterLbName := strings.ToLower(fmt.Sprintf("%s.%s", cgwTarget.GetShortCode(), lbName))
+				clusterLbName := strings.ToLower(fmt.Sprintf("%s-%s.%s", mcgTarget.ClusterName, mcgTarget.GetShortCode(), lbName))
```
mikenairn, line 234: Should this not be `cgwTarget.ClusterName`? mcgTarget has many cluster targets each with a unique name (id). This will likely not work with ocm.

pkg/common/k8s_utils.go
```diff
@@ -147,3 +155,18 @@ func GetLabel(obj metav1.Object, key string) string {
...
+	if clusterID != "" {
+		return nil, clusterID
+	}
+
+	ns := &corev1.Namespace{}
+	err := c.Get(ctx, client.ObjectKey{Name: clusterIDNamespace}, ns)
+	if err != nil {
+		return err, ""
+	}
+	clusterID = multicluster.ToBase36HashLen(string(ns.UID), clusterIDLength)
```
mikenairn, line 170: Probably better to move `ToBase36HashLen` to a more common package if we are using it outside of `multicluster`.

pkg/multicluster/gateway_wrapper.go
```diff
@@ -173,7 +174,7 @@ func (g *GatewayWrapper) GetClusterGateways() []ClusterGateway {
 		}
 		clusterGateways = append(clusterGateways, ClusterGateway{
 			Gateway:     gw,
-			ClusterName: clusterName,
+			ClusterName: g.ClusterID,
```
mikenairn, line 177: This one should not be changed. When a control plane has been used to generate the Gateway status the cluster Names are retrieved from the status. 
This will make all the cluster specific endpoints identical and break the OCM integration.

examples/dnspolicy/script.sh
```diff
@@ -0,0 +1,5 @@
+kubectl create ns gateway
+envsubst < examples/dnspolicy/managedzone.yaml | kubectl apply -f -
+envsubst < examples/dnspolicy/gateway.yaml | kubectl apply -f -
+envsubst < examples/dnspolicy/dnspolicy.yaml | kubectl apply -f -
+envsubst < examples/dnspolicy/application.yaml | kubectl apply -f -
```
mikenairn, line 5: All these example files seem specific to your setup e.g. `pb.hcpapps.net`. Should they be getting checked in?

pkg/multicluster/target.go
```diff
@@ -111,7 +111,7 @@ func (t *ClusterGatewayTarget) GetName() string {
 }
 
 func (t *ClusterGatewayTarget) GetShortCode() string {
-	return ToBase36hash(t.GetName())
+	return fmt.Sprintf("%s-%s", t.ClusterName, ToBase36HashLen(t.Namespace+"-"+t.Name, 7))
```
maleck13, line 114: what is clusterName here? is the id?

maleck13 (commented):
changes look good. Just wonder are you intending on the clusterName being plain?

controllers/dnspolicy_controller_multi_cluster_test.go
```diff
@@ -194,46 +203,46 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
 									"ManagedZoneRef": HaveField("Name", "mz-example-com"),
 									"Endpoints": ConsistOf(
 										PointTo(MatchFields(IgnoreExtras, Fields{
-											"DNSName":       Equal("2w705o.lb-" + lbHash + ".test.example.com"),
+											"DNSName":       Equal(cgwTarget.GetShortCode() + "." + lbHash + ".test.example.com"),
```
mikenairn, line 206: If you use `cgwTarget.GetShortCode()`  in the assertion like this you are never really testing what you expect the DNSName to be. What if there is a bug in  `cgwTarget.GetShortCode()` and it's not generating what we expect?

pkg/multicluster/target.go
```diff
@@ -36,7 +36,7 @@ func (t *GatewayTarget) GetName() string {
 }
 
 func (t *GatewayTarget) GetShortCode() string {
-	return ToBase36hash(t.GetName())
+	return ToBase36HashLen(t.GetName(), 7)
```
mikenairn, line 39: Why a length 7? Was 6 not long enough?

pkg/multicluster/target.go
```diff
@@ -111,7 +111,7 @@ func (t *ClusterGatewayTarget) GetName() string {
 }
 
 func (t *ClusterGatewayTarget) GetShortCode() string {
-	return ToBase36hash(t.GetName())
+	return fmt.Sprintf("%s-%s", t.ClusterName, ToBase36HashLen(t.Namespace+"-"+t.Name, 7))
 }
```
mikenairn, line 115: Does this not need to include the Gateway name, or are we saying per namespace is unique enough? This is being called [here ](https://github.com/Kuadrant/kuadrant-operator/pull/493/files#diff-75ea4f4ad086ceff450395450ca7845c9507114178820b7e625f9221c0a893b1R234) to generate the unique cluster specific endpoint, so it's just using `<clusterName>-<hash of namespace>`, where i thought we were doing `<clusterName>-<GW NS>-<GW Name>` 

Also, since you aren't hashing the result of `GetName`now, cluster names via a control plane (OCM) won't be hashed and will appear in plain text.

mikenairn (changes requested):

controllers/dnspolicy_controller_multi_cluster_test.go
mikenairn: Actually, there is a bug and this change is hiding it, some of these tests should fail because of https://github.com/Kuadrant/kuadrant-operator/pull/493#discussion_r1537350894 

You can see some of the cluster hashes were different previously, now they will all be the same.

examples/dnspolicy/script.sh
philbrookes: no, I'll remove them

controllers/dnspolicy_controller_multi_cluster_test.go
philbrookes: hmm, so what do we expect the hostnames to look like in multicluster? `<plaintext clusterid>-<gateway-hash>...` is what I was getting, is that correct?

pkg/multicluster/target.go
philbrookes: ah, this was unintended, I'll set it back.

pkg/multicluster/gateway_wrapper.go
philbrookes: discussing in https://github.com/Kuadrant/kuadrant-operator/pull/493#discussion_r1537354557

pkg/common/k8s_utils.go
philbrookes: good idea

controllers/dnspolicy_controller_multi_cluster_test.go
philbrookes: ok, I think I understand now, reviewing the code again.

pkg/multicluster/gateway_wrapper.go
```diff
@@ -173,7 +176,7 @@ func (g *GatewayWrapper) GetClusterGateways() []ClusterGateway {
 		}
 		clusterGateways = append(clusterGateways, ClusterGateway{
 			Gateway:     gw,
-			ClusterName: clusterName,
+			ClusterName: common.ToBase36HashLen(clusterName, common.ClusterIDLength),
```
mikenairn, line 179: Why does this have to be hashed here?

pkg/common/k8s_utils.go
```diff
@@ -147,3 +153,18 @@ func GetLabel(obj metav1.Object, key string) string {
...
+		return nil, clusterID
+	}
+
+	ns := &corev1.Namespace{}
+	err := c.Get(ctx, client.ObjectKey{Name: clusterIDNamespace}, ns)
+	if err != nil {
+		return err, ""
+	}
+	clusterID = ToBase36HashLen(string(ns.UID), ClusterIDLength)
+	return nil, clusterID
```
mikenairn, line 169: Why can't this not just return string(ns.UID) and the decision on hashing it or not can be done eslewhere?

philbrookes: yes ok, this makes sense, I'll do it that way

pkg/multicluster/target.go
```diff
@@ -107,11 +104,11 @@ func (t *ClusterGatewayTarget) GetWeight() int {
...
 
 func (t *ClusterGatewayTarget) GetName() string {
-	return t.ClusterName
+	return common.ToBase36HashLen(t.ClusterName, common.ClusterIDLength)
 }
 
 func (t *ClusterGatewayTarget) GetShortCode() string {
-	return ToBase36hash(t.GetName())
+	return common.ToBase36HashLen(t.Name+"-"+t.Namespace, 6)
```
mikenairn, line 111:
```suggestion
	return common.ToBase36HashLen(t.GetName(), common.ClusterIDLength)
```

pkg/multicluster/target.go
```diff
@@ -107,11 +104,11 @@ func (t *ClusterGatewayTarget) GetWeight() int {
 }
 
 func (t *ClusterGatewayTarget) GetName() string {
-	return t.ClusterName
+	return common.ToBase36HashLen(t.ClusterName, common.ClusterIDLength)
```
mikenairn, line 107:
```suggestion
		return t.ClusterName
```

examples/dnspolicy/script.sh
mikenairn, line 5: Are we keeping these?

controllers/dnspolicy_controller_single_cluster_test.go
```diff
@@ -165,93 +172,93 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
...
+		//								"RecordType":    Equal("CNAME"),
+		//								"SetIdentifier": Equal(""),
+		//								"RecordTTL":     Equal(externaldns.TTL(300)),
+		//							})),
+		//						),
+		//					}),
+		//				}),
+		//			))
+		//	}, TestTimeoutMedium, TestRetryIntervalMedium, ctx).Should(Succeed())
+		//})
```
mikenairn, line 261: Needs updated

controllers/dns_helper.go
```diff
@@ -232,7 +231,7 @@ func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTar
 			}
 
 			if len(ipValues) > 0 {
-				clusterLbName := strings.ToLower(fmt.Sprintf("%s.%s", cgwTarget.GetShortCode(), lbName))
+				clusterLbName := strings.ToLower(fmt.Sprintf("%s-%s.%s", cgwTarget.GetName(), cgwTarget.GetShortCode(), lbName))
```
mikenairn, line 234:
```suggestion
				clusterLbName := strings.ToLower(fmt.Sprintf("%s-%s.%s", cgwTarget.GetShortCode(), mcgTarget.GetShortCode(), lbName))
```

controllers/dnspolicy_dnsrecords.go
```diff
@@ -109,22 +112,26 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      dnsRecordName(gateway.Name, string(targetListener.Name)),
 			Namespace: managedZone.Namespace,
 			Labels:    commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), dnsPolicy),
+			Annotations: map[string]string{
+				"hostname": string(*targetListener.Hostname),
+			},
 		},
 		Spec: kuadrantdnsv1alpha1.DNSRecordSpec{
+			OwnerID: &clusterID,
```
mikenairn, line 126: Should we shorten this here? `common.ToBase36HashLen(clusterID, 6)`

controllers/dnspolicy_dnsrecords.go
```diff
@@ -109,22 +112,27 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw
...
+func (r *DNSPolicyReconciler) desiredDNSRecord(gateway *multicluster.GatewayWrapper, dnsPolicy *v1alpha1.DNSPolicy, targetListener gatewayapiv1.Listener, clusterGateways []multicluster.ClusterGateway, managedZone *kuadrantdnsv1alpha1.ManagedZone, clusterID string) (*kuadrantdnsv1alpha1.DNSRecord, error) {
+	clusterID = common.ToBase36HashLen(clusterID, common.ClusterIDLength)
 	dnsRecord := &kuadrantdnsv1alpha1.DNSRecord{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      dnsRecordName(gateway.Name, string(targetListener.Name)),
 			Namespace: managedZone.Namespace,
 			Labels:    commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), dnsPolicy),
+			Annotations: map[string]string{
+				"hostname": string(*targetListener.Hostname),
+			},
```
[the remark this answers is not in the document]
philbrookes: I don't see a `spec.rootHost`, but isn't the host of the listener on the gateway the right value to use here anyway? 

That's what the user is defining and we work from that down to IP addresses?

controllers/dnspolicy_controller_single_cluster_test.go
```diff
@@ -171,85 +180,97 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
...
+								})),
+								PointTo(MatchFields(IgnoreExtras, Fields{
+									"DNSName":       Equal(TestHostOne),
+									"Targets":       ConsistOf("klb.test.example.com"),
+									"RecordType":    Equal("CNAME"),
+									"SetIdentifier": Equal(""),
+									"RecordTTL":     Equal(externaldns.TTL(300)),
+								})),
+							),
 						}),
```
mikenairn, line 230: I would have probably have avoided re-factoring these tests as part of other work, but since you have half done it here, these would be better broken up like:

```
g.Expect(dnsRecord.Name).To(Equal(recordName))
g.Expect(dnsRecord.Spec.OwnerID).To(Equal(&clusterID))
g.Expect(dnsRecord.Spec.ManagedZoneRef).To(HaveField("Name", "mz-example-com"))
g.Expect(dnsRecord.Spec.Endpoints).To(ConsistOf(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":       Equal(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"Targets":       ConsistOf(TestIPAddressOne, TestIPAddressTwo),
		"RecordType":    Equal("A"),
		"SetIdentifier": Equal(""),
		"RecordTTL":     Equal(externaldns.TTL(60)),
	})),
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":          Equal("default." + "klb.test.example.com"),
		"Targets":          ConsistOf(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"RecordType":       Equal("CNAME"),
		"SetIdentifier":    Equal(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"RecordTTL":        Equal(externaldns.TTL(60)),
		"ProviderSpecific": Equal(externaldns.ProviderSpecific{{Name: "weight", Value: "120"}}),
	})),
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":          Equal("klb.test.example.com"),
		"Targets":          ConsistOf("default." + "klb.test.example.com"),
		"RecordType":       Equal("CNAME"),
		"SetIdentifier":    Equal("default"),
		"RecordTTL":        Equal(externaldns.TTL(300)),
		"ProviderSpecific": Equal(externaldns.ProviderSpecific{{Name: "geo-code", Value: "*"}}),
	})),
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":       Equal(TestHostOne),
		"Targets":       ConsistOf("klb.test.example.com"),
		"RecordType":    Equal("CNAME"),
		"SetIdentifier": Equal(""),
		"RecordTTL":     Equal(externaldns.TTL(300)),
	})),
))
```

or even maybe this for easier debugging when stuff starts failing:

```
g.Expect(dnsRecord.Name).To(Equal(recordName))
g.Expect(dnsRecord.Spec.OwnerID).To(Equal(&clusterID))
g.Expect(dnsRecord.Spec.ManagedZoneRef).To(HaveField("Name", "mz-example-com"))
g.Expect(dnsRecord.Spec.Endpoints).To(HaveLen(4))
g.Expect(dnsRecord.Spec.Endpoints).To(ContainElement(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":       Equal(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"Targets":       ConsistOf(TestIPAddressOne, TestIPAddressTwo),
		"RecordType":    Equal("A"),
		"SetIdentifier": Equal(""),
		"RecordTTL":     Equal(externaldns.TTL(60)),
	})),
))
g.Expect(dnsRecord.Spec.Endpoints).To(ContainElement(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":          Equal("default." + "klb.test.example.com"),
		"Targets":          ConsistOf(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"RecordType":       Equal("CNAME"),
		"SetIdentifier":    Equal(clusterID + "-" + gwHash + "." + "klb.test.example.com"),
		"RecordTTL":        Equal(externaldns.TTL(60)),
		"ProviderSpecific": Equal(externaldns.ProviderSpecific{{Name: "weight", Value: "120"}}),
	})),
))
g.Expect(dnsRecord.Spec.Endpoints).To(ContainElement(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":          Equal("klb.test.example.com"),
		"Targets":          ConsistOf("default." + "klb.test.example.com"),
		"RecordType":       Equal("CNAME"),
		"SetIdentifier":    Equal("default"),
		"RecordTTL":        Equal(externaldns.TTL(300)),
		"ProviderSpecific": Equal(externaldns.ProviderSpecific{{Name: "geo-code", Value: "*"}}),
	})),
))
g.Expect(dnsRecord.Spec.Endpoints).To(ContainElement(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":       Equal(TestHostOne),
		"Targets":       ConsistOf("klb.test.example.com"),
		"RecordType":    Equal("CNAME"),
		"SetIdentifier": Equal(""),
		"RecordTTL":     Equal(externaldns.TTL(300)),
	})),
))
```

mikenairn: That all said, i would avoid making any more changes to these in this PR and would prefer that re-factor to be done against main.

pkg/multicluster/target.go
```diff
@@ -111,7 +109,8 @@ func (t *ClusterGatewayTarget) GetName() string {
 }
 
 func (t *ClusterGatewayTarget) GetShortCode() string {
-	return ToBase36hash(t.GetName())
+	log.Log.V(1).Info("getting short code", "clusterName", t.GetName(), "clusterID", common.ToBase36HashLen(t.GetName(), common.ClusterIDLength))
```
mikenairn, line 112: Do we need this log? Probably better to log where ever it is being used. This might end up elsewhere, library code, possibly in dns operator itself at some point.

controllers/dnspolicy_dnsrecords.go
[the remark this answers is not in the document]
mikenairn: I mean set `dnsRecord.spec.rootHost` to `string(*targetListener.Hostname)`. Why do we need the annotation as well? What is it for?

controllers/dnspolicy_controller_single_cluster_test.go
philbrookes: yes, this makes sense, I might do this in a separate PR though, keen to get this closed out and move on to the AWS Health checks.

controllers/dnspolicy_dnsrecords.go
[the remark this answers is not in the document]
philbrookes: ah, now I understand, ok I'll update this.

controllers/dnspolicy_dnsrecords.go
```diff
@@ -109,22 +112,26 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw
...
+	rootHost := string(*targetListener.Hostname)
 	dnsRecord := &kuadrantdnsv1alpha1.DNSRecord{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      dnsRecordName(gateway.Name, string(targetListener.Name)),
 			Namespace: managedZone.Namespace,
 			Labels:    commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), dnsPolicy),
 		},
 		Spec: kuadrantdnsv1alpha1.DNSRecordSpec{
+			OwnerID:  &clusterID,
+			RootHost: &rootHost,
```
mikenairn, line 126:
```suggestion
func (r *DNSPolicyReconciler) desiredDNSRecord(gateway *multicluster.GatewayWrapper, dnsPolicy *v1alpha1.DNSPolicy, targetListener gatewayapiv1.Listener, clusterGateways []multicluster.ClusterGateway, managedZone *kuadrantdnsv1alpha1.ManagedZone) (*kuadrantdnsv1alpha1.DNSRecord, error) {
	ownerID := common.ToBase36HashLen(gateway.ClusterID, common.ClusterIDLength)
	rootHost := string(*targetListener.Hostname)
	dnsRecord := &kuadrantdnsv1alpha1.DNSRecord{
		ObjectMeta: metav1.ObjectMeta{
			Name:      dnsRecordName(gateway.Name, string(targetListener.Name)),
			Namespace: managedZone.Namespace,
			Labels:    commonDNSRecordLabels(client.ObjectKeyFromObject(gateway), dnsPolicy),
		},
		Spec: kuadrantdnsv1alpha1.DNSRecordSpec{
			OwnerID:  &ownerID,
			RootHost: &rootHost,
```

controllers/dnspolicy_dnsrecords.go
```diff
@@ -90,7 +93,7 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw
 			return nil
 		}
 
-		dnsRecord, err := r.desiredDNSRecord(gatewayWrapper.Gateway, dnsPolicy, listener, listenerGateways, mz)
+		dnsRecord, err := r.desiredDNSRecord(gatewayWrapper, dnsPolicy, listener, listenerGateways, mz, clusterID)
```
mikenairn, line 96:
```suggestion
		dnsRecord, err := r.desiredDNSRecord(gatewayWrapper, dnsPolicy, listener, listenerGateways, mz)
```

mikenairn (approved):

Issue #435, opened 2024-02-27
Refine & release the App Developer dashboard from the API Quickstart


david-martin:
The dashboard from the quickstart is sourced from https://github.com/Kuadrant/multicluster-gateway-controller/blob/api-upstream/config/prometheus-for-federation/api-dashboards/stitch.json

Requirements for the dashboard:

* Response times and latency metrics for individual API endpoints to optimize backend performance
* Breakdown of API errors by type and endpoint, helping in pinpointing issues for quicker debugging
* Data showing the most and least used endpoints to guide development focus and potential feature deprecation
* For improved usability, the viewed API/application can be switched between easily.

The dashboard should be included in the unified kuadrant release process such that:
- it can be sourced at a specific version
- it can be imported directly as is, or taken and modified (the raw json/yaml is easily available)

Dashboards will have tooltips on the panels created, to aid users on what the panels are meant to show to allow them gain more knowledge to be able to customize them.
Add new metrics & labels/dimensions to kuadrant components where needed to support the above dashboards

Recommendation is to follow through the API Quickstart and modify the dashboard in grafana directly.
Then export it as json when changes have been made.

PR #494, opened 2024-03-19
Refine platform eng dashboard

### LINKED ISSUE #434: Refine & release the Platform Engineer dashboard from the API Quickstart
The dashboard from the quickstart is sourced from https://github.com/Kuadrant/multicluster-gateway-controller/blob/api-upstream/config/prometheus-for-federation/api-dashboards/stitch_platform_eng_dashboard.json

* Network I/O metrics, including error rates, to identify potential bottlenecks or issues in data transmission
* Overview of current deployments, including success/failure rates, to monitor continuous delivery processes
* CPU, memory, and disk usage metrics for the underlying Kubernetes infrastructure to ensure adequate resource allocation.
* A specific API/application can be drilled down/linked to from this dashboard to give more granular data

david-martin:
Closes #434 
Depends on #461 

* New panels to show network i/o, cpu & memory by API
* Removed gatekeeper panels
* Added tooltips for all panels
* Added variables for namespace and httproute selection

Panels are grouped by 2 rows. Policies and APIs.

image

image

image

Deep links to the app developer dashboard will be added in #435  

To verify this PR, the single cluster getting started guide can be followed https://docs.kuadrant.io/getting-started-single-cluster/
Then run the commands from the observability README from #461.

You can then import the example dashboard json into grafana.
For data to show up on the dashboards, you can follow the guide at https://docs.kuadrant.io/kuadrant-operator/doc/user-guides/secure-protect-connect/ to create various gateway api and kuadrant resources.

You will also need to add the deployment name as a label to the HTTPRoute for some panels to show data.
e.g. `kubectl label httproute toystore -n default deployment=toystore`
Some panels will only show data if traffic has been sent recently (may take up to 5 mins for some panels to show aggregated data). You can leave a curl loop running, like the guide above suggests.

### DIFF
--- a/examples/dashboards/platform_engineer.json
+++ b/examples/dashboards/platform_engineer.json
@@ -124,7 +124,7 @@
             "mode": "thresholds"
           },
           "custom": {
-            "align": "auto",
+            "align": "center",
             "displayMode": "auto",
             "inspect": false
           },
@@ -153,10 +153,6 @@
               {
                 "id": "unit",
                 "value": "bool"
-              },
-              {
-                "id": "custom.width",
-                "value": 100
               }
             ]
           },
@@ -169,10 +165,6 @@
               {
                 "id": "unit",
                 "value": "bool"
-              },
-              {
-                "id": "custom.width",
-                "value": 92
               }
             ]
           },
@@ -181,24 +173,14 @@
               "id": "byName",
               "options": "Gateway Class"
             },
-            "properties": [
-              {
-                "id": "custom.width",
-                "value": 395
-              }
-            ]
+            "properties": []
           },
           {
             "matcher": {
               "id": "byName",
               "options": "Name"
             },
-            "properties": [
-              {
-                "id": "custom.width",
-                "value": 136
-              }
-            ]
+            "properties": []
           },
           {
             "matcher": {
--- a/examples/dashboards/platform_engineer.json
+++ b/examples/dashboards/platform_engineer.json
@@ -26,8 +26,8 @@
   "fiscalYearStartMonth": 0,
   "gnetId": 7630,
   "graphTooltip": 0,
-  "id": 15,
-  "iteration": 1700061613956,
+  "id": 33,
+  "iteration": 1710847753319,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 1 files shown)


R-Lawton:
👀

david-martin:
@R-Lawton I see you had some comments about this dashboard in https://github.com/Kuadrant/kuadrant-operator/pull/461#issuecomment-2007467718.
The dashboard has changed from what you may have seen in that PR.
Would you mind reviewing and see if your feedback has been addressed sufficiently in this PR?
Thanks

R-Lawton:
Looking at the dashboard with the new changes one or two things I'm thinking:

I think making the single stat values a bit larger would be good. For the tool tips on those panels, I'm thinking if i was looking at that for the first time with little context i would be thinking what does accepted or programmed mean so a link to a doc or a quick tldr might be worth adding.

The tables are showing up a bit squashed to one corner of the tables themselves, setting the panels to centre rather then auto should solve that. I made some of the column widths a bit more uniform rather than some being wide and some being skinny.  

I was thinking of making the API error graph the full length of the screen to rather then having a gap
Screenshot 2024-03-22 at 11 53 00
Screenshot 2024-03-22 at 11 53 07

david-martin:
@R-Lawton Thanks for the feedback.

I've widened the single stats and added links to the gateway api status condition type docs
image
image


The tables look much neater now with the centering and auto column width
image

For the API Error Rate graph, I'm not sure about changing that panel to be full width.
I kept the panels aligned below each other so they can more easily be compared for the same time range. For example, if you are seeing spikes on the API Traffic Summary graphs, you can move your view vertically down to the Error Rate panel and see if there's a correlation more easily.
I know that could be an argument for putting all 5 of the line chart panels aligned vertically, but my thinking there is most people would be viewing in a wide window, and the full width of the window for looks quite sparse. 

All this being said, the goal here is to give a reasonable starting point or inspiration for people to go and build their own dashboards that give the most important information for them, so I won't get too hung up on it as long as it subjectively looks good.

R-Lawton:
no problemo, Looks really good, the tables look great, the only things is i think there might be overrides on the gateways table that are causing the big gaps and the centring to still look pushed to the side. 

Ok yeah i see what your saying that makes sense to me. Yeah your right these are jumping off points for users

david-martin:
@R-Lawton you're right.
Fixed up the Gateways table now too
image

R-Lawton:
Looks great noice job 
/lgtm

R-Lawton (approved):

PR #499, opened 2024-03-20
Fixes of Secure, connect and protect guide

trepel:
## Overview

Small fixes of "Secure, connect and protect" guide. What was done:
1) `dnsrecord.kuadrant.io` needs to be used since it conflicts with `dnsrecord.ingres.operator.openshift.io`
2) added `--context kind-kuadrant-local ` just for the sake of consistency
3) added missing `response` config to AuthPolicy, same is used in https://docs.kuadrant.io/kuadrant-operator/doc/user-guides/authenticated-rl-for-app-developers/
4) added `when` selector. If not present Bob won't get the extra request compared to Alice since both limits will be applied and the most restrictive wins
5) added note that the `--resolve` flag might not be used if DNS was properly configured

## Verification Steps
Eye review might suffice since the changes are relatively simple. Optionally you can just follow the guide.


doc/user-guides/secure-protect-connect.md
```diff
@@ -353,6 +361,10 @@ spec:
       - limit: 1
         duration: 3
         unit: second
+      when:
```
eguzki, line 364: that's `general-user` limit, hence should not be constrained to `bob` user

eguzki (changes requested):

doc/user-guides/secure-protect-connect.md
trepel: the operator is `neq` so it is actually constrained to everyone else but `bob` user. This constraint had to be added because this limit would have been applied to `bob` user which needs to be avoided.

maleck13 (commented):
Looks good to me but will wait for @eguzki to confirm on the RLP

doc/user-guides/secure-protect-connect.md
eguzki: well, `bob` is also a member of the `general-user` group, isn't it?

trepel: I thought that `general-user` was just an arbitrary name of the limit and it had no other meaning/function. There is no `general-user` group or anything like that afaict

trepel: I can force-push a rename to something more accurate but I would appreciate some suggestions :) Smt along the lines of  `all-but-bob`.

eguzki: I do not care about the name of the limit either :) 

What the policy is saying is that for `bob` there are two limits: 1) one shared limit of 1 request every 3 seconds and 2) one dedicated (only for bob) limit of 2 request every 3 seconds. The most restrictive applies, so the effective limit for bob is 1 request every 3 seconds. The policy is also saying that for anybody the limit is (shared) 1 request every 3 seconds. 

Yes, indeed, the policy does not make sense from a business perspective. It wants to show that you can define limits per user. So, it makes more sense to me to have something like this:

```yaml
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    "general-user":
      rates:
      - limit: 1
        duration: 3
        unit: second
      counters:
      - metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid
      when:
      - selector: metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid
        operator: neq
        value: bob
    "bob-limit":
      rates:
      - limit: 2
        duration: 3
        unit: second
      when:
      - selector: metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid
        operator: eq
        value: bob
```

This policy states the following:
* Unauthenticated users are not rate limited. (The previous authZ stage should reject unauthenticated users)
* There is one default limit for authenticated users, which is 1 request every 3 seconds. That's the default "plan". Each user has its own quota... it is not a shared counter. Same limit definition, dedicated counters.
* `bob` is a premium user that paid some subscription and he gets a plan of 2 requests every 3 seconds

trepel: I see, this is a good idea! If I am not mistaken the only difference is the addition of
```
      counters:
      - metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid
```
to the `general-user` limit which makes it "per user". I will update the PR

eguzki: yeah, it would be nice to remove the 

```yaml
when:
      - selector: metadata.filter_metadata.envoy\.filters\.http\.ext_authz.identity.userid
        operator: neq
        value: bob
```
for scalability (think about thousands of premium users. But for that case, we need something else, like groups and plans and tagging those for cleaner policy spec. But for now, for the sake of simplicity of the example, it is good enough.

trepel: I force pushed the counter addition but I kept the "not bob" condition in place.

eguzki: > @eguzki It does not work if that "not bob" condition is removed from `general-user` limit . If it is removed then the `general-user` limit is applied on bob user as well so bob does not get his extra request - since the `general-user` limit is more restrictive it wins.

I know, hence my comment "It would be nice..."

eguzki (approved):

doc/user-guides/secure-protect-connect.md
trepel: Ah, sorry, I did not get that "would". Thanks for approval.

PR #503, opened 2024-03-21
AP Defaults

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/462

Adds the `defaults` field to the AuthPolicy spec. This field is mutually exclusive with the pre-existing fields for defining auth rules in the bare spec.

# Verification
The explicit default path is already tests by integration tests but it you want to verify manually:
* Follow the guide at https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/authenticated-rl-for-app-developers.md and ensure you can complete it successfully
* Repeat the guide but at step 3, use the following AuthPolicy that defines explicit defaults instead:
```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  defaults:
    rules:
      authentication:
        "api-key-users":
          apiKey:
            selector:
              matchLabels:
                app: toystore
            allNamespaces: true
          credentials:
            authorizationHeader:
              prefix: APIKEY
      response:
        success:
          dynamicMetadata:
            "identity":
              json:
                properties:
                  "userid":
EOF
```
* Verify you can complete the guide successfully
* Try creating AuthPolicy with both implicit and explicit defaults
```
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  defaults:
    rules:
      authentication:
        "api-key-users":
          apiKey:
            selector:
              matchLabels:
                app: toystore
            allNamespaces: true
          credentials:
            authorizationHeader:
              prefix: APIKEY
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
EOF
```
* Verify policy is rejected due to:
```
The AuthPolicy "toystore" is invalid: spec: Invalid value: "object": Implicit and explicit defaults are mutually exclusive
```


api/v1beta2/authpolicy_types.go
```diff
@@ -141,6 +141,18 @@ type AuthPolicySpec struct {
...
+	// +optional
+	Defaults *CommonSpec `json:"defaults"`
+
+	// CommonSpec defines implicit default values for this policy and for policies inheriting this policy.
+	// CommonSpec is mutually exclusive with explicit defaults defined by Defaults.
+	CommonSpec `json:""`
+}
+
+// CommonSpec contains common shared fields for defaults and overrides
+type CommonSpec struct {
```
KevFan, line 155: This gives issues with CEL due to and size of the CRD

```
The CustomResourceDefinition "authpolicies.kuadrant.io" is invalid: 
* metadata.annotations: Too long: must have at most 262144 bytes
* spec.validation.openAPIV3Schema.properties[spec].properties[rules].properties[authentication].additionalProperties.properties[routeSelectors].items.properties[matches].items.properties[path].x-kubernetes-validations[10].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema
* spec.validation.openAPIV3Schema.properties[spec].properties[rules].properties[authorization].additionalProperties.properties[routeSelectors].items.properties[matches].items.properties[path].x-kubernetes-validations[10].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema
* spec.validation.openAPIV3Schema.properties[spec].properties[rules].properties[callbacks].additionalProperties.properties[routeSelectors].items.properties[matches].items.properties[path].x-kubernetes-validations[10].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema
* spec.validation.openAPIV3Schema.properties[spec].properties[rules].properties[metadata].additionalProperties.properties[routeSelectors].items.properties[matches].items.properties[path].x-kubernetes-validations[10].rule: Forbidden: contributed to estimated rule cost total exceeding cost limit for entire OpenAPIv3 schema
* spec.validation.openAPIV3Schema: Forbidden: x-kubernetes-validations estimated rule cost total for entire OpenAPIv3 schema exceeds budget by factor of 1.7x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)
make: *** [install] Error 1
```

So we might want to rethink of whether we truely want the 2 pathways to achieving defaults 🤔

KevFan: Added a workaround at https://github.com/Kuadrant/kuadrant-operator/pull/503/commits/6a01a4b870343c293c70159822517adf96819043 decreasing the max number of route selectors allowed, but would likely hit this problem again for overrides also which would decrease this number again

Boomatang (commented):
Verification worked as expected. Unsure what the best manual verification steps are for the CEL expression that have being add. 

Added a couple of comments.

controllers/authpolicy_authconfig.go
```diff
@@ -112,39 +112,39 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 	authConfig.Spec.Hosts = hosts
 
 	// named patterns
-	if namedPatterns := ap.Spec.NamedPatterns; len(namedPatterns) > 0 {
+	if namedPatterns := ap.GetNamedPatterns(); len(namedPatterns) > 0 {
```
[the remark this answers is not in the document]
KevFan: > I have a thought/concern, which I am not sure what the best resolution is. There is nothing stop someone from accessing `Spec.NamedPatterns` directly. This could lead to errors later. One way would be to force access through methods by making the fields in the Spec private. In this case the methods would want to be on the Spec struct and not the AP struct. Access would then look like `ap.Spec.GetNamedPatterns()`.
> 
> This would need to apply to all data fields in the Spec and not just the `NamedPatterns`. It should also apply to the spec in #456.

Hmm, I dont think this will quite work as intended. If we make fields private / not exported, I guess we need setters for all the fields also, while also having setters for the defaults fields. The will work for using the struct via code.

The other one is since the field is not exported, the json field will not be available via `kubectl get policy x -o json` commands which I think makes this not viable unless I'm missing something.  

> 
> What do you think? Personal if I have access to the fields of a struct at some point I will access them directly.

Yeah, unfortunately I think is an artefact of having multiple pathways of achieving the same result. I guess the assumption is that if people are acessing the fields directly via code, they know which pathway they are using, either `spec.X` or `spec.defaults.x` but the getters will be the best way to get the data without needing to know which pathway is used

api/v1beta2/authpolicy_types.go
```diff
@@ -57,7 +57,7 @@ type CommonAuthRuleSpec struct {
 	// At least one selected HTTPRoute rule must match to trigger the auth rule.
 	// If no route selectors are specified, the auth rule will be evaluated at all requests to the protected routes.
 	// +optional
-	// +kubebuilder:validation:MaxItems=15
+	// +kubebuilder:validation:MaxItems=8
```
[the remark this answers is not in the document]
KevFan: Sure, can add a test. Though, I'm expecting this value to decrease again when the overrides field is added in https://github.com/Kuadrant/kuadrant-operator/issues/464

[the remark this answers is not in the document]
Boomatang: Cool, and thinking out loud. If we need to increase that number in the future, is the workaround to create a webhook to do the validation and drop the use of the CEL?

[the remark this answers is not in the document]
KevFan: Added integration test for this. 

I don't think just simply switching to using validation webhook would work, as I believe this more so relates to using a slice of `RouteSelectors` that the fields itself also have CEL validation for and how frequently we have this slice of `RouteSelectors` in the various structs used within the CRD. This will fail the CEL cost estimation when applying the CRD even if switch to using a validating webhook 🤔 

https://github.com/Kuadrant/kuadrant-operator/blob/4f6a138d4b670381f6ef83409f5bffc2824c19ff/api/v1beta2/route_selectors.go#L13-L24

api/v1beta2/authpolicy_types.go
```diff
@@ -161,13 +184,13 @@ type AuthPolicySpec struct {
 
 	// The auth rules of the policy.
 	// See Authorino's AuthConfig CRD for more details.
-	AuthScheme AuthSchemeSpec `json:"rules,omitempty"`
+	AuthScheme *AuthSchemeSpec `json:"rules,omitempty"`
```
KevFan, line 187: Switched to pointer so that this will not show on json `spec` as empty object `{}` when empty which would be a bit confusing to a user, especially if `spec.defaults` is used instead 🤔 Also makes the mutual exclusive CEL rule a bit easier by switching to pointer

controllers/authpolicy_authconfig.go
```diff
@@ -187,8 +194,10 @@ func authorinoSpecsFromConfigs[T, U any](configs map[string]U, extractAuthorinoS
 }
 
 func mergeConditionsFromRouteSelectorsIntoConfigs(ap *api.AuthPolicy, route *gatewayapiv1.HTTPRoute, authConfig *authorinoapi.AuthConfig) (*authorinoapi.AuthConfig, error) {
+	commonSpec := ap.Spec.CommonSpec()
+
```
KevFan, line 198: There's no `nil` check for `AuthScheme` here as this function is only called from `r.desiredAuthConfig()` which has been updated with this `nil` check for `AuthScheme`. This function will not be called in this case

Makefile
```diff
@@ -406,14 +406,15 @@ kind-load-bundle: ## Load image to local cluster
 ##@ Deployment
 
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+	# Use server side apply, otherwise will hit into this issue https://medium.com/pareture/kubectl-install-crd-failed-annotations-too-long-2ebc91b40c7d
+	$(KUSTOMIZE) build config/crd | kubectl apply --server-side -f -
```
KevFan, line 410: Not quite sure will this be an issue for OLM upgrading CRDs. I'm wishful that this will not be an issue for OLM, and this is just a local issue 🤔

api/v1beta2/authpolicy_types.go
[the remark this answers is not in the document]
guicassolato: It's because we're importing Gateway API types and, with those, their CEL validation as well.

I guess our best options to avoid this hard limit are:
1. wait for https://github.com/kubernetes-sigs/gateway-api/pull/2593 to be implemented and then simplify the route selectors
2. redefine the HTTPRouteMatch types ourselves without the validation – we could do this because the validation that matters is the one enforced on the HTTPRoute, not on the policy; on the policy, the only consequence of an "invalid" route selector is that the selector will not select any route rule.

doc/reference/authpolicy.md
```diff
@@ -24,9 +25,20 @@
...
+| `patterns`       | Map<String: [NamedPattern](#namedpattern)>                                                                                                  | No           | Implicit default named patterns of lists of `selector`, `operator` and `value` tuples, to be reused in `when` conditions and pattern-matching authorization rules.                                                                                                                              |
+| `when`           | [][PatternExpressionOrRef](https://docs.kuadrant.io/authorino/docs/features/#common-feature-conditions-when)                                | No           | List of implicit default additional dynamic conditions (expressions) to activate the policy. Use it for filtering attributes that cannot be expressed in the targeted HTTPRoute's `spec.hostnames` and `spec.rules.matches` fields, or when targeting a Gateway.                                |
```
guicassolato, line 34: The `when` conditions at this level will be a problem when we get to Tier 2 of D/O and introduce _constraints_.

Makefile
guicassolato: The AuthPolicy CRD itself has reached ~600 Kb already with this change. IIRC, by default, the maximum request size accepted by etcd is 1.5 Mb – actual value size between 2 Mb (default) and 8 Mb (recommended maximum.)

guicassolato (approved):
LGTM. Nice work, @KevFan!

PR #517, opened 2024-04-03
Fix local setup for podman users

david-martin:
To verify, when using podman only:

```
make local-cluster-setup
```

prior to this change, you may see an error like this:

```
./utils/docker-network-ipaddresspool.sh kind /Users/davidmartin/work/kuadrant-operator/bin/yq | kubectl apply -n metallb-system -f -
Error: template: inspect:1:23: executing "inspect" at <.IPAM.Config>: can't evaluate field IPAM in type interface {}
error: no objects passed to apply
make[1]: *** [install-metallb] Error 1
make: *** [local-cluster-setup] Error 2
```

### DIFF
--- a/utils/docker-network-ipaddresspool.sh
+++ b/utils/docker-network-ipaddresspool.sh
@@ -17,9 +17,11 @@ SUBNET=""
 
 # Try podman version of cmd first. docker alias may be used for podman, so network
 # command will be different
+set +e
 if command -v podman &> /dev/null; then
   SUBNET=`podman network inspect -f '{{range .Subnets}}{{if eq (len .Subnet.IP) 4}}{{.Subnet}}{{end}}{{end}}' $networkName`
 fi
+set -e
 
 # Fallback to docker version of cmd
 if [[ -z "$SUBNET" ]]; then
--- a/utils/docker-network-ipaddresspool.sh
+++ b/utils/docker-network-ipaddresspool.sh
@@ -13,8 +13,20 @@ YQ="${2:-yq}"
 
 ## Parse kind network subnet
 ## Take only IPv4 subnets, exclude IPv6
-SUBNET=$(docker network inspect $networkName --format '{{json .IPAM.Config }}' | \
-    ${YQ} '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet')
+SUBNET=""
+
+# Try podman version of cmd first. docker alias may be used for podman, so network
+# command will be different
+if command -v podman &> /dev/null; then
+  SUBNET=`podman network inspect -f '{{range .Subnets}}{{if eq (len .Subnet.IP) 4}}{{.Subnet}}{{end}}{{end}}' $networkName`
+fi
+
+# Fallback to docker version of cmd
+if [[ -z "$SUBNET" ]]; then
+  SUBNET=`docker network inspect $networkName -f '{{ (index .IPAM.Config 0).Subnet }}'`
+fi
+
+# Neither worked, error out
 if [[ -z "$SUBNET" ]]; then
    echo "Error: parsing IPv4 network address for '$networkName' docker network"
    exit 1
@@ -22,7 +34,7 @@ fi
 
 # shellcheck disable=SC2206
 subnetParts=(${SUBNET//./ })
-cidr="${subnetParts[0]}.${subnetParts[1]}.200.0/24"
+cidr="${subnetParts[0]}.${subnetParts[1]}.0.252/30"
 
 cat <<EOF | ADDRESS=$cidr ${YQ} '(select(.kind == "IPAddressPool") | .spec.addresses[0]) = env(ADDRESS)'
 ---


utils/docker-network-ipaddresspool.sh
```diff
@@ -13,16 +13,21 @@ YQ="${2:-yq}"
 
 ## Parse kind network subnet
 ## Take only IPv4 subnets, exclude IPv6
-SUBNET=$(docker network inspect $networkName --format '{{json .IPAM.Config }}' | \
-    ${YQ} '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet')
+SUBNET=""
+if command -v podman &> /dev/null; then
+  SUBNET=`podman network inspect -f '{{range .Subnets}}{{if eq (len .Subnet.IP) 4}}{{.Subnet}}{{end}}{{end}}' kind`
```
mikenairn, line 18: Should `kind` here be `$networkName`?

mikenairn:
@david-martin How are you deciding if you are using docker or podman when running local-setup? Is it not common to just alias docker to podman?

david-martin:
> How are you deciding

@mikenairn 

`if command -v podman &> /dev/null; then`

> Is it not common to just alias docker to podman

Yes, that's in the podman migration docs https://podman-desktop.io/docs/migrating-from-docker/emulating-docker-cli-with-podman

utils/docker-network-ipaddresspool.sh
```diff
@@ -13,16 +13,21 @@ YQ="${2:-yq}"
...
 ## Parse kind network subnet
 ## Take only IPv4 subnets, exclude IPv6
-SUBNET=$(docker network inspect $networkName --format '{{json .IPAM.Config }}' | \
-    ${YQ} '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet')
+SUBNET=""
+if command -v podman &> /dev/null; then
+  SUBNET=`podman network inspect -f '{{range .Subnets}}{{if eq (len .Subnet.IP) 4}}{{.Subnet}}{{end}}{{end}}' $networkName`
+elif command -v docker &> /dev/null; then
+  SUBNET=`docker network inspect $networkName -f '{{ (index .IPAM.Config 0).Subnet }}'`
+fi
```
mikenairn, line 21: Is this not assuming if i have podman installed that i want to use it? I have both installed but use docker, this will try and inspect the podman network first and fail no?

david-martin: I think that's why the CI job is failing actually.

If you have both installed, and don't have the 'kind' network setup in podman, it would give that error.
If you have just podman installed and a docker alias, this would work fine as all other commands use 'podman'.
If you have just podman installed and no alias for docker, the setup would fail a lot earlier.

If we switched to the docker way of doing things first, that would fail for anyone with just podman & the docker alias.
Maybe that's the right ordering though, and have it fallback to the podman version of the command if there's an error?

david-martin: I've made another change to try podman, then docker, then fail

mikenairn: I think this might work as long as you have never created the kind network using both podman and docker. On a Mac is it expected that you do not have docker installed? When you create a kind cluster are you telling it what container runtime to use or relying on it to auto detect?

This change is probably fine, but people should probably be aware that it's being made and that if you happen to have used both podman and docker at any time, especially with kind, that you could get odd results.

utils/docker-network-ipaddresspool.sh
```diff
@@ -13,16 +13,30 @@ YQ="${2:-yq}"
...
+# Neither worked, error out
 if [[ -z "$SUBNET" ]]; then
    echo "Error: parsing IPv4 network address for '$networkName' docker network"
    exit 1
 fi
 
 # shellcheck disable=SC2206
 subnetParts=(${SUBNET//./ })
-cidr="${subnetParts[0]}.${subnetParts[1]}.200.0/24"
+cidr="${subnetParts[0]}.${subnetParts[1]}.0.252/30"
```
mikenairn, line 39: I'm not sure what implications this has. I assume it was changed for a reason? @eguzki I think you were the last one looking at this, do you think this will cause any issues?

david-martin: Reason for changing is to be compatible with podman and docker.

With docker, the default subnet is typically a /16 e.g. `172.17.0.0/16`
With podman, it's landing as a /24 for me locally:

```
podman network inspect kind|grep subnet
          "subnets": [
                    "subnet": "fc00:f853:ccd:e793::/64",
                    "subnet": "10.89.0.0/24",
```

Without this change, the .200.0/24 subnet wouldn't be part of the network subnet.
With the change, a .0.252/30 subnet would be on both podman & docker networks. (albeit with just 4 ip addresses for load balancers, which may be fine for most local dev use cases)

eguzki: sorry guys, I missed this. 

I came back to this PR because these changes are causing one more side effect https://github.com/Kuadrant/kuadrant-operator/pull/557

Issue #441, opened 2024-02-27
Create a set of template Multiwindow, Multi-Burn-Rate Alerts, tailored to an API view


david-martin:
Multiwindow, Multi-Burn-Rate as per https://sre.google/workbook/alerting-on-slos/

Suggested SLIs

* Availability based on error rate (5xx responses) from an API as a whole
* Availability based on latency from an API as a whole

Both of these will include example prometheus recording rules for how to set up these alerts.

We can use https://sloth.dev/ to help generate the rules.

We should include prometheus unit tests for these and include manual testing if possible.

It will more than likely require istio or envoy metrics (and so be specific to that gateway api provider),
however that's reasonable as there are no standards around gateway api low level metrics.

It is important to have these be templated as we have no way of knowing the exact criteria for alerting that will suit every individual user.
We can use what we think are reasonable default values, and allow them to be templated where possible.

PR #522, opened 2024-04-08
Add business user dashboard

### LINKED ISSUE #436: Create a new dashboard for the Business User persona

* Number of requests over time to monitor adoption rates and peak usage times
* API response times and error rates to gauge the user experience
* Example SLIs/SLOs (depends on https://github.com/Kuadrant/kuadrant-operator/issues/441)
* Have a template variable for choosing which API(s) to show

david-martin:
Closes #436

- New dashboard focused on business user
- text area at top explaining dashboard
- repeating row of all panels for each API
- within the repeating row:
  - time series for req/s
  - time series for req/s broken down by path
  - heatmap comparing req/s for current selected range with previous range
  - table showing total requests for current range, previous and the trend up/down

SLIs are not included in this PR. That depends on #441 and will have to land after.

image

### DIFF
--- a/examples/dashboards/business_user.json
+++ b/examples/dashboards/business_user.json
@@ -1,0 +1,655 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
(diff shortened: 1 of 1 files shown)


jasonmadigan:
LGTM

I can see us building this out with more feedback, but this gives us a good starting point

jasonmadigan (approved):

PR #523, opened 2024-04-08
RLP atomic override

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/463

RLP atomic override. RLP targetting gateways that define an override block, will override the limits of HTTPRoute RLPs.

CEL validation was added to:
* make `defaults` and `override` block mutually exclusive
* make implicit limits block and `override` block mutually exclusive
* only allow `override` block for RLPs targetting Gateway resource

# Verification 
The scenario's is already generally tested with the integration tests added 

If you want to manually verify:
* Checkout this branch and deploy cluster
```sh
make local-setup
```
* Deploy Kuadrant CR
```
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

* Deploy toystore
```sh
kubectl wait --timeout=300s --for=condition=Available deployment toystore
```

* Create  HTTP Route
```sh
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
    - method: POST
      path:
        type: Exact
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

* Create GW RLP with defaults
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: gw-rlp
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  defaults:
    limits:
      "gateway":
        rates:
        - limit: 5
          duration: 10
          unit: second
EOF
```
* Port forward gateway service
```sh
kubectl port-forward -n istio-system service/istio-ingressgateway-istio 9080:80 2>&1 >/dev/null &
export GATEWAY_URL=localhost:9080
```
* Ensure HTTP Route is rate limit at 5 requests per 10 seconds
```sh
```
* Create HTTPRoute RLP with implicit default limits
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    "route":
      rates:
      - limit: 8
        duration: 10
        unit: second
EOF
```
* Ensure HTTP Route is rate limted at 8 requests per second (Route RLP defaults takes precedence over gateway defaults) (Note: you might need to port forward the gateway service again)
```sh
```
* Update GW RLP to set overrides instead
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: gw-rlp
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  overrides:
    limits:
      "gateway":
        rates:
        - limit: 5
          duration: 10
          unit: second
EOF
```
* Ensure HTTP Route is rate limits at 5 requests per second again (GW RLP Override takes precedence)
```sh
```
* Verify overrides to not allowed on RLPs targetting HTTPRoutes
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  overrides:
    limits:
      "create-toy":
        rates:
        - limit: 8
          duration: 10
          unit: second
EOF
```
* Verify explicit defaults and overrides blocks are mutually exclusive
```
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: gw-rlp
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  defaults:
    limits:
      "create-toy":
        rates:
        - limit: 1
          duration: 10
          unit: second
  overrides:
    limits:
      "create-toy":
        rates:
        - limit: 5
          duration: 10
          unit: second
EOF
```
* Verify implicit defaults and overrides blocks are mutually exclusive
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: gw-rlp
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  limits:
    "create-toy":
      rates:
      - limit: 1
        duration: 10
        unit: second
  overrides:
    limits:
      "create-toy":
        rates:
        - limit: 5
          duration: 10
          unit: second
EOF
```


controllers/ratelimitpolicy_limits.go
```diff
@@ -105,8 +107,38 @@ func (r *RateLimitPolicyReconciler) buildRateLimitIndex(ctx context.Context, rlp
...
 	}
 
 	return rateLimitIndex, nil
 }
+
+// applyOverrides checks for any overrides set for the RateLimitPolicy.
+// It iterates through the RateLimitPolicyList to find overrides for the provided target HTTPRoute.
+// If an override is found, it updates the limits in the RateLimitPolicySpec in accordingly.
+func (r *RateLimitPolicyReconciler) applyOverrides(ctx context.Context, rlp *kuadrantv1beta2.RateLimitPolicy, targetNetworkObject client.Object) error {
+	if route, ok := targetNetworkObject.(*gatewayapiv1.HTTPRoute); ok {
```
guicassolato, line 124: I think we could follow a different approach here that is indifferent to the fact that nowadays we only allow one policy of a kind targeting a given network resource at a time.

Instead, what about the following?
1. List all the policies that target the same resource, plus, if it's a route policy, all the policies that target any of the route's parent Gateways
2. Sort the list of policies by precedence – i.e. the higher the target is in the [hierarchy](https://gateway-api.sigs.k8s.io/geps/gep-713/#hierarchy), the higher the precedence; same level in the hierarchy, older policy beats newer ones
3. Drop all policies with lower precedence than `rlp`
4. Iterate from highest to lowest until finding a block of overrides, in which case replace the `rlp`'s limits with it and `break`

In the future, for adding support for the `merge` strategy, we invert the order of the list (iterating from lowest to highest) and go all the way through the list (i.e. without breaking at the first match.)

A remaining problem to solve – which is an issue in your current proposed implementation also – is what to do in case of multiple parent gateways with override policies targeting them. I suppose we'll have to add the host names in.

WDYT?

Boomatang: I am torn with my view on this. Yes the use of the DAG make it an easier problem. If code freeze is in two weeks I think it would be better to have a working solution sooner.

The conflict I have is with the hierarchy. Currently we only support polices attached to the gateway or the route. If the gateway has an atomic override the number of policies attached to route does not matter, the gateway always wins. 

The same goes for the multiply routes, we currently don't support that as a feature. While we possible should to me it is feeling like scope creep and solving a problem of tomorrow. A problem we are not sure even exists in the wild.

This is were my conflict is. I think we should be solving that problem, and that we should support polices attached to the gateway class. That is not the product of today, but of tomorrow.

guicassolato: I see your point, @Boomatang. The part that I don't like is the spreading of one decision across the entire code, i.e. the coupling of implementations at multiple levels to a rule otherwise enforced at a single point, possibly at a level far beyond. We are just increasing the number of points in the code that will need to change when the rule also shifts.

In this case, the "decision" (or "rule") we'd be coupling the implementation to or not is the current behaviour of 1 policy of a kind per target. I already lost count of how many places in the code we rely on that for something. And this one is particularly nasty because the rule itself is not even obvious at a glance. Instead, one who reads the code needs to have that click that "oh, right! this is because of only 1 policy per..."

Other examples include for sure the kinds of resources that can be targeted by a policy. Dozens of ifs and switch cases all over for this one too, when IMO in many cases the types could/should abstract the complexity. (BTW, you can git blame me on several of those.)

As for using the DAG, I may be overlooking the difficulties to add it to implementation here, so I'll let you and @KevFan decide. My initial intuition is that it could result in some DRYer code and arguably improve performance.

KevFan: Hmm, yeah might play around with this, although part of this ,at least for RLP, might be addressed/implemented by @eguzki already since he has draft PR introducing a new limitador limits controller which uses the DAG https://github.com/Kuadrant/kuadrant-operator/pull/527

KevFan: Pushed an implememtation using the DAG that is focused solely on the overriden portion of the reconcile logic based on the algorithm suggested (https://github.com/Kuadrant/kuadrant-operator/pull/527 would be the general refactor to use the DAG throughout)

KevFan: Currently doesn't account for:

> A remaining problem to solve – which is an issue in your current proposed implementation also – is what to do in case of multiple parent gateways with override policies targeting them. I suppose we'll have to add the host names in.

In this scenario, the oldest Gateway RLP's override block is used

pkg/library/gatewayapi/types.go
```diff
@@ -19,7 +19,31 @@ func (a PolicyByCreationTimestamp) Less(i, j int) bool {
...
+
+func (a PolicyByTargetRefKindAndCreationTimeStamp) Len() int      { return len(a) }
+func (a PolicyByTargetRefKindAndCreationTimeStamp) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
+func (a PolicyByTargetRefKindAndCreationTimeStamp) Less(i, j int) bool {
+	targetRef1 := a[i].GetTargetRef()
+	targetRef2 := a[j].GetTargetRef()
+
+	// Compare kind first
+	if targetRef1.Kind != targetRef2.Kind {
+		return targetRef1.Kind < targetRef2.Kind
```
guicassolato, line 39: Sorting alphabetically? I guess ATM we're lucky **G**ateway < **H**TTPRoute < **S**ervice.

controllers/ratelimitpolicy_limits.go
```diff
@@ -105,8 +115,91 @@ func (r *RateLimitPolicyReconciler) buildRateLimitIndex(ctx context.Context, rlp
...
+
+		// this policy is potentially affected by other policies from this gateway
+		if slices.Contains(policyKeys, client.ObjectKeyFromObject(rlp)) {
+			policies = append(policies, policyList...)
+		}
+	}
+
+	filteredPolicies := utils.Filter(policies, func(policy kuadrantgatewayapi.Policy) bool {
+		// HTTPRoute RLPs should only care about overrides from gateways
+		if kuadrantgatewayapi.IsTargetRefHTTPRoute(rlp.GetTargetRef()) {
```
guicassolato, line 151: For a gateway `rlp`, this will be false for all `policies`. Therefore, `filterPolicies` will be an empty list. Maybe avoid calling `applyOverrides` altogether when `rlp` is known to be at the top of the hierarchy (i.e. when `rlp` is a gateway policy)?

guicassolato (changes requested):
We're still missing overriding the rules in the wasm plugin config.

The reason why the verification steps seem to work is because, coincidentally, the limit is named "create-toy" in both policies. But, of course, we shouldn't expect that to be always the case. In fact, other differences between the override and the lower-tier policies, such as on `when` conditions and `routeSelectors`, would also cause divergences here.

guicassolato:
In the verification, I personally found useful checking the states of the WasmPlugin and Limitador CRs after each step of applying the policies:

```sh
kubectl get wasmplugin/kuadrant-istio-ingressgateway -n istio-system -o yaml
kubectl get limitador/limitador -n kuadrant-system -o yaml
```

pkg/library/gatewayapi/types.go
KevFan: Updated to sort by Gateway, then HTTPRoute and then alphebetically

controllers/ratelimitpolicy_limits.go
```diff
@@ -105,8 +124,137 @@ func (r *RateLimitPolicyReconciler) buildRateLimitIndex(ctx context.Context, rlp
...
+
+	sort.Sort(kuadrantgatewayapi.PolicyByTargetRefKindAndCreationTimeStamp(filteredPolicies))
+
+	for _, policy := range filteredPolicies {
+		p := policy.(*kuadrantv1beta2.RateLimitPolicy)
+		if p.Spec.Overrides != nil {
+			rlp.Spec.CommonSpec().Limits = p.Spec.Overrides.Limits
+			logger.V(1).Info("applying overrides from parent policy", "parentPolicy", client.ObjectKeyFromObject(p))
+			r.AffectedPolicyMap.SetAffectedPolicy(rlp, []client.ObjectKey{client.ObjectKeyFromObject(p)})
+			break
```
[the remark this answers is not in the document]
KevFan: No, I went with `break` because the applied overrides currently can only come from one gateway RLP, with the oldest having the precdence. In this case, we should break out the loop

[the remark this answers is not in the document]
Boomatang: Of course it is sorted by the time stamp. Makes perfect sense now.

guicassolato (approved):
Verification steps working. Good job, @KevFan!

🚀

PR #524, opened 2024-04-09
 Updated HTTPRoute for ToysStore with new label.

david-martin:
- Added a "deployment" label to the HTTPRoute resource so it can be tied to istio metrics

i.e. the `destination_workload ` and `destination_service_name ` labels on the `istio_requests_total` metric


jasonmadigan (approved):
LGTM

david-martin:
@jasonmadigan added a `service` label as well for matching with `destination_service_name` so aggregating can be done per httproute (which isn't possible with `destination_workload` in cases of 429s where it's set to `unknown`)

jasonmadigan (approved):

PR #525, opened 2024-04-09
AuthPolicy Atomic Overrides

Boomatang:
# AuthPolicy Atomic Overrides
Closes: #464 

### Base Setup
Create local cluster from this branch
```shell
make local-setup
```
    
Apply Kuadrant CR.
```shell
echo "  
kind: Kuadrant
metadata:
  name: kuadrant
  namespace: kuadrant-system
spec: {}
"
```

Install the toy store app
```shell
```

Add HTTP Route
```shell
echo "
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - path:
        type: Exact
        value: "/toy"
      method: GET
    backendRefs:
    - name: toystore
      port: 80
" | kubectl apply -f -
```

Create secrets for Auth. The secrets are doubled to be created in two namespaces.
```shell
echo "
apiVersion: v1
kind: Secret
metadata:
  name: alice-key
  namespace: default
  labels:
    app: toystore
  annotations:
stringData:
  api_key: IAMALICE
type: Opaque
---
apiVersion: v1
kind: Secret
metadata:
  name: alice-key
  namespace: istio-system
  labels:
    app: toystore
  annotations:
stringData:
  api_key: IAMALICE
type: Opaque
" | kubectl apply -f -
```

Set up gateway url
```shell
```

Set up a watch on AuthConfigs.
```shell
watch kubectl get authconfig -A
```

### Case 1
There is an existing AuthPolicy attached to a HTTPRoute and the system admin attaches a AuthPolicy to the Gateway with atonic overrides.

Create HTTPRoute AuthPolicy. Expect watch to contain one AuthConfig `ap-default-toystore`
```shell
echo "
kind: AuthPolicy
metadata:
  name: toystore
  namespace: default
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: TOYSTORE
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
" | kubectl apply -f - 
```

Check auth is working. Note the name of the Authorization key, `TOYSTORE`. Expected result is `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Add the Gateway AuthPolicy that contains overrides. Expect AuthConfig to change to `ap-istio-system-gateway`
```shell    
echo "
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  overrides:
    rules:
      authentication:
        "api-key-users":
          apiKey:
            selector:
              matchLabels:
                app: toystore
            allNamespaces: true
          credentials:
            authorizationHeader:
              prefix: APIKEY
      response:
        success:
          dynamicMetadata:
            "identity":
              json:
                properties:
                  "userid":
" | kubectl apply -f -
```

Curl the endpoint again with Route Key. Expected `401`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

This show the Gateway overrides are working.

### Case 2
In this case the Gateway AuthPolicy has being defined before any route AuthPolicies have being attached.

Set the cluster back to the configuration as per the [Base Setup](#base-setup).

Install the gateway AuthPolicy. Expected AuthConfig on watch: `ap-istio-system-gateway`
```shell    
echo "
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  overrides:
    rules:
      authentication:
        "api-key-users":
          apiKey:
            selector:
              matchLabels:
                app: toystore
            allNamespaces: true
          credentials:
            authorizationHeader:
              prefix: APIKEY
      response:
        success:
          dynamicMetadata:
            "identity":
              json:
                properties:
                  "userid":
" | kubectl apply -f -
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Apply Auth Policy to route. Expected AuthConfig on watch: `ap-istio-system-gateway`
```shell
echo "
kind: AuthPolicy
metadata:
  name: toystore
  namespace: default
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: TOYSTORE
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
" | kubectl apply -f - 
```
Curl the endpoint with Route Key. Expected `401`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

This show the Gateway overrides are working for routes added after the initial overrides.

### Case 3
The cluster has exiting AuthPolicies attached to the routes and gateways. The system admin later defines the overrides on the gateway AuthPolicy.

Set the cluster back to the configuration as per the [Base Setup](#base-setup).

Create the Route and Gateway AuthPolices. Note that the Gateway AuthPolicy does not define any overrides. Expected AuthConfig in watch: `ap-default-toystore` 
```shell
echo "
kind: AuthPolicy
metadata:
  name: toystore
  namespace: default
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: TOYSTORE
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
---
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
" | kubectl apply -f - 
```
    
Curl the endpoint with Route Key. Expected `200`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `401`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Update the Gateway AuthPolicy to have `overrides`. Expected AuthConfig on watch: `ap-istio-system-gateway`.
```shell    
echo "
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  overrides:
    rules:
      authentication:
        "api-key-users":
          apiKey:
            selector:
              matchLabels:
                app: toystore
            allNamespaces: true
          credentials:
            authorizationHeader:
              prefix: APIKEY
      response:
        success:
          dynamicMetadata:
            "identity":
              json:
                properties:
                  "userid":
" | kubectl apply -f -
```

Curl the endpoint with Route Key. Expected `401`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

This proves that existing Gateway AuthPolicies can be updated to contain `overrides`.

### Case 4
The cluster in this case is configure to have `overrides` on the Gateway from the start. At a later time the system admin removes the `overrides` from the gateway AuthPolicy.

Set the cluster back to the configuration as per the [Base Setup](#base-setup).

Create the Route and Gateway AuthPolices. Note that the Gateway AuthPolicy does define overrides. Expected AuthConfig in watch: `ap-istio-system-gateway`
```shell
echo "
kind: AuthPolicy
metadata:
  name: toystore
  namespace: default
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: TOYSTORE
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
---
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  overrides:
      rules:
        authentication:
          "api-key-users":
            apiKey:
              selector:
                matchLabels:
                  app: toystore
              allNamespaces: true
            credentials:
              authorizationHeader:
                prefix: APIKEY
        response:
          success:
            dynamicMetadata:
              "identity":
                json:
                  properties:
                    "userid":
" | kubectl apply -f - 
```

Curl the endpoint with Route Key. Expected `401`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `200`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Update the Gateway AuthPolicy to remove the overrides. Expect AuthConfig in watch: `ap-default-toystore`
```shell
echo "
kind: AuthPolicy
metadata:
  name: gateway
  namespace: istio-system
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
    response:
      success:
        dynamicMetadata:
          "identity":
            json:
              properties:
                "userid":
" | kubectl apply -f - 
```

Curl the endpoint with Route Key. Expected `200`
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: TOYSTORE IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

Curl endpoint with key defined in the Gateway AuthPolicy, `APIKEY`. Expected `401`.
```shell
curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Authorization: APIKEY IAMALICE' -H 'Host: api.toystore.com' http://$GATEWAY_URL/toy
```

This proves AuthPolicies can be updated to remove `overrides`.````

### DIFF
--- a/api/v1beta2/authpolicy_types.go
+++ b/api/v1beta2/authpolicy_types.go
@@ -24,19 +24,19 @@ type AuthSchemeSpec struct {
 	// Authentication configs.
 	// At least one config MUST evaluate to a valid identity object for the auth request to be successful.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	Authentication map[string]AuthenticationSpec `json:"authentication,omitempty"`
 
 	// Metadata sources.
 	// Authorino fetches auth metadata as JSON from sources specified in this config.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	Metadata map[string]MetadataSpec `json:"metadata,omitempty"`
 
 	// Authorization policies.
 	// All policies MUST evaluate to "allowed = true" for the auth request be successful.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	Authorization map[string]AuthorizationSpec `json:"authorization,omitempty"`
 
 	// Response items.
@@ -47,7 +47,7 @@ type AuthSchemeSpec struct {
 	// Callback functions.
 	// Authorino sends callbacks at the end of the auth pipeline to the endpoints specified in this config.
 	// +optional
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	Callbacks map[string]CallbackSpec `json:"callbacks,omitempty"`
 }
 
@@ -104,13 +104,13 @@ type ResponseSpec struct {
 type WrappedSuccessResponseSpec struct {
 	// Custom success response items wrapped as HTTP headers.
 	// For integration of Authorino via proxy, the proxy must use these settings to inject data in the request.
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	Headers map[string]HeaderSuccessResponseSpec `json:"headers,omitempty"`
 
 	// Custom success response items wrapped as HTTP headers.
 	// For integration of Authorino via proxy, the proxy must use these settings to propagate dynamic metadata.
 	// See https://www.envoyproxy.io/docs/envoy/latest/configuration/advanced/well_known_dynamic_metadata
-	// +kubebuilder:validation:MaxProperties=14
+	// +kubebuilder:validation:MaxProperties=10
 	DynamicMetadata map[string]SuccessResponseSpec `json:"dynamicMetadata,omitempty"`
 }
 
@@ -144,8 +144,18 @@ type CallbackSpec struct {
 // +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.defaults) || !has(self.defaults.rules) || !has(self.defaults.rules.response) || !has(self.defaults.rules.response.success) || !has(self.defaults.rules.response.success.headers) || !self.defaults.rules.response.success.headers.exists(x, has(self.defaults.rules.response.success.headers[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
 // +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.defaults) || !has(self.defaults.rules) || !has(self.defaults.rules.response) || !has(self.defaults.rules.response.success) || !has(self.defaults.rules.response.success.dynamicMetadata) || !self.defaults.rules.response.success.dynamicMetadata.exists(x, has(self.defaults.rules.response.success.dynamicMetadata[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
 // +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.defaults) || !has(self.defaults.rules) || !has(self.defaults.rules.callbacks) || !self.defaults.rules.callbacks.exists(x, has(self.defaults.rules.callbacks[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// RouteSelectors - explicit overrides validation
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.routeSelectors)",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.authentication) || !self.overrides.rules.authentication.exists(x, has(self.overrides.rules.authentication[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.metadata) || !self.overrides.rules.metadata.exists(x, has(self.overrides.rules.metadata[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.authorization) || !self.overrides.rules.authorization.exists(x, has(self.overrides.rules.authorization[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.response) || !has(self.overrides.rules.response.success) || !has(self.overrides.rules.response.success.headers) || !self.overrides.rules.response.success.headers.exists(x, has(self.overrides.rules.response.success.headers[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.response) || !has(self.overrides.rules.response.success) || !has(self.overrides.rules.response.success.dynamicMetadata) || !self.overrides.rules.response.success.dynamicMetadata.exists(x, has(self.overrides.rules.response.success.dynamicMetadata[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.callbacks) || !self.overrides.rules.callbacks.exists(x, has(self.overrides.rules.callbacks[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
 // Mutual Exclusivity Validation
 // +kubebuilder:validation:XValidation:rule="!(has(self.defaults) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit and explicit defaults are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit defaults and explicit overrides are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && has(self.defaults))",message="Explicit overrides and explicit defaults are mutually exclusive"
 type AuthPolicySpec struct {
 	// TargetRef identifies an API object to apply policy to.
 	// +kubebuilder:validation:XValidation:rule="self.group == 'gateway.networking.k8s.io'",message="Invalid targetRef.group. The only supported value is 'gateway.networking.k8s.io'"
@@ -156,6 +166,11 @@ type AuthPolicySpec struct {
 	// Defaults are mutually exclusive with implicit defaults defined by AuthPolicyCommonSpec.
 	// +optional
 	Defaults *AuthPolicyCommonSpec `json:"defaults,omitempty"`
+
+	// Overrides define explicit override values for this policy.
+	// Overrides are mutually exclusive with explicit and implicit defaults defined by AuthPolicyCommonSpec.
+	// +optional
+	Overrides *AuthPolicyCommonSpec `json:"overrides,omitempty"`
 
 	// AuthPolicyCommonSpec defines implicit default values for this policy and for policies inheriting this policy.
 	// AuthPolicyCommonSpec is mutually exclusive with explicit defaults defined by Defaults.
--- a/api/v1beta2/authpolicy_types.go
+++ b/api/v1beta2/authpolicy_types.go
@@ -262,6 +262,10 @@ type AuthPolicy struct {
 	Status AuthPolicyStatus `json:"status,omitempty"`
 }
 
+func (ap *AuthPolicy) IsAtomicOverride() bool {
+	return ap.Spec.Overrides != nil
+}
+
 func (ap *AuthPolicy) TargetKey() client.ObjectKey {
 	ns := ap.Namespace
 	if ap.Spec.TargetRef.Namespace != nil {
@@ -348,6 +352,10 @@ func (ap *AuthPolicy) DirectReferenceAnnotationName() string {
 func (ap *AuthPolicySpec) CommonSpec() *AuthPolicyCommonSpec {
 	if ap.Defaults != nil {
 		return ap.Defaults
+	}
+
+	if ap.Overrides != nil {
+		return ap.Overrides
 	}
 
 	return &ap.AuthPolicyCommonSpec
--- a/api/v1beta2/authpolicy_types.go
+++ b/api/v1beta2/authpolicy_types.go
@@ -156,6 +156,7 @@ type CallbackSpec struct {
 // +kubebuilder:validation:XValidation:rule="!(has(self.defaults) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit and explicit defaults are mutually exclusive"
 // +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit defaults and explicit overrides are mutually exclusive"
 // +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && has(self.defaults))",message="Explicit overrides and explicit defaults are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && self.targetRef.kind == 'HTTPRoute')",message="Overrides are not allowed for policies targeting a HTTPRoute resource"
 type AuthPolicySpec struct {
 	// TargetRef identifies an API object to apply policy to.
 	// +kubebuilder:validation:XValidation:rule="self.group == 'gateway.networking.k8s.io'",message="Invalid targetRef.group. The only supported value is 'gateway.networking.k8s.io'"
--- a/controllers/authpolicy_authconfig.go
+++ b/controllers/authpolicy_authconfig.go
@@ -65,8 +65,17 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 
 	switch obj := targetNetworkObject.(type) {
 	case *gatewayapiv1.HTTPRoute:
+		ok, err := routeGatewayHasAuthOverrides(ctx, obj, r.Client())
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			logger.V(1).Info("targeted gateway has authpolicy with atomic overrides, skipping authorino authconfig for the HTTPRoute authpolicy")
+			utils.TagObjectToDelete(authConfig)
+			r.OverriddenPolicyMap.SetOverriddenPolicy(ap)
+			return authConfig, nil
+		}
 		route = obj
-		var err error
 		hosts, err = kuadrant.HostnamesFromHTTPRoute(ctx, obj, r.Client())
 		if err != nil {
 			return nil, err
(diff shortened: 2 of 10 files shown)


controllers/authpolicy_authconfig.go
```diff
@@ -65,8 +65,17 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 
 	switch obj := targetNetworkObject.(type) {
 	case *gatewayapiv1.HTTPRoute:
+		ok, err := routeGatewayHasAuthOverrides(ctx, obj, r.Client())
```
guicassolato, line 68: Please see https://github.com/Kuadrant/kuadrant-operator/pull/523/files#r1559122844

controllers/authpolicy_authconfig.go
```diff
@@ -65,8 +65,17 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
...
+		if err != nil {
+			return nil, err
+		}
+		if ok {
+			logger.V(1).Info("targeted gateway has authpolicy with atomic overrides, skipping authorino authconfig for the HTTPRoute authpolicy")
+			utils.TagObjectToDelete(authConfig)
+			r.OverriddenPolicyMap.SetOverriddenPolicy(ap)
```
guicassolato, line 75: Nice. I wonder if we could do something similar (or exactly the same!) for the RLP as well.

The RLP doesn't have the same 1:1 relation to the underlying resources as the AP does with the AuthConfigs, so it would somewhere else of course. But useful for setting the status of the policy later on nevertheless.

api/v1beta2/authpolicy_types.go
```diff
@@ -144,8 +144,18 @@ type CallbackSpec struct {
...
+// +kubebuilder:validation:XValidation:rule="self.targetRef.kind != 'Gateway' || !has(self.overrides) || !has(self.overrides.rules) || !has(self.overrides.rules.callbacks) || !self.overrides.rules.callbacks.exists(x, has(self.overrides.rules.callbacks[x].routeSelectors))",message="route selectors not supported when targeting a Gateway"
 // Mutual Exclusivity Validation
 // +kubebuilder:validation:XValidation:rule="!(has(self.defaults) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit and explicit defaults are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && (has(self.routeSelectors) || has(self.patterns) || has(self.when) || has(self.rules)))",message="Implicit defaults and explicit overrides are mutually exclusive"
+// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && has(self.defaults))",message="Explicit overrides and explicit defaults are mutually exclusive"
```
KevFan, line 158: I'm not too sure this makes sense for `AuthPolicy` even if we only support atomic default/overrides at the moment, since the `AuthPolicyCommonSpec` has multiple fields aside from auth rules. For example, perhaps  user setting `defaults.routeselectors` and `override.rules` set valid, where inheriting policies will default to a some routerselctor but must use inherited auth rules from gateway policy  🤔  

Anyhow, just a thought, since if we keep it as is, the implementation logic is probably simpler

Boomatang: This might be a misunderstanding. It was my understanding that in the first implementation was to set defaults and overrides for the spec and not the individual components with in the spec. The diagrams in the [RFC](https://github.com/Kuadrant/architecture/blob/main/rfcs/0009-defaults-and-overrides.md#examples-c---override-policy-entirely-replacing-other-at-lower-level) lead me to believe it was at the spec level. 

If I was wrong in my understanding then that changes a lot about how this implementation would work.

guicassolato: > It was my understanding that in the first implementation was to set defaults and overrides for the spec and not the individual components with in the spec.

This is correct. Which means there's very little use case for `defaults` and `overrides` in a same policy. At least until we introduce merges, conditions and/or start allowing multiple policies of a kind to target a same resource, I suppose users will be limited to either defaults or overrides set at the level of a Gateway.

KevFan: Cool, thanks 👍 In that case, please disregard my comment

KevFan, line 158: In the RLP equivalent, I've opted to add a CEL validation to only allow `overrides` block for policies targeting a gateway. Not sure do we want to do the same here

```suggestion
// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && has(self.defaults))",message="Explicit overrides and explicit defaults are mutually exclusive"
// +kubebuilder:validation:XValidation:rule="!(has(self.overrides) && self.targetRef.kind != 'Gateway')",message="Overrides are only allowed for policies targeting a Gateway resource"
```

controllers/authpolicy_authconfig.go
KevFan: Yes, I think we can and probably should do something similar for RLPs also 👍

controllers/authpolicy_authconfig.go
```diff
@@ -174,6 +183,35 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
...
+		ap := &api.AuthPolicy{}
+		err = c.Get(ctx, client.ObjectKey{Name: anno[1], Namespace: anno[0]}, ap)
+		if err != nil {
+			return false, err
+		}
+
+		override := ap.IsAtomicOverride()
+		if override {
+			return true, nil
+		}
```
KevFan, line 210: nit: unnessary if block ?
```suggestion
		return ap.IsAtomicOverride(), nil
```

guicassolato: I suppose you want to continue in the loop until the end, because this parent may not have an override, but the next one yes?

KevFan: Oh, right, yes, I missed that, please disregard this comment

api/v1beta2/authpolicy_types.go
Boomatang: I was thinking about this, and I felt if we blocked the overrides on the route policies, we should block the defaults on the gateway policies. I do think we should be consistent on what we allow on both sets of policies and that it is easier to add access than remove it at a later date. What do you think @KevFan

Boomatang: I am working off my understand of the [hierarchy chart](https://gateway-api.sigs.k8s.io/geps/gep-2649/?h=#hierarchy), the defaults sets value for the resources above in the tree and overrides set values on resources lower in the tree. As we don't support setting these values on resources about the gateway, we don't need to allow defaults on it. 

I did check the RFC0009 and it would allow setting the defaults as we currently have it but that seems to be different than the definition in the gateway-api GEP2649. I know that GEP is being up dated, so maybe that definition is also being updated.

guicassolato: Not really. Both defaults and overrides are set for the resources lower in the tree. The arrows (up or down) tell which level have precedence in case of conflicting rules set at multiple levels. (The arrow point from higher precedence to lower.)

Boomatang: During the sync we cleared this up. I will be implementing the CEL to disable overrides on the policy attached to the HttpRoutes.

KevFan (approved):
Verified, changes looks good to me 👍

controllers/authpolicy_authconfig.go
```diff
@@ -174,6 +183,35 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
 	return mergeConditionsFromRouteSelectorsIntoConfigs(ap, route, authConfig)
 }
 
+// routeGatewayHasAuthOverrides return true when the gateway which a route is attached to has an attached authPolicy that defines atomic overrides
+func routeGatewayHasAuthOverrides(ctx context.Context, route *gatewayapiv1.HTTPRoute, c client.Client) (bool, error) {
+	for idx := range route.Spec.ParentRefs {
+		parentRef := route.Spec.ParentRefs[idx]
+		gw := &gatewayapiv1.Gateway{}
+		err := c.Get(ctx, client.ObjectKey{Name: string(parentRef.Name), Namespace: string(*parentRef.Namespace)}, gw)
```
guicassolato, line 191:
```suggestion
		namespace := ptr.Deref(parentRef.Namespace, gatewayapiv1.Namespace(route.GetNamespace()))
		err := c.Get(ctx, client.ObjectKey{Name: string(parentRef.Name), Namespace: string(namespace)}, gw)
```

controllers/authpolicy_authconfig.go
```diff
@@ -174,6 +183,35 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
...
+			return false, err
+		}
+
+		annotation, ok := gw.GetAnnotations()[common.AuthPolicyBackRefAnnotation]
+		if !ok {
+			continue
+		}
+		anno := strings.Split(annotation, "/")
+		ap := &api.AuthPolicy{}
+		err = c.Get(ctx, client.ObjectKey{Name: anno[1], Namespace: anno[0]}, ap)
```
guicassolato, line 202: Coudl also use

```suggestion
		ap := &api.AuthPolicy{}
		err = c.Get(ctx, utils.NamespacedNameToObjectKey(annotation, gw.Namespace), ap)
```

controllers/authpolicy_controller.go
```diff
@@ -167,6 +167,44 @@ func (r *AuthPolicyReconciler) reconcileResources(ctx context.Context, ap *api.A
...
+			}
+
+			ref := &api.AuthPolicy{}
+			err = r.Client().Get(ctx, policyKey, ref)
+			if err != nil {
+				return err
+			}
+
+			refNetworkObject := &gatewayapiv1.HTTPRoute{}
+			err = r.Client().Get(ctx, ref.TargetKey(), refNetworkObject)
```
guicassolato, line 197: Maybe add a comment here that this only works because of single policy of a kind allowed per target object. Since the gateway policy was skipped in line 186, all the other refs are of policies targeting routes. In the future, that may no longer be the case.

controllers/authpolicy_authconfig.go
```diff
@@ -174,6 +183,35 @@ func (r *AuthPolicyReconciler) desiredAuthConfig(ctx context.Context, ap *api.Au
...
+		if err != nil {
+			return false, err
+		}
+
+		override := ap.IsAtomicOverride()
+		if override {
```
guicassolato, line 208: nit: Do we need the variable?

```suggestion
		if ap.IsAtomicOverride() {
```

controllers/authpolicy_controller.go
Boomatang: I see your point, but I don't think a comment is worth it. I would assume there are other spots in the code base that assume only one type of resources is attached to the gateway. When the day comes when we support it, I would expect that would be a large change to the whole code base and not just here.

guicassolato (approved):
This looks very good, @Boomatang. Thank you!


Leaving here almost as a self reminder. In the future, we can revisit this and try to leverage more the `gatewayapi.Topology` and `gatewayapi.TopologyIndexes` (aka [DAG](https://github.com/Kuadrant/kuadrant-operator/issues/530)) as an alternative to nested API reads such as: https://github.com/Kuadrant/kuadrant-operator/blob/7efe3c9fea55d87c1879cee8372e6badfc9b05ba/controllers/authpolicy_authconfig.go#L192-L204

guicassolato:
Some additional verification steps I tried, for the CEL validation piece:

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication: {}
  overrides:
    rules:
      authentication: {}
EOF
The AuthPolicy "my-policy" is invalid:
* spec: Invalid value: "object": Implicit defaults and explicit overrides are mutually exclusive
* spec: Invalid value: "object": Overrides are not allowed for policies targeting a HTTPRoute resource
```

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  defaults:
    rules:
      authentication: {}
  overrides:
    rules:
      authentication: {}
EOF
The AuthPolicy "my-policy" is invalid:
* spec: Invalid value: "object": Explicit overrides and explicit defaults are mutually exclusive
* spec: Invalid value: "object": Overrides are not allowed for policies targeting a HTTPRoute resource
```

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
    namespace: istio-system
  rules:
    authentication: {}
  overrides:
    rules:
      authentication: {}
EOF
The AuthPolicy "my-policy" is invalid: spec: Invalid value: "object": Implicit defaults and explicit overrides are mutually exclusive
```

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
    namespace: istio-system
  defaults:
    rules:
      authentication: {}
  overrides:
    rules:
      authentication: {}
EOF
The AuthPolicy "my-policy" is invalid: spec: Invalid value: "object": Explicit overrides and explicit defaults are mutually exclusive
```

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
    namespace: istio-system
  overrides:
    rules:
      authentication: {}
    routeSelectors:
    - matches:
      - method: GET
EOF
The AuthPolicy "my-policy" is invalid: spec: Invalid value: "object": route selectors not supported when targeting a Gateway
```

```
$ kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: my-policy
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
    namespace: istio-system
  overrides:
    rules:
      authentication:
        anonymous-access:
          anonymous: {}
          routeSelectors:
          - matches:
            - method: GET
EOF
The AuthPolicy "my-policy" is invalid: spec: Invalid value: "object": route selectors not supported when targeting a Gateway
```

PR #527, opened 2024-04-09
Rate limiting limits controller

eguzki:
### What

Custom controller for Limitador limits reading from DAG Gateway API topology and RateLimitPolicies

Limitador cluster envoy filter controller reading from DAG Gateway API topology and RateLimitPolicies

Kuadrant rate limiting controllers no longer depends on *back references* annotation: `kuadrant.io/ratelimitpolicies`. Instead, controllers rely on the DAG Gateway API topopology to read cluster state. Kuadrant rate limiting controller responsible for the limitador's limits is also based on this approach. So is the limitador cluster envoy filter controller

Fix reconciliation event blind spots due to race conditions or timing issues. The reconciled resource reflects latest updates from the cluster (Gateways, Routes and policies).

Work being done as part of #530


controllers/rate_limiting_limits_controller.go
```diff
@@ -0,0 +1,258 @@
...
+		kuadrantgatewayapi.WithGateways(utils.Map(gwList.Items, ptr.To)),
+		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To)),
+		kuadrantgatewayapi.WithPolicies(policies),
+		kuadrantgatewayapi.WithLogger(logger),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	topologyindexes := kuadrantgatewayapi.NewTopologyIndexes(t)
```
eguzki, line 178: TODO: Common function to load the topology for all routes, gateways and policies.

eguzki:
Closing in favor of https://github.com/Kuadrant/kuadrant-operator/pull/893

I leave the branch undeleted because there are some integration tests that can be backported

Issue #414, opened 2024-02-06
RLP Enforced condition


KevFan:
Part of https://github.com/Kuadrant/architecture/issues/38

Related issues: https://github.com/Kuadrant/kuadrant-operator/issues/140

Implement the `Enforced` condition type for RateLimitPolicy

KevFan:
Some thoughts on this:

In order to check that the RateLimitPolicy is enforced in Limitador, there is multiple options for this:

1. Compare the limits config map on k8s cluster to the one loaded into the limitador pod 
    * Easier option
    * Naive assumption that limits are and will be loaded into Limitador correctly without any issues

or 

2. A deeper check, compare the limits config map entries with the limits loaded into Limitador redis using the `/limits/{namespace}` API provided by limitador itself. 
    * More difficult option
    * Extra requests made to Limitador for this check (can potentially impact Limitador performance by these requests)

Option 1 and 2 require shelling into the Limitador pod in order to compare / access the Limitador API unless we want expose a service endpoint for Limitador in Limitador operator if we want to go with option 2 at https://github.com/Kuadrant/limitador-operator/blob/main/pkg/limitador/k8s_objects.go#L51

Either or, I feel this logic should exist in Limtiador Operator itself to check if the Limits are correctly loaded into Limtiador. 
Kuadrant Operator should then be able infer this via the Limitador CR and set whether the RateLimitPolicy is enforced (similar to how AuthPolicy can infer the enforcement via AuthConfig https://github.com/Kuadrant/kuadrant-operator/pull/411).

Interested in @eguzki @didierofrivia @alexsnaps thoughts on this and on what's the preferred approach 🤔

didierofrivia:
I'd only say that we shouldn't go to the storage option of Limitador to check this, since that is going to far for the Operator. The first option, as simple as it sounds, could be a good first approach IMHO. Then we could decide to evolve it to a more thorough probe... However, I agree with you that it should be Limitador Operator where the logic should reside.

eguzki:
My take on this:

The RLP controller should rely on the Limitador CR status. 

[...]

PR #533, opened 2024-04-11
[wip] feat: rlp enforced condition

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/414#issuecomment-1992165890

With https://github.com/Kuadrant/limitador-operator/pull/127, we can use Limitador `Ready` condition to naviely assume that limits are enforced at least to limitador


KevFan:
Will be addressed as part of https://github.com/Kuadrant/kuadrant-operator/pull/523

Issue #468, opened 2024-03-13
AuthPolicy discoverability


guicassolato:

guicassolato:
Implemented in #536

PR #536, opened 2024-04-11
PolicyAffected condition for AuthPolicy and RateLimitPolicy

### LINKED ISSUE #467: RateLimitPolicy discoverability

guicassolato:
- [x] PolicyAffected status condition added to the Gateway objects for AuthPolicy and RateLimitPolicy
- [ ] PolicyAffected status condition added to HTTPRoute objects for AuthPolicy and RateLimitPolicy

Closes #467 and #468


controllers/kuadrant_status.go
```diff
@@ -196,7 +195,7 @@ func BuildPolicyAffectedCondition(conditionType string, policyObject runtime.Obj
 		return condition
 	}
 
-	condition.Message = fmt.Sprintf("policy success. Object affected by policy %s in namespace %s with name %s ", policyObject.GetObjectKind().GroupVersionKind().String(), objectMeta.GetNamespace(), objectMeta.GetName())
+	condition.Message = fmt.Sprintf("policy success. Object affected by policy %s in namespace %s with name %s", policyObject.GetObjectKind().GroupVersionKind().String(), objectMeta.GetNamespace(), objectMeta.GetName())
```
guicassolato, line 198: @mikenairn @maleck13, how important would you say having the name of the policy in the message here is?

I guess for gateway policies – which is always the case of DNSPolicy and TLSPolicy – it's probably good?

This doesn't work well though when the gateway can be affected by more than one policy, such as with AuthPolicies and RateLimitPolicies, that can affect gateways by targeting the routes as well, or, in general, when more than one policy of a kind can target the Gateway (not allowed today).

Currently, the message contains the name of the last policy which caused the PolicyAffected condition to be set (to either true or false.) We also have the full list of policies that affect a gateway stored in the back ref annotation of the gateway object.

I guess my actual question is: are we OK with only the last policy mentioned in the message for each and all kinds of policies or I may change the message to make it more generic without mentioning the name of the policy?

pkg/library/reconcilers/target_ref_reconciler.go
```diff
@@ -209,3 +211,35 @@ func (r *TargetRefReconciler) ReconcileGatewayPolicyReferences(ctx context.Conte
...
+	for i, gw := range gatewayDiffs {
+		previous := gw.DeepCopy()
+		gwCondition := condition.DeepCopy()
+		gwCondition.ObservedGeneration = gw.Generation
+		meta.SetStatusCondition(&gatewayDiffs[i].Status.Conditions, *gwCondition)
+		if !reflect.DeepEqual(previous.Status.Conditions, gw.Status.Conditions) {
+			if err := r.Client.Status().Update(ctx, gw.Gateway); err != nil {
+				return err
+			}
+			// TODO(guicassolato): for inherited policies only – add the policyaffetced status condition on all routes accepted by the gateway
```
guicassolato, line 227: This is assuming that, no matter at level the policy targets an object (gateway _or_ route), all objects (gateways _and_ routes) will be marked as affected by the policy.

Is that really what we want or we want to set this condition only on the object targeted by the policy plus its children (but never upwards)?

cc @maleck13 @R-Lawton @david-martin

pkg/library/reconcilers/target_ref_reconciler.go
```diff
@@ -209,3 +213,144 @@ func (r *TargetRefReconciler) ReconcileGatewayPolicyReferences(ctx context.Conte
...
+	})
+	if i < 0 {
+		logger.V(1).Info("cannot find parent status, creating new one")
+		route.Status.RouteStatus.Parents = append(route.Status.RouteStatus.Parents, gatewayapiv1.RouteParentStatus{
+			ControllerName: controllerName,
+			ParentRef: gatewayapiv1.ParentReference{
+				Kind:      ptr.To(gatewayapiv1.Kind(gateway.Kind)),
+				Name:      gatewayapiv1.ObjectName(gateway.GetName()),
+				Namespace: ptr.To(gatewayapiv1.Namespace(gateway.GetNamespace())),
+			},
```
guicassolato, line 320: This will create in the route a parent status block with `controllerName: kuadrant.io/policy-controller` and `targetRef` pointing to the affected gateway.

One can only assume what the actual `controllerName` of the gateway is, but this information is omitted here.

Should we instead reuse the same parent status block of the gateway controller and not have one specifically for Kuadrant?

controllers/kuadrant_status.go
maleck13: Interesting question. I don't yet have a strong opinion. If the policy is truly affecting the gateway and not being overridden, does it not make sense to have multiple conditions perhaps at the listener and top level? I guess it comes down to what we want to achieve with the condition. Is it that we want to indicate this gateway is affected by at least one policy or do we want a pointer to say these specific policies are affecting this gateway. Sorry for the late reply

pkg/library/reconcilers/target_ref_reconciler.go
maleck13: I would go for the latter. It makes more sense to me this way

guicassolato: Ok. I need to change a few things then. Thanks!

controllers/target_status_controller.go
```diff
@@ -0,0 +1,371 @@
...
+
+const PolicyAffectedConditionPattern = "kuadrant.io/%sAffected" // Policy kinds are expected to be named XPolicy
+
+// TargetStatusReconciler reconciles a the status stanzas of objects targeted by Kuadrant policies
+type TargetStatusReconciler struct {
+	*reconcilers.BaseReconciler
+}
+
+func (r *TargetStatusReconciler) Reconcile(eventCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Logger().WithValues("gateway", req.NamespacedName, "request id", uuid.NewString())
```
guicassolato, line 53: Thinking about adding `request id` at all reconcilers. WDYT?

Boomatang: I was also thinking about this but wondering should it follow the guidelines set in opentelemetry, but I am not sure if our current approach fits into that style if parent ids.

Boomatang:
Everything looks good to me. That said I could only run the verification steps for the AuthPolicies. I have a local issue with RateLimiting, unrelated to the changes in this PR. You will need to find someone to do the full verification.

controllers/target_status_controller.go
guicassolato: Actually, I may have seen `reconcileID` in some other log trail. No longer sure if this `request id` will be useful in the end.

pkg/library/reconcilers/target_ref_reconciler.go
guicassolato: Changed so direct policies (TLS/DNS) only add to the statuses of the objects they target (Gateways only) and inherited (auth/RL) to the objects they target plus children beneath.

controllers/kuadrant_status.go
guicassolato: Ended applying the following:
- If _only_ a policy targeting the object indirectly exists (i.e. Gateway policy with no route policy) → this policy will be referred in the PolicyAffected message

- If a policy targeting the object directly exists → this policy will be referred in the PolicyAffected message even if the policy is being overridden by another. To confirm if a policy referred in PolicyAffected message is being enforced, the user must check the status of the policy object itself.

guicassolato:
Verification steps updated in the description of the PR to:
- simplify the topology (2 routes instead of 3);
- add a step for the TLSPolicy (DNSPolicy should also work);
- include `curl`s to the app (to check as well the enforcement of the policy – optional.)

controllers/target_status_controller_test.go
```diff
@@ -0,0 +1,560 @@
...
+				policy.Name = "gateway-auth"
+				policy.Spec.TargetRef = gatewayapiv1alpha2.PolicyTargetReference{
+					Group:     gatewayapiv1.GroupName,
+					Kind:      "Gateway",
+					Name:      testGatewayName,
+					Namespace: ptr.To(gatewayapiv1.Namespace(testNamespace)),
+				}
+			})
+			Expect(k8sClient.Create(context.Background(), policy)).To(Succeed())
+			Eventually(policyAcceptedAndTargetsAffected(policy), 30*time.Second, 5*time.Second).Should(BeTrue())
```
KevFan, line 197: Should the toystore route name also be passed here and possibly a few other places to check that GW policies put the affected condition to underlying untargeted routes ? 🤔 
```suggestion
			Eventually(policyAcceptedAndTargetsAffected(policy, testHTTPRouteName), 30*time.Second, 5*time.Second).Should(BeTrue())
```

KevFan (approved):
Verified, works perfectly 💯  
Have a minor comment / question on integration test that I'm happy for it to be addressed at a later PR if it's an issue since it's very minor 👍

controllers/target_status_controller_test.go
guicassolato: Nice catch! Fixed.

guicassolato:
I’m merging this in spite of the integration test failing. It’s unrelated to the PR. In fact, [similar ones are currently failing in main](https://github.com/Kuadrant/kuadrant-operator/actions/runs/8798283409/job/24144947686) as well.

PR #537, opened 2024-04-11
fix direct reference annotation reconciliation

eguzki:
### What

Found issue with the policies direct references in network resources used to prevent multiple policies (of the same kind) to target the same network resource.

The issue can be reproduced by:
* Crate  RLP A -> Route A
* Change RLP A target to Route B: RLP A -> Route B
* Route A annotation `kuadrant.io/ratelimitpolicy` is never deleted and therefore, new policies cannot target Route A.

### DIFF
--- a/controllers/helper_test.go
+++ b/controllers/helper_test.go
@@ -391,6 +391,99 @@ func testRLPIsAccepted(rlpKey client.ObjectKey) func() bool {
 	}
 }
 
+func testRLPIsNotAccepted(rlpKey client.ObjectKey) func() bool {
+	return func() bool {
+		existingRLP := &kuadrantv1beta2.RateLimitPolicy{}
+		err := k8sClient.Get(context.Background(), rlpKey, existingRLP)
+		if err != nil {
+			logf.Log.V(1).Info("ratelimitpolicy not read", "rlp", rlpKey, "error", err)
+			return false
+		}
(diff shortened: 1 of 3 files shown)


guicassolato (approved):
I haven't tried the verification steps, but this looks good... and needed.

Only risk I can think of is that there's no rollback in case of failure mid-step 2 or during step 3. However, and to be fair, this is pretty much the case of most of our reconciliation functions ATM.

🚀

eguzki:
I have been looking at the integration tests failing.  The test failed because the tear down (`AfterEach`) expired. It has nothing to do with the test itself (one regarding validation of targets happen in the same namespace).

The process happened as follow:
* The tear down method deleted the namespace. That triggered deletion of all the resources without any guarantee on the order. 
* First, the controller responsible of the `kuadrant.io/namespace` annotation, got Kuadrant CR deletion event. Thus, the controller removed that annotation from the gateway.
* Then, the rate limit policy controller got event for the deletion of the policy. The method handling removal of the policy tried to read the kuadrant namespace to get limitador's namespace and reconcile limitador's limits. An error was thrown and the method handling the removal of the policy could not finish, never removing the finalizer. Hence, the tear down method timed out. 

I think I have fixed the issue in https://github.com/Kuadrant/kuadrant-operator/pull/527. I will make sure the controllers, on handling removal, they do not rely on some order when objects are deleted like when a namespace is deleted.

PR #538, opened 2024-04-11
fix kuadrant CR finalizer

### LINKED ISSUE #363: Not having any available Istio instalation prevents deletion of Kuadrant resource
Suppose you do not have any Istio installation available and no OSSM 2.x present. In that case, the Kuadrant resource will have a status saying `maista.io/v1` resources are missing and if you try to delete the resource, it will get stuck.

The workaround is to either fix your installation (TODO verify) or remove finalizer from a resource

I believe it is because of a missing try/catch for OSSM 2.X resource, which raises an error about missing resources.

eguzki:
### What

Fix Kuadrant CR cleaning up resources

When Kuadrant CR was deleted, the clean up process failed and the finalizer was never removed. The issue is specifically when the istio config map was not found, error was raised and the finalizer was never removed. 

Fixes #363

The whole  auth provider registration in istio/maistra code deserves a refactor. More :bomb: may arise 

### Verification Steps

* Setup the environment:   
                         
```sh                    
make local-setup         
```        

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

Delete Istio

```
kubectl delete namespace istio-system
```

```
kubectl delete namespace istio-operator
```

Delete Kuadrant CR

```
kubectl delete kuadrant kuadrant -n kuadrant-system
```
The delete command should not be stuck and finish successfully.

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -338,10 +338,13 @@ func (r *KuadrantReconciler) getIstioConfigObjects(ctx context.Context, logger l
 
 	istioConfigMap := &corev1.ConfigMap{}
 	if err := r.GetResource(ctx, client.ObjectKey{Name: controlPlaneConfigMapName(), Namespace: controlPlaneProviderNamespace()}, istioConfigMap); err != nil {
-		logger.V(1).Info("failed to get istio configMap", "key", istKey, "err", err)
-		return configsToUpdate, err
+		if !apierrors.IsNotFound(err) {
+			logger.V(1).Info("failed to get istio configMap", "key", istKey, "err", err)
+			return configsToUpdate, err
+		}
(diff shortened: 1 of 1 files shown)


KevFan (approved):
Verified, LGTM 👍

KevFan:
Build catalog is failing due to `quay.io/kuadrant/kuadrant-operator-bundle:fix-kuadrant-cr-finalizer` is pushed as bundle image but catalog build is expecting `quay.io/kuadrant/kuadrant-operator-bundle:fix-kuadrant-CR-finalizer`. Probably related to the branch name using uppercase letters. Happy for this to merge even with this workflow failing since it's unrelated

eguzki:
Got it, never use upper case letters in branches

PR #526, opened 2024-04-09
adding refined app developer dashboard and telemetry piece

R-Lawton:
Refined the pre-existing app developer dashboard
What's new:
* Split the requests from one singular table and panels for both error requests and successful request into there own sections
* Added in tooltips for all panels
* Moved the location of panels so errors show at the top followed by success followed by API summary
* Added variables for API name and API endpoint path selection
# Verification
1. To populate the panels run the single cluster getting started guide https://docs.kuadrant.io/getting-started-single-cluster/

2. Then run the commands from the observability README https://github.com/Kuadrant/kuadrant-operator/tree/main/config/observability#readme
You will need to add the label `kubectl label httproute toystore -n default deployment=toystore`

Screenshot 2024-04-09 at 17 24 25
Screenshot 2024-04-09 at 17 24 34

PR #540, opened 2024-04-11
Refine the app developer dashboard

### LINKED ISSUE #435: Refine & release the App Developer dashboard from the API Quickstart

david-martin:
Follow up on #526
Closes #435 

In addition to the changes add in #526, also includes:

- repeating row per API/HTTPRoute
- more compact view of all data already included
- consistent colours for success/error/%
- %ile color gradient
- text box at top to explain dashboard & row repeating
- tooltips updated

image

### DIFF
--- a/config/observability/kustomization.yaml
+++ b/config/observability/kustomization.yaml
@@ -19,6 +19,8 @@ resources:
 # scrape config.
 # See https://github.com/prometheus-operator/prometheus-operator/issues/3071#issuecomment-763746836
   - additional-scrape-configs.yaml
+  #https://istio.io/latest/docs/reference/config/telemetry/#MetricSelector-IstioMetric
+  - telemetry.yaml
 
 patchesStrategicMerge:
   - cluster_role.yaml
--- a/config/observability/telemetry.yaml
+++ b/config/observability/telemetry.yaml
@@ -1,0 +1,28 @@
+apiVersion: telemetry.istio.io/v1alpha1
+kind: Telemetry
+metadata:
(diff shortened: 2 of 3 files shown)


thomasmaas:
Screenshot 2024-04-12 at 17 16 19

I know this has already been merged, wondering if for a next iteration it might make sense to combine these boxes in 1 column, and maybe add a percentage?

```
total
0,43 r/s (380 r, 100 %)
success
0,43 r/s (321 r, 84 %)
error
0,43 r/s (59 r, 6 %)
```

david-martin:
Thanks for the suggestion @thomasmaas.
It would definitely look less cluttered without the title duplication.
We're at the mercy of what you can do with grafana panels, though I do think there's a way to have multiple queries in a single state that appear horizontal or vertical. Could be a way of doing that. Or a table view, though we moved away from that as the single stats allow customisation on the text color.

Would you mind creating a new issue for that if you feel strongly. (There's actually a shortcut to reference your comment in a new issue that I haven't used before. Might be useful)

Thanks

PR #542, opened 2024-04-12
Tidy up PE dashboard

david-martin:
- consistent colouring with #540 
- add panels for healthy/unhealty gateways
- aggregate into 2 rows (gateways & apis) instead of 3 rows
- add text box at top with summary of dashboard and httproute labelling info
- remove the deployments table as it felt out of place
- use shared crosshair for time series
- use hue gradient in time series
- use 'all' tooltip mode on time series

I couldn't get it to use different shades of yellow only for the errors time series. I went with a yellow to red option, which may be the best option there to try keep close to a yellow colour for errors.

image

### DIFF
--- a/examples/dashboards/platform_engineer.json
+++ b/examples/dashboards/platform_engineer.json
@@ -25,92 +25,45 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "gnetId": 7630,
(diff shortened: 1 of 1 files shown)


R-Lawton (approved):
Looks great

PR #543, opened 2024-04-12
propagate health check to DNS Records

philbrookes:


examples/dnspolicy/application.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: gateway.networking.k8s.io/v1
```
maleck13, line 1: is this example unique to dnspolicy?

philbrookes: it's an example of the minimum setup that results in DNS Records from a DNS Policy.

maleck13 (approved):

PR #548, opened 2024-04-16
[WIP] Feature - Distributed DNS

philbrookes:
update manifests and bundle
upgrade controller-gen to 0.14.0 (#460)
dns: Bump DNS Operator version
propagate health check to DNS Records


controllers/dns_helper.go
```diff
@@ -134,7 +125,7 @@ func (dh *dnsHelper) setEndpoints(mcgTarget *multicluster.GatewayTarget, dnsReco
 	}
 
 	sort.Slice(endpoints, func(i, j int) bool {
-		return endpoints[i].SetID() < endpoints[j].SetID()
+		return getSetID(endpoints[i]) < getSetID(endpoints[j])
```
maleck13, line 128: why did we change from SetID to getSetID ?

.github/workflows/test.yaml
```diff
@@ -1,6 +1,7 @@
 name: Test
 
 on:
+  merge_group:
```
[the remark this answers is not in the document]
mikenairn: This should be removed, it was required to use the merge queue on that branch

controllers/dns_helper.go
mikenairn: We switched from out own implementation of Endpoint to an external dns one which doesn't have a `SetID()`. That said, it does have [Key](https://github.com/kubernetes-sigs/external-dns/blob/master/endpoint/endpoint.go#L305) method which we could use instead of adding  `getSetID`

controllers/dns_helper.go
```diff
@@ -258,50 +249,50 @@ func (dh *dnsHelper) getLoadBalancedEndpoints(mcgTarget *multicluster.GatewayTar
 		endpoints = append(endpoints, clusterEndpoints...)
 
 		//Create lbName CNAME (lb-a1b2.shop.example.com -> default.lb-a1b2.shop.example.com)
-		endpoint = createOrUpdateEndpoint(lbName, []string{geoLbName}, kuadrantdnsv1alpha1.CNAMERecordType, string(geoCode), DefaultCnameTTL, currentEndpoints)
+		ep = createOrUpdateEndpoint(lbName, []string{geoLbName}, kuadrantdnsv1alpha1.CNAMERecordType, string(geoCode), DefaultCnameTTL, currentEndpoints)
 
 		//Deal with the default geo endpoint first
 		if geoCode.IsDefaultCode() {
-			defaultEndpoint = endpoint
+			defaultEndpoint = ep
```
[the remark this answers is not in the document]
mikenairn: Clash with the endpoint import name I'd guess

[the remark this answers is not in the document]
mikenairn: That said, it 's `externaldns`, so maybe not. Does seem unnecessary.

maleck13 (approved):
Looks like  a few format and import things to fix but the changes look reasonable to me

.github/workflows/test.yaml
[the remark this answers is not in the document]
philbrookes: I suspect @mikenairn added this.

controllers/dns_helper.go
philbrookes: this was because it changed from our own endpoint implementation with a `setID` function, to a externaldns endpoint that doesn't have that.

controllers/dns_helper.go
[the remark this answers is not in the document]
philbrookes: the var name `endpoint` was colliding with an import.

controllers/dns_helper.go
philbrookes: I've switched this to use Key().SetIdentifer now.

controllers/dns_helper.go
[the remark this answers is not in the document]
philbrookes: I've reset these var names.

controllers/dns_helper.go
mikenairn: @philbrookes You need to use the value returned by Key() as the key not just the `SetIdentifier`

controllers/dnspolicy_controller_multi_cluster_test.go
```diff
@@ -46,6 +54,16 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
 			Gateway
 		Expect(k8sClient.Create(ctx, gateway)).To(Succeed())
 
+		ownerID = common.ToBase36HashLen(fmt.Sprintf("%s-%s-%s", ownerID, gateway.Name, gateway.Namespace), utils.ClusterIDLength)
+
+		//clusterOneIDHash = common.ToBase36HashLen(fmt.Sprintf("%s-%s-%s", TestClusterNameOne, gateway.Name, gateway.Namespace), utils.ClusterIDLength)
+		//clusterTwoIDHash = common.ToBase36HashLen(fmt.Sprintf("%s-%s-%s", TestClusterNameTwo, gateway.Name, gateway.Namespace), utils.ClusterIDLength)
```
mikenairn, line 60: comments can be removed

controllers/dns_helper.go
```diff
@@ -103,25 +101,14 @@ func gatewayDNSRecordLabels(gwKey client.ObjectKey) map[string]string {
...
-	var endpoints []*kuadrantdnsv1alpha1.Endpoint
+	var endpoints []*externaldns.Endpoint
 
-	//Health Checks currently modify endpoints so we have to keep existing ones in order to not lose health check ids
-	currentEndpoints := make(map[string]*kuadrantdnsv1alpha1.Endpoint, len(dnsRecord.Spec.Endpoints))
+	//Health Checks currently modify endpoints, so we have to keep existing ones in order to not lose health check ids
+	currentEndpoints := make(map[string]*externaldns.Endpoint, len(dnsRecord.Spec.Endpoints))
 	for _, endpoint := range dnsRecord.Spec.Endpoints {
-		currentEndpoints[endpoint.SetID()] = endpoint
+		currentEndpoints[endpoint.Key().SetIdentifier] = endpoint
```
mikenairn, line 111:
```suggestion
		currentEndpoints[endpoint.Key()] = endpoint
```

controllers/dns_helper.go
```diff
@@ -134,7 +121,7 @@ func (dh *dnsHelper) setEndpoints(mcgTarget *multicluster.GatewayTarget, dnsReco
 	}
 
 	sort.Slice(endpoints, func(i, j int) bool {
-		return endpoints[i].SetID() < endpoints[j].SetID()
+		return endpoints[i].Key().SetIdentifier < endpoints[j].Key().SetIdentifier
```
mikenairn, line 124:
```suggestion
		return endpoints[i].Key() < endpoints[j].Key()
```

mikenairn (changes requested):

controllers/dns_helper.go
philbrookes: I can't do this, I get an error, that's why I wondered if you meant that I should use the setIdentifier here.

philbrookes:
moved : https://github.com/Kuadrant/kuadrant-operator/pull/550

PR #511, opened 2024-03-26
Podman fixes for v0.6.2

jasonmadigan:
- **Podman fixes**
- **added podman support for docker-network-ipaddresspool.sh**
- Based atop of `release-v0.6.1` branch

PR #557, opened 2024-04-19
Make selection of subnet using docker more flexible.

Boomatang:
This removes the assumption that the first item in IPAM.Config is the IPv4 entry. However the assumption that only the IPv4 has a value for the gateway is being made.

### DIFF
--- a/utils/docker-network-ipaddresspool.sh
+++ b/utils/docker-network-ipaddresspool.sh
@@ -29,7 +29,7 @@ set -e
 
 # Fallback to docker version of cmd
 if [[ -z "$SUBNET" ]]; then
-  SUBNET=$(docker network inspect $networkName | ${YQ} e '.[] | select(.IPAM.Config[] | has("Gateway")) | .IPAM.Config[] | select(has("Gateway")) | .Subnet' -)
+  SUBNET=$(docker network inspect $networkName --format '{{json .IPAM.Config}}' | ${YQ} '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet')
 fi
 # Neither worked, error out
 if [[ -z "$SUBNET" ]]; then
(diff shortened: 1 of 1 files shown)


eguzki:
We had 

```bash
## Parse kind network subnet
## Take only IPv4 subnets, exclude IPv6
SUBNET=$(docker network inspect $networkName --format '{{json .IPAM.Config }}' | \
Feb 14, 2024
[docker-network-ipaddresspool.sh: little enhancements](https://github.com/Kuadrant/kuadrant-operator/commit/a825617c9c344efe48bdad4b6905b06ac7332b8b)
    ${YQ} '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet')
if [[ -z "$SUBNET" ]]; then
   echo "Error: parsing IPv4 network address for '$networkName' docker network"
   exit 1
fi
```

in the first place and it was changed on https://github.com/Kuadrant/kuadrant-operator/pull/517 when adding podman local setup. Maybe @david-martin has a good reason.

Maybe add it back?

eguzki:
When I run 
```
docker network inspect kind | yq e '.[] | select(.IPAM.Config[] | has("Gateway")) | .IPAM.Config[] | select(has("Gateway")) | .Subnet' -
```

I get

```
172.18.0.0/16
fc00:f853:ccd:e793::/64
```

When I run 

```
docker network inspect kind --format '{{json .IPAM.Config }}' | \
    yq '.[] | select( .Subnet | test("^((25[0-5]|(2[0-4]|1\d|[1-9]|)\d)\.?\b){4}/\d+$")) | .Subnet'
```
I get
```
172.18.0.0/16
```

david-martin:
> in the first place and it was changed on #517 when adding podman local setup. Maybe @david-martin has a good reason

The change landed on a working branch [here](https://github.com/Kuadrant/kuadrant-operator/pull/511/files#diff-5b998a9878a3a66cef715c09332ca6dbeac34a5dc204d399b24f3a8cdbcf1d6bL13) before the changes landed back on main.
@jasonmadigan do you recall the reason for the change to the `network inspect` cmd? I think the suggested cmd came from the kind docs or metallb docs?

Either way, reverting the docker version of the cmd sounds reasonable if it fixes an issue and still gives the desired behaviour.

eguzki (approved):

PR #565, opened 2024-04-23
tests: speed & other minor improvements

KevFan:
Some refactoring to improve testing speed and other minor changes:

* Remove deploying toystore app
  * This is not used at all and deleting a namespace with it deployed takes a lot longer than without (i.e. this can compound for each test and therefore increase the overall test suite runtime)
* Deploy single kuadrant install in BeforeAll/AfterAll in seperate namespace
  * Installing kuadrant takes time and is generally shared between tests so there is no need to set it each time in `BeforeEach`
  * Tests with these stages must be mark as `Ordered`. For now this is okay until we are able to run the tests concurrently
* Use spec context and timeouts per spec
  * the `BeforeEach` time also uses this same timeout (https://onsi.github.io/ginkgo/#the-spectimeout-and-nodetimeout-decorators). `Eventually` will use this context as timeout and should exit the assertion quicker which makes the overall test complete faster  
* Remove unnecessart docker build in test workflow
  * Image is never deployed as part of integration test as the test suite itself runs the controller reconcilers

Integreation test time in now ~34 mins (down from over an hour previously)

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -61,7 +61,6 @@ import (
 var cfg *rest.Config
 var k8sClient client.Client
 var testEnv *envtest.Environment
-var appNamespace string
 
 func testClient() client.Client { return k8sClient }
 
@@ -253,24 +252,17 @@ var _ = BeforeSuite(func(ctx SpecContext) {
 
 	Expect(err).NotTo(HaveOccurred())
 
-	CreateNamespaceWithContext(ctx, &appNamespace)
-
-	err = ApplyResources(filepath.Join("..", "examples", "toystore", "toystore.yaml"), k8sClient, appNamespace)
-	Expect(err).ToNot(HaveOccurred())
-
 	go func() {
 		defer GinkgoRecover()
 		err = mgr.Start(ctrl.SetupSignalHandler())
 		Expect(err).ToNot(HaveOccurred())
 	}()
-
 })
 
 var _ = AfterSuite(func(ctx SpecContext) {
 	By("tearing down the test environment")
 	err := testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
-	DeleteNamespaceCallbackWithContext(ctx, &appNamespace)
 }, NodeTimeout(3*time.Minute))
 
 func TestMain(m *testing.M) {
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -60,6 +60,7 @@ import (
 var cfg *rest.Config
 var k8sClient client.Client
 var testEnv *envtest.Environment
+var appNamespace string
 
 func testClient() client.Client { return k8sClient }
 
@@ -69,7 +70,7 @@ func TestAPIs(t *testing.T) {
 	RunSpecs(t, "Controller Suite")
 }
 
-var _ = BeforeSuite(func() {
+var _ = BeforeSuite(func(ctx SpecContext) {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
 	By("bootstrapping test environment")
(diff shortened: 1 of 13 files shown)


controllers/suite_test.go
```diff
@@ -251,6 +253,11 @@ var _ = BeforeSuite(func() {
 
 	Expect(err).NotTo(HaveOccurred())
 
+	CreateNamespaceWithContext(ctx, &appNamespace)
+
+	err = ApplyResources(filepath.Join("..", "examples", "toystore", "toystore.yaml"), k8sClient, appNamespace)
```
guicassolato, line 258: Today, we probably don't even need to deploy the app. I reckon our tests won't issue requests to the service. In fact, I believe the HTTPRoutes we create in the tests don't even include backendRefs.

In case we _do_ want to keep deploying the app and eventually start linking the test routes to it, by moving the app to its own namespace, we need to make sure routes created in the test namespaces include backendRefs pointing to the service created here in the `appNamespace` namespace.

controllers/suite_test.go
```diff
@@ -251,19 +253,23 @@ var _ = BeforeSuite(func() {
...
 
+	CreateNamespaceWithContext(ctx, &appNamespace)
+
 	go func() {
 		defer GinkgoRecover()
 		err = mgr.Start(ctrl.SetupSignalHandler())
 		Expect(err).ToNot(HaveOccurred())
 	}()
 
+	ApplyKuadrantCR(appNamespace)
```
guicassolato, line 264: Won't having this here also affect the tests that need to create and delete Kuadrant CRs, such as the ones for `KuadrantReconciler` (if we had any 😕)

I liked the approach of having this in the `BeforeAll` and then deleting the CR in `AfterAll` funcs added to the DNS controller tests. WDYT?

KevFan: Yeah, we can do this is the `BeforeAll` /`AfterAll` funcs instead. Only thing with this is we'd need to have the tests in an `Ordered` manner - https://onsi.github.io/ginkgo/#setup-in-ordered-containers-beforeall-and-afterall so we'll lose running the tests in a random order, which may not be ideal either on making sure there's no dependency / leak between tests 🤔 If we are good with this, I can do this in the `BeforeAll/AfterAll` instead

guicassolato: My 2¢ – I guess "ordered" is not as much of an issue as "sequential" is, and we already have sequential tests. So I suppose _random_ is adding just a marginal unit of chance for writing better tests. Not being able to run tests _concurrently_ is our biggest constraint IMO.

KevFan: Yeah, that's true, I'll push this change to use  do this in `BeforeAll/AfterAll` instead

Boomatang: Do you have to do the `BeforeAll` with the `AfterAll```? What I am thinking is if there is a `BeforeEach` that checks if kuadrant is installed and if it is missing install it. Then after the all specs have ran you run the `AfterAll`. Not sure if this would get away from needing the ordered flag but in theory it should still allow us to run tests concurrently. "Maybe"

KevFan: No, I don't think you have to the `BeforeAll/AfterAll` but if you use any one of them, them it needs to marked as Ordered. Using `BeforeEach`doesn't sound like a nice solution to me 🤔 I guess the dificulty is that most tests require a kuadrant instance, some doesn't matter, some does not want an install, and some ideally wants to set up the kuadrant instance in their own testing scenarios 🤔 This is an intermediate solution anyway to just speed things up a good margin still while we look into getting the tests to run concurrently in the future

KevFan:
image

Ran integration tests with the latest changes on CI 6 times already and all passed which suggests the changes might be stable now.

Average runtime in now 34mins with these changes

KevFan:
Last 10 integration test runs has all passed so the changes here are stable at least

image

guicassolato (approved):
🏅 Gold! Awesome job, @KevFan! Thank you!

PR #574, opened 2024-04-24
docs: limitation on multiple network resources with identical hostnames

### LINKED ISSUE #547: Document limitations on multiple routes with identical host name
Due to the importance of host names in the data plane for selecting the policies to enforce at the requests, Kuadrant only provides support for targeting HTTPRoutes that do not declare nor inherit host names identically declared or inherited by other routes. This includes routes that are directly targeted with a policy as well as routes that inherit policies from the gateway, which must also not declare host names that identically conflict with other routes.

To be able to support routes with identical host names (declared or inherited from the gateway), the policy lookup would have to include as well other attributes of the HTTP request (path, headers, etc), allowing for a perfect mapping between _route honored by the gateway_ and _policy set targeting the route_ (directly or indirectly.) Arguably, the lookup key might be needed to expand even beyond HTTP attributes.

Keeping it at the abstract level, the key used to index and lookup for the _set of policy rules to enforce_ should deterministically map 1:1 to the _subset of the network traffic to protect_ by those rules. Such subset of the network would not be the same as any other subset desired to keep unprotected or protected under different policy rules.

Often, the subset of the network is defined in the form of a HTTPRoute object, possibly narrowed to a subset of its host names and internal HTTP matching rules. In general, with more kinds of network resources supported in the targetRef of Kuadrant policies in the future, anything that delimits a subset of the traffic, such as a service name, a pod selector, a namespace, etc, would qualify as a unique identifier to index and to trigger policy rules in the data plane.

While a more robust deterministic mapping between any generically defined subset of the network and its corresponding policies to enforce is not implemented, routes with identical host names cannot be properly supported.

The problem of conflicting host names affects especially the enforcement of AuthPolicy and RateLimitPolicy resources, not having been reported for the DNSPolicy and TLSPolicy so far. To exemplify, consider the following topology:

```
            gateway <--- authpolicy-2
            (*.io)     ratelimitpolicy-2
           /      \
     route-a      route-b
     (app.io)     (app.io)
      /foo          /bar
        ^
        |
  authpolicy-1
ratelimitpolicy-1
```

On requests to `app.io/foo`, both `authpolicy-1` and `ratelimitpolicy-1` are enforced as expected. Whereas on requests to `app.io/bar`, unexpectedly, neither `authpolicy-2` nor `ratelimitpolicy-2` will be enforced.

This is because, from both the auth service's point of view and the wasm-shim's point of view, `app.io` is linked to `authpolicy-1` and to `ratelimitpolicy-1` respectively. `/foo` works only as a secondary condition in those decision points, _after_ `authpolicy-1` and `ratelimitpolicy-1` have already been selected. By not falling back to `authpolicy-2`/`ratelimitpolicy-2` (linked to `*.io`) on requests to `/bar`, no policy is enforced and the request is deemed authorized/unlimited.

This problem would not undergo with more control over the request that is sent to the protection services. Therefore, it is, to an extend, an easier problem to solve for rate limiting today than it is for auth, due to greater control over the request that is sent to Limitador (performed by the [wasm-shim](https://github.com/Kuadrant/wasm-shim)), compared to no control at all over the request to Authorino (limited by the implementation of Istio [AuthorizationPolicy](https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy)).

While Kuadrant does not have full control over the routing that triggers the request to ext-authz and no control at all over the payload that is sent to Authorino, the indexation and triggering of auth policies is based only on the host names. Other HTTP attributes (path, headers, verb, etc) can only determine whether a particular policy already selected by host name will be enforced or ignored. This second level of decision is implemented in the form of ["top-level conditions"](https://docs.kuadrant.io/authorino/docs/features/#common-feature-conditions-when). Nevertheless, a police, once selected, if ignored, does not cause other secondary policies to be searched and triggered alternatively.

Consistent behavior is implemented in the wasm-shim, which also uses the host as lookup key to find a configured RateLimitPolicy. Then, the other conditions are employed at a second layer, to decide whether to call Limitador. But similarly to what happens in Authorino, if the next conditions do not match, no alternate RateLimitPolicy is searched.

Multiple approaches are currently under consideration to overcome this limitation of Kuadrant, with some known caveats and different levels of effort and complexity to implement. These include:
1. Modifications in the lookup system of the auth service to enable falling back to more generic policies in case of top-level conditions mismatch and analogous work in the wasm-shim for rate limiting → can only cover for a portion of the known use cases, redundant with existing features of the auth service, 2 sprints of work (Authorino and wasm-shim) to do the exact same thing
2. Modifications in the wasm-shim to support indexing beyond host names and replacement of Istio AuthorizationPolicy API for triggering external authorization by own implementation in the wasm-shim as well, thus making possible to leverage the auth service's [already existing alternative to host names](https://docs.kuadrant.io/authorino/docs/architecture/#host-lookup) → most viable option in the long-term, though with increased dependency on wasm for integration; still flawed against non-deterministic decision at the gateway regarding which route to honor if not based on HTTP attributes
3. Modifications in the wasm-shim to support indexing beyond host names and work together with Istio to add support for [Envoy per-route external authorization](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto#extensions-filters-http-ext-authz-v3-extauthzperroute) configs in the Istio AuthorizationPolicy API → hard to implement in Istio; likely to face the same challenge later on with other Gateway API implementation (e.g. Envoy Gateway); could solve for auth but not completely for rate limiting
4. Merging of policies into single bigger internal configs, whose original rules are wrapped under more specific conditions → risk of conflicting rules and conditions, may require changes in the APIs (unique names for policy rules across policies)

guicassolato:
Closes #547

### Verification steps

Albeit this is a docs PR, if you want to try the limitation it describes, use the following steps.

#### Setup

```sh
make local-setup
```

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

```sh
kind: Gateway
metadata:
  name: istio-ingressgateway
spec:
  gatewayClassName: istio
  listeners:
  - name: http
    allowedRoutes:
      namespaces:
        from: All
    port: 80
    hostname: "*.io"
    protocol: HTTP
EOF
```

```sh

kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: route-a
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - app.io
  rules:
  - matches:
    - path:
        value: /foo
    backendRefs:
    - name: toystore
      port: 80
---
kind: HTTPRoute
metadata:
  name: route-b
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - app.io
  rules:
  - matches:
    - path:
        value: /bar
    backendRefs:
    - name: toystore
      port: 80
EOF
```

```sh
kubectl port-forward -n istio-system service/istio-ingressgateway-istio 9080:80 2>&1 >/dev/null &
```

#### Scenario 1: 2 route policies

AuthPolicy:

```sh
kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: policy-1
spec:
  targetRef:
    kind: HTTPRoute
    name: route-a
  rules:
    response:
      success:
        headers:
          "auth-policy":
            plain:
              value: policy-1
EOF
```

```sh
kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: policy-2
spec:
  targetRef:
    kind: HTTPRoute
    name: route-b
  rules:
    response:
      success:
        headers:
          "auth-policy":
            plain:
              value: policy-2
EOF
```

```sh
kubectl get authpolicies -o jsonpath='{.items[*].status.conditions}' | jq
```

```sh
curl --silent -H 'Host: app.io' http://localhost:9080/foo | jq -r .headers.HTTP_AUTH_POLICY
# policy-1
```

```sh
curl --silent -H 'Host: app.io' http://localhost:9080/bar | jq -r .headers.HTTP_AUTH_POLICY
# null
```

RateLimitPolicy:

```sh
kubectl apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: policy-1
spec:
  targetRef:
    kind: HTTPRoute
    name: route-a
  limits:
    limit-1:
      rates:
      - limit: 1
        duration: 5
        unit: second
EOF
```

```sh
kubectl apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: policy-2
spec:
  targetRef:
    kind: HTTPRoute
    name: route-b
  limits:
    limit-2:
      rates:
      - limit: 2
        duration: 5
        unit: second
EOF
```

```sh
kubectl get ratelimitpolicies -o jsonpath='{.items[*].status.conditions}' | jq
```

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: app.io' http://localhost:9080/foo | grep -E --color "\b(429)\b|$"; sleep 1; done
# 200
# 200
# 200
# 200
```

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: app.io' http://localhost:9080/bar | grep -E --color "\b(429)\b|$"; sleep 1; done
# 200
# 200
# 429
# 429
```

#### Scenario 2: 1 route policy and 1 gateway policy

(Clean up all policies created in Scenario 1)

AuthPolicy:

```sh
kubectl apply -f -<<EOF
kind: AuthPolicy
metadata:
  name: policy-1
spec:
  targetRef:
    kind: HTTPRoute
    name: route-a
  rules:
    response:
      success:
        headers:
          "auth-policy":
            plain:
              value: policy-1
EOF
```

```sh
kubectl apply -n istio-system -f -<<EOF
kind: AuthPolicy
metadata:
  name: policy-2
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    response:
      success:
        headers:
          "auth-policy":
            plain:
              value: policy-2
EOF
```

```sh
kubectl get authpolicies --all-namespaces -o jsonpath='{.items[*].status.conditions}' | jq
```

```sh
curl --silent -H 'Host: app.io' http://localhost:9080/foo | jq -r .headers.HTTP_AUTH_POLICY
# policy-1
```

```sh
curl --silent -H 'Host: app.io' http://localhost:9080/bar | jq -r .headers.HTTP_AUTH_POLICY
# null
```

RateLimitPolicy:

```sh
kubectl apply -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: policy-1
spec:
  targetRef:
    kind: HTTPRoute
    name: route-a
  limits:
    limit-1:
      rates:
      - limit: 1
        duration: 5
        unit: second
EOF
```

```sh
kubectl apply -n istio-system -f -<<EOF
kind: RateLimitPolicy
metadata:
  name: policy-2
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  limits:
    limit-2:
      rates:
      - limit: 2
        duration: 5
        unit: second
EOF
```

```sh
kubectl get ratelimitpolicies --all-namespaces -o jsonpath='{.items[*].status.conditions}' | jq
```

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: app.io' http://localhost:9080/foo | grep -E --color "\b(429)\b|$"; sleep 1; done
# 200
# 429
# 429
# 429
```

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: app.io' http://localhost:9080/bar | grep -E --color "\b(429)\b|$"; sleep 1; done
# 200
# 200
# 200
# 200
```

### DIFF
--- a/doc/auth.md
+++ b/doc/auth.md
@@ -224,14 +224,6 @@ spec:
 └───────────────────┘             └────────────────────┘
 ```
 
-#### Multiple HTTPRoutes with the same hostname
-
-Kuadrant currently does not support concurrent AuthPolicies targeting the exact same hostname.
-
-In case 2 or more AuthPolicies target resources that specify the same exact hostnames, the first AuthPolicy created claims the hostname; all the other AuthPolicies will not be enforced for the conflicting hostname and their status will be reported as not ready.
-
-This limitation only applies to identical hostnames. Different AuthPolicies can still be declared for different hostnames, as well as sets and strict subsets of hostnames.
-
 #### Hostnames and wildcards
 
 If an AuthPolicy targets a route defined for `*.com` and another AuthPolicy targets another route for `api.com`, the Kuadrant control plane will not merge these two AuthPolicies. Rather, it will mimic the behavior of gateway implementation by which the "most specific hostname wins", thus enforcing only the corresponding applicable policies and auth rules.
--- a/doc/auth.md
+++ b/doc/auth.md
@@ -357,7 +357,87 @@ Check out the following user guides for examples of protecting services with Kua
 * One HTTPRoute can only be targeted by one AuthPolicy.
 * One Gateway can only be targeted by one AuthPolicy.
 * AuthPolicies can only target HTTPRoutes/Gateways defined within the same namespace of the AuthPolicy.
-* 2+ AuthPolicies cannot target network resources that define/inherit the same exact hotname.
+* 2+ AuthPolicies cannot target network resources that define/inherit the same exact hostname.
+
+#### Limitation: Multiple network resources with identical hostnames
+
+Kuadrant currently does not support multiple AuthPolicies simultaneously targeting network resources that declare identical hostnames. This includes multiple HTTPRoutes that specify the same hostnames in the `spec.hostnames` field, as well as HTTPRoutes that specify a hostname that is identical to a hostname specified in a listener of one of the route's parent gateways or HTTPRoutes that don't specify any hostname at all thus inheriting the hostnames from the parent gateways. In any of these cases, **a maximum of one AuthPolicy targeting any of those resources that specify identical hostnames is allowed**.
+
+Moreover, having **multiple resources that declare identical hostnames** may lead to unexpected behavior and therefore **should be avoided**.
+
+This limitation is rooted at the underlying components configured by Kuadrant for the implementation of its polcies and the lack of information in the data plane regarding the exact route that honored by the API gateway in cases of conflicting hostnames.
+
+To exemplify one way this limitation can impact deployments, consider the following topology:
+
+```
+                 ┌──────────────┐
+                 │   Gateway    │
+                 ├──────────────┤
+          ┌─────►│ listeners:   │◄──────┐
+          │      │ - host: *.io │       │
+          │      └──────────────┘       │
+          │                             │
+          │                             │
+┌─────────┴─────────┐        ┌──────────┴────────┐
+│     HTTPRoute     │        │     HTTPRoute     │
+│     (route-a)     │        │     (route-b)     │
+├───────────────────┤        ├───────────────────┤
+│ hostnames:        │        │ hostnames:        │
+│ - app.io          │        │ - app.io          │
+│ rules:            │        │ rules:            │
+│ - matches:        │        │ - matches:        │
+│   - path:         │        │   - path:         │
+│       value: /foo │        │       value: /bar │
+└───────────────────┘        └───────────────────┘
+          ▲                            ▲
+          │                            │
+    ┌─────┴──────┐               ┌─────┴──────┐
+    │ AuthPolicy │               │ AuthPolicy │
+    │ (policy-1) │               │ (policy-2) │
+    └────────────┘               └────────────┘
+```
+
+In the example above, with the `policy-1` resource created before `policy-2`, `policy-1` will be enforced on all requests to `app.io/foo` while `policy-2` will be rejected. I.e. `app.io/bar` will not be secured. In fact, the status conditions of `policy-2` shall recflect `Enforced=false` with message _"AuthPolicy has encountered some issues: AuthScheme is not ready yet"_.
+
+Notice the enforcement of `policy-1` and no enforcement of `policy-2` is the opposite behavior as the [analogous problem with the Kuadrant RateLimitPolicy](rate-limiting.md#limitation-multiple-network-resources-with-identical-hostnames).
+
+A slightly different way the limitation applies is when two or more routes of a gateway declare the exact same hostname and a gateway policy is defined with expectation to set default rules for the cases not covered by more specific policies. E.g.:
+
+```
+                                    ┌────────────┐
+                         ┌──────────┤ AuthPolicy │
+                         │          │ (policy-2) │
+                         ▼          └────────────┘
+                 ┌──────────────┐
+                 │   Gateway    │
+                 ├──────────────┤
+          ┌─────►│ listeners:   │◄──────┐
+          │      │ - host: *.io │       │
+          │      └──────────────┘       │
+          │                             │
+          │                             │
+┌─────────┴─────────┐        ┌──────────┴────────┐
+│     HTTPRoute     │        │     HTTPRoute     │
+│     (route-a)     │        │     (route-b)     │
+├───────────────────┤        ├───────────────────┤
+│ hostnames:        │        │ hostnames:        │
+│ - app.io          │        │ - app.io          │
+│ rules:            │        │ rules:            │
+│ - matches:        │        │ - matches:        │
+│   - path:         │        │   - path:         │
+│       value: /foo │        │       value: /bar │
+└───────────────────┘        └───────────────────┘
+          ▲
+          │
+    ┌─────┴──────┐
+    │ AuthPolicy │
+    │ (policy-1) │
+    └────────────┘
+```
+
+Once again, requests to `app.io/foo` will be protected under AuthPolicy `policy-1`, while requests to `app.io/bar` will **not** be protected under any policy at all, unlike expected gateway policy `policy-2` enforced as default. Both policies will report status condition as `Enforced` nonetheless.
+
+To avoid these problems, use different hostnames in each route.
 
 ## Implementation details
 
--- a/doc/auth.md
+++ b/doc/auth.md
@@ -365,7 +365,7 @@ Kuadrant currently does not support multiple AuthPolicies simultaneously targeti
 
 Moreover, having **multiple resources that declare identical hostnames** may lead to unexpected behavior and therefore **should be avoided**.
 
-This limitation is rooted at the underlying components configured by Kuadrant for the implementation of its polcies and the lack of information in the data plane regarding the exact route that honored by the API gateway in cases of conflicting hostnames.
+This limitation is rooted at the underlying components configured by Kuadrant for the implementation of its policies and the lack of information in the data plane regarding the exact route that is honored by the API gateway at each specific request, in cases of conflicting hostnames.
 
 To exemplify one way this limitation can impact deployments, consider the following topology:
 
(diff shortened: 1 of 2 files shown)


doc/auth.md
```diff
@@ -300,7 +300,87 @@ Check out the following user guides for examples of protecting services with Kua
...
+* 2+ AuthPolicies cannot target network resources that define/inherit the same exact hostname.
+
+#### Limitation: Multiple network resources with identical hostnames
+
+Kuadrant currently does not support multiple AuthPolicies simultaneously targeting network resources that declare identical hostnames. This includes multiple HTTPRoutes that specify the same hostnames in the `spec.hostnames` field, as well as HTTPRoutes that specify a hostname that is identical to a hostname specified in a listener of one of the route's parent gateways or HTTPRoutes that don't specify any hostname at all thus inheriting the hostnames from the parent gateways. In any of these cases, **a maximum of one AuthPolicy targeting any of those resources that specify identical hostnames is allowed**.
+
+Moreover, having **multiple resources that declare identical hostnames** may lead to unexpected behavior and therefore **should be avoided**.
+
+This limitation is rooted at the underlying components configured by Kuadrant for the implementation of its polcies and the lack of information in the data plane regarding the exact route that honored by the API gateway in cases of conflicting hostnames.
```
trepel, line 311: typo: polcies -> policies
I am not 100% sure about that "...exact route that honored by...". Should not it be "...that is honored by..."

doc/auth.md
```diff
@@ -300,7 +300,87 @@ Check out the following user guides for examples of protecting services with Kua
...
+└───────────────────┘        └───────────────────┘
+          ▲                            ▲
+          │                            │
+    ┌─────┴──────┐               ┌─────┴──────┐
+    │ AuthPolicy │               │ AuthPolicy │
+    │ (policy-1) │               │ (policy-2) │
+    └────────────┘               └────────────┘
+```
+
+In the example above, with the `policy-1` resource created before `policy-2`, `policy-1` will be enforced on all requests to `app.io/foo` while `policy-2` will be rejected. I.e. `app.io/bar` will not be secured. In fact, the status conditions of `policy-2` shall recflect `Enforced=false` with message _"AuthPolicy has encountered some issues: AuthScheme is not ready yet"_.
```
trepel, line 343: typo: recflect -> reflect

doc/rate-limiting.md
```diff
@@ -253,6 +253,87 @@ Check out the following user guides for examples of rate limiting services with
...
+
+#### Limitation: Multiple network resources with identical hostnames
+
+Kuadrant currently does not support multiple RateLimitPolicies simultaneously targeting network resources that declare identical hostnames. This includes multiple HTTPRoutes that specify the same hostnames in the `spec.hostnames` field, as well as HTTPRoutes that specify a hostname that is identical to a hostname specified in a listener of one of the route's parent gateways or HTTPRoutes that don't specify any hostname at all thus inheriting the hostnames from the parent gateways. In any of these cases, **a maximum of one RateLimitPolicy targeting any of those resources that specify identical hostnames is allowed**.
+
+Moreover, having **multiple resources that declare identical hostnames** may lead to unexpected behavior and therefore **should be avoided**.
+
+This limitation is rooted at the underlying components configured by Kuadrant for the implementation of its polcies and the lack of information in the data plane regarding the exact route that honored by the API gateway in cases of conflicting hostnames.
```
trepel, line 264: same as above: polcies typo and "that honored by" question which I am not sure about

doc/rate-limiting.md
```diff
@@ -253,6 +253,87 @@ Check out the following user guides for examples of rate limiting services with
...
+└───────────────────┘        └───────────────────┘
+          ▲                            ▲
+          │                            │
+ ┌────────┴────────┐           ┌───────┴─────────┐
+ │ RateLimitPolicy │           │ RateLimitPolicy │
+ │   (policy-1)    │           │   (policy-2)    │
+ └─────────────────┘           └─────────────────┘
+```
+
+In the example above, with the `policy-1` resource created before `policy-2`, `policy-2` will be enforced on all requests to `app.io/bar` while `policy-1` will be **not** be enforced at all. I.e. `app.io/foo` will not be rate-limited. Nevertheless, both policies will report status condition as `Enforced`.
```
trepel, line 296: "...will be not be enforced..." -> redundant "be"?

doc/rate-limiting.md
```diff
@@ -253,6 +253,87 @@ Check out the following user guides for examples of rate limiting services with
...
+└───────────────────┘        └───────────────────┘
+          ▲
+          │
+ ┌────────┴────────┐
+ │ RateLimitPolicy │
+ │   (policy-1)    │
+ └─────────────────┘
+```
+
+Once again, both policies will report status condition as `Enforced`. However, in this case, only `policy-1` will be enforced on requests to `app.io/foo`, while `policy-2` will be **not** be enforced at all. I.e. `app.io/bar` will not be not rate-limited. This is same behavior as the [analogous problem with the Kuadrant AuthPolicy](auth.md#limitation-multiple-network-resources-with-identical-hostnames).
```
trepel, line 334: "...will be not be enforced..." -> redundant "be"?

trepel (commented):
LGTM, just a few small typos.

I have noticed that both documents already contain a short stanza about httproutes with identical hostnames. You might consider removing them since the text that is added as part of this PR is much better.

https://github.com/Kuadrant/kuadrant-operator/blob/2d641c3276d23a0a66e1a26c8e6798aacd5dd8ce/doc/auth.md#multiple-httproutes-with-the-same-hostname

https://github.com/Kuadrant/kuadrant-operator/blob/2d641c3276d23a0a66e1a26c8e6798aacd5dd8ce/doc/rate-limiting.md#multiple-httproutes-with-the-same-hostname

KevFan (approved):
Verified limitation 👍 

Looks good to me, other than the typos / suggestions found by @trepel

PR #577, opened 2024-04-25
fix: auth policy test flake

KevFan:


guicassolato (approved):
This may not fix the flaky tests overall, but it's definitely correct.

PR #581, opened 2024-04-25
feat: test speed up

KevFan:
Integration tests are failing by timeout at 1hour mark due to this is the default suite timeout for ginkgo

Introducing some changes to speed up general running of the tests to decrease test run time rather than increasing suite timeout:
* [tests: remove applying toystore](https://github.com/Kuadrant/kuadrant-operator/commit/a62b6828bab0c1ccdfdaf6c76e574937c6a3a837)
* [tests: kuadrant install in BeforeAll/AfterAll](https://github.com/Kuadrant/kuadrant-operator/commit/c00e889f944955c1769d35b9aa232c65769418b3)

### DIFF
--- a/controllers/authpolicy_controller_test.go
+++ b/controllers/authpolicy_controller_test.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


mikenairn:
Anything that speeds the tests up and makes them less flaky sounds good to me I will let others more familiar with the Auth and RLP tests that are changing approve however.

I have a test related PR [here](https://github.com/Kuadrant/kuadrant-operator/pull/583) to make it possible to run the DNS and TLS policy tests in parallel. This makes a significant difference in how long the test suite takes. We should probably try and get to a point where the vast majority, if not all, the test suite can be run in parallel and enable it on the ci job. Using `Ordered` prevents those tests from being run in parallel, but if it makes it quicker in the mean time then lets go with it.

KevFan:
Superseded by https://github.com/Kuadrant/kuadrant-operator/pull/565  as its seem to be running stably now which contains these changes

PR #584, opened 2024-04-26
Increase metallb useable ips

mikenairn:
closes #541 

The number of IPs available to metallb is currently really small (4) which means you can only create a small number of gateways (3, istio takes one) before it stops dishing out new IPs.

These changes increase the number of ips to 16 and assumes the network will always give us a range allowing the use of all 256 ips, 0.0.0.[0-255] (This appears to be the case for kind networks created with podman(10.89.0.0/24) and docker(172.18.0.0/16)).

Also adds an option to specify an offest (default is 0) when creating a kuadrant cluster with local-setup allowing multiple instances to be created using a different range of IPs.

```
make local-setup KIND_CLUSTER_NAME=kuadrant-local-1 SUBNET_OFFSET=0
make local-setup KIND_CLUSTER_NAME=kuadrant-local-2 SUBNET_OFFSET=1
```

| subnet | offset | address | ip range |
| --- | --- | --- | --- |
| 10.89.0.0/24 | 0 | 10.89.0.0/28 |  10.89.0.0 - 10.89.0.15 | 
| 10.89.0.0/24 | 1 | 10.89.0.16/28 |  10.89.0.16 - 10.89.0.31 | 
| 10.89.0.0/24 | 2 | 10.89.0.32/28 |  10.89.0.32 - 10.89.0.47 | 
etc...

Note: Needs someone with a Mac to test it still works

Issue #495, opened 2024-03-20
Update Quick Start and Associated docs


maleck13:
**What**

As part of implementing distributed DNS, we will also need to update the quick start and follow on walkthrough for multi-cluster to use this setup.

- [x] Update multi-cluster quickstart and scripts https://docs.kuadrant.io/getting-started-multi-cluster/
- [x] Add a new multi-cluster walkthrough similar to https://docs.kuadrant.io/multicluster-gateway-controller/docs/how-to/multicluster-gateways-walkthrough/

maleck13:
@david-martin I am going to assign you and I to this one

maleck13:
I am looking at the walkthrough doc and I believe you are on the quickstart?

PR #590, opened 2024-04-29
Add multi cluster support to quickstart-setup script

david-martin:
Part of #495 
Depends on #584

The changes in the PR are intended to give the same behaviour for the single cluster setup as documented in https://docs.kuadrant.io/getting-started-single-cluster/
The changes also allow the script to be used subsequently to add more clusters (as workers) to your local environment.

Changes:

* Remove the cluster deletion on subsequent re-runs of the script. Any cluster deletion has to be done by the user with the `kind` cli themselves.
* If a cluster named 'kuadrant-local' is already running locally, this is seen as a hub cluster (as well as a worker). The script then asks if the user wants to do a multi cluster setup.
* When in multi cluster setup mode, all new clusters have a suffix of a number, and are worker only. For example, `kuadrant-local-1`, `kuadrant-local-2` etc...

TODO:

* [x] test that a distributed gateway example works in a multi cluster setup.
* [x] how to use docker run for yq cmds (see TODO in utils/docker-network-ipaddresspool.sh)

### DIFF
--- a/doc/user-guides/secure-protect-connect.md
+++ b/doc/user-guides/secure-protect-connect.md
@@ -2,7 +2,7 @@
 
 ## Pre-requisites
 
-- Completed the [Single-cluster Quick Start](https://docs.kuadrant.io/getting-started-single-cluster/)
+- Completed the [Single-cluster Quick Start](https://docs.kuadrant.io/getting-started-single-cluster/) or [Multi-cluster Quick Start](https://docs.kuadrant.io/getting-started-multi-cluster/)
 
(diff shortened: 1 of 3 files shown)


jasonmadigan:
👀

jasonmadigan:
Gave this a spin yesterday, and it works well. Verified multiple clusters seem to be setup OK, and also seeing the dns-operator add each GW IP to records.

jasonmadigan (approved):

PR #593, opened 2024-04-30
kuadrant wasm server component

eguzki:
Requires https://github.com/Kuadrant/wasm-shim/pull/52

### What

Kuadrant's Wasm Server is an HTTP-based static file server that serves
Kuadrant's Wasm module binaries. The wasm module integrates with the gateway in the data plane via
the [Wasm Network filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/network_filters/wasm_filter).
The source code of the compiled Wasm binaries is hosted at
[Kuadrant's Wasm-Shim project](https://github.com/Kuadrant/wasm-shim).

Currently, at runtime, the istio control plane downloads an [oci wasm image](https://istio.io/latest/docs/reference/config/proxy_extensions/wasm-plugin/). Usually from cluster external image repo like quay.io. This clearly opens a risky door to inject malicious code. 

### Why

Envoy dynamically loads the Wasm module during runtime, and it is used to extend Envoy's
capabilities. Kuadrant cannot inject a local .wasm file that is accessible by the Envoy proxy.
Instead, Kuadrant tells Envoy to download the Wasm module from a source using existing
Envoy control plane APIs.
The main goal of Kuadrant's Wasm Server component is to be the source of the Wasm module running
inside the same Kubernetes cluster. This way, Envoy does not need to fetch the Wasm module from
external sources, which adds security protection against malicious code injection.

This architecture enables so-called *offline* or *disconnected* installs,
which allow having the entire cluster disconnected from the internet,
at least regarding the Wasm module.
> Disconnected install is itself a full feature and we did not tested that yet. 

### How

The Wasm Server is implemented using [nginx](https://nginx.org) HTTP server.
The compiled [Kuadrant's Wasm module](https://github.com/Kuadrant/wasm-shim) is being copied
to a docker image containing the nginx server configured to serve it on the `/plugin.wasm` path.

#### Istio

The following sequence diagram shows the workflow when Envoy is managed by [Istio](https://istio.io/)

```mermaid
sequenceDiagram
    autonumber
    box transparent Kubernetes cluster
    participant K as Kuadrant
    participant W as Wasm Server
    participant WP as WasmPlugin
    participant I as Istio
    participant E as Envoy
    end
    K->>+W: sha256?
    W->>-K: sha256
    K->>WP: Wasm Server internal URL, sha256
    WP->>I: Wasm Server internal URL, sha256
    I->>W: Fetch Wasm binary, verify sha256 checksum
    I->>E: Push Wasm binary
    I->>E: Setup Wasm filter
```

## Verification Steps

* Setup the environment:   
                         
```sh                    
make local-env-setup   
```        

* run the operator manually with custom wasm-server image. (It will allow verifying the upgrading of the Wasm image)

```sh
RELATED_IMAGE_WASMSHIM=quay.io/kuadrant/wasm-server:wasm-server make run
```

* Request an instance of Kuadrant:

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```
* Deploy toystore

```sh
```

Create a HTTPRoute to route traffic to the service via Istio Ingress Gateway:

```sh
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
    - method: POST
      path:
        type: Exact
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

```sh
```

Verify the route works:

```sh
curl -H 'Host: api.toystore.com' http://$GATEWAY_URL/toys -i
# HTTP/1.1 200 OK
```
* Enforce rate limiting on requests to the Toy Store API
```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    "create-toy":
      rates:
      - limit: 5
        duration: 10
        unit: second
      routeSelectors:
      - matches: # selects the 2nd HTTPRouteRule of the targeted route
        - method: POST
          path:
            type: Exact
            value: "/toys"
EOF
```

* Check wasm plugin has been created and it contains the url of local service of the Wasm Server
```sh
kubectl get wasmplugin kuadrant-istio-ingressgateway -n istio-system -o jsonpath="{.spec.url}"
```
It should return 
```
http://wasm-server-kuadrant.kuadrant-system.svc.cluster.local/plugin.wasm
```

* Check wasm plugin has been created and it contains the *sha256* checksum of the Wasm binary served by the Wasm Server
```sh
kubectl get wasmplugin kuadrant-istio-ingressgateway -n istio-system -o jsonpath="{.spec.sha256}"
```
It should return some sha256 value (may not be the same)
```
d5bd8b7a21e2369e130a9a2bd80345df30102b08d4fe070c1896a75769e7e2c0
```
* Check Wasm Server configmap is created and it contains the *sha256* checksum of the Wasm binary served by the Wasm Server
```sh
kubectl get configmap wasm-server-kuadrant -n kuadrant-system -o jsonpath="{.data.rate-limit-wasm-sha256}"
```
It should return some sha256 value (may not be the same)
```sh
d5bd8b7a21e2369e130a9a2bd80345df30102b08d4fe070c1896a75769e7e2c0
```
* Run requests (5 out of 10 allowed)
```sh
```

* Let's verify upgrade of the Wasm Server version
There is no need to stop http client... it should not be affected and rate limiting should work all the time

Cancel operator process (CTRL+C) and run new version of the Wasm Server
```sh
RELATED_IMAGE_WASMSHIM=quay.io/kuadrant/wasm-server:wasm-server-v2 make run
```

* Check wasm plugin has been updated and it contains the **new** *sha256* checksum of the Wasm binary served by the Wasm Server
```sh
kubectl get wasmplugin kuadrant-istio-ingressgateway -n istio-system -o jsonpath="{.spec.sha256}"
```
It should return some sha256 value (may not be the same)
```
f835ebeaaa49cbba135cc4175644af343feb3a06ec46a91e69281f09048bad3f
```
* Check Wasm Server configmap has been updated and it contains the **new** *sha256* checksum of the Wasm binary served by the Wasm Server
```sh
kubectl get configmap wasm-server-kuadrant -n kuadrant-system -o jsonpath="{.data.rate-limit-wasm-sha256}"
```
It should return some sha256 value (may not be the same)
```sh
f835ebeaaa49cbba135cc4175644af343feb3a06ec46a91e69281f09048bad3f
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,9 +159,9 @@ WASM_SHIM_VERSION ?= latest
 shim_version_is_semantic := $(call is_semantic_version,$(WASM_SHIM_VERSION))
 
 ifeq (true,$(shim_version_is_semantic))
-RELATED_IMAGE_WASMSHIM ?= oci://quay.io/kuadrant/wasm-shim:v$(WASM_SHIM_VERSION)
+RELATED_IMAGE_WASMSHIM ?= quay.io/kuadrant/wasm-server:v$(WASM_SHIM_VERSION)
 else
-RELATED_IMAGE_WASMSHIM ?= oci://quay.io/kuadrant/wasm-shim:$(WASM_SHIM_VERSION)
+RELATED_IMAGE_WASMSHIM ?= quay.io/kuadrant/wasm-server:$(WASM_SHIM_VERSION)
 endif
 
 all: build
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ to operate the cluster (Istio's) ingress gateway to provide API management with 
 
 ### Kuadrant components
 
-| CRD                                                                  | Description                                                                                                                                                                                                                                                                               |
+| Component                                                            | Description                                                                                                                                                                                                                                                                               |
 |----------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | Control Plane                                                        | The control plane takes the customer desired configuration (declaratively as kubernetes custom resources) as input and ensures all components are configured to obey customer's desired behavior.<br> This repository contains the source code of the kuadrant control plane              |
 | [Kuadrant Operator](https://github.com/Kuadrant/kuadrant-operator)   | A Kubernetes Operator to manage the lifecycle of the kuadrant deployment                                                                                                                                                                                                                  |
@@ -39,6 +39,7 @@ to operate the cluster (Istio's) ingress gateway to provide API management with 
 | [Authorino Operator](https://github.com/Kuadrant/authorino-operator) | A Kubernetes Operator to manage Authorino instances                                                                                                                                                                                                                                       |
 | [Limitador Operator](https://github.com/Kuadrant/limitador-operator) | A Kubernetes Operator to manage Limitador instances                                                                                                                                                                                                                                       |
 | [DNS Operator](https://github.com/Kuadrant/dns-operator)             | A Kubernetes Operator to manage DNS records in external providers                                                                                                                                                                                                                         |
+| [Wasm Server](doc/wasm-server.md)                                    | Static file server to serve Wasm binaries internally. Lifecycle managed by the Kuadrant Operator.                                                                                                                                                                                         |
 
 ### Provided APIs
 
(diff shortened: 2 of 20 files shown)


pehala:
I like the idea of not pulling `wasm-shim` from external source, but I am not quite sure why we need yet another component that will serve one static file. Would it make sense to serve it instead directly in `kuadrant-operator`?

eguzki:
> e. Would it make sense to serve it instead dire

Good question. Having the kuadrant operator serve the Wasm binary directly instead was the initial idea I tried. Serving it from limitador was yet another initial idea. It turns out that it's not straightforward. The kuadrant controller responsible of the wasmplugin needs to be aware of its own location (service and namespace). It can be done adding env vars updating the operator deployments. However, I consider that having a new deployment (+ service) being a nginx with the Wasm binary embedded is easier to deploy, manage upgrades and monitor. The kuadrant operator deploys the service at the namespace where the kuadrant CR is being created and it knows the service address when configuring the WasmPlugin object. Open to discussion, I guess.

eguzki:
discarded

Issue #586, opened 2024-04-29
Extend the Partially Enforced status on DNSPolicy


makslion:
## What 
If the DNSPolicy is not enforced on all DNS Records we are marking it as "Partially Enforced" (in the status-conditions). We want to know the exact records that are not affected by the policy. 

## How 
Change logic [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_status.go#L108) to go through all the DNSRecords and put their names in the [message](https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/library/kuadrant/apimachinery_status_conditions.go#L111) field for the status. I.E. "DNS Policy has been partially enforced. Not enforced on: [name]"

makslion:
This issue is blocked by a [bug](https://github.com/Kuadrant/dns-operator/issues/165) that prevents correct configuration of integration tests

PR #594, opened 2024-04-30
use correct status type when enforcing policy

makslion:
Now `Enforced` will be false even on partially enforced records. If there are no ready records it will still say  "partially enforced" (which implies it is enforced on some of them which is not true). This will be addressed in #586 
Adding a check to one of the tests to ensure we are not marking policy as `Enforced - True` when the record is not ready. 

Aslo a tinny refactor - moved some vars around for convenience


KevFan (approved):
Changes looks good to me 👍

philbrookes (approved):

guicassolato:
This is incorrect and should be reverted

PR #578, opened 2024-04-25
Update dnspolicy crd validations

mikenairn:
closes #554 

[fix: DNSPolicy CRD required validations](https://github.com/Kuadrant/kuadrant-operator/commit/85e99a8b8e3b86899b4c5f6a5aafb5d81f85fb40)

The validations to ensure required fields were marked as required were incorrect.
Any fields that are not explicitly marked with `+optional` and do not have `omitempty` are considered required.

[Update loadbalanced routingStrategy CRD validation](https://github.com/Kuadrant/kuadrant-operator/commit/fdf78056bff073cfd2cf94e85d91e6a2c3537847)

Update the CRD validations to make the "loadBalancing" field required if the routingStrategy is "loadbalanced" and make the "defaultWeight" and "defaultGeo" fields mandatory within the "loadBalancing" section.

Minimum valid specs are now:

Simple:
```
spec:
  targetRef:
    name: prod-web
    group: gateway.networking.k8s.io
    kind: Gateway
  routingStrategy: simple
```

LoadBalanced:
```
spec:
  targetRef:
    name: prod-web
    group: gateway.networking.k8s.io
    kind: Gateway
  routingStrategy: loadbalanced
  loadBalancing:
    geo:
      defaultGeo: foo
    weighted:
      defaultWeight: 100
```

PR #595, opened 2024-04-30
remove out of date fields from DNS Policy API

philbrookes:


api/v1alpha1/dnspolicy_types.go
```diff
@@ -239,33 +234,17 @@ func (l *DNSPolicyList) GetItems() []kuadrant.Policy {
...
-
-	return nil
+	// +kubebuilder:validation:Required
+	Endpoint string `json:"endpoint"`
+	// +kubebuilder:validation:Required
+	Port *int `json:"port"`
+	// +kubebuilder:validation:Required
+	Protocol *string `json:"protocol"`
+	// +kubebuilder:validation:Required
+	FailureThreshold *int `json:"failureThreshold"`
```
mikenairn, line 244: Should probably avoid using pointers if its required. Also, you don't need these lines `+kubebuilder:validation:Required` see https://github.com/Kuadrant/kuadrant-operator/pull/578

api/v1alpha1/dnspolicy_types.go
```diff
@@ -239,37 +226,10 @@ func (l *DNSPolicyList) GetItems() []kuadrant.Policy {
...
-	}
-
-	if s.Protocol == nil {
-		protocol := "HTTPS"
-		s.Protocol = &protocol
-	}
+	Endpoint         string `json:"endpoint"`
+	Port             int    `json:"port"`
+	Protocol         string `json:"protocol"`
+	FailureThreshold int    `json:"failureThreshold"`
```
mikenairn, line 232: You might want to consider adding some CEL validations for these, presumably there are values that we should just not accept for some of these.

mikenairn: Also, a description of each would be good since it will end up as the description in the generated CRD.

mikenairn (approved):
Looks fine. Thought these were removed long ago tbh. Couple of suggestions, but nothing that need hold up the merge.

api/v1alpha1/dnspolicy_types.go
philbrookes: nice, added.

api/v1alpha1/dnspolicy_types.go
```diff
@@ -239,37 +226,20 @@ func (l *DNSPolicyList) GetItems() []kuadrant.Policy {
...
+	// Port to connect to the host on
+	// +kubebuilder:validation:Minimum:=1
+	Port int `json:"port"`
+	// Protocol to use when connecting to the host, valid values are "HTTP" or "HTTPS"
+	// +kubebuilder:validation:Enum:=HTTP;HTTPS
+	Protocol string `json:"protocol"`
+	// FailureThreshold is a limit of consecutive failures that must occur for a host
+	// to be considered unhealthy
+	// +kubebuilder:validation:Minimum:=1
+	FailureThreshold int `json:"failureThreshold"`
```
mikenairn, line 242: :+1:

mikenairn (approved):

maleck13:
@philbrookes you need to regen the bundle also we will want to ensure docs call out the removal of these fields as this is a breaking change from the previous release 0.6.x

philbrookes:
> @philbrookes you need to regen the bundle also we will want to ensure docs call out the removal of these fields as this is a breaking change from the previous release 0.6.x

yes, I'm working on a docs update already in a separate branch. AFAIK I have regenerated the bundle, is something missing?

PR #597, opened 2024-05-01
dnsrecord condition only propagate owned records status

philbrookes:
All listed records were propagating to all dns policies.

This PR does 2 things:
- rename ProbeConditions to RecordConditions
- only propagates owned DNS Record conditions to the policy.

### DIFF
--- a/api/v1alpha1/dnspolicy_types.go
+++ b/api/v1alpha1/dnspolicy_types.go
@@ -133,7 +133,7 @@ type DNSPolicyStatus struct {
 	HealthCheck *HealthCheckStatus `json:"healthCheck,omitempty"`
 
 	// +optional
-	ProbeConditions map[string][]metav1.Condition `json:"probeConditions,omitempty"`
+	RecordConditions map[string][]metav1.Condition `json:"recordConditions,omitempty"`
 }
 
 func (s *DNSPolicyStatus) GetConditions() []metav1.Condition {
--- a/controllers/dnspolicy_status.go
+++ b/controllers/dnspolicy_status.go
@@ -81,6 +81,8 @@ func (r *DNSPolicyReconciler) calculateStatus(ctx context.Context, dnsPolicy *v1
 	}
 
 	recordsList := &kuadrantdnsv1alpha1.DNSRecordList{}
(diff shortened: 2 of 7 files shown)


api/v1alpha1/dnspolicy_types.go
```diff
@@ -133,7 +133,7 @@ type DNSPolicyStatus struct {
 	HealthCheck *HealthCheckStatus `json:"healthCheck,omitempty"`
 
 	// +optional
-	ProbeConditions map[string][]metav1.Condition `json:"probeConditions,omitempty"`
+	RecordConditions map[string][]metav1.Condition `json:"recordConditions,omitempty"`
```
mikenairn, line 136: Makes more sense than `probeConditions`. I do wonder how we intend to this for other policies though, i had expected a more policy generic way of reporting status about controller sub resources.

controllers/dnspolicy_status.go
```diff
@@ -90,7 +92,15 @@ func (r *DNSPolicyReconciler) calculateStatus(ctx context.Context, dnsPolicy *v1
 
 	meta.SetStatusCondition(&newStatus.Conditions, *enforcedCondition)
 
-	propagateRecordConditions(recordsList, newStatus)
+	for _, record := range recordsList.Items {
+		for _, reference := range record.GetOwnerReferences() {
+			if reference.Controller != nil && *reference.Controller && reference.Name == dnsPolicy.Name && reference.UID == dnsPolicy.UID {
+				controlledRecords.Items = append(controlledRecords.Items, record)
+			}
+		}
```
mikenairn, line 100: What's the significance of it being a controller reference or not? This also differs from how we find related sub resources [elsewhere](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_dnsrecords.go#L167), wondering which approach is more desirable, not sure if we use the owner ref in other polices or not.

philbrookes: I just copied this logic from [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_status.go#L103), happy to change it, if there's a better pattern, this does seem to work for me though.

mikenairn (approved):

PR #607, opened 2024-05-02
fix: do not reconcile status of target objects multiple times

### LINKED ISSUE #605: target status causes constant reconcilation when there is a conflicting policy
# Description
When one RLP (rlp0) is created and accepted and then another RLP (rlp1) is created that targets the same route. The target status reconciler will constantly update the `HTTPRoute` condition with being affected by rlp0 and then unaffected by rlp1 as it loops through all the policies and get the route. Since there's multiple policies, although conflicting, the generated affected condition is not the same and so constantly updates the route and triggers the reconcile of other controllers.
-  https://github.com/Kuadrant/kuadrant-operator/blob/5834a9dfbf27dcc61549b9505457bd32ff1b404c/controllers/target_status_controller.go#L142-L152

# How to replicate
* Create HTTPRoute
```
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
EOF
```
* Create RLPs
```
kubectl apply -f - <<EOF
---
# Accepted
kind: RateLimitPolicy
metadata:                
  name: rlp0
spec:   
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:   
    "create-toy":
      rates:
      - limit: 5
        duration: 10
        unit: second
---
# Conflict
kind: RateLimitPolicy
metadata:                
  name: rlp1
spec:   
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:   
    "create-toy":
      rates:
      - limit: 5
        duration: 10
        unit: second
EOF
```
* Watch route status shifting between affected and unaffected condition
```
watch -n 0.1 "kubectl get httproutes -A -o yaml | yq '.items[].status'"
```
* Controller should be constantly reconciling from the logs

guicassolato:
Closes #605


KevFan (approved):
Verified, looks good to me 🏅

PR #608, opened 2024-05-02
Update limitador to 0.9.0

eguzki:
### What

Update Limitador to `v0.9.0`

Bring to the Kuadrant CRD Limtiador's new features:
* rate limit headers
* redis-cached -> options -> response-timeout
* telemetry
* ~tracing~ -> purposefully left out, for now.
* verbosity

Added some integrations test to increase coverage. 

Refactored reconciliation logic to enforce limitador CR configuration only when defined in the Kuadrant CR object. If not defined in the kuadrant CR object, the controller will not revert changes in the limitador object.


alexsnaps:
There are few things here that we deliberately left out... e.g. tracing. We purposefully decided to _not_ have them part of the API. This is fairly big change and I think we should be cautious about adding all these. It would also possibly mean needing to support these moving forward!

eguzki:
> There are few things here that we deliberately left out... e.g. tracing. We purposefully decided to _not_ have them part of the API. This is fairly big change and I think we should be cautious about adding all these. It would also possibly mean needing to support these moving forward!

removing tracing. Anything else?

alexsnaps:
Honestly, anything around the cached_redis makes me uncomfortable...

eguzki:

pkg/kuadranttools/limitador_mutators.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func LimitadorVerbosityMutator(desired, existing *limitadorv1alpha1.Limitador) bool {
+	update := false
+	if !reflect.DeepEqual(existing.Spec.Verbosity, desired.Spec.Verbosity) {
+		existing.Spec.Verbosity = desired.Spec.Verbosity
+		update = true
+	}
+	return update
+}
```
didierofrivia, line 119: Do you think it's possible to DRY these funcs a bit, with something like...

```go
func UpdateFieldIfNotEqual(existing, desired interface{}, fieldName string) bool {
	existingVal := reflect.ValueOf(existing).Elem().FieldByName(fieldName)
	desiredVal := reflect.ValueOf(desired).Elem().FieldByName(fieldName)

	if existingVal.IsValid() && desiredVal.IsValid() {
		if !existingVal.Equal(desiredVal) {
			existingVal.Set(desiredVal)
			return true
		}
	}
	return false
}

```

Since we are updating fields when are not matching the desired state, then we can populate the fields in an array and call it either in a for loop or reducer fn like...

```go
limitadorSpecFieldNames := make(string, 0)
limitadorSpecFieldNames = append(limitadorSpecFieldNames,"Affinity")

....

for _, fieldName := range limitadorSpecFieldNames {
		if UpdateFieldIfNotEqual(existing, desired, fieldName) {
			update = true
		}
	}
```

not sure if we can even use the limitador spec types for the iteration instead of strings and interface{}

Maybe worth to try... maybe not

controllers/kuadrant_controller.go
```diff
@@ -440,29 +440,31 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
-	}
-	if kObj.Spec.Limitador.PodDisruptionBudget != nil {
-		limitadorMutators = append(limitadorMutators, kuadranttools.LimitadorPodDisruptionBudgetMutator)
-	}
-	if kObj.Spec.Limitador.ResourceRequirements != nil {
-		limitadorMutators = append(limitadorMutators, kuadranttools.LimitadorResourceRequirementsMutator)
-	}
-	if kObj.Spec.Limitador.Verbosity != nil {
-		limitadorMutators = append(limitadorMutators, kuadranttools.LimitadorVerbosityMutator)
+	if kObj.Spec.Limitador != nil {
```
didierofrivia, line 443: If you feel the DRYing is worthy, maybe the reflect lib can help with iteration...

```go
if kObj.Spec.Limitador != nil {
fields := reflect.ValueOf(kObj.Spec.Limitador)
for i := 0; i < v.NumField(); i++ {
// here check if the value `v.Field(i).Interface()` is nil and call the mutator with the desired and expected
}
```

didierofrivia:
It looks good, maybe there's a pattern for crafting the mutators other than the suggested, we could put a pin and look into it later on too.

pkg/kuadranttools/limitador_mutators.go
eguzki: It seems a great idea to avoid DRY. However, I have few little issues with the proposed implementation

* What is the semantics of that `Equal` method? Each field may need different comparing method. 
* The decision call of which fields needs to be "updated if not equal", belongs to the caller, not to the mutator. 
* TBH: I am not very fan of the reflection lib, only when strictly needed.

didierofrivia (approved):
Let's leave the DRYing pending until we know a common pattern for comparing specs. 👍🏼

Issue #551, opened 2024-04-17
DNSPolicy: Update all integration tests to validate the structure of the produced DNSrecord endpoints 


mikenairn:
**What**

Add a validation function that can be called by every dnspolicy integration test to check the integrity of the DNSRecord endpoints being generated. It should test whether one dnsName (i.e. \<clusterHash\>-\<gwHash\>.klb.test.example.com) can be reached from another dnsName (test.example.com) by inspecting the dnsName and targets of each endpoint in the record.  

Example pass( test.example.com -> clusterHash-gwHash.klb.test.example.com):

1. test.example.com (klb.test.example.com) 
2. klb.test.example.com (default.klb.test.example.com) 
3. default.klb.test.example.com(clusterHash-gwHash.klb.test.example.com) 
4. clusterHash-gwHash.klb.test.example.com (127.0.0.1)

Example fail( test.example.com -> clusterHash-gwHash.klb.test.example.com):

1. test.example.com (klb.test.example.com) 
2. klb.test.example.com (default.klb.test.example.com) 
3. clusterHash-gwHash.klb.test.example.com (127.0.0.1)

**How**

- [ ] Add validation function that takes in a `from` -> `to` dnsName as input and returns true or false
- [ ] Update all dnsPolicy integration tests that generate a DNSRecord to call this function

PR #613, opened 2024-05-03
GH-551 check traversability of generated endpoints

makslion:
Adding a new helper function that checks if endpoints on the DNSRecord are traversable or not. And using that function in existing integration suie

### DIFF
--- a/controllers/dnspolicy_controller_multi_cluster_test.go
+++ b/controllers/dnspolicy_controller_multi_cluster_test.go
@@ -79,7 +79,8 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
 		gwHash = common.ToBase36HashLen(gateway.Name+"-"+gateway.Namespace, 6)
 
 		//Set multi cluster gateway status
-		Eventually(func() error {
+		Eventually(func(g Gomega) {
+			g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(gateway), gateway)).To(Succeed())
 			gateway.Status.Addresses = []gatewayapiv1.GatewayStatusAddress{
 				{
 					Type:  ptr.To(multicluster.MultiClusterIPAddressType),
@@ -116,7 +117,7 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
 					Conditions:     []metav1.Condition{},
 				},
 			}
-			return k8sClient.Status().Update(ctx, gateway)
+			g.Expect(k8sClient.Status().Update(ctx, gateway)).To(Succeed())
 		}, TestTimeoutMedium, TestRetryIntervalMedium).ShouldNot(HaveOccurred())
 
 		recordName = fmt.Sprintf("%s-%s", TestGatewayName, TestListenerNameOne)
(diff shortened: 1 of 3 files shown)


controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
...
+					// target is not out IP. Use target as a host and check for existence of Endpoints with such a DNSName
+					for _, target := range ep.Targets {
+						// if at least one returns as true allTargetsFound will be locked in true
+						allTargetsFound = allTargetsFound || testEndpointsTraversable(endpoints, target, []string{ip})
+					}
+				}
+			}
+		}
+		// at least one endpoint that matches host targeted IP
+		return allTargetsFound
```
mikenairn, line 564: Does this not return on the first found target?

controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
...
+				// target is our IP
+				if slices.Contains(ep.Targets, ip) {
+					return true
+				} else {
+					// target is not out IP. Use target as a host and check for existence of Endpoints with such a DNSName
+					for _, target := range ep.Targets {
+						// if at least one returns as true allTargetsFound will be locked in true
+						allTargetsFound = allTargetsFound || testEndpointsTraversable(endpoints, target, []string{ip})
+					}
+				}
```
mikenairn, line 560: Should probably break here

controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
 	}
 }
 
+// testEndpointsTraversable consumes an array of endpoints and returns a boolean
+// indicating presence of that path from host to all IPs
+func testEndpointsTraversable(endpoints []*endpoint.Endpoint, host string, ips []string) bool {
```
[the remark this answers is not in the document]
makslion: those are always an array of IP addresses :thinking:  why rename?

controllers/helper_test.go
makslion: We should really check all of the endpoints that match the current host. [Here](https://github.com/Kuadrant/kuadrant-operator/blob/45337827e3b3b41a081ed7639b596381848f3310/controllers/dnspolicy_controller_multi_cluster_test.go#L229-L243), for example, we will fail if we break after the first match if we happened to look for a wrong IP

controllers/helper_test.go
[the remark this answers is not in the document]
mikenairn: Why does it have to be limited to checking for ip targets?

controllers/helper_test.go
makslion: No, it returns "at least one emdpoint has IP" after we went though the list of all endpoints

controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
...
+			// the target on the endpoint exists as an endpoint
+			if ep.DNSName == host {
+				// target is our IP
+				if slices.Contains(ep.Targets, ip) {
+					return true
+				} else {
+					// target is not out IP. Use target as a host and check for existence of Endpoints with such a DNSName
+					for _, target := range ep.Targets {
+						// if at least one returns as true allTargetsFound will be locked in true
+						allTargetsFound = allTargetsFound || testEndpointsTraversable(endpoints, target, []string{ip})
```
mikenairn, line 558: Probably no point calling `testEndpointsTraversable` unless the endpoint is a CNAME type

controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
...
+			if testEndpointsTraversable(record.Spec.Endpoints, hostname, IP) {
+				return true
+			}
+		}
+		// no records are traversable for this key/value pair
+		return false
+	}
+	// no hostnames to check
+	return false
+}
```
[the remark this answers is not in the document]
makslion: some of the tests deal with the list of records. It is either this or adjusting some of the tests. This check is not expensive so i went with this approach to allow for more flexibility in the tests. (and i just spotted that if the first record is traversable this func returns `true` without checking the second).

controllers/helper_test.go
makslion: if `endpoint.Target` is an IP it will just check it it `if ep.DNSName == IP` and happily fail. Otherwise there should be a regex check of some sort to tell if that is IP or not :thinking: I don't have any strong feelings about one way or another. Let me see what I can do

controllers/helper_test.go
```diff
@@ -538,6 +540,52 @@ func testBuildManagedZone(name, ns, domainName string) *kuadrantdnsv1alpha1.Mana
...
+// indicating presence of that path from host to all IPs
+func testEndpointsTraversable(endpoints []*endpoint.Endpoint, host string, ips []string) bool {
+	for _, ip := range ips {
+		allTargetsFound := false
+		for _, ep := range endpoints {
+			// the target on the endpoint exists as an endpoint
+			if ep.DNSName == host {
+				// target is our IP
+				if slices.Contains(ep.Targets, ip) {
+					return true
```
mikenairn, line 553: Why does this return here?

controllers/helper_test.go
mikenairn: ok, i see

controllers/helper_test.go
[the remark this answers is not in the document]
mikenairn: I would probably have changed the tests tbh. If this started failing it will probably be hard enough to debug the issue without having arrays of records you need to look at.

Originally we had been doing assertions on the arrays of records, and there are obviously still some tests that are, but we should really change those since it was proving difficult to debug.

controllers/helper_test.go
mikenairn: I'm not saying it won't work, it's just pointless so why do it. If you hit an A record with multiple targets it will call that function x amount of times for no reason. You don't need to do regexp either,  since you already know the type of endpoint `ep.RecordType`

controllers/helper_test.go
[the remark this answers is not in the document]
makslion: right now it is "I am looking for IPs. Havent found them but still have where to look / found them so succeed / haven't found them and novhere to look so fail". If I will allow it to look for targets there will be two paths: "I looking for a CNAME. Found it so look for it's target / didn't found it - fail" and "I'm looking for an IPs" one. The difference would be an extra `if` statement that will decide if to say it is traversable or use the current `target` as a new host (the later is done now by just recurring). And this will, actually, lead to more calls (more expecive to run) if I'm not mistaken.

controllers/helper_test.go
makslion: Changed. Now should be more obvious

controllers/helper_test.go
makslion: this is the match for the IP in questions. Means we found an A record that contains the target IP and the fact that we got to the record means there is a path. We should return here to the higher levels. 
Although you led me to an error - if the second IP is not found we have already returned from the highest level of the recursion: fale positive change. Changed

controllers/helper_test.go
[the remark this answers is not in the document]
makslion: Rewrote tests. Now only one way of checking traversability

controllers/dnspolicy_controller_multi_cluster_test.go
```diff
@@ -141,41 +142,50 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
...
+							PointTo(MatchFields(IgnoreExtras, Fields{
+								"DNSName":       Equal(TestHostOne),
+								"Targets":       ContainElements(TestIPAddressOne, TestIPAddressTwo),
+								"RecordType":    Equal("A"),
+								"SetIdentifier": Equal(""),
+								"RecordTTL":     Equal(externaldns.TTL(60)),
+							})),
+						),
+					}),
+				}))
```
mikenairn, line 169:
```suggestion
				g.Expect(dnsRecord.Name).To(Equal(recordName))
				g.Expect(dnsRecord.Spec.OwnerID).To(Equal(&ownerID))
				g.Expect(dnsRecord.Spec.ManagedZoneRef.Name).To(Equal("mz-example-com"))
				g.Expect(dnsRecord.Spec.Endpoints).To(ConsistOf(
					PointTo(MatchFields(IgnoreExtras, Fields{
						"DNSName":       Equal(TestHostOne),
						"Targets":       ContainElements(TestIPAddressOne, TestIPAddressTwo),
						"RecordType":    Equal("A"),
						"SetIdentifier": Equal(""),
						"RecordTTL":     Equal(externaldns.TTL(60)),
					})),
				))
```

The main reason these should be re-factored is that when they fail having so much done in a single `Expect` produces a really hard to read error, in some cases so large it doesn't even print all of it. Not doing a single expect on the entire list(`recordList.Items`) is an improvement, but it can be improved again by just doing an expect on each part of the dnsrecord we care about. 

This re-factor should probably be done in it's own PR where there are no additional changes to the tests or code.

mikenairn: Easier to read: 

```.go
g.Expect(dnsRecord.Name).To(Equal(recordName))
g.Expect(dnsRecord.Spec.OwnerID).To(Equal(&ownerID))
g.Expect(dnsRecord.Spec.ManagedZoneRef.Name).To(Equal("mz-example-com"))
g.Expect(dnsRecord.Spec.Endpoints).To(ConsistOf(
	PointTo(MatchFields(IgnoreExtras, Fields{
		"DNSName":       Equal(TestHostOne),
		"Targets":       ContainElements(TestIPAddressOne, TestIPAddressTwo),
		"RecordType":    Equal("A"),
		"SetIdentifier": Equal(""),
		"RecordTTL":     Equal(externaldns.TTL(60)),
	})),
))
```

makslion: #652 to track it

controllers/helper_test.go
makslion: @mikenairn are you ok with the changes made?

mikenairn:
This doesn't really do what was asked for in #551, and you have gone a different route of checking the eventual target IP(s) are reachable. This may or may not be useful, but we probably want to consider why #551 was created in the first place. 

There were a number of PRs being opened that were changing the tests to suit the results they were seeing, in particular when it came to generated hashes in the dns names, and even though our current tests do check all endpoints in most cases, the verbosity of those tests makes it very difficult to be sure that things weren't changing that shouldn't be unless you really looked for it in reviews. To make this less likely to happen we created #551 to add validation to make sure one dns name could always be reached from another i.e. test.example.com -> clusterHash-gwHash.klb.test.example.com, and it was to be a generic enough function that would allow us to use it wherever it was needed, so could be used to check the geo dns names were still connected for example i.e. test.example.com -> geo1.klb.test.example.com. This would allow the addition of explicit steps to assert the structure of the endpoints where it is important to do so.

The solution given here is to make sure that all the IPs given are reachable from a given host, but it doesn't really check how they come to be accessible, you could make that test pass and still have an endpoints list that we don't really desire. We already check the endpoint target values explicitly, what advantage are we gaining by doing this? It doesn't really tell me that my geo target is still accessible from the root host, just that the IP is a target somewhere somehow.

This also highlights that we have no tests checking for CNAME targets. It is the case that an endpoint list can be created that has no IPs, but rather it creates a CNAME record pointing to a loadbalancer dns name. The current solution of checking the ip's only wouldn't allow for a test that is checking for a known CNAME target, if we want to check targets it should at least allow this.

makslion:
The last comment was discussed in the call. 

It was decided to keep this test checking for targets on the endpoint but not restrain it to A records only. Now it can consume an array of targets (destinations) to reach 
Note that it cannot guarantee the presence of the endpoint but only the presence of the target on the endpoint

mikenairn (approved):

controllers/helper_test.go
```diff
@@ -164,6 +166,37 @@ func testBuildInMemoryCredentialsSecret(name, ns string) *corev1.Secret {
 	}
 }
 
+// testEndpointsTraversable consumes an array of endpoints and returns a boolean
+// indicating presence of that path from host to all destinations
+// this function DOES NOT report a presence of an endpoint with one of destinations DNSNames
+func testEndpointsTraversable(endpoints []*endpoint.Endpoint, host string, destinations []string) bool {
```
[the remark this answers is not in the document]
makslion: it will create confusion since we have `endpoint.targets` and a list of `targets` we are trying to reach. They were called `targets` and `endpointTargets` for a while but this does not help. But yes, i really want to avoid calling them `targets` as much as I can: the function is awkward to read as is and creating similar names won't help.

PR #619, opened 2024-05-06
add catalogsource to install doc

maleck13:
add docs to install via catalogsource directly


alexsnaps (approved):
Unsure if we want to merge this yet necessarily, but yeah...

maleck13:
I will merge for now. If we get the operator released to operator hub will revert

PR #622, opened 2024-05-07
Doc format

maleck13:
some missing code quotes

### DIFF
--- a/doc/user-guides/secure-protect-connect-single-multi-cluster.md
+++ b/doc/user-guides/secure-protect-connect-single-multi-cluster.md
@@ -418,10 +418,12 @@ Use `kuadrantctl` to generate your `HTTPRoute`.
 
 NOTE: The sample OAS has some placeholders for namespaces and domains. You will inject valid values into these placeholders based on your previous environment variables.
 
-Generate the resource from your OAS, (`envsubst` will replace the placeholders):
+Generate the resource from your OAS, (`envsubst` will replace the placeholders) and apply it:
 
 ```bash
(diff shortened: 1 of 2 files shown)


doc/user-guides/secure-protect-connect-single-multi-cluster.md
```diff
@@ -351,6 +351,15 @@ You can use `curl` to hit your endpoint. You should see a `403` Because this exa
 curl -k -w "%{http_code}" https://$(kubectl get httproute test -n ${gatewayNS} -o=jsonpath='{.spec.hostnames[0]}')
 ```
 
+
+### Opening up the Gateway for other namespaces:
+
+As you have now configured the Gateway, secured it with Kuadrant policies and tested it, you can now open it up for use by other teams in other namespaces:
+
+```bash
+kubectl patch gateway ${gatewayName} -n ${gatewayNS} --type='json' -p='[{"op": "replace", "path": "/spec/listeners/0/allowedRoutes/namespaces/from", "value":"All"}]'
```
pehala, line 360: Wouldnt it be better to include it at original Gateway creation section?

maleck13: The original gateway limits HTTPRoutes to the same ns as it is not protected. It is only at this point you have applied the policies and tested connectivity and know it is safe for other namespaces to use the Gateway. At least this is the concept it is intended to convey

jasonmadigan (approved):
lgtm.

I can push this to 0.7.0 docs after (a bit manual right now)

pehala (approved):

eguzki:
@maleck13 do you still want to push for this?

PR #623, opened 2024-05-07
disable HTTPRoute indexer when CRDs not present

eguzki:
`mgr.GetFieldIndexer().IndexField()` returns error when HTTPRoute CRD not present, hence the operator dies when the GatewayAPI CRDS not installed in the cluster. 

While this is reasonable, it is causing failure on OLM catalog release tests. This should not be backported to `main`.

### DIFF
--- a/controllers/rate_limiting_wasmplugin_controller.go
+++ b/controllers/rate_limiting_wasmplugin_controller.go
@@ -343,6 +343,16 @@ func (r *RateLimitingWASMPluginReconciler) routeFromRLP(ctx context.Context, t *
 // to prevent creating the same index field multiple times, the function is declared private to be
 // called only by this controller
 func addHTTPRouteByGatewayIndexer(mgr ctrl.Manager, baseLogger logr.Logger) error {
+	ok, err := kuadrantgatewayapi.IsGatewayAPIInstalled(mgr.GetRESTMapper())
+	if err != nil {
+		return nil
+	}
+
+	if !ok {
+		baseLogger.Info("GatewayAPI CRDs not found. Disabling HTTPRoute indexer")
+		return nil
+	}
+
 	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &gatewayapiv1.HTTPRoute{}, HTTPRouteGatewayParentField, func(rawObj client.Object) []string {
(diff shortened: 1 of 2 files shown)


eguzki:
Upgrade tested and works

maleck13 (approved):
ok to get us out of a bind.

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -343,6 +343,16 @@ func (r *RateLimitingWASMPluginReconciler) routeFromRLP(ctx context.Context, t *
 // to prevent creating the same index field multiple times, the function is declared private to be
 // called only by this controller
 func addHTTPRouteByGatewayIndexer(mgr ctrl.Manager, baseLogger logr.Logger) error {
+	ok, err := kuadrantgatewayapi.IsGatewayAPIInstalled(mgr.GetRESTMapper())
```
guicassolato, line 346: I think we should move this function out of this reconciler. The index is also used at https://github.com/Kuadrant/kuadrant-operator/blob/d4c57789d1e449930d51ada7e0e569d5bdb1487e/controllers/target_status_controller.go#L206

eguzki: When the CRDs are not present, the indexer is not created, hence the query using the index will not work. That's expected. This is in order to pass the tests. When the operator is being installed by users and customers the CRDs should be there to have all working as expected.

eguzki: BTW, this is already happening https://github.com/Kuadrant/kuadrant-operator/pull/527/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R136 in that PR.

WE just do not need it for this patch.

guicassolato: I understand what this is and my comment has nothing to do with the patch itself. Just pointing out this function does not belong to this file (anymore).

eguzki: It should have never been there. But it was my first indexer implementation :) (and somebody approved)

PR #626, opened 2024-05-07
Sail default for quickstarts

jasonmadigan:
- Default quickstart-setup.sh to Istio installs via Sail
- Also needs a docs update to remove ref to the env-var.

### DIFF
--- a/hack/quickstart-setup.sh
+++ b/hack/quickstart-setup.sh
@@ -55,13 +55,6 @@ fi
 if [ -z $KUADRANT_REF ]; then
   KUADRANT_REF=${KUADRANT_REF:="main"}
 fi
-if [ -z $MGC_REF ]; then
-  MGC_REF=${MGC_REF:="main"}
-fi
-
-if [ -z $ISTIO_INSTALL_SAIL ]; then
-  ISTIO_INSTALL_SAIL=${ISTIO_INSTALL_SAIL:=false}
-fi
 
 export TOOLS_IMAGE=quay.io/kuadrant/mgc-tools:latest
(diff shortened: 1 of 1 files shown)


hack/quickstart-setup.sh
```diff
@@ -450,21 +446,10 @@ success "Gateway API installed successfully."
...
 info "Installing Istio as a Gateway API provider... 🛫"
-if [ "$ISTIO_INSTALL_SAIL" = true ]; then
-  info "Installing Istio via Sail"
-  kubectl apply -k ${KUADRANT_ISTIO_KUSTOMIZATION}
-  kubectl -n istio-system wait --for=condition=Available deployment istio-operator --timeout=300s
-  kubectl apply -f ${KUADRANT_REPO_RAW}/config/dependencies/istio/sail/istio.yaml
-else
-  # Create CRD first to prevent race condition with creating CR
-  info "Generating Istio configuration... 🛠️"
-  kubectl kustomize ${MGC_ISTIO_KUSTOMIZATION} >${TMP_DIR}/doctmp
```
david-martin, line 461: with this removed, you can remove `MGC_ISTIO_KUSTOMIZATION` & `MGC_REF` vars as well

david-martin:
/lgtm
/approve

verified locally

maleck13:
I think before we do this we need to figure out how we keep sail upto date. Really we should be testing against the latest sail upstream at the moment we are pinned to an older version for stability

jasonmadigan:
@maleck13 Adam working a bit on this in https://github.com/Kuadrant/kuadrant-operator/pull/696/files

maleck13:
This is already the case now

PR #627, opened 2024-05-07
Fixing broken hyperlinks in user guides

pmccarthy:
**What**
Currently the hyperlinks listed in the secure, protect and connect user guide do not work due to an additional `-` character in the url


jasonmadigan (approved):
LGTM

PR #628, opened 2024-05-07
ci: skip test workflow when only doc changes

mikenairn:
Copied from https://github.com/Kuadrant/multicluster-gateway-controller/blob/main/.github/workflows/ci.yaml#L10


didierofrivia (approved):
📚

PR #631, opened 2024-05-08
Add observability target

david-martin:
To verify, run the `quickstart-setup.sh` script and verify the observability stack comes up correctly (can access grafana, and the gateway api dashboards show data)

There is also a new make target `deploy-observability`.
To verify this, run `make local-cluster-setup` followed by `make deploy-observability`, and check the the observability stack as above.


ehearneRedHat:
@david-martin Taking a look now...

ehearneRedHat (approved):
Verification steps were successful.

PR #632, opened 2024-05-08
Return early if the AuthPolicy being filtered is marked for deletion

Boomatang:
Improves filter function and returns early when searching for AuthPolicies attached to the Gateway.

### DIFF
--- a/controllers/authpolicy_authconfig.go
+++ b/controllers/authpolicy_authconfig.go
@@ -203,9 +203,13 @@ func routeGatewayAuthOverrides(t *kuadrantgatewayapi.Topology, ap *api.AuthPolic
 	// 1. targets a gateway
 	// 2. is not the current AP that is being assessed
 	// 3. is an overriding policy
+	// 4. is not marked for deletion
 	affectedPolicies = utils.Filter(affectedPolicies, func(policy kuadrantgatewayapi.Policy) bool {
 		p, ok := policy.(*api.AuthPolicy)
 		if !ok {
+			return false
+		}
+		if p.DeletionTimestamp != nil {
 			return false
 		}
 		return kuadrantgatewayapi.IsTargetRefGateway(policy.GetTargetRef()) &&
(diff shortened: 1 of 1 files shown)


controllers/authpolicy_authconfig.go
```diff
@@ -203,11 +203,15 @@ func routeGatewayAuthOverrides(t *kuadrantgatewayapi.Topology, ap *api.AuthPolic
...
 	affectedPolicies = utils.Filter(affectedPolicies, func(policy kuadrantgatewayapi.Policy) bool {
 		p, ok := policy.(*api.AuthPolicy)
 		if !ok {
 			return false
 		}
+		if p.DeletionTimestamp != nil {
+			return false
+		}
 		return kuadrantgatewayapi.IsTargetRefGateway(policy.GetTargetRef()) &&
 			ap.GetUID() != policy.GetUID() && p.IsAtomicOverride()
```
guicassolato, line 216:
```suggestion
			ap.GetUID() != policy.GetUID() && p.IsAtomicOverride() && p.DeletionTimestamp == nil
```

?

Boomatang: So I did have it that way to start. My reasoning for not doing it was why do the extra function calls when it can return early if there is a deletion time stamp.

guicassolato: Make it the first condition then?

```suggestion
		return p.DeletionTimestamp == nil && 
			kuadrantgatewayapi.IsTargetRefGateway(policy.GetTargetRef()) &&
			ap.GetUID() != policy.GetUID() &&
			p.IsAtomicOverride()
```

I'm usually in favour of safeguard clauses but in this case where the return is straightforward, the less `if`s the better, no? In fact, if really we want to push it:

```suggestion
		return ok &&
			p.DeletionTimestamp == nil && 
			kuadrantgatewayapi.IsTargetRefGateway(policy.GetTargetRef()) &&
			ap.GetUID() != policy.GetUID() &&
			p.IsAtomicOverride()
```

Boomatang: Moving the `ok` check here makes more sense also. I am kinda unsure about the readability, personally I don't like to have logic in return statements. It makes using a debugger harder. But in this case this might be the cleaner option.

guicassolato (approved):

PR #634, opened 2024-05-09
Add thanos to hub quickstart cluster & remote write from all clusters

david-martin:
- adds installation of thanos to the first/hub cluster
- configures grafana in the first/hub cluster with thanos-query as a data source
- configures every cluster to remote write metrics to the hub thanos instance

To verify, the PR branch `thanos` will need to be referenced until changes are back on main.

- Setup a hub cluster with `KUADRANT_REF=thanos ./hack/quickstart-setup.sh`
- then at least 1 more cluster with `KUADRANT_REF=thanos ./hack/quickstart-setup.sh` again.

In the first/hub cluster, you should see metrics from both clusters in the Grafana explore view. e.g. the `up{app="metallb",component="controller"}` metric should have 1 entry for each cluster.
To access grafana in the hub, you'll need to port-forward to the service with `kubectl --context kind-kuadrant-local -n monitoring port-forward svc/grafana 3000:3000` then access it at http://127.0.0.1:3000/explore

### DIFF
--- a/hack/quickstart-setup.sh
+++ b/hack/quickstart-setup.sh
@@ -515,8 +515,8 @@ fi
 
 # Install observability stack
 info "Installing observability stack in ${KUADRANT_CLUSTER_NAME}..."
-kubectl kustomize ${KUADARNT_OBSERVABILITY_KUSTOMIZATION} | docker run --rm -i ryane/kfilt -i kind=CustomResourceDefinition | kubectl apply --server-side -f -
-kubectl kustomize ${KUADARNT_OBSERVABILITY_KUSTOMIZATION} | docker run --rm -i ryane/kfilt -x kind=CustomResourceDefinition | kubectl apply -f -
+kubectl kustomize ${KUADARNT_OBSERVABILITY_KUSTOMIZATION} | $CONTAINER_RUNTIME_BIN run --rm -i ryane/kfilt -i kind=CustomResourceDefinition | kubectl apply --server-side -f -
+kubectl kustomize ${KUADARNT_OBSERVABILITY_KUSTOMIZATION} | $CONTAINER_RUNTIME_BIN run --rm -i ryane/kfilt -x kind=CustomResourceDefinition | kubectl apply -f -
 kubectl kustomize ${KUADRANT_DASHBOARDS_KUSTOMIZATION} | kubectl apply --server-side -f -
 success "observability stack installed successfully."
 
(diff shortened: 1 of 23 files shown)


hack/quickstart-setup.sh
```diff
@@ -504,13 +506,26 @@ info "Deploying Kuadrant sample configuration..."
...
+# Install thanos on hub cluster
+if [ "$HUB" -eq 1 ]; then
+  info "Installing thanos in ${KUADRANT_CLUSTER_NAME}... (as hub cluster)"
+  kubectl apply -k ${KUADARNT_THANOS_KUSTOMIZATION}
+  success "thanos installed successfully."
+fi
+
 # Install observability stack
 info "Installing observability stack in ${KUADRANT_CLUSTER_NAME}..."
 kubectl kustomize ${KUADARNT_OBSERVABILITY_KUSTOMIZATION} | docker run --rm -i ryane/kfilt -i kind=CustomResourceDefinition | kubectl apply --server-side -f -
```
R-Lawton, line 518: The script fails at this point for me saying docker command not found. I do have a alias but maybe the pipe is causing weird things with it. Its not a change from this pr that's causing the issue but Im thinking we should use the CONTAINER_RUNTIME_BIN env from the top of the script instead

david-martin: Agreed.
Pushed up a fix for that

R-Lawton (approved):
LGTM works as expected can 
Screenshot 2024-05-10 at 12 33 27

ehearneRedHat (commented):
I know this PR has been approved and merged, but I wanted to document an issue I had when trying to review this PR and have since found the solution.

My system runs on FedoraOS, and without changing any settings, works fine for using just single cluster quickstart. However, I had issues earlier when trying to run multicluster quickstart and have discovered why.

My issue was that I could not get past the `INFO: Waiting for cert-manager deployments to be ready` part of the script. 

The reason for this, was that inotify did not have enough resources to create more than the pods in a single cluster.

Kind has [this page](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files) which explains why this issue occurs. After I ran:

```bash
sudo sysctl fs.inotify.max_user_watches=524288
sudo sysctl fs.inotify.max_user_instances=512
```
After deleting the faulty cluster, and re-running the quickstart as seen in the description, I was able to verify the changes working as expected.

Commenting this for future reference. :)

Issue #572, opened 2024-04-24
TLSPolicy Enforced Status


maleck13:
**What**

Based on https://github.com/Kuadrant/architecture/blob/main/rfcs/0004-policy-status.md we want to have some consistent status in place. 

One of those status conditions is the `Ready` condition which is already present in TLSPolicy. 

But we want to add a new condition `Enforced` this reflects the state of enforcement of the policy

- If there is no error and everything is in place then we move enforced to true
- If there is some form of error we mark enforced as unknown

**Done**

- [ ] Add logic to set the Enforced condition. Likely this is quite similar to the ready condition 
- [ ] Check certmanager status on certificates to see is ready enough to say everything (including secret is ready)
- [ ] Add new tests where appropriate 
- [ ] Update the TLSPolicy reference doc to call out the new status conditions 
- [ ] Ensure certificate issuer is present and `Ready`

PR #635, opened 2024-05-09
feat: tlspolicy enforced condition

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/572

Add enforced condition for TLSPolicy. This is based on whether the underlying `Issuer` and `Certificates` are in a `Ready` state.

# Verification
Happy path is integration tested but if you want to verify manually

* Checkout this branch
* Deploy
```
make local-setup
```
* Deploy policy 
```
kind: Gateway
metadata:
  name: istio-ingressgateway
spec:
  gatewayClassName: istio
  listeners:
  - name: http
    hostname: '*.toys.io'
    port: 443
    protocol: HTTPS
    tls:
      mode: Terminate
      certificateRefs:
      - name: toys
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All    
---
kind: Issuer
metadata:
  name: selfsigned-issuer
spec:
  selfSigned: {}
---
kind: TLSPolicy
metadata:
  name: gw-tls
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: selfsigned-issuer
EOF
```
* Check that the TLSPolicy has an enforced condition
```
kubectl get tlspolicy -A -o yaml | yq '.items[].status'
```
image

* Check that there is a new column for the enforced condition
```
kubectl get tlspolicy -A -o wide 
```

image

### DIFF
--- a/controllers/tlspolicy_status.go
+++ b/controllers/tlspolicy_status.go
@@ -18,20 +18,28 @@ package controllers
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"slices"
 
+	certmanv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
 	"github.com/kuadrant/kuadrant-operator/api/v1alpha1"
 	"github.com/kuadrant/kuadrant-operator/pkg/library/kuadrant"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/reconcilers"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
 )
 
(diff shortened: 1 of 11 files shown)


maleck13 (approved):
Changes look good. 
I am starting to wonder if we need an enforced status per listener though. Doesn't need to be part of this PR

controllers/tlspolicy_status.go
```diff
@@ -67,5 +75,88 @@ func (r *TLSPolicyReconciler) calculateStatus(tlsPolicy *v1alpha1.TLSPolicy, spe
...
+	}
+
+	if len(gwDiffObj.GatewaysWithValidPolicyRef) == 0 {
+		return errors.New("no valid gateways found")
+	}
+
+	for _, gw := range gwDiffObj.GatewaysWithValidPolicyRef {
+		expectedCertificates := r.expectedCertificatesForGateway(ctx, gw.Gateway, tlsPolicy)
+
+		for _, cert := range expectedCertificates {
```
KevFan, line 146: Not sure if this is what we want but there's no error if `len(expectedCerificates) == 0` since this is already allowed at:
- https://github.com/Kuadrant/kuadrant-operator/blob/bca513031565aafea05cb8ea0469cacc906e369a/controllers/tlspolicy_certmanager_certificates.go#L115-L119
- https://github.com/Kuadrant/kuadrant-operator/blob/bca513031565aafea05cb8ea0469cacc906e369a/controllers/tlspolicy_certmanager_certificates.go#L36-L39

so I've kept it the same. In this case a TLSPolicy is still marked as `enforced: true` even if there are no `Certificates` created

maleck13: I think that is ok. If there are no expected certificates then we have done everything we can to enforce

Issue #325, opened 2023-11-21
Envoy Gateway Support 


alexsnaps:
While there isn't first class support for [Wasm](https://github.com/envoyproxy/gateway/issues/2025) extensions yet in Envoy Gateway, it might be feasible to use [`EnvoyPatchPolicy`](https://gateway.envoyproxy.io/latest/design/envoy-patch-policy/) to achieve some of our goals right now, possibly with the use of the [Extension Manager](https://gateway.envoyproxy.io/latest/design/extending-envoy-gateway/#registering-extensions-in-envoy-gateway). 

This issue's goal is to track the work and/or dependencies so we can support Envoy Gateway as Gateway API provider.

## Initial Tasks

 - [x] #567
 - [x] Spike into seeing what's missing to add support for _Policy Attachment_ to Envoy Gateway, just like it does for `ExtensionFilter`s.
 - [x]  #645 
 - [x]  #681
 - [x] #686

## Tasks
- [x] #678
- [ ] Rate limit policy implementation: Rate Limit EnvoyExtensionPolicy controller
- [ ] Authorization policy implementation: SecurityPolicy controller

## Further questions

 - [x] How would [EG calling back into the Kuadrant Operator](https://gateway.envoyproxy.io/v0.6.0/design/extending-envoy-gateway/) about topology changes be addressed and what would it mean for our [DAG proposal](https://github.com/Kuadrant/architecture/issues/29)? 
 - [x] Are there issues in having out-of-order changes handled by EG and Kuadrant mutations of the config thru [`EnvoyPatchPolicy`](https://gateway.envoyproxy.io/latest/design/envoy-patch-policy/) 
 - [x] Are there unresolved dependencies in EG for us to get this done at all?

eguzki:
Added https://github.com/eguzki/poc-kuadrant-envoy-gateway/issues/2 to report findings

eguzki:
> How would [EG calling back into the Kuadrant Operator](https://gateway.envoyproxy.io/v0.6.0/design/extending-envoy-gateway/) about topology changes be addressed and what would it mean for our https://github.com/Kuadrant/architecture/issues/29?

[...]

PR #645, opened 2024-05-13
Have Istio optional

eguzki:
### What

Part of https://github.com/Kuadrant/kuadrant-operator/issues/325

This is a preparation work for on-boarding [EnvoyGateway](https://gateway.envoyproxy.io/) on following up PRs. It is, essentially,  moving Istio from being the **required** GatewayAPI provider to **optional** gateway provider. Among other things, being *optional* means that when Istio is not installed in the cluster, the kuadrant operator does not die. Additionally, make room for on-boarding other GatewayAPI providers. 

- [X] Make Istio EnvoyFilter controller optional. When  istio envoy filter CRD or GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make Istio Wasmplugin controller optional. When  istio WasmPlugin CRD or GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make kuadrant controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make ratelimitpolicy controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make authpolicy controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make dnspolicy controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make tlspolicy controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make target status controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [X] Make ratelimit policy enforcedstatus controller optional. When GatewayAPI CRD is not installed, the controller does not start. 
- [x] Make Istio optional. When istio is not installed, the kuadrant operator still runs. 
- [X] At golang level, helper functions working with Istio APIs in `github.com/kuadrant/kuadrant-operator/pkg/istio` package. Controllers should be the only one importing that package.
- [X] kuadrant status reports when all the supported GatewayAPI providers are not found.
- [x] (Integration) Tests refactor. 

| Golang package | Required environment | Makefile env setup target | Makefile test run target |
| --- | --- | --- | --- |
| `github.com/kuadrant/kuadrant-operator/tests/bare_k8s` | no gateway provider, no GatewayAPI CRDs. Just Kuadrant API and Kuadrant dependencies. | `make local-k8s-env-setup` | `make test-bare-k8s-integration` |
| `github.com/kuadrant/kuadrant-operator/tests/gatewayapi` | no gateway provider. GatewayAPI CRDs, Kuadrant API and Kuadrant dependencies. | `make local-gatewayapi-env-setup` | `make test-gatewayapi-env-integration` |
| `github.com/kuadrant/kuadrant-operator/controllers` | at least one gatewayapi provider. It can be any: istio, envoygateway, ...  | `make local-env-setup GATEWAYAPI_PROVIDER=[ISTIO\|SAIL]` (Default *Istio*) | `make test-integration` |
| `github.com/kuadrant/kuadrant-operator/tests/istio` | GatewayAPI CRDs, Istio, Kuadrant API and Kuadrant dependencies.  | `make local-env-setup GATEWAYAPI_PROVIDER=ISTIO` | `make test-istio-env-integration` |

#### Defined behavior
| GatewayAPI CRDs | Istio | Kuadrant CR status |
| ---|---|---|
| NO | NO | Not reconciled |
| YES | NO | Error: GatewayAPI provider not found |
| YES | YES | OK |

### Verification Steps: run kuadrant in cluster without GatewayAPI provider

* run environment
```sh
make local-gatewayapi-env-setup
```

* Run the operator
```sh
make run 
```

* Create Kuadrant CR
```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```
Kudrant CR status reports not ready status
```yaml
kubectl get kuadrant kuadrant -n kuadrant-system -o jsonpath='{.status}' | yq e -P
conditions:
  - lastTransitionTime: "2024-05-27T21:25:38Z"
    message: GatewayAPI provider not found
    reason: GatewayAPIPRoviderNotFound
    status: "False"
    type: Ready
observedGeneration: 1
```


controllers/suite_test.go
```diff
@@ -20,45 +20,25 @@ package controllers
...
-	"github.com/kuadrant/kuadrant-operator/pkg/library/reconcilers"
 	"github.com/kuadrant/kuadrant-operator/pkg/log"
-	//+kubebuilder:scaffold:imports
 )
 
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
+// This test suite will be run on bare k8s with GatewayAPI CRDS and Kuadrant CRDs installed
```
eguzki, line 40: Another option is to have these integration tests run on dedicated `github.com/kuadrant/kuadrant-operator/tests/gatewayapi` package.  No strong opinion on this.

eguzki: outdated

pkg/istio/external_authorizer.go
```diff
@@ -0,0 +1,283 @@
...
+			logger.Info("remove external authorizer from istio meshconfig")
+			if err = cl.Update(ctx, config.GetConfigObject()); err != nil {
+				return true, err
+			}
+		}
+	}
+	return true, nil
+}
+
+func unregisterExternalAuthorizerOSSM(ctx context.Context, cl client.Client, kObj *kuadrantv1beta1.Kuadrant) error {
```
guicassolato, line 79: It looks like we're extracting too much. These OSSM functions do not belong to the `istio` package IMO.

The decision of whether to activate Istio or OSSM should continue to happen in the base controller or some abstraction from where we fork towards one path or the other.

eguzki: ok, moving back the logic to register either in istio or OSSM to the controller.

eguzki: done!

KevFan (approved):
Verified, looks good to me anyway 👍

controllers/kuadrant_controller.go
```diff
@@ -495,6 +495,15 @@ func (r *KuadrantReconciler) reconcileAuthorino(ctx context.Context, kObj *kuadr
...
 func (r *KuadrantReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	ok, err := kuadrantgatewayapi.IsGatewayAPIInstalled(mgr.GetRESTMapper())
+	if err != nil {
+		return err
+	}
+	if !ok {
+		r.Logger().Info("Kuadrant controller disabled. GatewayAPI was not found")
+		return nil
+	}
+
```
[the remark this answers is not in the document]
guicassolato: This is a great point. I would try to avoid the restart if we can. If we are to keep the pod running anyway, at least it could retry the conditions in a loop, setup the reconcilers automatically in case the state changes.

controllers/authpolicy_authconfig.go
```diff
@@ -235,8 +235,8 @@ func getAffectedPolicies(t *kuadrantgatewayapi.Topology, ap *api.AuthPolicy) []k
 	return affectedPolicies
 }
 
-// authConfigName returns the name of Authorino AuthConfig CR.
-func authConfigName(apKey client.ObjectKey) string {
+// AuthConfigName returns the name of Authorino AuthConfig CR.
+func AuthConfigName(apKey client.ObjectKey) string {
```
guicassolato, line 239: Are we planning on moving this out to another package in another iteration? Is that the reasoning for exporting the function?

make/development-environments.mk
```diff
@@ -0,0 +1,122 @@
+##@ Deployment
+
+GATEWAYAPI_PROVIDER = ISTIO
+GATEWAYAPI_PROVIDER_LOWERCASE = $(shell echo $(GATEWAYAPI_PROVIDER) | tr A-Z a-z)
```
guicassolato, line 4: Can we make the expected value lowercase? We're not using `ISTIO` (uppercase) anywhere, are we?

```suggestion
GATEWAYAPI_PROVIDER = istio
```

make/integration-tests.mk
```diff
@@ -0,0 +1,47 @@
...
+.PHONY: test-integration
+test-integration: clean-cov generate fmt vet ginkgo ## Requires kubernetes cluster with at least one GatewayAPI provider installed.
+	mkdir -p $(PROJECT_PATH)/coverage/integration
+#	Check `ginkgo help run` for command line options. For example to filtering tests.
+	$(GINKGO) \
+		--coverpkg $(INTEGRATION_COVER_PKGS) \
+		--output-dir $(PROJECT_PATH)/coverage/integration \
+		--coverprofile cover.out \
+		-tags integration \
+		./controllers/...
```
guicassolato, line 47: Can we please keep `$(INTEGRATION_TEST_SUITE_PATHS)`? I use that a lot to filter the tests to run locally.

guicassolato: It can also be a more generic `INTEGRATION_TESTS_EXTRA_ARGS` that defaults to empty, if you prefer.


```suggestion
		$(INTEGRATION_TESTS_EXTRA_ARGS) ./controllers/...
```

I'd run it with, e.g.:

```sh
make test-integration INTEGRATION_TESTS_EXTRA_ARGS="--focus-file ./controllers/ratelimitpolicy_controller_test.go"
```

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
eguzki: Yeah, ideally the controller should keep trying until it finds it.

I thought about implementing it. But I wanted to discuss first the implementation

controllers/authpolicy_authconfig.go
eguzki: my bad. I initially moved the tests to another package. Fix it.

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
guicassolato: It can be done in another PR.

controllers/authpolicy_authconfig.go
eguzki: done

make/development-environments.mk
eguzki: :+1:  done

make/integration-tests.mk
eguzki: you know you can do that at Ginkgo level, right?

https://onsi.github.io/ginkgo/#focused-specs

I use `focused` feature a lot for doing exactly that.

eguzki:
rebasing as there are conflicts with some files

make/integration-tests.mk
guicassolato: Sure, but in that case I have to write the entire command myself, i.e. plus tags, progress, verbosity, etc, whereas this used be a handy extension on top of the standard way to run the integration tests.

pkg/istio/mesh_config.go
```diff
@@ -17,12 +17,115 @@ import (
...
+func NewKuadrantAuthorizer(namespace string) *KuadrantAuthorizer {
+	return &KuadrantAuthorizer{
+		extensionProvider: createKuadrantAuthorizer(namespace),
+	}
+}
+
+// GetExtensionProvider Returns the Istio MeshConfig ExtensionProvider for Kuadrant
+func (k *KuadrantAuthorizer) GetExtensionProvider() *istiomeshv1alpha1.MeshConfig_ExtensionProvider {
+	return k.extensionProvider
+}
```
guicassolato, line 48: I thought we were keeping these interfaces and functions at the abstract level. Does this mean that having a Kuadrant Authorizer became a Istio-specific thing?

eguzki: yes, it is an istio specific thing.

guicassolato: I don't follow. We won't deploy Authorino when integrating with Envoy Gateway?

eguzki: Yes, we will. But this "kuadrant authorizer" is not deploying authorino. The kuadrant controller is. The "kuadrant authorizer" is to register external auth in istio control plane. AFAIK, EnvoyGateway does not require any registration step in the control plane of EG. @adam-cattermole can confirm.

guicassolato: I understand, @eguzki. My point is that this abstraction ("Kuadrant Authorizer config") has not been as much of an abstraction as it could be or [originally intended](https://github.com/Kuadrant/kuadrant-operator/pull/192#discussion_r1219630839).

At the level of the Kuadrant CR/Kuadrant deployment, _configuring the Kuadrant Authorizer_ means ensuring that the Kuadrant deployment has all the configuration it needs for ext_authz requests to be sent to an instance of Authorino in the data plane, when handling traffic for the gateways in the scope of this Kuadrant deployment and it's applicable.

Because so far we only support Istio-based Gateway API controllers, it's easy overlook the difference, but I'm arguing for the decoupling of the abstraction that ensures the external authorizer is configured for a given Gateway API controller from the implementation details that get it done.

IWO, these interfaces belong to the Kuadrant controller and should be fixed to not know as much as they do today about Istio IMHO. In fact, they should not know anything at all about Istio. Only the Istio-specific implementations should know that the way to get the external authorizer configured (in their case) is by patching a "mesh config." Perhaps for other Gateway API controllers (e.g. EG), configuring the external authorizer at the level of the Kuadrant CR means something else; it could even be a no-op – but even that should be hidden behind the abstraction IMO.

Now that we are decoupling the controller from Istio, it looked like the right opportunity to fix this.

make/integration-tests.mk
eguzki: added

pkg/istio/mesh_config.go
eguzki: That abstraction is not meant to decouple from gateway API provider. It is totally coupled to Istio. 

```go
type ConfigWrapper interface {
	GetConfigObject() client.Object
	GetMeshConfig() (*istiomeshv1alpha1.MeshConfig, error)
	SetMeshConfig(*istiomeshv1alpha1.MeshConfig) error
}

type KuadrantAuthorizer struct {
	extensionProvider *istiomeshv1alpha1.MeshConfig_ExtensionProvider
}
```

I think the intended design wanted to abstract from the different "flavors" of Istio. But actually, IMHO, there is no abstraction actually. It is something like: "if istioOperator -> do this; else if istio configmap -> do that; else if ossm -> do other thing; else panic". 

Anyway, I get your point and I will try something I had in mind.

guicassolato: Prefer to stab it in separate PR to unblock this one? It's all refactoring in the end.

guicassolato (approved):
/lgtm

Nice work, @eguzki. Let's unblock this and let to tackle the issue of the loosely coupled Gateway API controllers with the future refactoring for sure to come.

PR #596, opened 2024-04-30
availability and latency slos

R-Lawton:
Creating SLO for both availability and latency using sloth. 

##Verification 
#### Unit tests
Run promtool tests:
`make alerts-tests`
You should see success for both alerts
#### Sloth generation
Install Sloth SLI https://sloth.dev/introduction/install/
Run `generate -i examples/alerts/sloth/latency.yaml --default-slo-period=28d`
The generated yaml should match the current alert for latency and when you change the file to availability you should see the same.

PR #648, opened 2024-05-14
Add info about sloth dashboard

david-martin:
Since #596 has merged, let's update the docs to add more info about these slo alerts and the sloth dashboard that can be used with them.

image


R-Lawton (approved):
lgtm

PR #654, opened 2024-05-16
docs: fix command in the single cluster guide

guicassolato:


david-martin (approved):
nice catch @guicassolato

Issue #580, opened 2024-04-25
ci: build catalog job can fail for branches with certain names


mikenairn:
**What**

The Build catalog job can fail in certain cases see https://github.com/Kuadrant/kuadrant-operator/actions/runs/8831513068/job/24259023192?pr=549

Looks like it fails due to the branch name having upper case characters in it.

KevFan:
The issue looks to be that when pushing images using branch name as tags, the pushed tag is always lowercased even though the branch name itself may have a uppercase. This causes a mismatch between the tag used by the workflow vs what is actually pushed/created in quay.io.

A possible solution to this may be to lowercase the tag input first and this this lowercased tag throughout the workflow instead. Possible ways to do this may be https://github.com/orgs/community/discussions/25768

PR #655, opened 2024-05-16
elliminate premature termination of DNS controller logic

makslion:
## What 
If we specify the first listener in the GW spec to be the one without attached routes we never get to the second and never create DNS records for it (even if there are attached routes). This PR fixes it. 
I'm not including integration test for this case - the suite has over 150 cases now and I do not consider this scenario to be critical for testing on each pull request

### DIFF
--- a/controllers/dnspolicy_controller_single_cluster_test.go
+++ b/controllers/dnspolicy_controller_single_cluster_test.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/kuadrant/kuadrant-operator/api/v1alpha1"
 	"github.com/kuadrant/kuadrant-operator/pkg/common"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/kuadrant"
 	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
 )
 
@@ -48,6 +49,7 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 		Expect(k8sClient.Create(ctx, managedZone)).To(Succeed())
 
 		gateway = NewGatewayBuilder(TestGatewayName, gatewayClass.Name, testNamespace).
+			WithHTTPListener("foo", "foo.example.com").
 			WithHTTPListener(TestListenerNameOne, TestHostOne).
 			WithHTTPListener(TestListenerNameWildcard, TestHostWildcard).
 			Gateway
@@ -71,6 +73,12 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 				},
 			}
 			gateway.Status.Listeners = []gatewayapiv1.ListenerStatus{
+				{
+					Name:           "foo",
+					SupportedKinds: []gatewayapiv1.RouteGroupKind{},
+					AttachedRoutes: 0,
+					Conditions:     []metav1.Condition{},
+				},
 				{
 					Name:           TestListenerNameOne,
 					SupportedKinds: []gatewayapiv1.RouteGroupKind{},
@@ -122,6 +130,21 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 		})
 
 		It("should create dns records", func() {
+
+			Eventually(func(g Gomega, ctx context.Context) {
+
+				g.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(dnsPolicy), dnsPolicy)).To(Succeed())
+				g.Expect(dnsPolicy.Status.Conditions).To(
+					ContainElement(MatchFields(IgnoreExtras, Fields{
+						"Type":    Equal(string(kuadrant.PolicyConditionEnforced)),
+						"Status":  Equal(metav1.ConditionFalse),
+						"Reason":  Equal(string(kuadrant.PolicyReasonEnforced)),
+						"Message": Equal("DNSPolicy has been partially enforced"),
+					})),
+				)
+
+			}, TestTimeoutMedium, TestRetryIntervalMedium, ctx).Should(Succeed())
+
 			Eventually(func(g Gomega, ctx context.Context) {
 				recordList := &kuadrantdnsv1alpha1.DNSRecordList{}
 				err := k8sClient.List(ctx, recordList, &client.ListOptions{Namespace: testNamespace})
--- a/controllers/dnspolicy_dnsrecords.go
+++ b/controllers/dnspolicy_dnsrecords.go
@@ -90,7 +90,7 @@ func (r *DNSPolicyReconciler) reconcileGatewayDNSRecords(ctx context.Context, gw
 			if err := r.dnsHelper.deleteDNSRecordForListener(ctx, gatewayWrapper, listener); client.IgnoreNotFound(err) != nil {
 				return fmt.Errorf("failed to delete dns record for listener %s : %w", listener.Name, err)
 			}
-			return nil
+			continue
 		}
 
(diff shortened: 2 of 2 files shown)


makslion:
For verification you could use [this](https://github.com/Kuadrant/kuadrant-operator/tree/GH-589-test) branch and run ```kustomize build test | k apply -f -``` to easily reproduce the bug

mikenairn:
> I'm not including integration test for this case - the suite has over 150 cases now and I do not consider this scenario to be critical for testing on each pull request

It seems fairly straightforward to force this error so not sure why we can't ensure we have it covered in our tests. It doesn't have to be an explicit test case if that is what you are concerned about, you can update the existing ones to include another listener with a status that has 0 AttachedRoutes. 

https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_controller_single_cluster_test.go#L50-L88

```go
gateway = NewGatewayBuilder(TestGatewayName, gatewayClass.Name, testNamespace).
	WithHTTPListener("foo", "foo.example.com"). //<-- Listener with no attached routes
	WithHTTPListener(TestListenerNameOne, TestHostOne).
	WithHTTPListener(TestListenerNameWildcard, TestHostWildcard).
	Gateway
Expect(k8sClient.Create(ctx, gateway)).To(Succeed())

clusterHash = common.ToBase36HashLen(clusterUID, utils.ClusterIDLength)
ownerID = common.ToBase36HashLen(fmt.Sprintf("%s-%s-%s", clusterUID, gateway.Name, gateway.Namespace), utils.ClusterIDLength)

gwHash = common.ToBase36HashLen(gateway.Name+"-"+gateway.Namespace, 6)

//Set single cluster gateway status
Eventually(func() error {
	gateway.Status.Addresses = []gatewayapiv1.GatewayStatusAddress{
		{
			Type:  ptr.To(gatewayapiv1.IPAddressType),
			Value: TestIPAddressOne,
		},
		{
			Type:  ptr.To(gatewayapiv1.IPAddressType),
			Value: TestIPAddressTwo,
		},
	}
	gateway.Status.Listeners = []gatewayapiv1.ListenerStatus{
		{
			Name:           "foo",
			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
			AttachedRoutes: 0,
			Conditions:     []metav1.Condition{},
		},
		{
			Name:           TestListenerNameOne,
			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
			AttachedRoutes: 1,
			Conditions:     []metav1.Condition{},
		},
		{
			Name:           TestListenerNameWildcard,
			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
			AttachedRoutes: 1,
			Conditions:     []metav1.Condition{},
		},
	}
	return k8sClient.Status().Update(ctx, gateway)
}, TestTimeoutMedium, TestRetryIntervalMedium).Should(Succeed())
```

[This](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_controller_single_cluster_test.go#L124) will fail without the fix.

makslion:
> > I'm not including integration test for this case - the suite has over 150 cases now and I do not consider this scenario to be critical for testing on each pull request
> 
> It seems fairly straightforward to force this error so not sure why we can't ensure we have it covered in our tests. It doesn't have to be an explicit test case if that is what you are concerned about, you can update the existing ones to include another listener with a status that has 0 AttachedRoutes.
> 
> https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_controller_single_cluster_test.go#L50-L88
> 
> ```go
> gateway = NewGatewayBuilder(TestGatewayName, gatewayClass.Name, testNamespace).
> 	WithHTTPListener("foo", "foo.example.com"). //<-- Listener with no attached routes
> 	WithHTTPListener(TestListenerNameOne, TestHostOne).
> 	WithHTTPListener(TestListenerNameWildcard, TestHostWildcard).
> 	Gateway
> Expect(k8sClient.Create(ctx, gateway)).To(Succeed())
> 
> clusterHash = common.ToBase36HashLen(clusterUID, utils.ClusterIDLength)
> ownerID = common.ToBase36HashLen(fmt.Sprintf("%s-%s-%s", clusterUID, gateway.Name, gateway.Namespace), utils.ClusterIDLength)
> 
> gwHash = common.ToBase36HashLen(gateway.Name+"-"+gateway.Namespace, 6)
> 
> //Set single cluster gateway status
> Eventually(func() error {
> 	gateway.Status.Addresses = []gatewayapiv1.GatewayStatusAddress{
> 		{
> 			Type:  ptr.To(gatewayapiv1.IPAddressType),
> 			Value: TestIPAddressOne,
> 		},
> 		{
> 			Type:  ptr.To(gatewayapiv1.IPAddressType),
> 			Value: TestIPAddressTwo,
> 		},
> 	}
> 	gateway.Status.Listeners = []gatewayapiv1.ListenerStatus{
> 		{
> 			Name:           "foo",
> 			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
> 			AttachedRoutes: 0,
> 			Conditions:     []metav1.Condition{},
> 		},
> 		{
> 			Name:           TestListenerNameOne,
> 			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
> 			AttachedRoutes: 1,
> 			Conditions:     []metav1.Condition{},
> 		},
> 		{
> 			Name:           TestListenerNameWildcard,
> 			SupportedKinds: []gatewayapiv1.RouteGroupKind{},
> 			AttachedRoutes: 1,
> 			Conditions:     []metav1.Condition{},
> 		},
> 	}
> 	return k8sClient.Status().Update(ctx, gateway)
> }, TestTimeoutMedium, TestRetryIntervalMedium).Should(Succeed())
> ```
> 
> [This](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_controller_single_cluster_test.go#L124) will fail without the fix.

I was mostly thinking about explicitly testing for it, but this could work as well, i guess :thinking:  added

controllers/dnspolicy_controller_single_cluster_test.go
```diff
@@ -71,6 +71,12 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 				},
 			}
 			gateway.Status.Listeners = []gatewayapiv1.ListenerStatus{
+				{
+					Name:           "foo",
+					SupportedKinds: []gatewayapiv1.RouteGroupKind{},
+					AttachedRoutes: 0,
+					Conditions:     []metav1.Condition{},
+				},
```
mikenairn, line 79: Just adding the status won't replicate the issue, the code you modified is inside an iteration of the spec listeners so you will also need to add it there so it goes into the check for attached routes and returns when it sees it's 0, see the example [here](https://github.com/Kuadrant/kuadrant-operator/pull/655#issuecomment-2120383882)  

```.go
gateway = NewGatewayBuilder(TestGatewayName, gatewayClass.Name, testNamespace).
	WithHTTPListener("foo", "foo.example.com"). //<-- Listener with no attached routes
	WithHTTPListener(TestListenerNameOne, TestHostOne).
	WithHTTPListener(TestListenerNameWildcard, TestHostWildcard).
	Gateway
```

mikenairn:
Build Images job is failing due to this https://github.com/Kuadrant/kuadrant-operator/issues/580. We can ignore it on this PR.

controllers/dnspolicy_controller_single_cluster_test.go
makslion: Done. Also added a check for the enforced condition on the policy since this was the original cause for creating the issue

mikenairn (approved):

PR #658, opened 2024-05-22
tests: Use inmemory provider backed managed zones

### LINKED ISSUE #601: DNSPolicy: Update integration tests to use inmemory dns provider
**What**

Update all DNSPolicy integration tests to use the [inmemory](https://github.com/Kuadrant/dns-operator/pull/103) dns provider.

**Why**

Will allow us to write more complete integration tests that can assert the status of resources under test (DNSRecord, ManagedZone and DNSPolicy etc..) since it will be working with a more "real" dns provider.

Note: We should just update the current tests here, extra tests and improvements to the current tests can be added once we have the basic test setup in place.

mikenairn:
closes #601 

Update all DNSPolicy integration tests to use the inmemory provider. If the dns operator is running on the target cluster managedzones and dnsrecords should work as if they are targeting a real DNS Provider.

Requires: https://github.com/Kuadrant/dns-operator/pull/142

### DIFF
--- a/controllers/dnspolicy_controller_multi_cluster_test.go
+++ b/controllers/dnspolicy_controller_multi_cluster_test.go
@@ -5,12 +5,14 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 	externaldns "sigs.k8s.io/external-dns/endpoint"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -28,6 +30,7 @@ import (
 var _ = Describe("DNSPolicy Multi Cluster", func() {
 
 	var gatewayClass *gatewayapiv1.GatewayClass
+	var dnsProviderSecret *corev1.Secret
 	var managedZone *kuadrantdnsv1alpha1.ManagedZone
 	var testNamespace string
 	var gateway *gatewayapiv1.Gateway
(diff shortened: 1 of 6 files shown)


controllers/dnspolicy_controller_multi_cluster_test.go
```diff
@@ -45,8 +48,21 @@ var _ = Describe("DNSPolicy Multi Cluster", func() {
...
+			err := k8sClient.Get(ctx, client.ObjectKeyFromObject(managedZone), managedZone)
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(managedZone.Status.Conditions).To(
+				ContainElement(MatchFields(IgnoreExtras, Fields{
+					"Type":               Equal(string(kuadrantdnsv1alpha1.ConditionTypeReady)),
+					"Status":             Equal(metav1.ConditionTrue),
+					"ObservedGeneration": Equal(managedZone.Generation),
+				})),
+			)
+		}, TestTimeoutMedium, time.Second).Should(Succeed())
```
mikenairn, line 65: Note: This will now expect the dns controller to be running on the target cluster. Up to now the dns policy integration tests here have not required that. If this becomes an issue we can remove the status check.

An alternative might be to only add status checks like this in the specs we want to do it in and label them as requiring a running dns operator on cluster allowing them to be excluded if needs be. However, RLP and Auth policy tests both also require operators to be running in order to create the kuadrant CR so we already have a requirement for running sub operators in the integration test suite anyway.

philbrookes (approved):

mikenairn:
/hold 

Will merge after https://github.com/Kuadrant/dns-operator/pull/142

philbrookes:
does this need an unhold? Why are the integration tests being skipped?

mikenairn:
> does this need an unhold? Why are the integration tests being skipped?

Test are doing weird things for some reason, triggered them again. Hold did nothing, prow isn't even enabled on this repo.

philbrookes (approved):

PR #659, opened 2024-05-22
Fix override RLPs on multiple gateway parents

guicassolato:
**Before:**

We are currently wrongly generating the wasm configs for cases of multiple gateway parents linked to routes, having override RateLimitPolicies targeting the gateways and an overridden route policy.

The overridden policy config ends up in the WasmPlugin and in the Limitador CR with mismatching `domain/namespace`, causing the override policy to never be enforced – in spite of the status reporting otherwise.

**After:**

This PR simplifies the configuration of overridden RateLimitPolicies.

A route RateLimitPolicy overridden at a particular gateway will no longer be added to the gateway's WasmPlugin config; rather, the override gateway policy will, with its `domain` value kept unchanged, as if all routes under the gateway were not targeted by policies.

In the Limitador CR, all RateLimitPolicies (override and overridden) will be added, even when some never get triggered (due to been overridden.)

#### Setup

```sh
make local-setup
```

```sh
kubectl apply -f - <<EOF
kind: Gateway
metadata:
  name: external-gw
  namespace: kuadrant-system
spec:
  gatewayClassName: istio
  addresses:
  - type: IPAddress
    value: "172.18.0.10"
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: api
    hostname: "*.external"
    port: 80
    protocol: HTTP
---
kind: Gateway
metadata:
  name: internal-gw
  namespace: kuadrant-system
spec:
  gatewayClassName: istio
  addresses:
  - type: IPAddress
    value: "172.18.0.11"
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
    name: api
    hostname: "*.internal"
    port: 80
    protocol: HTTP
EOF
```

```sh
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/main/examples/toystore/toystore.yaml

kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore-api
spec:
  parentRefs:
  - name: external-gw
    namespace: kuadrant-system
  - name: internal-gw
    namespace: kuadrant-system
  hostnames:
  - api.external
  - api.internal
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: "/cars"
    - path:
        type: PathPrefix
        value: "/dolls"
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
    backendRefs:
    - name: toystore
      port: 80
---
kind: HTTPRoute
metadata:
  name: toystore-www
spec:
  parentRefs:
  - name: external-gw
    namespace: kuadrant-system
  - name: internal-gw
    namespace: kuadrant-system
  hostnames:
  - www.external
  - www.internal
  rules:
  - backendRefs:
    - name: toystore
      port: 80
EOF
```

#### Send requests

Repeat the following requests as many times as needed after each state transition between creating and updating policies:

```sh
curl --resolve api.external:80:172.18.0.10 "http://api.external/cars"
```

```sh
curl --resolve api.internal:80:172.18.0.11 "http://api.internal/cars"
```

```sh
curl --resolve www.external:80:172.18.0.10 "http://www.external/cars"
```

```sh
curl --resolve www.internal:80:172.18.0.11 "http://www.internal/cars"
```

What to look at in the responses:

- _Without any policies_ → all requests should go through without any header modification or limits
- _After creating AuthPolicies_ → a special request header `authpolicy` is injected containing the name of the policy that was enforced
- _After creating RateLimitPolicies_ → both gateway policies (external: 3rp10s, internal: 5rp10s) must be enforced at both routes respectively for `*.external` and `*.internal` traffic
- _After modifying the external override gateway RateLimitPolicy to defaults_ → internal override gateway policy (5rp10s) must continue to be enforced at both routes for `*.internal` traffic; toystore api specific policy (1rp10s) must be enforced at `api.external` traffic, while external gateway override (3rp10s) continues to be enforced at the rest of `*.external` (i.e. `www.external`)

#### Create policies

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: external-gw-auth
  namespace: kuadrant-system
spec:
  targetRef:
    kind: Gateway
    name: external-gw
  overrides:
    rules:
      response:
        success:
          headers:
            "authpolicy":
              plain:
                value: external-gw-auth
---
kind: AuthPolicy
metadata:
  name: internal-gw-auth
  namespace: kuadrant-system
spec:
  targetRef:
    kind: Gateway
    name: internal-gw
  overrides:
    rules:
      response:
        success:
          headers:
            "authpolicy":
              plain:
                value: internal-gw-auth
---
kind: AuthPolicy
metadata:
  name: toystore-api-auth
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore-api
  rules:
    response:
      success:
        headers:
          "authpolicy":
            plain:
              value: toystore-api-auth
EOF
```

```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: external-gw-rl
  namespace: kuadrant-system
spec:
  targetRef:
    kind: Gateway
    name: external-gw
  overrides:
    limits:
      "external-gw-3rp10s":
        rates:
        - limit: 3
          duration: 10
          unit: second
---
kind: RateLimitPolicy
metadata:
  name: internal-gw-rl
  namespace: kuadrant-system
spec:
  targetRef:
    kind: Gateway
    name: internal-gw
  overrides:
    limits:
      "internal-gw-5rp10s":
        rates:
        - limit: 5
          duration: 10
          unit: second
---
kind: RateLimitPolicy
metadata:
  name: toystore-api-rl
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore-api
  limits:
    "toystore-api-1rp10s":
      rates:
      - limit: 1
        duration: 10
        unit: second
EOF
```

#### Change a gateway policy from `overrides` → `defaults`

```sh
kubectl apply -f - <<EOF
kind: RateLimitPolicy
metadata:
  name: external-gw-rl
  namespace: kuadrant-system
spec:
  targetRef:
    kind: Gateway
    name: external-gw
  defaults:
    limits:
      "external-gw-3rp10s":
        rates:
        - limit: 3
          duration: 10
          unit: second
EOF
```

### DIFF
--- a/controllers/rate_limiting_wasmplugin_controller.go
+++ b/controllers/rate_limiting_wasmplugin_controller.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 
 	"github.com/go-logr/logr"
+	"github.com/google/uuid"
 	istioextensionsv1alpha1 "istio.io/api/extensions/v1alpha1"
 	istioclientgoextensionv1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -61,7 +62,7 @@ type RateLimitingWASMPluginReconciler struct {
 // For more details, check Reconcile and its Result here:
 // - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
 func (r *RateLimitingWASMPluginReconciler) Reconcile(eventCtx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	logger := r.Logger().WithValues("Gateway", req.NamespacedName)
+	logger := r.Logger().WithValues("Gateway", req.NamespacedName, "request id", uuid.NewString())
 	logger.Info("Reconciling rate limiting WASMPlugin")
 	ctx := logr.NewContext(eventCtx, logger)
 
@@ -233,7 +234,10 @@ func (r *RateLimitingWASMPluginReconciler) topologyIndexesFromGateway(ctx contex
 }
 
 func (r *RateLimitingWASMPluginReconciler) wasmRateLimitPolicy(ctx context.Context, t *kuadrantgatewayapi.TopologyIndexes, rlp *kuadrantv1beta2.RateLimitPolicy, gw *gatewayapiv1.Gateway, affectedPolices []kuadrantgatewayapi.Policy) (*wasm.RateLimitPolicy, error) {
-	logger, _ := logr.FromContext(ctx)
+	// Skip the wasm config for this policy if the policy has been overridden by another policy
+	if r.overridden(ctx, rlp, affectedPolices) {
+		return nil, nil
+	}
 
 	route, err := r.routeFromRLP(ctx, t, rlp, gw)
 	if err != nil {
@@ -267,22 +271,6 @@ func (r *RateLimitingWASMPluginReconciler) wasmRateLimitPolicy(ctx context.Conte
 	routeWithEffectiveHostnames := route.DeepCopy()
 	routeWithEffectiveHostnames.Spec.Hostnames = hostnames
 
-	// Policy limits may be overridden by a gateway policy for route policies
-	if kuadrantgatewayapi.IsTargetRefHTTPRoute(rlp.GetTargetRef()) {
-		filteredPolicies := utils.Filter(affectedPolices, func(p kuadrantgatewayapi.Policy) bool {
-			return kuadrantgatewayapi.IsTargetRefGateway(p.GetTargetRef()) && p.GetUID() != rlp.GetUID()
-		})
-
-		for _, policy := range filteredPolicies {
-			p := policy.(*kuadrantv1beta2.RateLimitPolicy)
-			if p.Spec.Overrides != nil {
-				rlp.Spec.CommonSpec().Limits = p.Spec.Overrides.Limits
-				logger.V(1).Info("applying overrides from parent policy", "parentPolicy", client.ObjectKeyFromObject(p))
-				break
-			}
-		}
-	}
-
 	rules := rlptools.WasmRules(rlp, routeWithEffectiveHostnames)
 	if len(rules) == 0 {
 		// no need to add the policy if there are no rules; a rlp can return no rules if all its limits fail to match any route rule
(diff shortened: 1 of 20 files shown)


controllers/ratelimitpolicy_enforced_status_controller.go
```diff
@@ -177,7 +177,7 @@ func (r *RateLimitPolicyEnforcedStatusReconciler) hasErrCondOnSubResource(ctx co
 
 func (r *RateLimitPolicyEnforcedStatusReconciler) setCondition(ctx context.Context, logger logr.Logger, p *kuadrantv1beta2.RateLimitPolicy, conditions *[]metav1.Condition, cond metav1.Condition) error {
 	idx := utils.Index(*conditions, func(c metav1.Condition) bool {
-		return c.Type == cond.Type && c.Status == cond.Status && c.Reason == cond.Reason && c.Message == cond.Message
+		return c.Type == cond.Type && c.Status == cond.Status && c.Reason == cond.Reason
```
guicassolato, line 180: Testing the status message here was causing infinite loop of status reconciliation when a route policy is overridden by more than one gateway policy.

controllers/ratelimitpolicy_limits.go
```diff
@@ -101,112 +94,32 @@ func GetLimitador(ctx context.Context, k8sclient client.Client, rlp *kuadrantv1b
...
 	if err != nil {
 		return nil, err
 	}
 
-	rateLimitIndex := rlptools.NewRateLimitIndex()
-
-	for _, rlpKey := range rlpRefs {
-		if _, ok := rateLimitIndex.Get(rlpKey); ok {
-			continue
-		}
```
guicassolato, line 118: It looks like we were never updating the Limitador CR after the limit definitions changed in a previously reconciled RateLimitPolicy.

Also fix that.

guicassolato: Never mind. What I said is incorrect. This was actually to avoid duplication in the config. Either way, it's no longer needed.

controllers/rate_limiting_wasmplugin_controller.go
```diff
@@ -267,22 +271,6 @@ func (r *RateLimitingWASMPluginReconciler) wasmRateLimitPolicy(ctx context.Conte
...
-	// Policy limits may be overridden by a gateway policy for route policies
-	if kuadrantgatewayapi.IsTargetRefHTTPRoute(rlp.GetTargetRef()) {
-		filteredPolicies := utils.Filter(affectedPolices, func(p kuadrantgatewayapi.Policy) bool {
-			return kuadrantgatewayapi.IsTargetRefGateway(p.GetTargetRef()) && p.GetUID() != rlp.GetUID()
-		})
-
-		for _, policy := range filteredPolicies {
-			p := policy.(*kuadrantv1beta2.RateLimitPolicy)
-			if p.Spec.Overrides != nil {
-				rlp.Spec.CommonSpec().Limits = p.Spec.Overrides.Limits
```
guicassolato, line 279: Here lies the bug. We are adding the overridden policy to the config, only with the limit definitions replaced with the ones from the override. Except that this will never find a match in the Limitador CR.

guicassolato: Hmmm. A minute has passed and I'm already having second doubts about this fix...

Maybe what we're doing in the WasmPlugin is not wrong after all, but rather what we do in the Limitador CR is?

WDYT @eguzki @KevFan

guicassolato: The difference is basically if we want a single counter for all routes or one counter per route.

guicassolato: I've given this another iteration after speaking with @eguzki.

We cannot take the simpler approach of falling back to the override limit definitions, otherwise we'd be sharing the same counters between all overridden routes. Instead, we really want to duplicate limit definitions for each independent `domain`/`namespace`, even if the configuration of the counters end up being identical.

With that, the actual fix became:
1. keeping the wasm config unchanged – only minor refactoring now in this PR;
2. duplicating the entries in the limits config (Limitador CR) for each gateway for an overridden policy that affects multiple gateways

(2) required changing the hash that ensures unique limit names (conditions), both in the wasm config and Limitador CR. Another option could be adding the name of the gateway again to domain/namespace (as it once was the case long ago.)

KevFan (approved):
Verified, working as expected 🏅 Changes looks good to me 👍

controllers/ratelimitpolicy_enforced_status_controller.go
guicassolato: I've fixed this so the message is comparable too.

KevFan (approved):

Issue #559, opened 2024-04-23
Example dashboards feedback


david-martin:
- [x] #665
- [x] #666
- [x] Change the Business User dashboard heatmap to use blue shades instead of red (avoid association with it being 'bad') https://github.com/Kuadrant/kuadrant-operator/pull/660
- [x] #667
- [x] Show duplicate Gateways, Policies and Routes in an intuitive way when more than 1 exists across clusters with the same name (e.g. assuming the same name & namespace means they are 'multi cluster'). Will need a 'cluster' label in remote write config. https://github.com/Kuadrant/kuadrant-operator/pull/664
- [x] unhealthy gateway stat panel not detecting gateway with 'accepted=false' (see 2nd screenshot) https://github.com/Kuadrant/kuadrant-operator/pull/660
- [x] Fix any m:m query errors that occur when more than 1 cluster has the same resource name & namespace (see screenshot below) https://github.com/Kuadrant/kuadrant-operator/pull/637

*m:m issue with unhealthy gateways stat panel*
![image](https://github.com/Kuadrant/kuadrant-operator/assets/878251/53e9229f-715c-4aa9-9723-38183cf2956c)

*unhealthy gateway*
<img width="898" alt="image" src="https://github.com/Kuadrant/kuadrant-operator/assets/878251/1896d44c-3765-4b27-9356-680cf47009a0">

query
```
count(gatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~"${gateway_namespace}"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~"${gateway_namespace}",type="Programmed"})
or ignoring(type)
gatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~"${gateway_namespace}"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~"${gateway_namespace}",type="Accepted"}))
```
metric
```
api_gateway_status{container="kube-state-metrics", customresource_group="gateway.networking.k8s.io", customresource_kind="Gateway", customresource_version="v1beta1", exported_namespace="ingress-gateway", instance="10.128.0.231:8081", job="kube-state-metrics-kuadrant", name="prod-web", namespace="monitoring", prometheus="openshift-user-workload-monitoring/user-workload", receive="true", replica="thanos-receive-ingestor-default-0", tenant_id="default-tenant", type="Programmed"} | 0
```

[...]

PR #660, opened 2024-05-23
Dashboard feedback

david-martin:
2 fixes from #559

- unhealthy gateways query fix
image

- blue/black for histogram instead of green/red
image

### DIFF
--- a/examples/dashboards/platform_engineer.json
+++ b/examples/dashboards/platform_engineer.json
@@ -750,7 +750,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "count(gatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~\"${gateway_namespace}\"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~\"${gateway_namespace}\",type=\"Programmed\"})\nor ignoring(type)\ngatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~\"${gateway_namespace}\"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~\"${gateway_namespace}\",type=\"Accepted\"}))",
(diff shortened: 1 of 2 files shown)


david-martin:
rebased.
@R-Lawton if you have some time to review, I'd appreciate it.

R-Lawton:
👀

examples/dashboards/platform_engineer.json
```diff
@@ -750,7 +750,7 @@
           },
           "editorMode": "code",
           "exemplar": false,
-          "expr": "count(gatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~\"${gateway_namespace}\"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~\"${gateway_namespace}\",type=\"Programmed\"})\nor ignoring(type)\ngatewayapi_gateway_info unless (gatewayapi_gateway_info{exported_namespace=~\"${gateway_namespace}\"} * on(name, exported_namespace, instance) group_left gatewayapi_gateway_status{exported_namespace=~\"${gateway_namespace}\",type=\"Accepted\"}))",
```
R-Lawton, line 753: Could you explain a bit more about how the <1 is replacing the unless. I cant work it out in my head. Unless is like complement so before it was counting gatewayapi_gateway_info without gatewayapi_gateway_info(.....) so now we are counting only the timeseries we were originally not wanting

david-martin: Before, I set up the query to include all gateways (`gatewayapi_gateway_info`), unless the status of Programmed was true ,
or all gateways unless unless the status of Accepted was True,
then count all the results.

This wasn't working as expected, so i changed it look for gateways where the status of Programmed was not True (i.e. has a value of 0, which is <1). I could have used an `== 0` or `< 1` check here I think, but I wanted to make sure any value < 1 got included.

```promql
gatewayapi_gateway_status{exported_namespace=~\"${gateway_namespace}\",type="Programmed"}
    < 1
```

Same for Accepted.
The `or ignoring(type)` ensures that in case a gateway is *not* Programmed and *not* Accepted, it doesn't get included twice.
Then all that is wrapped in a count.

R-Lawton: Ahhh ok im with you now makes sense

R-Lawton (approved):

PR #677, opened 2024-05-29
Revert "use correct status type when enforcing policy"

guicassolato:
Reverts Kuadrant/kuadrant-operator#594

The Enforced status of policies that are _partially_ enforced should be `True`, not `False`.


mikenairn:
This will need removed https://github.com/Kuadrant/kuadrant-operator/pull/655/files#diff-8ca7bb2b5ec65645cb6668b37f1e07e74cb4fdbc880ded9b176b34303da44b3cR134-R146, @makslion fyi

guicassolato:
Superseded by #677

PR #678, opened 2024-05-30
envoygateway dev environment install

eguzki:
### What

Epic issue https://github.com/Kuadrant/kuadrant-operator/issues/325

Deploy kuadrant on k8s cluster powered by EnvoyGateway

```
make local-setup GATEWAYAPI_PROVIDER=envoygateway
```

Let's hit that gateway (the steps only work on linux)

```sh
INGRESS_HOST=$(kubectl get gtw eg -n envoy-gateway-system -o jsonpath='{.status.addresses[0].value}')
INGRESS_PORT=$(kubectl get gtw eg -n envoy-gateway-system -o jsonpath='{.spec.listeners[?(@.name=="http")].port}')
```

```sh
curl --verbose --resolve www.example.com:${INGRESS_PORT}:${INGRESS_HOST} http://www.example.com:${INGRESS_PORT}/get
```
The expected response is 404 (no route yet)

```
* Added www.example.com:80:172.18.0.252 to DNS cache
* Hostname www.example.com was found in DNS cache
*   Trying 172.18.0.252:80...
* Connected to www.example.com (172.18.0.252) port 80 (#0)
> GET /get HTTP/1.1
> Host: www.example.com
> User-Agent: curl/7.81.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< date: Fri, 12 Apr 2024 09:28:19 GMT
< content-length: 0
< 
* Connection #0 to host www.example.com left intact
```

> Note that the target branch of this PR is not `main`, instead, it is `envoygateway` to collect all required changes to support entirely envoygateway gatewayapi provider

### DIFF
--- a/config/dependencies/envoy-gateway/gateway/gateway.yaml
+++ b/config/dependencies/envoy-gateway/gateway/gateway.yaml
@@ -1,0 +1,14 @@
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: eg
+spec:
+  gatewayClassName: eg
+  listeners:
+    - name: http
+      protocol: HTTP
+      port: 80
+      allowedRoutes:
+        namespaces:
+          from: All
--- a/make/alerts.mk
+++ b/make/alerts.mk
@@ -1,12 +1,9 @@
 ##@ Alerts
-export WORKDIR ?= $(shell pwd)
 export IMAGE ?= quay.io/prometheus/prometheus
-export AVAILABILITY_SLO_RULES ?= $(WORKDIR)/examples/alerts/slo-availability.yaml
-export LATENCY_SLO_RULES ?= $(WORKDIR)/examples/alerts/slo-latency.yaml
-export UNIT_TEST_DIR ?= $(WORKDIR)/examples/alerts/tests
-export OS = $(shell uname | tr '[:upper:]' '[:lower:]')
-export ARCH = $(shell uname -m | tr '[:upper:]' '[:lower:]')
-export SLOTH = $(WORKDIR)/bin/sloth
+export AVAILABILITY_SLO_RULES ?= $(PROJECT_PATH)/examples/alerts/slo-availability.yaml
+export LATENCY_SLO_RULES ?= $(PROJECT_PATH)/examples/alerts/slo-latency.yaml
+export UNIT_TEST_DIR ?= $(PROJECT_PATH)/examples/alerts/tests
+export SLOTH = $(PROJECT_PATH)/bin/sloth
 export ALERTS_SLOTH_INPUT_DIR = /examples/alerts/sloth
 export ALERTS_SLOTH_OUTPUT_DIR = /examples/alerts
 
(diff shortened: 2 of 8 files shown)


config/dependencies/envoy-gateway/gateway/gateway.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: eg
```
adam-cattermole, line 5: When we eventually merge the overarching feature we might want to see if we can deploy both `istio-ingressgateway` and envoy gateway `eg` with the same name/namespace so that all of the examples/quickstarts "just work" for both.

make/envoy-gateway.mk
```diff
@@ -0,0 +1,42 @@
...
+
+## Targets to help install and configure EG
+
+EG_CONFIG_DIR = config/dependencies/envoy-gateway
+EG_NAMESPACE = envoy-gateway-system
+
+# egctl tool
+EGCTL=$(PROJECT_PATH)/bin/egctl
+EGCTL_VERSION ?= v1.0.0
+OS = linux
```
adam-cattermole, line 12: Should we get this from the system instead? `$(uname -s)` or something?

adam-cattermole (approved):
Looks good just some minor comments

config/dependencies/envoy-gateway/gateway/gateway.yaml
eguzki: yeah. we might need to pick a generic name for both. It cannot be either `istio-ingressgateway` or `eg`.

make/envoy-gateway.mk
eguzki: added automatic detection of OS and ARCH. EG has a custom bash script to do egctl installation: `https://gateway.envoyproxy.io/get-egctl.sh`. Unfortunately, currently it has some limitations, like it requires `sudo`. So, until they enhance their script, we can use our own.

eguzki:
@adam-cattermole you approved the PR, however afterwards I added the automatic detection of OS and ARCH. Additionally, I rebased to resolve existing conflicts with current HEAD of `main`. Technically, a new approval is needed. Thanks.

adam-cattermole:
Sorry had a brief look at this and forgot to get back to it.

make/envoy-gateway.mk
```diff
@@ -0,0 +1,58 @@
...
+endif
+ifneq ($(filter armv5%,$(UNAME_P)),)
+    ARCH = armv5
+endif
+ifneq ($(filter armv6%,$(UNAME_P)),)
+    ARCH = armv6
+endif
+ifneq ($(filter armv7%,$(UNAME_P)),)
+    ARCH = arm
+endif
```
adam-cattermole, line 28: Any reason not to use `uname -m` the same way it's done in the `get-egctl.sh`?

https://github.com/envoyproxy/gateway/blob/main/tools/hack/get-egctl.sh#L12-L25

On my system:
```
$ uname -p
unknown
```

eguzki: from `uname` manpage,
```
       -p, --processor
              print the processor type (non-portable)
       -s, --kernel-name
              print the kernel name
       -m, --machine
              print the machine hardware name
```

they all give `x86_64` in my machine.

using `uname (GNU coreutils) 9.4`

> I also just noticed that the ARCH var is already set in alerts.mk so it's currently getting set regardless.

good catch. I will refactor a bit.

adam-cattermole: Can't say for sure what the difference between `-p` and `-m` is in my env.
```
$ uname --version
uname (GNU coreutils) 9.5

$ uname -p
unknown

$ uname -m
x86_64
```

eguzki: done, can you try getting egctl again?

make/alerts.mk
```diff
@@ -1,12 +1,9 @@
 ##@ Alerts
-export WORKDIR ?= $(shell pwd)
 export IMAGE ?= quay.io/prometheus/prometheus
-export AVAILABILITY_SLO_RULES ?= $(WORKDIR)/examples/alerts/slo-availability.yaml
```
eguzki, line 4: @R-Lawton you are the original owner of this script. heads up as I introduced small changes.

on a different note, why exporting makefile variables? are they being used in sub-processes? If not, they are not required.

make/envoy-gateway.mk
adam-cattermole: Nice yep all looks good now!

PR #679, opened 2024-05-30
Partially Enforced sets Enforced status to true

makslion:
Redo of the change made earlier about the partially enforced state. 
Not the "Partially Enforced" will count as Enforced (will set the Enforced Status to true)

### DIFF
--- a/pkg/library/kuadrant/apimachinery_status_conditions.go
+++ b/pkg/library/kuadrant/apimachinery_status_conditions.go
@@ -107,14 +107,12 @@ func AcceptedCondition(p Policy, err error) *metav1.Condition {
 func EnforcedCondition(policy Policy, err PolicyError, allSubresourcesReady bool) *metav1.Condition {
 	// Enforced
 	message := fmt.Sprintf("%s has been successfully enforced", policy.Kind())
-	status := metav1.ConditionTrue
 	if !allSubresourcesReady {
 		message = fmt.Sprintf("%s has been partially enforced", policy.Kind())
-		status = metav1.ConditionFalse
 	}
 	cond := &metav1.Condition{
 		Type:    string(PolicyConditionEnforced),
-		Status:  status,
+		Status:  metav1.ConditionTrue,
 		Reason:  string(PolicyReasonEnforced),
 		Message: message,
 	}
--- a/controllers/dnspolicy_controller_single_cluster_test.go
+++ b/controllers/dnspolicy_controller_single_cluster_test.go
@@ -137,7 +137,7 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 				g.Expect(dnsPolicy.Status.Conditions).To(
 					ContainElement(MatchFields(IgnoreExtras, Fields{
 						"Type":    Equal(string(kuadrant.PolicyConditionEnforced)),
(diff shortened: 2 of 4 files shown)


pkg/library/kuadrant/apimachinery_status_conditions.go
```diff
@@ -110,7 +110,6 @@ func EnforcedCondition(policy Policy, err PolicyError, allSubresourcesReady bool
 	status := metav1.ConditionTrue
```
guicassolato, line 110: Do we still need the variable?

makslion:
@guicassolato I wanted to wait until checks have passed before posting it in the Slack discussion. For me, it would make sense to close #677 as we don't really want to revert the change. What was fixed in that PR is still valid - it just revealed another issue. I'm using the old issue just for tracking purposes. 
The initial issue was about not using the correct status to check subresources of the policy: we will always assume they are ready. Now we are checking it properly. 
What was discussed is related to what we mark policy if not all subresources are ready. 

A brief recap of the conversation that lead to this: 
Initially, it was decided to have `Enforced - False` on `Partially Enforced` as a way to highlight that the policy is not fine; or rather subresources that are affected by the policy are not ok. E.g. DNS Policy targeting GW with 2 listeners. We create a DNS Record per listener and one of the records is not ready due to misconfiguration. It was debated that if we mark `Enforced - true` it will mislead to the belief that all DNS records are ok. 

What was discussed in the Slack thread makes more sense tbh. With this change, we will mark the DNS Policy as not enforced only if we fail to retrieve the list of DNS Records or if there is no single DNS Record controlled by the policy and in a ready state. In all other cases the `Enforced` will always be `true`. 
The rest of the policies will be fine and will stop being `Enforced - false` once overridden.

guicassolato:
Thanks @makslion. I'll close #677 then.

pkg/library/kuadrant/apimachinery_status_conditions.go
```diff
@@ -107,14 +107,12 @@ func AcceptedCondition(p Policy, err error) *metav1.Condition {
...
 		message = fmt.Sprintf("%s has been partially enforced", policy.Kind())
-		status = metav1.ConditionFalse
 	}
 	cond := &metav1.Condition{
 		Type:    string(PolicyConditionEnforced),
-		Status:  status,
+		Status:  metav1.ConditionTrue,
 		Reason:  string(PolicyReasonEnforced),
 		Message: message,
 	}
```
mikenairn, line 118: From what i understand this was the important bit to revert from https://github.com/Kuadrant/kuadrant-operator/pull/677/files#diff-c62f06c8c31cd4ef9b0a3f839225773a42ea10c75dff00563b42b413ec425107, but we still need this https://github.com/Kuadrant/kuadrant-operator/pull/677/files#diff-06c8a42ee2952e17c9b90724b17324a7dfb178dbea00e606ce3c6eb62b3bd73eR118, so i think this is fine.

mikenairn (approved):

pkg/library/kuadrant/apimachinery_status_conditions.go
makslion: Yes, just the `Status` is the problem. The rest should stay

PR #680, opened 2024-05-30
cert manager api as dependency

eguzki:
### What

Remove cert-manager operator as dependency and, instead, add cert-manager API as dep.

On openshift, the upstream cert-manager operator can conflict with RH build of the cert manager operator.

### Verification steps

* Build custom bundle https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/development.md#build-kuadrant-operator-bundle-image
```
# manifests
make bundle VERSION=0.8.0

# bundle image
make bundle-build BUNDLE_IMG=quay.io/kuadrant/kuadrant-operator-bundle:my-bundle

# push bundle image
make bundle-push BUNDLE_IMG=quay.io/kuadrant/kuadrant-operator-bundle:my-bundle
```

* Build custom catalog https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/development.md#build-custom-catalog

````
make catalog BUNDLE_IMG=quay.io/kuadrant/kuadrant-operator-bundle:my-bundle
make catalog-build CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:my-catalog
make catalog-push CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:my-catalog
````
* Create kind cluster
```
make kind-create-cluster
```
* Deploy OLM system
```
make install-olm
```
* Deploy kuadrant operator using OLM. Since the certificate manager API is not installed, the operation should fail.
```
make deploy-catalog CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:my-catalog
```
The subscription status eventually reports `constraints not satisfiable`
```
kubectl get subscriptions kuadrant -n kuadrant-system -o jsonpath='{.status.conditions}' | yq e -P
```
Condition type `CatalogSourcesUnhealthy` should eventually report that `all available catalogsources are healthy`. It takes up to few minutes.
```yaml
- lastTransitionTime: "2024-05-30T15:31:33Z"
  message: all available catalogsources are healthy
  reason: AllCatalogSourcesHealthy
  status: "False"
  type: CatalogSourcesUnhealthy
```
OLM should have resolved the dependency by installing the cert-manager operator.

```
❯ kubectl get clusterserviceversions.operators.coreos.com | grep cert-manager
cert-manager.v1.14.2        cert-manager         1.14.2    cert-manager.v1.13.3             Succeeded
```

### DIFF
--- a/bundle/metadata/dependencies.yaml
+++ b/bundle/metadata/dependencies.yaml
@@ -11,7 +11,33 @@ dependencies:
     value:
       packageName: dns-operator
       version: "0.0.0"
-  - type: olm.package
+  - type: olm.gvk
     value:
-      packageName: cert-manager
-      version: "1.14.2"
+      group: cert-manager.io
+      kind: CertificateRequest
+      version: v1
+  - type: olm.gvk
+    value:
+      group: cert-manager.io
+      kind: Certificate
+      version: v1
+  - type: olm.gvk
+    value:
+      group: acme.cert-manager.io
+      kind: Challenge
+      version: v1
+  - type: olm.gvk
+    value:
+      group: cert-manager.io
+      kind: ClusterIssuer
+      version: v1
+  - type: olm.gvk
+    value:
+      group: cert-manager.io
+      kind: Issuer
+      version: v1
+  - type: olm.gvk
+    value:
+      group: acme.cert-manager.io
+      kind: Order
+      version: v1
--- a/bundle/metadata/dependencies.yaml
+++ b/bundle/metadata/dependencies.yaml
@@ -14,17 +14,7 @@ dependencies:
   - type: olm.gvk
     value:
       group: cert-manager.io
-      kind: CertificateRequest
-      version: v1
-  - type: olm.gvk
-    value:
-      group: cert-manager.io
       kind: Certificate
-      version: v1
-  - type: olm.gvk
-    value:
-      group: acme.cert-manager.io
-      kind: Challenge
       version: v1
   - type: olm.gvk
     value:
@@ -35,9 +25,4 @@ dependencies:
     value:
       group: cert-manager.io
       kind: Issuer
-      version: v1
-  - type: olm.gvk
-    value:
-      group: acme.cert-manager.io
-      kind: Order
       version: v1
--- a/config/deploy/olm/catalogsource.yaml
+++ b/config/deploy/olm/catalogsource.yaml
@@ -6,4 +6,6 @@ spec:
   sourceType: grpc
   image: quay.io/kuadrant/kuadrant-operator-catalog:latest
   displayName: Kuadrant Operators
+  grpcPodConfig:
+    securityContextConfig: restricted
   publisher: grpc
--- a/utils/generate-catalog.sh
+++ b/utils/generate-catalog.sh
@@ -4,11 +4,6 @@
 
 set -euo pipefail
 
-### CONSTANTS
-# Used as well in the subscription object
-DEFAULT_CHANNEL=preview
-###
-
 OPM="${1?:Error \$OPM not set. Bye}"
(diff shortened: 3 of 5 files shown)


bundle/metadata/dependencies.yaml
```diff
@@ -11,7 +11,33 @@ dependencies:
     value:
       packageName: dns-operator
       version: "0.0.0"
-  - type: olm.package
+  - type: olm.gvk
```
eguzki, line 14: According to 
https://github.com/Kuadrant/kuadrant-operator/blob/dc98cab7d131931351b32cdcf792d26442240ad0/controllers/tlspolicy_controller.go#L50-L53

```
//+kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;watch;
//+kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;watch;
//+kubebuilder:rbac:groups="cert-manager.io",resources=certificates,verbs=get;list;watch;create;update;patch;delete
```

`CertificateRequest`, `Order` and `Challenge` GVKs are not being used by the TLSPolicy controller. Should those also be added as deps? @maleck13 @mikenairn

config/deploy/olm/catalogsource.yaml
```diff
@@ -6,4 +6,6 @@ spec:
   sourceType: grpc
   image: quay.io/kuadrant/kuadrant-operator-catalog:latest
   displayName: Kuadrant Operators
+  grpcPodConfig:
+    securityContextConfig: restricted
```
didierofrivia, line 10: Is this something we need to override in OLM for cert manager to work? What does the `securityContextConfig` adds to the grpc Pod?

utils/generate-catalog.sh
```diff
@@ -4,19 +4,14 @@
...
-
 OPM="${1?:Error \$OPM not set. Bye}"
 YQ="${2?:Error \$YQ not set. Bye}"
 BUNDLE_IMG="${3?:Error \$BUNDLE_IMG not set. Bye}"
 REPLACES_VERSION="${4?:Error \$REPLACES_VERSION not set. Bye}"
 LIMITADOR_OPERATOR_BUNDLE_IMG="${5?:Error \$LIMITADOR_OPERATOR_BUNDLE_IMG not set. Bye}"
 AUTHORINO_OPERATOR_BUNDLE_IMG="${6?:Error \$AUTHORINO_OPERATOR_BUNDLE_IMG not set. Bye}"
 DNS_OPERATOR_BUNDLE_IMG="${7?:Error \$DNS_OPERATOR_BUNDLE_IMG not set. Bye}"
-CHANNELS="${8:-$DEFAULT_CHANNEL}"
+CHANNELS="${8?:Error \$CHANNELS not set. Bye}"
```
didierofrivia, line 14: No channels, no catalog. We might need to review the release process of setting channel(s) and default channels

eguzki: the default channel is set at the makefile level. The script only requires one. I have moved the default from the script to the makefile. The script needs one channel value and fails if not provided. The default value is meaningful outside the script only.

didierofrivia (approved):
Considering that we are ignoring CertificateRequest, Order and Challenge GVKs since are not being used... LGTM! 🥇

bundle/metadata/dependencies.yaml
mikenairn: I wouldn't have thought we needed to add GVKs for things this controller doesn't deal with directly, but then again I'm not totally sure what OLM expects you to do here, if anything. Why do we want to change it from specifying a known package to specifying gvks that any package could be providing?

mikenairn (approved):

PR #687, opened 2024-06-04
Update makefile to support standalone Podman

Boomatang:
Changes allow podman to be used as the standalone container engine. 

This PR introduces a new environment variable call `CONTAINER_ENGINE` which defaults to docker. The variable is used in the building of images and tells kind what engine to use.

Loading images to the kind clusters has being changed from `kind load docker-image ...` to `kind load image-archive ...` which allows supporting both docker and podman.  This does have the extra step of creating the archives for the images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ SHELL = /usr/bin/env bash -o pipefail
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 
+# Container Engine to be used for building image and with kind
+CONTAINER_ENGINE ?= docker
+
 # VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,6 +6,9 @@ SHELL = /usr/bin/env bash -o pipefail
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 PROJECT_PATH := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 
+# Container Engine to be used for building image and with kind
+CONTAINER_ENGINE ?= docker
```
eguzki, line 10: nitpick: `CONTAINER_ENGINE = docker` is enough as the variable has not been defined in the same makefile. Note that it still allows overriding it with `make target CONTAINER_ENGINE=podman`

Makefile
```diff
@@ -314,16 +317,22 @@ run: generate fmt vet ## Run a controller from your host.
...
 kind-load-image: ## Load image to local cluster
-	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
+	@mkdir -p ./tmp
+	@if [ -f ./tmp/image.tar ]; then rm ./tmp/image.tar; fi
+	$(CONTAINER_ENGINE) save -o ./tmp/image.tar $(IMG)
+	KIND_EXPERIMENTAL_PROVIDER=$(CONTAINER_ENGINE) $(KIND) load image-archive ./tmp/image.tar $(IMG) --name $(KIND_CLUSTER_NAME)
 
 kind-load-bundle: ## Load image to local cluster
-	$(KIND) load docker-image $(BUNDLE_IMG) --name $(KIND_CLUSTER_NAME)
+	@mkdir -p ./tmp
```
eguzki, line 332: AFAIK make target `kind-load-bundle` is not being used. And if it was being used, I would not DRY. instead:

```make
kind-load-bundle: ## Load image to local cluster    
    $(MAKE) kind-load-image IMG=$(BUNDLE_IMG)
```

I would remove the target if not being used.

make/development-environments.mk
```diff
@@ -49,7 +49,12 @@ namespace: ## Creates a namespace where to deploy Kuadrant Operator
 .PHONY: local-deploy
 local-deploy: ## Deploy Kuadrant Operator from the current code
 	$(MAKE) docker-build IMG=$(IMAGE_TAG_BASE):dev
-	$(KIND) load docker-image $(IMAGE_TAG_BASE):dev --name $(KIND_CLUSTER_NAME)
+
+	@mkdir -p ./tmp
```
eguzki, line 53: call `kind-load-image` ?

Makefile
```diff
@@ -314,16 +317,22 @@ run: generate fmt vet ## Run a controller from your host.
...
-	docker build -t $(IMG) .
+	$(CONTAINER_ENGINE) build -t $(IMG) .
 
 docker-push: ## Push docker image with the manager.
-	docker push $(IMG)
+	$(CONTAINER_ENGINE) push $(IMG)
 
 kind-load-image: ## Load image to local cluster
-	$(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
+	@mkdir -p ./tmp
```
eguzki, line 326: This is how I would do this. wdyt?
```make
kind-load-image: ## Load image to local cluster
       $(KIND) load docker-image $(IMG) --name $(KIND_CLUSTER_NAME)
       $(eval TMP_DIR := $(shell mktemp -d))
       docker save -o $(TMP_DIR)/image.tar $(IMG) \
               && KIND_EXPERIMENTAL_PROVIDER=$(CONTAINER_ENGINE) $(KIND) load image-archive $(TMP_DIR)/image.tar $(IMG) --name $(KIND_CLUSTER_NAME) ; \
         EXITVAL=$$? ; \
               rm -rf $(TMP_DIR) ;\
         exit $${EXITVAL}
```

eguzki (changes requested):

Makefile
Boomatang: I am going to say this does not work in manner that is useful. If you use `make target CONTAINER_ENGINE=podman` works but, at least in fish, setting the `CONTAINER_ENGINE` as a environment variable and then calling `make target` does not pass the correct value. While the `?=` does allow this to work.

eguzki (approved):

PR #695, opened 2024-06-10
Set gateway-api-state-metrics version to 0.4.0 (latest at this time) …

### LINKED ISSUE #667: Add `allowedRoutes.namespaces` to Gateways table in platform engineer dashboard view
Checking with @maleck13 on this feedback item, but I think it might be `allowedRoutes.namespaces` from https://gateway-api.sigs.k8s.io/concepts/api-overview/?h=allowedroutes#restricting-route-attachment

e.g.

```yaml
kind: Gateway
metadata:
  name: prod-web
spec:
  gatewayClassName: example
  listeners:
  - protocol: HTTP
    port: 80
    name: prod-web-gw
    allowedRoutes:
      namespaces:
        from: Same
```

david-martin:
…to pickup new field introduced in https://github.com/Kuadrant/gateway-api-state-metrics/pull/43

Fixes #667

To show this new field, I think it makes sense to have a new panel to show Gateway Listeners, including all the fields of a listener that are available via state metrics.
Here's where the new panel fits in, with a slight change to the layout of the PE dashbaord. (lot of fields, so needs scrolling)
image

And here's the panel showing all the fields when stretched out to accomodate them without scrolling (including the new 'allowed_routes_namespaces_from' field)
image

### DIFF
--- a/config/observability/kustomization.yaml
+++ b/config/observability/kustomization.yaml
@@ -3,8 +3,8 @@ kind: Kustomization
 
 resources:
   - github.com/prometheus-operator/kube-prometheus?ref=release-0.13
-  - github.com/Kuadrant/gateway-api-state-metrics?ref=main
-  - github.com/Kuadrant/gateway-api-state-metrics/config/examples/dashboards?ref=main
+  - github.com/Kuadrant/gateway-api-state-metrics?ref=0.4.0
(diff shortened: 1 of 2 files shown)


jasonmadigan (approved):
LG

PR #696, opened 2024-06-10
Use sail dep from istio-ecosystem

adam-cattermole:


adam-cattermole:
I've moved the gateway api v1.1 changes out to it's own PR and will rebase this on top #780

adam-cattermole:
This is awaiting the release of the sail-operator [v0.1 milestone](https://github.com/istio-ecosystem/sail-operator/milestone/1) to be able to consume the helm charts.

config/dependencies/istio/sail/istio.yaml
```diff
@@ -1,15 +1,12 @@
-apiVersion: operator.istio.io/v1alpha1
+apiVersion: sailoperator.io/v1alpha1
 kind: Istio
 metadata:
   name: default
 spec:
-  version: v1.20.0
+  # Supported values v1.22.4|v1.23.0
+  version: v1.22.4
```
adam-cattermole, line 7: Worth noting that this release of the sail operator only supports newer versions of Istio than the v1.20.0 that we are on for the `istioctl` install

adam-cattermole:
I think this is ready for review with the caveat that this is moving to a pre-release version of the sail-operator. If we decide to merge before the final release we'll need a follow up to update the version

adam-cattermole:
I've updated this to use the v0.1.0 release https://github.com/istio-ecosystem/sail-operator/releases/tag/0.1.0

config/dependencies/istio/sail/istio.yaml
adam-cattermole: In fact there seems to be an issue with `v1.22.4` on the 0.1.0 release which was not present in the rc. The default profile is missing. This works with `v1.23.0` however so I've updated in the latest commit

adam-cattermole:
As most of the changes here are no longer relevant I'm going to close this and open a new pulling in only the changes to install the sail-operator release and I'll test it on top of the ongoing changes in the sotw work.

PR #707, opened 2024-06-17
add metrics for DNS events

### LINKED ISSUE #453: dnspolicy: Emit a metric for Policy ready status
In order to correctly monitor the kuadrant-operator and spot if it is behaving abnormally with regards to the DNS Records, we need to emit an event whenever the kuadrant-operator:
- Creates a new DNS Record
- Deletes an existing DNS Record
- Makes changes to a DNS Record that modify it's spec.endpoints value.
- [ ] Add a metric to flag a clash in DNS record types from the DNS Operator (zone id, desired record type and current record type,  dns name)
- [ ] Add a metric that records a dnspolicy is not ready
- [ ] Add doc describing these metrics and how to use them

philbrookes:
Emit metrics for:
- DNS Policy ready state

ping @mikenairn @eguzki 

closes #453


philbrookes:
ping @mikenairn @eguzki @maleck13 could I get a review here, when you have a chance?

pkg/library/reconcilers/base_reconciler.go
```diff
@@ -146,31 +160,54 @@ func (b *BaseReconciler) ReconcileResource(ctx context.Context, obj, desired cli
...
+		eh.OnUpdate(ctx, obj, err)
+	}
+	return err
 }
 
 func (b *BaseReconciler) DeleteResource(ctx context.Context, obj client.Object, options ...client.DeleteOption) error {
 	logger, _ := logr.FromContext(ctx)
 	logger.Info("delete object", "kind", strings.Replace(fmt.Sprintf("%T", obj), "*", "", 1), "name", obj.GetName(), "namespace", obj.GetNamespace())
-	return b.Client().Delete(ctx, obj, options...)
+	if obj.GetDeletionTimestamp() != nil {
```
philbrookes, line 193: Added a check here, to not reissue delete requests on items already deleting.

mikenairn:
Since it was deemed that this was necessary for DNSPolicy, would the same not be true for all policies? Presumably this is intended to be used along some alert or dashboard that signals if CRUD operations are happening too frequently?

philbrookes:
> Since it was deemed that this was necessary for DNSPolicy, would the same not be true for all policies? Presumably this is intended to be used along some alert or dashboard that signals if CRUD operations are happening too frequently?

I think this was deemed necessary for DNS Policy as it could result in excessive traffic to the DNS Provider API, which is not a problem with the other policies?

controllers/dnspolicy_controller.go
```diff
@@ -199,5 +202,35 @@ func (r *DNSPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	r.RegisterEventHandler(&DNSMetricsEventHandler{})
 	return ctrlr.Complete(r)
 }
+
+const (
+	dnsRecordNameLabel      = "dns_record_name"
+	dnsRecordNamespaceLabel = "dns_record_namespace"
+	dnsRecordEventTypeLabel = "dns_record_event_type"
+	dnsPolicyNameLabel      = "dns_policy_name"
+	dnsPolicyNamespaceLabel = "dns_policy_namespace"
```
maleck13, line 214: this is probably not needed. The dns record is always is the same ns as the policy

philbrookes: this is for the dns policy ready metric.

controllers/dnspolicy_controller.go
```diff
@@ -199,5 +202,35 @@ func (r *DNSPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	dnsPolicyReadyLabel     = "dns_policy_ready"
+)
+
+var (
+	dnsRecordEvent = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "dns_record_event",
```
maleck13, line 221: is the naming here consistent with other metrics? I think we should prob have a naming convention for metrics we emit WDYT

controllers/dnspolicy_dnsrecords.go
```diff
@@ -20,6 +20,43 @@ import (
...
+	if err != nil {
+		return
+	}
+	dnsRecord, ok := obj.(*kuadrantdnsv1alpha1.DNSRecord)
+	if !ok {
+		return
+	}
+	dnsRecordEvent.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace, "delete").Inc()
+}
+func (m *DNSMetricsEventHandler) OnStatusUpdate(_ context.Context, _ client.Object, _ error) {
```
maleck13, line 55: what are these for?

philbrookes: just seemed like the right thing to do, if others want to add event handlers in the future, I don't currently need them for anything.

controllers/dnspolicy_controller.go
philbrookes: as far as I can tell, these are the only 2 metrics in the kuadrant-operator right now?

controllers/dnspolicy_status.go
```diff
@@ -58,13 +58,25 @@ func (r *DNSPolicyReconciler) reconcileStatus(ctx context.Context, dnsPolicy *v1
...
 }
 
+func (r *DNSPolicyReconciler) emitConditionMetrics(dnsPolicy *v1alpha1.DNSPolicy) {
+	readyStatus := meta.FindStatusCondition(dnsPolicy.Status.Conditions, ReadyConditionType)
+	if readyStatus == nil {
+		dnsPolicyReady.WithLabelValues(dnsPolicy.Name, dnsPolicy.Namespace, "false").Set(1)
+	} else {
+		dnsPolicyReady.WithLabelValues(dnsPolicy.Name, dnsPolicy.Namespace, string(readyStatus.Status)).Set(1)
+	}
+}
```
maleck13, line 78: hmm this one seems odd. Wont I end up with a metric as `false` 1 and `true` 1 ?

maleck13: wondering what I am using this for? Is if "true" == 0 for > x time then alert?

maleck13: If I have {name:"test", namesapce:"test"} 0 | 1 would we acheive the same thing?

pkg/library/reconcilers/base_reconciler.go
```diff
@@ -61,9 +70,14 @@ func NewBaseReconciler(
 		apiClientReader: apiClientReader,
 		logger:          logger,
 		recorder:        recorder,
+		eventHandlers:   []EventHandler{},
```
maleck13, line 73: Are we concerned about multiple registrations happening leading to possible multiple increments? Should we ensure uniqueness ?

controllers/dnspolicy_status.go
david-martin: Yeah, this probably wont behave the way you are thinking.
Metrics, and each combination of label values, should be thought of as existing all the time.
What changes is the value of those metric label combinations.
Check out https://prometheus.io/docs/practices/instrumentation/#avoid-missing-metrics

If you set a falsey metric to 1 when it's false, you should also set to to 0 when it's not false.
Otherwise you might end up with odd behaviour in graphs and alerts where your DNSPolicy shows up as 'not ready' and also as 'ready'

Another way to think about it is restarting your controller should result in the same metrics and values being emitted. 
If you do actually want a metric to no longer be scraped, theres [Delete](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.Delete) and other functions,

controllers/dnspolicy_controller.go
david-martin: There's a good guide on naming metrics at https://prometheus.io/docs/practices/naming/

controllers/dnspolicy_controller.go
```diff
@@ -199,5 +202,35 @@ func (r *DNSPolicyReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+			Help: "DNS Record events",
+		},
+		[]string{dnsRecordNameLabel, dnsRecordNamespaceLabel, dnsRecordEventTypeLabel})
+	dnsPolicyReady = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "dns_policy_ready",
```
david-martin, line 227: When dealing with the state of resources, I've found the patterns in the kube-state-metrics project to be quite good.
Check out the `kube_pod_status_ready` metrics (and variations) at https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md

controllers/dnspolicy_status.go
philbrookes: ok, so I need to emit 1 when ready and otherwise 0? That makes sense to me. Rereading the above I'm not actually clear on what I was thinking.

controllers/dnspolicy_dnsrecords.go
```diff
@@ -20,6 +20,43 @@ import (
...
+
+func (m *DNSMetricsEventHandler) OnCreate(_ context.Context, obj client.Object, err error) {
+	if err != nil {
+		return
+	}
+	dnsRecord, ok := obj.(*kuadrantdnsv1alpha1.DNSRecord)
+	if !ok {
+		return
+	}
+	dnsRecordEvent.WithLabelValues(dnsRecord.Name, dnsRecord.Namespace, "create").Inc()
```
david-martin, line 33: This combo of labels feels odd for create & delete.
Will the counter only be incremented 1 time ever as the record with that name & namespace combo only be created or deleted once?

Maybe a better option is to have an '_info' like metric be created/deleted (and have a value of 1) when the record is created/deleted.
See `kube_pod_info` in https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md for an example

controllers/dnspolicy_controller.go
philbrookes: I'm not sure I'm following how to reconcile this with the above conversation. I was under the impression the current approach was inferior as it has the status as a label (so then, what is the gauge gauging), but the kube_pod_status_ready has the status as a label. Am I missing something? 

Or, maybe an alternative question: When does the `kube_pod_status_ready` gauge increment?

controllers/dnspolicy_dnsrecords.go
philbrookes: I was under the impression that doing this would allow a query like: "show me how many dns record created events there are every 5 seconds for the last day", so it would highlight for example, 2 hours ago there was a huge jump in create and delete events.

pkg/library/reconcilers/base_reconciler.go
philbrookes: I am not convinced this is a good approach.
1) If you're registering the same event handler multiple times (e.g. registration during a reconcile loop), you have a need to refactor your code, and this misbehaving would highlight that - where preventing multiple registrations would mask the problem
2) It's not impossible to want multiple of the same type of event handler, but that are configured differently

controllers/dnspolicy_dnsrecords.go
david-martin: Because of the name & namespace label being included, you'll end up with a new time series for each dnsrecord.
Each series will never increment beyond 1.
For example, if you create 3 new records (record1, record2, record3), you'll get 3 new time series like this:

```
dns_record_event{name="record1",namespace="whatever"} 1
dns_record_event{name="record2",namespace="whatever"} 1
dns_record_event{name="record3",namespace="whatever"} 1
```

controllers/dnspolicy_controller.go
david-martin: Maybe an example set of metrics will help understand the pattern used by kube-state-metrics.
Here's a set for 1 pod:

```
kube_pod_status_ready{condition="false", namespace="demo-petstore", pod="petstore-6f58b6dd67-b2ncj"} 0
kube_pod_status_ready{condition="true", namespace="demo-petstore", pod="petstore-6f58b6dd67-b2ncj"} 1
kube_pod_status_ready{condition="unknown", namespace="demo-petstore", pod="petstore-6f58b6dd67-b2ncj"} 0
```
From this current state, we can say that the `petstore-6f58b6dd67-b2ncj` pod Ready condition is currently 'true'.
A time series exists for each condition, and the guage value changes to 0 to 1 depending on if that ready condition is the active one.

philbrookes: oh I see, so when it goes:
```
ready: false
- emit with label condition false: 1
- emit with label condition true: 0
- emit with label condition unknown: 0
ready: true
- emit with label condition false: 0
- emit with label condition true: 1
- emit with label condition unknown: 0
ready: absent
- emit with label condition false: 0
- emit with label condition true: 0
- emit with label condition unknown: 1
```

controllers/dnspolicy_dnsrecords.go
philbrookes: wouldn't it be:
```
dns_record_event{name="record1",namespace="whatever", event="create"} 1
dns_record_event{name="record2",namespace="whatever", event="create"} 1
dns_record_event{name="record3",namespace="whatever", event="create"} 1
```

david-martin: Yes.
Is this what you want?

From my understanding on how time series data is stored in prometheus, this would mean there's a new time series with the new label combination each time a new record is created and scraped for the first time. 
If you just want a count on creates, a single time series may be sufficient where the values are what's saved each scrape.

More info in https://www.robustperception.io/cardinality-is-key/

philbrookes: I suppose what I want is 2 things:
- be able to see a spike in creates or deletes
  - Determine if it's lots of unique records, or the same record deleting and recreating over and over
- see a spike in updates
  - Identify which records are updating a lot.

If you have recommendations on how I could best formulate this metric to achieve, I'm happy to follow any advice!

david-martin: I wonder if you can satisfy those use cases sufficiently with existing controller runtime metrics https://book.kubebuilder.io/reference/metrics-reference
Here's a screenshot from a dev cluster showing the rate of reconciles, grouped by controller (so, by resource).
image
and the query
```
sum(rate(controller_runtime_reconcile_total{job="dns-operator-controller-manager-metrics-service"}[5m])) by(controller)
```
If the rate of reconciles is "too high", an alert could fire.
You'd have to jump to logs to get higher resolution data on what's causing the spike.
Alertnatively, you go the route you are thinking and have that higher resolution data (dnsrecord names & namespaces) as labels on the time series data and have users take the potential storage hit.

There may be other useful labels in those controller runtime metrics as well, like `result` to show if the reconciles are mostly successes or errors. It might be worth exploring them if you haven't already.
Here's a screenshot of what that kind of grouping might look like
image

philbrookes: ok, then I think it makes sense to remove these event metrics for now, which also significantly simplifies this PR, and as you say, use the controller_runtime metrics to identify the spike in activity is coming from Kuadrant-operator, and use logs to work out what it's doing.

That sounds like a feasible solution to me, which also allows me to remove all the event handler code from the base_reconciler.

maleck13 (approved):
THis looks a much simpler solution
/lgtm

PR #708, opened 2024-06-18
444 add a gh action for deploying and verifying dashboards alerts load ok with quickstart

ehearneRedHat:
### What:

GitHub Action `Verify Dashboards and Alerts OK` created to test whether dashboards and alerts are deployed successfully.

### How:

Within the `Verify Dashboard and Alerts OK` workflow, it spins up the infrastructure needs for self-hosted runner via Terraform, then it runs `/hack/quickstart-setup.sh` first to run kuadrant-operator and in particular loads in the `Observability Stack` which contains our dashboards via Grafana, and our alerts via Prometheus:

- It exposes the grafana service endpoint in the kind cluster and performs HTTP GET requests via Grafana's local API to ensure it can detect the dashboards in `examples/dashboards/` . Once the requests are carried out, it kills the process ID associated with grafana service endpoint exposure. 
- It performs a very similar process with alerts in `examples/alerts/` via Prometheus. 

### Verify:

1. Create a branch referencing the commits from `444-add-a-gh-action-for-deploying-and-verifying-dashboards-alerts-load-ok-with-quickstart` using the origin `https://www.github.com/Kuadrant/kuadrant-operator`. 
2. Comment out the triggers in `.github/workflows/verify-dashboard-alerts.yaml` so that it gets triggered everytime there is a push. Then, commit and push the changes to observe the workflow. If the workflow runs quickstart, performs HTTP requests, it did its job correctly.
3. Revert the changes made and delete the branch.


.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,385 @@
...
+      run:
+        shell: bash -ieo pipefail {0}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run Quickstart
+        run: |
+          sudo chown $(whoami) .
+          kind delete clusters kuadrant-local
+          kind delete clusters kuadrant-local
```
david-martin, line 164: typo on duplicate cmd?

david-martin:
> the quickstart script is unable to set up networking related components of kuadrant-operator

What components cannot be set up using a built in runner?

self-hosted-runner.tf
```diff
@@ -0,0 +1,125 @@
...
+  key_name = var.aws_key_name
+  
+  tags = {
+    Name = "kuadrant-operator-self-hosted-runner"
+  }
+
+  // Security Group for SSH, HTTP, and HTTPS access
+  security_groups = ["ssh-http-https-access"]
+
+  # Uncomment when creating an AMI .
```
david-martin, line 39: When should a new AMI be created?

self-hosted-runner.tf
```diff
@@ -0,0 +1,125 @@
...
+  type        = string
+}
+
+variable "aws_key_name" {
+  description = "AWS Key Name"
+  type = string
+}
+
+resource "aws_instance" "self_hosted_runner" {
+  ami           = "ami-055032149717ffb30" # change to ami-0776c814353b4814d when creating an AMI.
```
david-martin, line 23: What's the reason for changing the ami when creating a new one?

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,385 @@
...
+          cat << 'EOF' > automate_setup.expect
+          #!/usr/bin/expect -f
+          set timeout -1
+          spawn bash hack/quickstart-setup.sh
+          expect "Are you ready to begin? (y/n)"
+          send "y\r"
+          expect "Do you want to set up a DNS provider for use with Kuadrant's DNSPolicy API? (y/n)"
+          send "n\r"
+          expect "Thank you for using Kuadrant! If you have any questions or feedback, please reach out to our community."
+          EOF
```
david-martin, line 177: Being able to run the quickstart script to completion is a very useful thing, outside of this specific workflow.
I'm not suggesting any changes at this time, but I'm wondering if you have given any thought on how this might be reused for setting up a CI test environment for use by different acceptance tests (like these dashboard & alerts tests). I'd imagine an acceptance test around a gateway and policies would be very useful to have in a workflow too.

david-martin:
Thanks for the awesome work on this @ehearneRedHat 
It definitely adds a lot of value and I'm inclined to see it merged soon.
I'd like to understand a few things a little better first though (questions inline) so we know how best to maintain the workflows being added.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: I found that running the command just once often proceeded in the container environment unable to delete, proceeding with an error. By performing this action twice, even if it fails the first time, it can delete the second time. I'm unsure as to why this behavior exists... Even if it succeeds the first time, it doesn't throw an error and proceeds. :)

self-hosted-runner.tf
ehearneRedHat: The new AMI should be created when dependencies are updated, such as kind, kubernetes, and podman. :)

self-hosted-runner.tf
ehearneRedHat: The commented AMI ID is a ubuntu AMI. This image allows us to have a clean environment to set up our dependencies so we can use the newly created AMI later on to save time on setting up our testing environment.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: It did cross my mind as to how it could be a universal way to run quickstart. I didn't implement it that way for this PR as I didn't know how such a huge PR would be perceived.  :D 

I think it definitely makes sense to have an implementation that can be called for multiple CI/CD pipelines. :)

ehearneRedHat:
> What components cannot be set up using a built in runner?

When running the quickstart in a GitHub hosted runner I am met with plenty of errors pertaining to missing network libraries, and even when these libraries are installed, because the quickstart requests to run a container in privileged mode, it cannot, as the runner itself is not privileged.

Therefore in order to run the network components, the runner had to be self hosted, and to avoid costs surrounding a self hosted runner powerful enough to run all the components within the quickstart, Terraform was used as keeping such a runner up for a prolonged period of time would be quite costly.

If you have any further questions please let me know. :D 

> Thanks for the awesome work on this @ehearneRedHat . 

No problem, @david-martin ! It was great to have had the opportunity to learn so much about this side of runners. :)

ehearneRedHat:
@pehala @jsmolar @david-martin 

New build of `ubuntu-latest` now seems to run quickstart script just fine in GitHub hosted runner. Therefore, there is no longer a need to run workflow in self-hosted runner. 

Therefore, verification steps have been updated, and redundant files have now been removed.

.github/workflows/verify-dashboards-alerts.yaml
pehala: My suggestion would be to add `quiet` flag feature to quickstart itself and the ability to specify everything through CLI args. Of course, this would be different issue and PR, but I think it would be worth it to implement that

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+      run:
+        shell: bash -eo pipefail {0}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run Quickstart
+        run: |
+          sudo chown $(whoami) .
+
+          sudo apt-get update -y
```
pehala, line 29: I would add separate step to setup dependencies

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+    runs-on: ubuntu-latest
+    defaults: 
+      run:
+        shell: bash -eo pipefail {0}
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run Quickstart
+        run: |
+          sudo chown $(whoami) .
```
pehala, line 27: Why is this needed?

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+          send "n\r"
+          expect "Thank you for using Kuadrant! If you have any questions or feedback, please reach out to our community."
+          EOF
+          chmod +x automate_setup.expect
+          ./automate_setup.expect
+
+      - name: Port forward grafana and check if contains dashboards.
+        run: |
+          # Port forward Grafana
+          kubectl -n monitoring wait --for=condition=available deployment grafana --timeout=600s
```
pehala, line 50: I would split Grafana setup to a separate step to keep tests and setup separate

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Run Quickstart
+        run: |
+          sudo chown $(whoami) .
+
+          sudo apt-get update -y
+          sudo apt-get install -y golang expect
+
+          export ISTIO_INSTALL_SAIL=true
+
+          cat << 'EOF' > automate_setup.expect
```
pehala, line 34: Does those tests even require quickstart? You are only checking presence of both alerts and dashboads, you are never actually using Kuadrant

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: Port forward Prometheus and check if contains alert rules.
+        run: |
+          kubectl -n monitoring wait --for=condition=ready pod prometheus-k8s-0 --timeout=600s
+          # Port forward Prometheus
+          kubectl -n monitoring port-forward service/prometheus-k8s 9090:9090 &
+          prometheus_process_id=$!
+          echo "Successfully port forwarded Prometheus service."
+
+          sleep 1
```
pehala, line 102: Why? If it is really needed, are we sure sleep 1 is always enough? Isnt there a better way to check?

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+      - name: Port forward grafana and check if contains dashboards.
+        run: |
+          # Port forward Grafana
+          kubectl -n monitoring wait --for=condition=available deployment grafana --timeout=600s
+          kubectl -n monitoring port-forward service/grafana 3000:3000 &
+          grafana_process_id=$!
+          echo "Successfully port forwarded Grafana service."
+
+          sleep 1
```
pehala, line 55: Why? If it is really needed, are we sure sleep 1 is always enough? Isnt there a better way to check?

.github/workflows/verify-dashboards-alerts.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+          echo "Grafana contains dashboards $app_developer, $business_user and $platform_engineer. Continuing to Prometheus..."
+
+          # Close Grafana port-forward.
+          kill $grafana_process_id
+          echo "Stoppped port forwarding Grafana."
+
+      - name: Port forward Prometheus and check if contains alert rules.
+        run: |
+          kubectl -n monitoring wait --for=condition=ready pod prometheus-k8s-0 --timeout=600s
```
pehala, line 96: I would separete setup and test phases

pehala (commented):
It looks much better now! Thanks for updating it

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: @pehala I think this makes sense. Would you prefer it this PR waited for those changes ? Therefore, I could implement it directly ?

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: It seemed that when I didn't use istio flag that the mgc folder being brought in had perm issues. With the new istio flag, this isn't needed at all. Nice catch !

ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/actions/runs/9837193514/job/27154462311

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: @pehala I guess it is a simple way of being able to run grafana and prometheus. You are right that there are several components not in use.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: It is there to give the system a chance before performing the request. Even though the process completed, when I didn't include it gave me endpoint errors.

There may be a possibility that sleep 1 may not be enough on a particular run. However, from the extensive workflow runs I have ran over the past few weeks, I haven't seen a case where it hasn't been enough time. 

There could be a better way to check... for example, pinging the endpoint until the expected response is met, and then performing the following HTTP GET requests. I'll implement now.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/pull/708#discussion_r1668332829

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: Makes sense. Will implement now.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: Seems to run fine without sleep when jobs are separated.

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/pull/708#discussion_r1668397947

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: I have determined that it is possible to just set up the observability stack on its own, meaning only the required components would be ran. It would mean no dependency on the quickstart script and the runner-generated expect script. Implementing now...

ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/actions/runs/9838597648/job/27158825631

.github/workflows/verify-dashboards-alerts.yaml
ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/pull/708#discussion_r1668422907

ehearneRedHat: Marking as resolved as new changes make waiting for quickstart changes redundant.

ehearneRedHat: https://github.com/Kuadrant/kuadrant-operator/issues/743

pehala (approved):

ehearneRedHat:
:tada: :smile:

david-martin (approved):
Thanks for taking the time to address all the comments and feedback @ehearneRedHat 
This solves the original problem in a standard way that should be relatively easy to maintain.

david-martin:
Merging with known test flake unrelated to the changes in this PR (see https://kubernetes.slack.com/archives/C05J0D0V525/p1720453954262009)

Issue #712, opened 2024-06-24
Auth controller to DAG 1.0


Boomatang:
Ensure Auth controller uses that DAG as expaected

guicassolato:
Superseded by #478

PR #713, opened 2024-06-26
Auth controller refactor to use DAG

Boomatang:
closes: #712 

This should be a drop in replacement. No new features have being added. It is expected to work as before.

### DIFF
--- a/controllers/authpolicy_authconfig.go
+++ b/controllers/authpolicy_authconfig.go
@@ -17,10 +17,8 @@ import (
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 	gatewatapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 
-	"github.com/kuadrant/kuadrant-operator/api/v1beta2"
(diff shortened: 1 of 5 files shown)


controllers/authpolicy_authconfig.go
```diff
@@ -235,6 +232,55 @@ func getAffectedPolicies(t *kuadrantgatewayapi.Topology, ap *api.AuthPolicy) []k
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err = json.Unmarshal(data, out); err != nil {
+		return nil, err
+	}
+
+	return out, nil
+}
```
KevFan, line 247: Is there a reason to do a marhsall/unmarshall to convert the type to `*api.AuthPolicy` instead of type casting since it implements `kuadrantgatewayapi.Policy`? Seems like an more expensive operation compared to type casting 🤔

Boomatang: Turned out my issue with casting was due to missing the `*` in the type. Now that is fixed I removed this function and put the casting inline. Thanks for calling this out.

Boomatang:
Close this as it will move to the policy machinery soon

PR #716, opened 2024-06-26
chore: Remove DNSRecord ownerID generation

mikenairn:
The `ownerID` field on DNSRecord is now optional, the dns operator will ensure a valid unique ownerID for all DNSRecords if none is explicitly set.

Requires: https://github.com/Kuadrant/dns-operator/pull/170

### DIFF
--- a/tests/common/dnspolicy/dnspolicy_controller_test.go
+++ b/tests/common/dnspolicy/dnspolicy_controller_test.go
@@ -541,10 +541,10 @@ var _ = Describe("DNSPolicy controller", func() {
 							"Type":    Equal(string(kuadrant.PolicyConditionEnforced)),
 							"Status":  Equal(metav1.ConditionTrue),
 							"Reason":  Equal(string(kuadrant.PolicyReasonEnforced)),
-							"Message": Equal("DNSPolicy has been partially enforced"),
+							"Message": Equal("DNSPolicy has been successfully enforced"),
 						})),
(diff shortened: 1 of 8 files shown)


tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -541,10 +541,10 @@ var _ = Describe("DNSPolicy controller", func() {
 							"Type":    Equal(string(kuadrant.PolicyConditionEnforced)),
 							"Status":  Equal(metav1.ConditionTrue),
 							"Reason":  Equal(string(kuadrant.PolicyReasonEnforced)),
-							"Message": Equal("DNSPolicy has been partially enforced"),
+							"Message": Equal("DNSPolicy has been successfully enforced"),
 						})),
 				)
-			}, tests.TimeoutMedium, time.Second).Should(Succeed())
+			}, tests.TimeoutLong, time.Second).Should(Succeed())
```
mikenairn, line 547: @makslion This test was sometimes failing. My guess is that previously one of the DNSRecords was never becoming ready due to https://github.com/Kuadrant/dns-operator/issues/165 so it was consistently `partially enforced`. Now though they can both become ready so we should just wait for it to be fully enforced. Presumably that should happen relatively quickly in the test setup.

maleck13 (approved):
/lgtm

PR #722, opened 2024-06-28
Refactor patrially enforced condition for dnspolicy

### LINKED ISSUE #586: Extend the Partially Enforced status on DNSPolicy

### LINKED ISSUE #685: Set Enforced to false when DNSPolicy does not succeed for all listener
When a DNSPolicy is attached to a Gateway with a single listener and corresponding HTTPRoute is created, DNSRecord for the only listener fails like in the example. In this case when DNSPolicy fails for all listeners the condition should be `Enforced: false`. Now it is `Enforced: true` with message that the policy is partially enforced which it is not.

```
status:
  conditions:
    - lastTransitionTime: '2024-06-03T12:21:42Z'
      message: DNSPolicy has been accepted
      reason: Accepted
      status: 'True'
      type: Accepted
    - lastTransitionTime: '2024-06-03T12:21:42Z'
      message: DNSPolicy has been partially enforced
      reason: Enforced
      status: 'True'
      type: Enforced
  observedGeneration: 1
  recordConditions:
    '*.tju51nm88888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888888.example.com':
      - lastTransitionTime: '2024-06-03T12:21:42Z'
        message: 'The DNS provider failed to ensure the record: Failure in zone example.com. [Id: /hostedzone/ZONEID] when submitting change: InvalidChangeBatch: FATAL problem: DomainLabelTooLong (Domain label is too long) encountered with ''tju51nm888888888888888888888888888888888888888888888888888888888'''
        observedGeneration: 1
        reason: ProviderError
        status: 'False'
        type: Ready
```

makslion:
closes #586 
fixes #685  

Changing the enforced logic around DNSPolicy. 
Now it works as one would expect. Also updated relevant integration tests. 
Added a new test to cover all possible `Enforced` conditions. It is arguable if we need such a hefty test, that is, more-less, a copy of an already existent one. My reasoning is that unexpected `Enforced` behaviour caused a few bugs and I'd rather have tests run an extra 20-ish seconds but be confident in the scope of the test.

### DIFF
--- a/controllers/dnspolicy_status.go
+++ b/controllers/dnspolicy_status.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"errors"
+	"fmt"
 	"slices"
 	"strings"
 
@@ -34,6 +35,7 @@ import (
 
 	"github.com/kuadrant/kuadrant-operator/api/v1alpha1"
 	"github.com/kuadrant/kuadrant-operator/pkg/library/kuadrant"
+	"github.com/kuadrant/kuadrant-operator/pkg/library/utils"
 )
 
 var NegativePolarityConditions []string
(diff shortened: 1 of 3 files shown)


tests/common/dnspolicy/dnspolicy_controller_single_cluster_test.go
```diff
@@ -97,12 +96,6 @@ var _ = Describe("DNSPolicy Single Cluster", func() {
 				},
 			}
 			gateway.Status.Listeners = []gatewayapiv1.ListenerStatus{
-				{
-					Name:           "foo",
-					SupportedKinds: []gatewayapiv1.RouteGroupKind{},
-					AttachedRoutes: 0,
-					Conditions:     []metav1.Condition{},
-				},
```
mikenairn, line 105: This was added for a reason https://github.com/Kuadrant/kuadrant-operator/pull/655/files#r1611332705

makslion: It has no effect now. No listeners -> no DNS Records created -> no effect on the status of the policy as long as there are other records. It was `partially enforced` for the wrong reason.

makslion: There is now a different [test](https://github.com/Kuadrant/kuadrant-operator/pull/722/files#diff-76eb28eef00ca4741889476a78254e8513eb9407be3a095fcee80a360f06a58fR403-R407) for the `PartiallyEnforced` case

mikenairn: This was nothing to do with enforced. It was ensuring that we had a listener with 0 attached routes which before that PR was broken. Adding that was ensuring we had covered that scenario in our tests so we didn't break it again. i.e if i reverted [this](https://github.com/Kuadrant/kuadrant-operator/pull/655/files#diff-ad34d851ae1426ca68b2179d94856acd67b5af507fcd3b3ff275cd8121e53db6R93) for some reason, the tests would fail.

makslion: Hm, i did forget about that bug. Brought it back. Now it will be checking, at least, for the correct status

controllers/dnspolicy_status.go
```diff
@@ -82,55 +84,63 @@ func (r *DNSPolicyReconciler) calculateStatus(ctx context.Context, dnsPolicy *v1
...
+		return kuadrant.EnforcedCondition(dnsPolicy, kuadrant.NewErrUnknown(dnsPolicy.Kind(), errors.New("policy is not enforced on any DNSRecord: no routes attached for listeners")), false)
+	}
+
+	// filter not ready records
+	notReadyRecords := utils.Filter(recordsList.Items, func(record kuadrantdnsv1alpha1.DNSRecord) bool {
+		for _, condition := range record.Status.Conditions {
+			if condition.Type == string(kuadrantdnsv1alpha1.ConditionTypeReady) && condition.Status == metav1.ConditionFalse {
+				return true
 			}
 		}
```
mikenairn, line 123:
```suggestion
		if meta.IsStatusConditionFalse(record.Status.Conditions, string(kuadrantdnsv1alpha1.ConditionTypeReady)) {
			return true
		}
```

Any reason not to do this?

mikenairn (commented):
Couple of comments/questions but i think the states of the enforced condition looks good as i understand it:

Policy + No DNSRecords = Enforced false
Policy + DNSRecords all ready = Enforced true
Policy + DNSRecords none ready = Enforced false
Policy + DNSRecord, some ready = Partially Enforced 

It does however look like it's basing these decisions on just listing DNSRecord resources that are owned by the policy, is this enough? Should we not be checking the DNSRecord for a corresponding listener exists and is ready? What would happen in the case where we have two listeners but for some reason a record wasn't being created for one of them? In this case would the Enforced condition be set to true even though there is a record missing?

[TLS](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/tlspolicy_status.go#L144) does something different here, i don't think its fully right either btw as it seems to do it only for one certificate, but it gets the list of expected certificates and then tries to find them, if one is missing it's going to cause it be not enforced.

tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -320,84 +160,270 @@ var _ = Describe("DNSPolicy controller", func() {
 			}, tests.TimeoutMedium, time.Second).Should(Succeed())
 		}, testTimeOut)
 
-		It("should have accepted condition with status true", func(ctx SpecContext) {
-			By("creating a valid Gateway")
+		It("should conflict DNS Policies of different strategy on the same host", func(ctx SpecContext) {
```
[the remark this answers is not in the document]
makslion: good point. Moved back

controllers/dnspolicy_status.go
makslion: good suggestion. Changed

makslion:
> Couple of comments/questions but i think the states of the enforced condition looks good as i understand it:
> 
> Policy + No DNSRecords = Enforced false Policy + DNSRecords all ready = Enforced true Policy + DNSRecords none ready = Enforced false Policy + DNSRecord, some ready = Partially Enforced
> 
> It does however look like it's basing these decisions on just listing DNSRecord resources that are owned by the policy, is this enough? Should we not be checking the DNSRecord for a corresponding listener exists and is ready? What would happen in the case where we have two listeners but for some reason a record wasn't being created for one of them? In this case would the Enforced condition be set to true even though there is a record missing?
> 
> [TLS](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/tlspolicy_status.go#L144) does something different here, i don't think its fully right either btw as it seems to do it only for one certificate, but it gets the list of expected certificates and then tries to find them, if one is missing it's going to cause it be not enforced.

It is a good point - what we consider to be an "Enforced" DNS Policy. When [initially](https://github.com/Kuadrant/kuadrant-operator/pull/449) making the change the ide was that as soon as DNS Record is published to the provider the policy is enforced. 
Also one could create a situation when we don't have DNSRecord CR but it was never removed from MZ - de facto policy is enforced since there are records in the provider but the controller will never know about it as it only checks for the presence of the CR. 
I'm not sure what is the best way to approach it. Would it be feasible to create an issue to address this problem @mikenairn ?

mikenairn:
> I'm not sure what is the best way to approach it. Would it be feasible to create an issue to address this problem @mikenairn ?

The current approach wasn't started in this PR so yes we can create a follow on issue to look into changing that behaviour. Can you create that follow on issue before closing this issue out?

mikenairn (approved):

makslion:
I've created #792 as a follow-up @mikenairn

PR #727, opened 2024-07-03
remove cert-manager as dependency

eguzki:
### What

While this older https://github.com/Kuadrant/kuadrant-operator/pull/680 change was removing cert-manager operator as dependency and, instead, add cert-manager API as dependency, this PR is completely removing cert-manager as dependency for the operator. 

Additionally, TLS policy status reports with some understandable error when the cert-manager is not available in the cluster.

### Why

The cert-manager is, in fact, a direct dependency of the kuadrant controller (regarding the TLS policy) and indirectly for authorino (required for the webhook deployment). Unfortunately, the dependency declaration of the cert-manager, either as operator or as API, creates unsolvable (from our side) issues in Openshift. 

And the issue comes from the [RH build of the cert-manager operator ](https://github.com/openshift/cert-manager-operator). This operator, only available for openshift, provides the same APIs as the upstream community version of the [cert-manager operator](https://github.com/cert-manager/cert-manager). That overlapping APIs, was an issue when kuadrant had dependency on the operator. When the openshift cluster had the RH build of the cert-manager operator installed, the kuadrant requirement could not be met as upstream cert-manager was unable to install. On https://github.com/Kuadrant/kuadrant-operator/pull/680 we fixed this issue changing the dependency type from operator to API level. Unfortunately, there are still issues. And the main issue is that The RH build of the cert-manager cannot be installed in the same namespace as the kuadrant operator. The reason being that the kuadrant operator is cluster wide and the RH cert-manager is namespace scoped operator. The [OperatorGroup](https://olm.operatorframework.io/docs/concepts/crds/operatorgroup/) does not enable installing different scoped operators in the same namespace. 

Even if the RH cert manager is installed in another NS A, when installing kuadrant in some NS K, OLM tries to install yet another RH cert-manager operator instance in the NS K. Which fails for the reasons explained above. 

Extracted from the installplan resource:

```yaml
clusterServiceVersionNames:             
- dns-operator.v0.0.0                   
- limitador-operator.v0.9.0             
- cert-manager-operator.v1.13.1         
- authorino-operator.v0.11.1            
- kuadrant-operator.v0.8.0 
```

My guess (just a guess) is that OLM thinks that even if the GVK exists, nobody is watching for it in the specified namespace, hence it tries to install an operator that fulfill the needs in the specified namespace. 

Interesting to note that even if the community cert-manager is installed before Kuadrant, the RH cert-manager will still try to be installed in the same namespace as kuadrant. Not sure OLM is behaving as it should. We might have found a bug here. If we re-add cert-manager APIs as dependencies, we need to verify that when the upstream operator is pre-installed, the RH build operator is not being installed  in openshift.

> read public slack channel discussion about this topic [here](https://kubernetes.slack.com/archives/C05J0D0V525/p1718604400142519)

### The Plan

The plan now is to have the cert-manager API as documentation requirement. The pre-requisite for kuadrant to work as expected is that the cert-manager APIs are in place in the cluster. 

The RH build of the cert-manager is planning cluster wide scope feature. Tracked here: https://github.com/openshift/cert-manager-operator/pull/188. 

Once a release of RH cert-manager that supports all namespaces is available, we can move back to having it as a dependency. One test we will need to do after bringing the cert-manager dependency back: verify that when the upstream operator is pre-installed, the RH build operator is not being installed  in openshift. Created issue for this https://github.com/Kuadrant/kuadrant-operator/issues/729

### Verification steps

* Create kind cluster
```
make kind-create-cluster
```

* Deploy OLM system
```
make install-olm
```

* Deploy Gateway API CRDs (for these verification steps no need to install istio)
```
make gateway-api-install
```
* Deploy kuadrant operator using OLM. The catalog has already been created for you out of the changes of this branch. But feel free to [create your own catalog](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/development.md#build-custom-olm-catalog).
```
make deploy-catalog CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:fix-cert-manager-deps
```

The operators should be installed correctly. Wait until the CSV reports all the operators succeeded.
```
❯ kubectl get clusterserviceversions.operators.coreos.com 
NAME                        DISPLAY              VERSION   REPLACES                         PHASE
authorino-operator.v0.0.0   Authorino Operator   0.0.0                                      Succeeded
dns-operator.v0.0.0         DNS Operator         0.0.0                                      Succeeded
kuadrant-operator.v0.0.0    Kuadrant Operator    0.0.0     kuadrant-operator.v0.0.0-alpha   Succeeded
limitador-operator.v0.0.0   Limitador            0.0.0                                      Succeeded
```

This is the key change of this PR: the **cert-manager** has **NOT** been installed

Let's test the TLSPolicy on this cert-manager-less scenario:

* Create a namespace:
```shell
kubectl create namespace my-gateways
```

* Create an ingress gateway

```sh
kubectl -n my-gateways apply -f - <<EOF
kind: Gateway
metadata:
  name: prod-web
spec:
  gatewayClassName: istio
  listeners:
    - allowedRoutes:
        namespaces:
          from: All
      name: api
      hostname: "*.toystore.local"
      port: 443
      protocol: HTTPS
      tls:
        mode: Terminate
        certificateRefs:
          - name: toystore-local-tls
            kind: Secret
EOF
```
* Create a Kuadrant `TLSPolicy` to configure TLS:
```sh
kubectl apply -n my-gateways -f - <<EOF
kind: TLSPolicy
metadata:
  name: prod-web
spec:
  targetRef:
    name: prod-web
    kind: Gateway
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: selfsigned-issuer
EOF
```

The TLSPolicy controller should be disabled with the following error log

```
❯ k logs deployment/kuadrant-operator-controller-manager -n kuadrant-system | grep TLSPolicy
{"level":"error","ts":"2024-07-15T09:38:19Z","logger":"kuadrant-operator.tlspolicy","msg":"CertManager CRD was not installed","group":"cert-manager.io","kind":"Certificate","version":"v1","stacktrace":"github.com/kuadrant/kuadrant-operator/pkg/library/gatewayapi.IsCertManagerInstalled\n\t/workspace/pkg/library/gatewayapi/utils.go:162\ngithub.com/kuadrant/kuadrant-operator/controllers.(*TLSPolicyReconciler).SetupWithManager\n\t/workspace/controllers/tlspolicy_controller.go:205\nmain.main\n\t/workspace/main.go:212\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:267"}
{"level":"info","ts":"2024-07-15T09:38:19Z","logger":"kuadrant-operator.tlspolicy","msg":"TLSPolicy controller disabled. CertManager was not found"}
```

The TLSPolicy instance status should be empty as the controller is not enabled. There is an open issue https://github.com/Kuadrant/kuadrant-operator/issues/730 for adding some understandable error when the cert-manager is not available in the cluster.

### DIFF
--- a/doc/install/install-openshift.md
+++ b/doc/install/install-openshift.md
@@ -5,7 +5,7 @@ NOTE: You must perform these steps on each OpenShift cluster that you want to us
 ## Prerequisites
 
 - OpenShift Container Platform 4.14.x or later with community Operator catalog available.
-- AWS account with Route 53 and zone. 
+- AWS account with Route 53 and zone.
 - Accessible Redis instance.
 
 
@@ -27,9 +27,27 @@ Before you can use Kuadrant, you must install Gateway API v1 as follows:
 kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml
 ```
 
-### Step 3 - Install and configure Istio with the Sail Operator
+### Step 3 - Install cert-manager
 
-Kuadrant integrates with Istio as a Gateway API provider. You can set up an Istio-based Gateway API provider by using the Sail Operator. 
+Before you can use Kuadrant, you must install cert-manager.
+
+Install one of the different flavours of the Cert-Manager.
+
+#### Install community version of the cert-manager
+
+Consider [installing cert-manager via OperatorHub](https://cert-manager.io/docs/installation/operator-lifecycle-manager/),
+which you can do from the OpenShift web console.
+
+More installation options at [cert-manager.io](https://cert-manager.io/docs/installation/)
+
+#### Install cert-manager Operator for Red Hat OpenShift
+
+You can install the [cert-manager Operator for Red Hat OpenShift](https://docs.openshift.com/container-platform/4.16/security/cert_manager_operator/cert-manager-operator-install.html)
+by using the web console.
+
+### Step 4 - Install and configure Istio with the Sail Operator
+
+Kuadrant integrates with Istio as a Gateway API provider. You can set up an Istio-based Gateway API provider by using the Sail Operator.
 
 #### Install Istio
 
@@ -46,9 +64,9 @@ apiVersion: operators.coreos.com/v1
 metadata:
   name: sail
   namespace: istio-system
-spec: 
-  upgradeStrategy: Default  
----  
+spec:
+  upgradeStrategy: Default
+---
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
@@ -97,11 +115,11 @@ Wait for Istio to be ready as follows:
 kubectl wait istio/default -n istio-system --for="condition=Ready=true"
 ```
 
-### Step 4 - Optional: Configure observability and metrics
+### Step 5 - Optional: Configure observability and metrics
 
 Kuadrant provides a set of example dashboards that use known metrics exported by Kuadrant and Gateway components to provide insight into different components of your APIs and Gateways. While not essential, it is best to set up an OpenShift monitoring stack. This section provides links to OpenShift and Thanos documentation on configuring monitoring and metrics storage.
 
-You can set up user-facing monitoring by following the steps in the OpenShift documentation on [configuring the monitoring stack](https://docs.openshift.com/container-platform/latest/observability/monitoring/configuring-the-monitoring-stack.html). 
+You can set up user-facing monitoring by following the steps in the OpenShift documentation on [configuring the monitoring stack](https://docs.openshift.com/container-platform/latest/observability/monitoring/configuring-the-monitoring-stack.html).
 
 If you have user workload monitoring enabled, it is best to configure remote writes to a central storage system such as Thanos:
 
@@ -128,7 +146,7 @@ If you have Grafana installed in your cluster, you can import the [example dashb
 For example installation details, see [installing Grafana on OpenShift](https://cloud.redhat.com/experts/o11y/ocp-grafana/). When installed, you must add your Thanos instance as a data source to Grafana. Alternatively, if you are using only the user workload monitoring stack in your OpenShift cluster, and not writing metrics to an external Thanos instance, you can [set up a data source to the thanos-querier route in the OpenShift cluster](https://docs.openshift.com/container-platform/4.15/observability/monitoring/accessing-third-party-monitoring-apis.html#accessing-metrics-from-outside-cluster_accessing-monitoring-apis-by-using-the-cli).
 
 
-### Step 5 - Create secrets for your credentials 
+### Step 6 - Create secrets for your credentials
 
 Before installing the Kuadrant Operator, you must enter the following commands to set up secrets that you will use later:
 
@@ -153,8 +171,8 @@ spec:
   updateStrategy:
     registryPoll:
       interval: 45m
-EOF      
-```      
+EOF
+```
 
 #### AWS Route 53 credentials for TLS
 
@@ -173,8 +191,8 @@ Set the Redis credentials for shared multicluster counters for the Kuadrant Limi
 
 ```bash
 kubectl -n kuadrant-system create secret generic redis-config \
-  --from-literal=URL=$REDIS_URL  
-```  
+  --from-literal=URL=$REDIS_URL
+```
 
 #### AWS Route 53 credentials for DNS
 
@@ -189,9 +207,9 @@ kubectl -n ingress-gateway create secret generic aws-credentials \
   --type=kuadrant.io/aws \
   --from-literal=AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
   --from-literal=AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY
-```  
+```
 
-### Step 6 - Install the Kuadrant Operator 
+### Step 7 - Install the Kuadrant Operator
 
 To install the Kuadrant Operator, enter the following command:
 
@@ -214,10 +232,10 @@ apiVersion: operators.coreos.com/v1
 metadata:
   name: kuadrant
   namespace: kuadrant-system
-spec: 
+spec:
   upgradeStrategy: Default
 EOF
-```  
+```
 
 Wait for the Kuadrant Operators to be installed as follows:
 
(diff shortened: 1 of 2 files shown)


maleck13:
This makes sense to me @eguzki I guess once a release that supports all namespaces is available we can move back to having it as a dependency?

maleck13 (approved):
/lgtm 

@mikenairn might be worth a look over also?

mikenairn:
Seems fine, but should we not consider just adding this as a pattern for all polices that have a dependency on CRDs not provided by the kuadrant operator i.e. all of them? I know we control the other operators currently required, and therefore less likely to be an issue,  but a consistent message on any policy that can't work due to missing CRDs might be beneficial no? Presumably this could be an issue for these operators also if they happened to be installed on cluster already, at least it seems we can't be sure what OLM is going to do.

eguzki:
> This makes sense to me @eguzki I guess once a release that supports all namespaces is available we can move back to having it as a dependency?

yes @maleck13 ! That's the plan. I missed adding this to the description. (Actually, I did but was lost in the ether). Updating the description, thanks

eguzki:
Added verification steps. I like doing that, only just for the record and future reference.

eguzki:
Once this PR is merged, @maleck13 can you add the new "documentation requirement"? Or give me a hint about where should I add that.

azgabur:
I can verify this fixes the automatic installation of RH cert-manager on Openshift. 
I think this PR should include adding installation of community cert-manager to the quickstart-setup.sh script so everything works like before on Kind. 
As well as updating documentation (https://docs.kuadrant.io/dev/kuadrant-operator/doc/install/install-openshift/) with the extra step of installing cert-manager manually. But this can be done in separate PR.

mikenairn (approved):
Changes look fine, but as pointed out in chat, some of the changes here might be an issue https://github.com/Kuadrant/kuadrant-operator/pull/715

eguzki:
> Changes look fine, but as pointed out in chat, some of the changes here might be an issue #715

Indeed, if the TLS policy controller starts watching not exiting types, the controller needs to be disabled, and therefore the status would not be populated. I proposed an independent common status controller to fulfill that need in the following issue https://github.com/Kuadrant/kuadrant-operator/issues/730

jasonmadigan:
makes sense

> but a consistent message on any policy that can't work due to missing CRDs might be beneficial no?

Could be done as some other issue, but I think so yes; if someone installs our operator with some kube distro with a UI, and doesn't install stuff that put other CRDs we need for certain policies in place, we don't really surface that well right now. easy for users to end up with a broken install, with little idea as to why.

maleck13:
I think it should probably go in here https://docs.kuadrant.io/0.8.0/kuadrant-operator/doc/install/install-openshift/

doc/install/install-openshift.md
```diff
@@ -243,9 +261,9 @@ spec:
     storage:
       redis-cached:
         configSecretRef:
-          name: redis-config 
-EOF
```
eguzki, line 247: My IDE does not like trailing spaces :) And me neither.

eguzki:
@maleck13 kindly asking for review regarding the last commit about the update on the openshift installation doc.

doc/install/install-openshift.md
```diff
@@ -27,9 +27,27 @@ Before you can use Kuadrant, you must install Gateway API v1 as follows:
...
 
-### Step 3 - Install and configure Istio with the Sail Operator
+### Step 3 - Install cert-manager
 
-Kuadrant integrates with Istio as a Gateway API provider. You can set up an Istio-based Gateway API provider by using the Sail Operator. 
+Before you can use Kuadrant, you must install cert-manager.
+
+Install one of the different flavours of the Cert-Manager.
+
+#### Install community version of the cert-manager
```
maleck13, line 36: do we want to specify a version or min version?

maleck13:
Looks good. Minor comment but not critical

doc/install/install-openshift.md
eguzki: We did not specify version or require min version until now. 

Is there really a min version of cert-manager that kuadrant supports? @mikenairn ?

eguzki: Actually, we did when the dependency was based on the operator, not on the API: 

From https://github.com/Kuadrant/kuadrant-operator/pull/680/files#diff-14e7feda952796eb377d3d14ab302dccca03b212ce86cfddea587f95a8351cebL16-L17

```yaml
- type: olm.package
    value:
      packageName: cert-manager
      version: "1.14.2"
```

Adding that.

eguzki:
rebased to fix conflicts with current `main` and added cert-manager min version to doc.

Please review at your leisure @mikenairn @maleck13

doc/install/install-openshift.md
mikenairn: The version of cert manager we are currently testing with, at least from this repo, is "1.12.1" https://github.com/Kuadrant/kuadrant-operator/blob/main/config/dependencies/cert-manager/kustomization.yaml#L2. Not sure what min version we support really, but I'd probably just say that version.

controllers/tlspolicy_status.go
```diff
@@ -87,6 +88,18 @@ func (r *TLSPolicyReconciler) calculateStatus(ctx context.Context, tlsPolicy *v1
...
+		return kuadrant.AcceptedCondition(tlsPolicy, err)
+	}
+	if !ok {
+		return kuadrant.AcceptedCondition(tlsPolicy, errors.New("cert-manager APIs not available"))
+	}
+
+	return kuadrant.AcceptedCondition(tlsPolicy, specErr)
+}
+
 func (r *TLSPolicyReconciler) enforcedCondition(ctx context.Context, tlsPolicy *v1alpha1.TLSPolicy, targetNetworkObject client.Object) *metav1.Condition {
```
mikenairn, line 103: Is this relevant now since https://github.com/Kuadrant/kuadrant-operator/pull/715. The controller no longer gets started if the cert manager CRDs are missing so should never get here.

didierofrivia (approved):
This is actually the way we've chosen to go with the Helm Charts too. Same issue with namespacing is affecting Helm Chart dependencies, so it'd be explicitly stated as a dependency in the docs.

doc/install/install-openshift.md
eguzki: done

controllers/tlspolicy_status.go
eguzki: yeah, reverting the changes done for the status reporting as they are useless. Pending tasks for a good reporting captured in https://github.com/Kuadrant/kuadrant-operator/issues/730

eguzki: done

eguzki:
@mikenairn some changes were reverted. The status logic does not change.

mikenairn (approved):

PR #737, opened 2024-07-05
Envoygateway authpolicy

adam-cattermole:


make/integration-tests.mk
```diff
@@ -45,7 +45,7 @@ test-gatewayapi-env-integration: clean-cov generate fmt vet ginkgo ## Requires k
 test-istio-env-integration: clean-cov generate fmt vet ginkgo ## Requires kubernetes cluster with GatewayAPI and Istio installed.
 	mkdir -p $(PROJECT_PATH)/coverage/istio-integration
 #	Check `ginkgo help run` for command line options. For example to filtering tests.
-	$(GINKGO) \
+	GATEWAYAPI_PROVIDER=istio $(GINKGO) \
```
eguzki, line 48: just wondering if `ldflags` would be a better alternative. Ginkgo supports passing same flags as `go test`.

Something like 

```
ginkgo -ldflags "-X 'github.com/kuadrant/kuadrant-operator/tests.GatewayClassName=istio'" 
```

pkg/library/mappers/envoysecuritypolicy_to_kuadrant.go
```diff
@@ -0,0 +1,53 @@
...
+
+type SecurityPolicyToKuadrantEventMapper struct {
+	opts MapperOptions
+}
+
+func NewSecurityPolicyToKuadrantEventMapper(o ...MapperOption) *SecurityPolicyToKuadrantEventMapper {
+	return &SecurityPolicyToKuadrantEventMapper{opts: Apply(o...)}
+}
+
+func (m *SecurityPolicyToKuadrantEventMapper) Map(ctx context.Context, obj client.Object) []reconcile.Request {
```
eguzki, line 22: For simplicity, I would just list the kuadrant CRs and return namespaced name of it. 

Other approach, would be to follow the topology. If the secpolicy is targeting a route, get the gateways and one of the gateways should have kuadrant reference in the annotations. If the sec policy is targeting a gateway, the gateway should have the kuadrant CR reference in the annotations.

controllers/envoysecuritypolicy_referencegrant_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		logger.V(1).Info(string(jsonData))
+	}
+
+	rg, err := r.securityPolicyReferenceGrant(ctx, kObj.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.ReconcileResource(ctx, &gatewayapiv1beta1.ReferenceGrant{}, rg, kuadrantenvoygateway.SecurityPolicyReferenceGrantMutator); err != nil && !apierrors.IsAlreadyExists(err) {
```
eguzki, line 67: Maybe catch conflic error. But "alreadyExists" error can happen?

controllers/envoysecuritypolicy_referencegrant_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	securityPolicyToKuadrantEventMapper := mappers.NewSecurityPolicyToKuadrantEventMapper(
+		mappers.WithLogger(r.Logger().WithName("securityPolicyToKuadrantEventMapper")),
+		mappers.WithClient(r.Client()),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kuadrantv1beta1.Kuadrant{}).
+		Owns(&gatewayapiv1beta1.ReferenceGrant{}).
```
eguzki, line 154: I cannot see the ownerref being set to the referencegrant

controllers/authpolicy_envoysecuritypolicy_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	kuadrantNamespace, err := kuadrant.GetKuadrantNamespace(gw)
+	if err != nil {
+		logger.Error(err, "failed to get kuadrant namespace")
+		return ctrl.Result{}, err
+	}
+
+	// reconcile security policies for gateways
+	for _, gwNode := range topology.Gateways() {
```
eguzki, line 74: according the reconciler setup, there is only one gw. No need to iterate. Is it needed?

controllers/authpolicy_envoysecuritypolicy_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		mappers.WithLogger(r.Logger().WithName("httpRouteToParentGatewaysEventMapper")),
+	)
+	apToParentGatewaysEventMapper := mappers.NewPolicyToParentGatewaysEventMapper(
+		mappers.WithLogger(r.Logger().WithName("authpolicyToParentGatewaysEventMapper")),
+		mappers.WithClient(r.Client()),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&gatewayapiv1.Gateway{}).
+		Owns(&egv1alpha1.SecurityPolicy{}).
```
eguzki, line 180: missing ownerref on the secpolicy?

controllers/authpolicy_istio_authorizationpolicy_controller.go
```diff
@@ -169,6 +185,48 @@ func (r *AuthPolicyReconciler) istioAuthorizationPolicy(ctx context.Context, ap
...
+	)
+
+	apToParentGatewaysEventMapper := mappers.NewPolicyToParentGatewaysEventMapper(
+		mappers.WithLogger(r.Logger().WithName("authPolicyToParentGatewaysEventMapper")),
+		mappers.WithClient(r.Client()),
+	)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&gatewayapiv1.Gateway{}).
+		Owns(&istiov1beta1.AuthorizationPolicy{}).
```
eguzki, line 218: missing ownerref?

eguzki (changes requested):
Looking good. Dropped some comments

.github/workflows/test.yaml
```diff
@@ -89,7 +90,7 @@ jobs:
           make env-setup GATEWAYAPI_PROVIDER=${{ matrix.gatewayapi-provider }} ISTIO_INSTALL_SAIL=${{ matrix.istio-type == 'sail' && true || false }}
       - name: Run integration tests
         run: |
-          make test-integration
+          make test-integration GATEWAYAPI_PROVIDER=${{ matrix.gatewayapi-provider }}
```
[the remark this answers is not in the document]
eguzki: now I see that the gateways class needs to be set :+1:

pkg/library/mappers/envoysecuritypolicy_to_kuadrant.go
adam-cattermole: Wouldn't this trigger our controller for updates to SecurityPolicies in the cluster which are not owned by kuadrant? We should at least check that the security policy is managed by us, and if so then it's simple enough to get the namespace of that kuadrant instance?

controllers/authpolicy_envoysecuritypolicy_controller.go
adam-cattermole: I iterated to handle case of 0 or 1 gateway instead of checking length > 0 before taking first index but I could do that instead if you prefer. I thought this way looked slightly cleaner

make/integration-tests.mk
adam-cattermole: This is an interesting suggestion, I'm not sure which one I prefer. I guess in the code instead of retrieving / checking for the presence of the env var we would check that the `GatewayClassName` is set. The only thing I can think is that the env var would be easier if you're ever running the ginkgo command directly but I assume we would always be using the Makefile

controllers/authpolicy_envoysecuritypolicy_controller.go
adam-cattermole: I'm a little unsure who should be the owner of the security policies here, I realise that given the format of this controller block it suggests it should be the gateway. Given that there is one sec policy per auth policy target the owner seems more likely to be the auth policy, or the target of the policy.

If the owner is the httproute/gateway that is targeted this actually solves one of the issues I'm facing regarding deletion (currently the sec policy remains if there's still an auth policy but the target of the policy is deleted). Does this break the model though if I set the owner of sec policies on route nodes to the route node within the controller for gateways?

make/integration-tests.mk
```diff
@@ -9,7 +9,7 @@ INTEGRATION_TEST_NUM_PROCESSES ?= 10
 test-bare-k8s-integration: clean-cov generate fmt vet ginkgo ## Requires only bare kubernetes cluster.
 	mkdir -p $(PROJECT_PATH)/coverage/bare-k8s-integration
 #	Check `ginkgo help run` for command line options. For example to filtering tests.
-	$(GINKGO) \
+	GATEWAYAPI_PROVIDER=$(GATEWAYAPI_PROVIDER) $(GINKGO) \
```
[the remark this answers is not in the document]
adam-cattermole: Removed

make/integration-tests.mk
```diff
@@ -27,7 +27,7 @@ test-bare-k8s-integration: clean-cov generate fmt vet ginkgo ## Requires only ba
 test-gatewayapi-env-integration: clean-cov generate fmt vet ginkgo ## Requires kubernetes cluster with GatewayAPI installed.
 	mkdir -p $(PROJECT_PATH)/coverage/gatewayapi-integration
 #	Check `ginkgo help run` for command line options. For example to filtering tests.
-	$(GINKGO) \
+	GATEWAYAPI_PROVIDER=$(GATEWAYAPI_PROVIDER) $(GINKGO) \
```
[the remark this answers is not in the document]
adam-cattermole: Removed

controllers/envoysecuritypolicy_referencegrant_controller.go
adam-cattermole: Fixed this

controllers/authpolicy_istio_authorizationpolicy_controller.go
adam-cattermole: Fixed this

controllers/envoysecuritypolicy_referencegrant_controller.go
adam-cattermole: I think this only ignores the error from where there's a race between the `Get` of the reference grant returning no reference grant, and it being created in the cluster before we call `Create`? But I guess in this case maybe we should error and try again

.github/workflows/test.yaml
```diff
@@ -192,8 +193,12 @@ jobs:
           fail_ci_if_error: false
           verbose: true
 
-  istio-integration-tests:
-    name: Integration Tests for github.com/kuadrant/kuadrant-operator/tests/istio
+  gatewayapi-provider-integration-tests:
```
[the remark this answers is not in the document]
eguzki: +1 to omit `github.com/kuadrant`

make/integration-tests.mk
eguzki: no strong opinion here. It is about having the value set at compile time vs runtime.

controllers/authpolicy_envoysecuritypolicy_controller.go
adam-cattermole: For now I've set the owner as the target gateway/httproute (which has allowed me to enable the deletion integration tests)

pkg/library/mappers/envoysecuritypolicy_to_kuadrant.go
eguzki: You are right, LGTM about this mapper. We can leverage the annotation `kuadrant.io/namespace`.

controllers/envoysecuritypolicy_referencegrant_controller.go
eguzki: But there is no race, is it? The only controller creating this refgrant is this very controller. So, if it does not exist, it will create, and if it exists, call the mutator to update if needed. I do not see how "alreadyExists" error might happen. I might be missing something, though.

.github/workflows/test.yaml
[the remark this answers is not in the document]
adam-cattermole: I've made this change and changed `<gatewayapi-provider>` to `[gatewayapi-provider]`

Will now be:
- `Integration Tests for kuadrant-operator/tests/[gatewayapi-provider] (envoygateway)`
- `Integration Tests for kuadrant-operator/tests/[gatewayapi-provider] (istioctl)`

I think this looks slightly better now

controllers/authpolicy_envoysecuritypolicy_controller.go
eguzki: It should always be 1, but this is good enough for now.

adam-cattermole: Right yep I see, only if the gateway is not programmed will it be 0

adam-cattermole:
@eguzki and I discussed offline the idea of changing the controller to be written in terms of `Kuadrants` rather than for gateways, and instead iterate over the policies within the topology (with the new changes from main). We then can set the owner ref as the auth policy to handle deletion of the security policies. I've pushed a change for this in the latest commit (888f81d) but I still see two issues:

1. The topology is built in terms of the gateways and routes that are affected by which policies. This means however that if a `SecurityPolicy` exists for an `AuthPolicy`, for which the target (gateway/route) is deleted, the policy is filtered from the topology and so there is not a straightforward way to delete the `SecurityPolicy` (hence the failing tests)
2. The use of the owner ref as the `AuthPolicy` is assuming that there is one `AuthPolicy` per target. There is actually only one `SecurityPolicy` per target object which is why I set the owner like this before. If we have multiple `AuthPolicies` targeting the same object - which one should be the "owner", and when the "owner" is deleted the `SecurityPolicy` will be removed even though there is still an `AuthPolicy` targeting the object.

In other words, I'm not sure the change in the latest commit (888f81d) actually solves any problems and is any more of an ideal solution than the previous version

controllers/authpolicy_controller.go
```diff
@@ -157,10 +156,6 @@ func (r *AuthPolicyReconciler) reconcileResources(ctx context.Context, ap *api.A
 		return err
 	}
 
-	if err := r.reconcileIstioAuthorizationPolicies(ctx, ap, targetNetworkObject, gatewayDiffObj); err != nil {
```
guicassolato, line 160: This will cause the AuthPolicy `Enforced` status condition to no longer be tied to the status of the internal resource that implements the call to ext authz, Istio AuthorizationPolicy in this case. We need some other way to do that now with the concurrent reconciliation of this resource, and the same for Envoy SecurityPolicies.

controllers/authpolicy_envoysecuritypolicy_controller.go
```diff
@@ -0,0 +1,208 @@
...
+			Labels: map[string]string{
+				kuadrant.KuadrantNamespaceAnnotation: kuadrantNamespace,
+			},
+		},
+		Spec: egv1alpha1.SecurityPolicySpec{
+			PolicyTargetReferences: egv1alpha1.PolicyTargetReferences{
+				TargetRefs: []gatewayapiv1alpha2.LocalPolicyTargetReferenceWithSectionName{
+					{
+						LocalPolicyTargetReference: gatewayapiv1alpha2.LocalPolicyTargetReference{
+							Group: gatewayapiv1.Group(targetNetworkObjectGvk.Group),
```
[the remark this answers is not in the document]
adam-cattermole: Yep, the security policies at the route level "override" (replace) any defined at the gateway level. Initially I intended to check for these cases and to for example only create at the gateway level if there are any children that do not have a security policy

controllers/authpolicy_controller.go
eguzki: There is one specific controller for the istio AuthorizationPolicy and one specific controller for the EnvoyGateway SecPolicy resources. Current functionality for Istio should remain exactly the same (if not improved).

eguzki (approved):

PR #741, opened 2024-07-08
adding gh actions for promq tests if alerts directory is modified

R-Lawton:


.github/workflows/test-alerts.yaml
```diff
@@ -0,0 +1,31 @@
+name: Test
```
david-martin, line 1: Could we use a more descriptive name for this?

```suggestion
name: Unit test example alerts
```

R-Lawton: This is the directory name for a want of a word for it, it follows the same format for the other but I'm happy to change it

david-martin: I couldn't find where this references a directory, but it does seem to set the Action name in the Github Actions UI at 
https://github.com/Kuadrant/kuadrant-operator/actions/workflows/test-alerts.yaml
image

It's the same name as another action/workflow, https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/test.yaml#L1, so we should probably name it better to make it more obvious what the action is for

david-martin: This is helpful
https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-workflow.json#L864-L867

R-Lawton: workflow sorry yes i couldnt think of the word to describe it. Are you saying update the test workflow for both the files that have that workflow? Or just this one. I feel like it make sense to have it under the same workflow as the other tests so all the tests have the same prefix but if readability is hard then sure let me update the naming for it

david-martin: I see what you mean here now in the checklist

image

Once 2 actions don't show up with the same name in the Actions list, I'm generally fine with where these tests ultimately end up.

PR #710, opened 2024-06-20
Kubernetes install guide

jasonmadigan:
Our single/multi-cluster quickstart scripts install Kuadrant in a kind cluster.

We have a standalone OpenShift install guide, but we should also have a Kubernetes one.
Screenshot 2024-06-20 at 13 16 08

TODO: an associated docs.kuadrant.io change update to include Kubernetes in the Installation block in the nav.

PR #742, opened 2024-07-08
Update guide to allow for different prereq cluster setup

david-martin:
related to changes in #710 that introduce a separate prereq guide for installing on an already existing kube cluster (rather than local kind clusters)


jasonmadigan (approved):
Makes sense, LGTM

Issue #743, opened 2024-07-08
`quiet` flag to allow for uninterrupted run of quickstart script


ehearneRedHat:
### What:

`quiet` flag to be added to `hack\quickstart-setup.sh` such that it will allow for an uninterrupted run of the script. 

This would mean deciding on whether we wish for the script to continue, if a DNS provider is to be used, and if multicluster setup should be allowed/denied by default. Also, deciding on what the use case for the `quiet` flag would be, such as for development or CI/CD , as it would give us a better picture of how we wish to implement. 

### How:

When implemented, the user could pass the `quiet` flag when running the script e.g. `quiet=true ./hack/quickstart-setup.sh` and the user wouldn't have to provide any other input to allow the script to run to completion.

ehearneRedHat:
https://github.com/Kuadrant/docs.kuadrant.io/issues/130

ehearneRedHat:
Closing as #744 is now closed .

PR #744, opened 2024-07-09
Quiet flag for uninterrupted single cluster run of quickstart script

ehearneRedHat:
### What: 
`KUADRANT_QUIET`, `DNS_PROVIDER` and `MULTICLUSTER` flag detection for running quickstart script uninterrupted has been added.

### How: 
It checks for whether `KUADRANT_QUIET` exists and if so, will run an uninterrupted run of quickstart in single cluster and no DNS provider mode. 

If `DNS_PROVIDER` is set alongside `KUADRANT_QUIET`, then it will run the quickstart script in single cluster mode, with dns provider. 

If `MULTICLUSTER` is set alongside `KUADRANT_QUIET` or `DNS_PROVIDER`, or both, it will run in multicluster mode with or without dns provider . 

### Why: 
Developers or System Administrators may find this useful so they do not need to use prompts to spin up Kuadrant quickly, such as for testing purposes in CI/CD pipelines. 

### Verify:

Run:
- `KUADRANT_QUIET=<whatever> ./hack/quickstart-setup.sh` 
- `KUADRANT_QUIET=<whatever> DNS_PROVIDER=<whatever> ./hack/quickstart-setup.sh`
- `KUADRANT_QUIET=<whatever> DNS_PROVIDER=<whatever> MULTICLUSTER=<whatever> ./hack/quickstart-setup.sh`
- `KUADRANT_QUIET=<whatever> MULTICLUSTER=<whatever> ./hack/quickstart-setup.sh`

Of course, other combinations exist for those env vars. The idea is that all these combinations run successfully and without error to make the script more automated.

### DIFF
--- a/hack/quickstart-setup.sh
+++ b/hack/quickstart-setup.sh
@@ -351,7 +351,7 @@ postSetup() {
   esac
 }
 
-if [[ -z "$QUIET" ]]; then
+if [[ -z "$KUADRANT_QUIET" ]]; then
   info "📘 Welcome to the Kuadrant Quick Start setup process"
 
   info "This script will guide you through setting up a local Kubernetes cluster with the following components:"
@@ -368,7 +368,7 @@ if [[ -z "$QUIET" ]]; then
   read -r -p "Are you ready to begin? (y/n) " yn </dev/tty
 else 
   yn="y"
-  info "QUIET flag detected. Proceeding without prompts in single cluster mode and without DNS Provider..."
+  info "KUADRANT_QUIET flag detected. Proceeding without prompts in single cluster mode and without DNS Provider..."
 fi
 
 case $yn in
@@ -392,7 +392,7 @@ check_dependencies
 
 info "Checking for existing Kubernetes clusters..."
 if cluster_exists "${KUADRANT_CLUSTER_NAME}"; then
-    if [[ -z "$QUIET" ]]; then
+    if [[ -z "$KUADRANT_QUIET" ]]; then
       echo "A cluster named '${KUADRANT_CLUSTER_NAME}' already exists."
       echo "This will be treated as a 'hub' cluster, with any new clusters being workers."
       read -r -p "Proceed with multi-cluster setup? (y/N): " proceed </dev/tty
@@ -422,14 +422,14 @@ if cluster_exists "${KUADRANT_CLUSTER_NAME}"; then
           exit 0
       fi
     else 
-      error "Aborting as single cluster already exists. Please delete the cluster first before running quickstart with QUIET flag."
+      error "Aborting as single cluster already exists. Please delete the cluster first before running quickstart with KUADRANT_QUIET flag."
       exit 0
     fi
 else
     info "No existing cluster named '${KUADRANT_CLUSTER_NAME}' found. Proceeding with initial setup."
 fi
 
-if [[ -z "$QUIET" ]]; then
+if [[ -z "$KUADRANT_QUIET" ]]; then
   echo "Do you want to set up a DNS provider for use with Kuadrant's DNSPolicy API? (y/n)"
   read -r SETUP_PROVIDER </dev/tty
 else
--- a/hack/quickstart-setup.sh
+++ b/hack/quickstart-setup.sh
@@ -351,21 +351,25 @@ postSetup() {
   esac
 }
 
-info "📘 Welcome to the Kuadrant Quick Start setup process"
+if [[ -z "$QUIET" ]]; then
+  info "📘 Welcome to the Kuadrant Quick Start setup process"
 
-info "This script will guide you through setting up a local Kubernetes cluster with the following components:"
-info "  - Docker or Podman (Container Runtime)"
-info "  - kind (Kubernetes IN Docker)"
-info "  - Kuadrant and its dependencies, including:"
-info "      * Gateway API"
-info "      * Istio"
-info "      * Cert-Manager"
-info "      * MetalLB"
-info "  - Optional DNS provider setup for Kuadrant's DNSPolicy API"
+  info "This script will guide you through setting up a local Kubernetes cluster with the following components:"
+  info "  - Docker or Podman (Container Runtime)"
+  info "  - kind (Kubernetes IN Docker)"
+  info "  - Kuadrant and its dependencies, including:"
+  info "      * Gateway API"
+  info "      * Istio"
+  info "      * Cert-Manager"
+  info "      * MetalLB"
+  info "  - Optional DNS provider setup for Kuadrant's DNSPolicy API"
 
-info "Please ensure you have an internet connection and local admin access to perform installations."
-
-read -r -p "Are you ready to begin? (y/n) " yn </dev/tty
+  info "Please ensure you have an internet connection and local admin access to perform installations."
+  read -r -p "Are you ready to begin? (y/n) " yn </dev/tty
+else 
+  yn="y"
+  info "QUIET flag detected. Proceeding without prompts in single cluster mode and without DNS Provider..."
+fi
 
 case $yn in
 [Yy]*)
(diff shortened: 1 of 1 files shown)


ehearneRedHat:
https://github.com/Kuadrant/kuadrant-operator/issues/743

david-martin:
I like the idea of the QUIET flag.
I wonder though if we can make it less restrictive on what it does when it's in quiet mode?

What about env vars for each of the prompts? For example, if DNS_PROVIDER is set, skip the prompt for that.

R-Lawton:
Having a quiet flag raised the question in my head is the quickstart to noisy to begin with and should we remove some extra prompts or combine prompts ? For example remove the are you ready prompt and combine the 2 DNS provider ones together so by inputting a provider that's a yes and have no also as a option

david-martin:
> For example remove the are you ready prompt and combine the 2 DNS provider ones together so by inputting a provider that's a yes and have no also as a option

sounds reasonable to make it easier for users and to simplify the script.

Boomatang:
I have thoughts on this and in general I think this is solving the symptom of the problem but not the problem. 

Firstly I think the flag name of `QUIET` is the wrong name. This seems to be more to do with running in single cluster mode than being quiet. Secondly I generally really dislike using environment variables with generic names in tooling that is expected to run on someone else's machine. Who is to say they do not have a `QUIET` variable set for some different process. They might not know why the quickstart guide is not working as expected, if they are wanting to set a multi cluster configuration.

Finally, if we look at the problem, you are trying to configure the script to only run the single cluster setup in a hands free mode (I am guessing to make your dev process easier). What stops a user wanting to run the multi cluster configuration in a hands free mode, maybe be in a pipeline for example? What I think the correct approach here should be to allow the user pass in and / or be auto read from the directory the script is ran from, a configuration file. This configuration file should hold the answer to any input the script expects and the script asks any question that the user forgot to set configuration parameter for. If the values are there, log that the value comes from the configuration file.

The use of a configuration file in this manner means that very hand holding questions can be asked to new user the first time they use the script. While at the same time allowing someone that runs the script regularly to have multiply configurations saved that they can run in a hands free experience.

ehearneRedHat:
@david-martin 
> What about env vars for each of the prompts? For example, if DNS_PROVIDER is set, skip the prompt for that.

Sounds good to me. It would allow for better flexibility .

ehearneRedHat:
@R-Lawton 
> remove the are you ready prompt and combine the 2 DNS provider ones together so by inputting a provider that's a yes and have no also as a option

Makes sense to me. :)

Boomatang:
There is one detail that maybe I was clear about in the first comment. The configuration file would only be used if it is present. So you don't make the script less convenient by needing to generate the configuration file. It should work with out it.

Also remember that a lot of our guides expect the user to create files, like `echo "some CR contains" | kubectl apply -f -`, so doing some thing like that for the quick start might be acceptable. That is just a thought and I am not suggesting that is what we do. I am not a fan of that pattern.

But it does raise a valid question for this PR. Where is this new flag documented. How is a new user meant to know what it does? If that involves reading and understand all the documentation then why not write it to build up a configuration file during that process? The same would go for `DNS_PROVIDER` environment variable that @david-martin is suggesting. That would need to be documented to say what it does, but also the inverse is true. What happens when the environment variable is not set would also need to be documented. 

One more thing that should be in the back of you mind when doing this is what happens to the cardinality caused by the introduction of the external configurations. I say external configurations as this affects both environment variable and configuration files. Take this example if `QUIET=true` and `DNS_PROVIDER=true`, does that mean the script should run in quiet mode but also ask the questions about what settings are need for the DNS provider?

ehearneRedHat:
> There is one detail that maybe I was clear about in the first comment. The configuration file would only be used if it is present. So you don't make the script less convenient by needing to generate the configuration file. It should work with out it.

This makes much more sense to me, thanks for clarifying that! I can now see the benefit of using such a configuration file, and will look to implement. 

> Also remember that a lot of our guides expect the user to create files, like `echo "some CR contains" | kubectl apply -f -`, so doing some thing like that for the quick start might be acceptable. That is just a thought and I am not suggesting that is what we do. I am not a fan of that pattern.

Thanks for pointing that out - I am now a fan of the idea with the idea that it is to complement, not to replace the quiet flag idea. 

> But it does raise a valid question for this PR. Where is this new flag documented. How is a new user meant to know what it does? If that involves reading and understand all the documentation then why not write it to build up a configuration file during that process? The same would go for `DNS_PROVIDER` environment variable that @david-martin is suggesting. That would need to be documented to say what it does, but also the inverse is true. What happens when the environment variable is not set would also need to be documented.

You are right, I will write up a ticket for documenting this new flag. I believe if we re-factor how we think about documenting, where this is an additional feature, and we can leave what we already have there in place, while adding this new section that outlines how to use these new environment variables or configuration, that it would require less work while still providing the user with the context.

Thinking about it, maybe it could also be another ticket for implementing the configuration too. That is to say, it could be something worked on after this PR, or simultaneously. I'll go ahead and write that one up, too. 

> One more thing that should be in the back of you mind when doing this is what happens to the cardinality caused by the introduction of the external configurations. I say external configurations as this affects both environment variable and configuration files. Take this example if `QUIET=true` and `DNS_PROVIDER=true`, does that mean the script should run in quiet mode but also ask the questions about what settings are need for the DNS provider?

Thanks for mentioning this! My understanding would be that when `DNS_PROVIDER` and `KUADRANT_QUIET` is set, that it would look for the related env variables associated with that provider, and if not set, the script would exit with a message to let the user know that those variables also need to be set so that the script can run correctly, therefore still complying with the script being ran without intervention. 

To summarise, I will continue on the path with creating the environment variables needed as mentioned above, while creating tickets for documenting the process and creating a configuration file to complement the quickstart . 

How does that sound ?

Boomatang:
Yes for the most of it. I disagree with the use of environment variable in general. While yes a configuration file would complement the environment variable, I still see the problem of using environment variables. If it was feature flags, ya, they go hand in hand with  configuration files. When I say feature flags an example call would look like `quickstart-setup.sh --quiet=true`. 

The other piece of information I will share, more for context then any else. Environment variables do not act the same every where. I use Fish for my shell. There is 5 main levels of environment variables. The first being local which is only accessible from within the fish function that defines it. The there is shell level access which allows access within that session shell only. So you can do `echo $HELLO_WORLD`. Then there is exported mode which allows scripts to access the environment variable in the current shell session. The there is universal mode which allow any shell session to access the variable. Finally there is the universal exported mode, which allows any script on in any shell session to access the environment variable. 

I should say, the plan you outlined in your summary sound good under the circumstances.

ehearneRedHat:
> Yes for the most of it. I disagree with the use of environment variable in general. While yes a configuration file would complement the environment variable, I still see the problem of using environment variables. If it was feature flags, ya, they go hand in hand with configuration files. When I say feature flags an example call would look like `quickstart-setup.sh --quiet=true`.

I can see the benefit of using feature flags, but wouldn't it nearly be over engineering the quickstart script if we were to start using them alongside env vars ? It could become confusing, and to mitigate, would mean re-designing the script entirely. Since we are already using env vars, I think it would be best to continue using them for now . 

> The other piece of information I will share, more for context then any else. Environment variables do not act the same every where. I use Fish for my shell. There is 5 main levels of environment variables. The first being local which is only accessible from within the fish function that defines it. The there is shell level access which allows access within that session shell only. So you can do `echo $HELLO_WORLD`. Then there is exported mode which allows scripts to access the environment variable in the current shell session. The there is universal mode which allow any shell session to access the variable. Finally there is the universal exported mode, which allows any script on in any shell session to access the environment variable.

I believe we defined exactly which shell to use at the top of the script, so it should use bash by default when running . I can see where it can present issue if using different shells, but since we have already defined the shell, I would expect the behaviour of running the script to be bash . 

> I should say, the plan you outlined in your summary sound good under the circumstances.

Thanks for letting me know! I will continue with that plan for now. :)

ehearneRedHat:
https://github.com/Kuadrant/docs.kuadrant.io/issues/129

ehearneRedHat:
PR closed as new issue replaces it.

Issue #530, opened 2024-04-10
GatewayAPI topology (DAG 1.0)


eguzki:
https://github.com/Kuadrant/kuadrant-operator/pull/447 introduced the GatewayAPI topology (DAG) implementation

This (epic) issue tracks the usage of GatewayAPI topology (DAG) implementation by the different kuadrant controllers

- [X]  Rate limiting wasmplugin controller #447
- [x] Rate limiting limits controller #527 
  - [x] #751 
  - [x] #752 
  - [x] #777 
- [x] AuthPolicy controller #712 #713 
- [X] DNS policy controller #638
- [ ] Controller adding kuadrant annotations on the gateways
- [x] Kuadrant controller #714 
- [ ] Limitador Cluster Envoy controller  #527  
- [X] RateLimitPolicy controller #638
- [ ] TLS Policy controller

#### The GatewayAPI Topology
The controller builds what it is named "The GatewayAPI Topology". It reads **ALL** Gateways, HTTPRoutes and policies (only those the controller cares about) in the cluster. With all that info, the controller creates a new instance of the Gateway API topology as well as a set of indexes (another "view" of the topology) that can be queried. Interesting to note here that [field indexers](https://pkg.go.dev/sigs.k8s.io/controller-runtime#hdr-Clients_and_Caches) can optionally be added to the manager to build internal index caches. One example being implemented is an index to query "give me all the HTTPRoutes parenting Gateway A". 

The current implementation of the "GatewayAPI Topology" does not use a k8s client to query Gateway API and kuadrant policies. Instead, fetching resources must be done externally and then the GatewayAPI topology receives a list of resources. It could be done otherwise and have generic function with a k8s client to build topology for a given policy of one kind. The main reason for the current implementation is to decouple GatewayAPI topology and the cluster. Meaning that I can build GatewayAPI topology from a set of resources filtered with some criteria. 

The implementation assumes "cheap" operations of reading **ALL** the gateways, routes and policies. That assumption comes from the controller runtime k8s client that holds a cache. At least for `list` and `get` operations. That cache is being kept up to date  in the background when there are updates/deletes in the objects triggered by kuadrant controllers or any other third party controllers or any other k8s API server user. Check [for this comment](https://github.com/Kuadrant/kuadrant-operator/pull/317#issuecomment-1963852398) with a verification of this "cheap" operation assumption. Indeed, the resource lists are cached across reconciliation events and controllers.

[...]

PR #751, opened 2024-07-09
enhanced gateway api topology

eguzki:
### What

* new topology option `WithAcceptedRoutesLinkedOnly()`: only accepted routes will be linked to the gateway
* new topology option `WithProgrammedGatewaysOnly()`: only programmed (valid) gateways can have routes and attached policies. Not programmed gateways will be isolated nodes in the topology.
* new public method `Policies()` that returns all policies included in the topology. Returned objects are `PolicyNodes` that allow to jump to topology nodes targeted by the policy (aka `PolicyTargetNode`).

This is part of a series of PR's that will end with a new controller that will only reconcile limitador limits configuration.

### DIFF
--- a/controllers/authpolicy_status.go
+++ b/controllers/authpolicy_status.go
@@ -169,6 +169,8 @@ func (r *AuthPolicyReconciler) generateTopology(ctx context.Context) (*kuadrantg
 	})
 
 	return kuadrantgatewayapi.NewTopology(
+		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
+		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
 		kuadrantgatewayapi.WithGateways(utils.Map(gwList.Items, ptr.To[gatewayapiv1.Gateway])),
 		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To[gatewayapiv1.HTTPRoute])),
 		kuadrantgatewayapi.WithPolicies(policies),
--- a/controllers/authpolicy_status.go
+++ b/controllers/authpolicy_status.go
@@ -170,7 +170,7 @@ func (r *AuthPolicyReconciler) generateTopology(ctx context.Context) (*kuadrantg
 
 	return kuadrantgatewayapi.NewTopology(
 		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
-		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
+		kuadrantgatewayapi.WithProgrammedGatewaysOnly(),
 		kuadrantgatewayapi.WithGateways(utils.Map(gwList.Items, ptr.To[gatewayapiv1.Gateway])),
 		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To[gatewayapiv1.HTTPRoute])),
 		kuadrantgatewayapi.WithPolicies(policies),
--- a/controllers/ratelimitpolicy_enforced_status_controller.go
+++ b/controllers/ratelimitpolicy_enforced_status_controller.go
@@ -200,6 +200,8 @@ func (r *RateLimitPolicyEnforcedStatusReconciler) buildTopology(ctx context.Cont
 	}
 
 	return kuadrantgatewayapi.NewTopology(
+		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
+		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
 		kuadrantgatewayapi.WithGateways(utils.Map(gatewayList.Items, ptr.To[gatewayapiv1.Gateway])),
 		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To[gatewayapiv1.HTTPRoute])),
 		kuadrantgatewayapi.WithPolicies(utils.Map(policyList.Items, func(p kuadrantv1beta2.RateLimitPolicy) kuadrantgatewayapi.Policy { return &p })),
--- a/controllers/ratelimitpolicy_enforced_status_controller.go
+++ b/controllers/ratelimitpolicy_enforced_status_controller.go
@@ -201,7 +201,6 @@ func (r *RateLimitPolicyEnforcedStatusReconciler) buildTopology(ctx context.Cont
 
 	return kuadrantgatewayapi.NewTopology(
 		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
-		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
 		kuadrantgatewayapi.WithGateways(utils.Map(gatewayList.Items, ptr.To[gatewayapiv1.Gateway])),
 		kuadrantgatewayapi.WithRoutes(utils.Map(routeList.Items, ptr.To[gatewayapiv1.HTTPRoute])),
 		kuadrantgatewayapi.WithPolicies(utils.Map(policyList.Items, func(p kuadrantv1beta2.RateLimitPolicy) kuadrantgatewayapi.Policy { return &p })),
--- a/pkg/library/gatewayapi/topology.go
+++ b/pkg/library/gatewayapi/topology.go
@@ -222,9 +222,9 @@ func WithPolicies(policies []Policy) TopologyOpts {
 	}
 }
 
-func WithProgrammedGatewaysOnly(programmedGatewaysOnly bool) TopologyOpts {
+func WithProgrammedGatewaysOnly() TopologyOpts {
 	return func(o *topologyOptions) {
-		o.programmedGatewaysOnly = programmedGatewaysOnly
+		o.programmedGatewaysOnly = true
 	}
 }
 
--- a/pkg/library/gatewayapi/topology.go
+++ b/pkg/library/gatewayapi/topology.go
@@ -328,7 +328,6 @@ func buildDAGEdges(opts *topologyOptions, gateways []gatewayDAGNode, routes []ht
 	edges := make([]edge, 0)
 
 	for _, route := range routes {
-
 		gatewayParentKeys := GetGatewayParentKeys(route.HTTPRoute)
 
 		if opts.linkAcceptedRoutesOnly {
@@ -359,7 +358,6 @@ func buildDAGEdges(opts *topologyOptions, gateways []gatewayDAGNode, routes []ht
 		for _, attachedPolicy := range attachedPolicies {
 			edges = append(edges, edge{parent: route, child: attachedPolicy})
 		}
-
 	}
 
 	for _, g := range effectiveGatewys {
--- a/pkg/library/gatewayapi/topology.go
+++ b/pkg/library/gatewayapi/topology.go
@@ -18,42 +18,135 @@ const (
 	typeField      dag.Field     = dag.Field("type")
 	gatewayLabel   dag.NodeLabel = dag.NodeLabel("gateway")
 	httprouteLabel dag.NodeLabel = dag.NodeLabel("httproute")
+	policyLabel    dag.NodeLabel = dag.NodeLabel("policy")
 )
 
-type RouteNode struct {
-	*gatewayapiv1.HTTPRoute
-
-	attachedPolicies []Policy
+type PolicyTargetNode interface {
+	GetGatewayNode() *GatewayNode
+	GetRouteNode() *RouteNode
 }
 
-func (r *RouteNode) AttachedPolicies() []Policy {
-	return r.attachedPolicies
+type PolicyNode struct {
+	policyDAGNode
+
+	graph *dag.DAG
+}
+
+func (p *PolicyNode) GetPolicy() Policy {
+	return p.Policy
+}
+
+func (p *PolicyNode) TargetRef() PolicyTargetNode {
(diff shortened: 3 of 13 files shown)


pkg/rlptools/overrides.go
```diff
@@ -37,7 +39,8 @@ func ApplyOverrides(topology *kuadrantgatewayapi.Topology, gateway *gatewayapiv1
 
 	for _, route := range topology.Routes() {
 		if !slices.Contains(kuadrantgatewayapi.GetRouteAcceptedGatewayParentKeys(route.HTTPRoute), client.ObjectKeyFromObject(gateway)) {
-			overriddenPolicies = append(overriddenPolicies, route.AttachedPolicies()...)
+			routePolicies := utils.Map(route.AttachedPolicies(), func(pNode kuadrantgatewayapi.PolicyNode) kuadrantgatewayapi.Policy { return pNode.Policy })
```
guicassolato, line 42: TO avoid adding another dependency:

```suggestion
			routePolicies := lo.Map(route.AttachedPolicies(), func(pNode kuadrantgatewayapi.PolicyNode, _ int) kuadrantgatewayapi.Policy { return pNode.Policy })
```

pkg/library/gatewayapi/topology.go
```diff
@@ -18,81 +18,192 @@ const (
...
+	programmedGatewaysOnly bool
+	linkAcceptedRoutesOnly bool
 }
 
 // TopologyOpts allows to manipulate topologyOptions.
 type TopologyOpts func(*topologyOptions)
 
+func WithAcceptedRoutesLinkedOnly() TopologyOpts {
+	return func(o *topologyOptions) {
+		o.linkAcceptedRoutesOnly = true
```
guicassolato, line 197: I guess we don't ever want non-accepted routes in the topology as we may occasionally want non-programmed gateways? I actually wonder what would be the use case for adding non-programmed gateways.

controllers/ratelimitpolicy_enforced_status_controller.go
```diff
@@ -200,6 +200,8 @@ func (r *RateLimitPolicyEnforcedStatusReconciler) buildTopology(ctx context.Cont
 	}
 
 	return kuadrantgatewayapi.NewTopology(
+		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
+		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
```
guicassolato, line 204: I understand a gateway that transitions from `Programmed: True` to `Programmed: False` will NOT get the statuses of its affecting policies updated then?

I'm thinking this line invoked for a gateway that has not been added to the topology: https://github.com/Kuadrant/kuadrant-operator/blob/4b9d60e49eae89460a963b9395f1e92a8f837332/controllers/ratelimitpolicy_enforced_status_controller.go#L60

pkg/library/gatewayapi/utils.go
```diff
@@ -165,3 +165,23 @@ func IsGatewayAPIInstalled(restMapper meta.RESTMapper) (bool, error) {
 
 	return false, err
 }
+
+// GetGatewayParentRefsFromRoute returns the list of parentRefs that are Gateway typed
+func GetGatewayParentRefsFromRoute(route *gatewayapiv1.HTTPRoute) []gatewayapiv1.ParentReference {
+	if route == nil {
+		return nil
+	}
+	return utils.Filter(route.Spec.ParentRefs, IsParentGateway)
```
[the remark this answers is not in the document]
eguzki: I do not see any reason to change when `utils.Filter()` does what it is expected to do. Anything I could miss?

pkg/rlptools/overrides.go
eguzki: same as https://github.com/Kuadrant/kuadrant-operator/pull/751#discussion_r1672144674

pkg/library/gatewayapi/topology.go
eguzki: there are some use cases I found that non-accepted routes should be linked to parent gateways: mappers. It allows getting events for "disconnected" policies, which status can be affected from the "disconnection".

controllers/ratelimitpolicy_enforced_status_controller.go
eguzki: good catch, thanks!

pkg/library/gatewayapi/topology.go
eguzki: >  I actually wonder what would be the use case for adding non-programmed gateways.

this is a use case https://github.com/Kuadrant/kuadrant-operator/pull/751#discussion_r1671913835

controllers/ratelimitpolicy_enforced_status_controller.go
eguzki: fixed

pkg/rlptools/overrides.go
eguzki: Maybe, for other PR, we could remove all `utils.Map` and `utils.Filter` and rewrite using [github.com/samber/lo](https://pkg.go.dev/github.com/samber/lo) (or something else). Then we can remove `utils` code and have the external dependency only.

pkg/library/gatewayapi/utils.go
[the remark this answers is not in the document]
guicassolato: `github.com/samber/lo` is immensely popular and IMO very good at this kind of generic widely used functions. Wherever we can leverage something like this from the community – provided it suits our needs of course – instead of keeping our own home-made version of the same thing, I personally would be in favour of.

At the same time, I don't want to create a fuss. I understand this is not the main goal of this PR. Just something to think about.

pkg/library/gatewayapi/topology.go
guicassolato: > there are some use cases I found that non-accepted routes should be linked to parent gateways

So why not giving the option here then?

guicassolato: Sorry. To clarify...

My point is that we have `WithProgrammedGatewaysOnly(bool)` for when we want to enforce programmed gateways only, i.e. `WithProgrammedGatewaysOnly(true)`. I don't see why one would use `WithProgrammedGatewaysOnly(false)` though – versus simply not calling `WithProgrammedGatewaysOnly` at all.

I guess that was your thinking for `WithAcceptedRoutesLinkedOnly`. Nobody will ever need `WithAcceptedRoutesLinkedOnly(false)`. The effect in this case is virtually the same as not using `WithAcceptedRoutesLinkedOnly`, right?

So why not designing `WithProgrammedGatewaysOnly` the same way? If there's a reason I struggling to see, then there may be for `WithAcceptedRoutesLinkedOnly` too perhaps?

IOW, why are the signatures of these 2 functions different?

eguzki: Makes all the sense. Fixing it

eguzki: done

controllers/authpolicy_status.go
```diff
@@ -169,6 +169,8 @@ func (r *AuthPolicyReconciler) generateTopology(ctx context.Context) (*kuadrantg
 	})
 
 	return kuadrantgatewayapi.NewTopology(
+		kuadrantgatewayapi.WithAcceptedRoutesLinkedOnly(),
+		kuadrantgatewayapi.WithProgrammedGatewaysOnly(true),
```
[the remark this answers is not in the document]
guicassolato: BTW, the described problem is not introduced by this PR, but neither it's fixed by it though. In fact, I'm afraid the change may even aggravate the problem.

The more fundamental problem is of course that, in this case, the topology is only marginal to the decisions we make. Most of the business logic for reconciling AuthPolicies still won't leverage the topology graph built by this function, while some already do.

By trying to fix _only_ the topology graph, we're effectively providing two views of world, thus splitting the business logic into the portion that creates configuration based on one reality and another based on a different reality.

I fear that, by trying to do the right thing here, we're actually bringing the system to an inconsistent state.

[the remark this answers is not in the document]
eguzki: :exploding_head: 

I am confused. I will kindly ask you to elaborate more on this because I do not get it. 

I will try to clarify some of your statements:

> BTW, the described problem is not introduced by this PR, but neither it's fixed by it though.

This PR does not pretend to change a bit the behavior of any of the controllers. All the tests pass. There is no problem being described and thus, no problem being fixed.

This PR is adding few more options to the DAG topology, yet keeping the controllers behavior as they are. The intention is to add more tools/capabilities on the topology lib to enable following up PR's introducing the Limitador  limits controller. 

The authcontroller in current `main` branch is building a topology with only programmed gateways and linking only accepted routes. So is doing the code with the new changes in this PR. Nothing changes. And if something changes, let me know and I will fix it. 

 > In fact, I'm afraid the change may even aggravate the problem.

First of all, what problem are you talking about? 

Configuration for "programmed gateways only" option was introduced by https://github.com/Kuadrant/kuadrant-operator/pull/715. This PR is adding (among other public methods) yet another configuration option which is opt out from "link accepted routes only". As you know, currently the topology has [hard-coded](https://github.com/Kuadrant/kuadrant-operator/blob/c8d02d0437956dc66c14f4a04327c435a1b66fa3/pkg/library/gatewayapi/topology.go#L198) linking only accepted routes. This may not be always desirable. This feature is not needed now, but will be needed by the limits controller. The reasoning being that the controller will  need a topology in its mappers which routes are linked to the gateway no matter they are accepted or not. But this is something I can explain further when I open PR for the limits controller.  

I do not see any problem being aggravated. Again, this is not changing any controller behavior. And if it does, I will fix it.

> By trying to fix only the topology graph, we're effectively providing two views of world, thus splitting the business logic into the portion that creates configuration based on one reality and another based on a different reality.

This is by no means fixing the topology. AFAIK, there is no issue with the current topology for the needs of the current controllers. This is about adding more configurability to the topology. Some business logic requires a type of topology with not programmed gateways as well as linked routes that have not been accepted. Specifically I am thinking on the mappers. Mappers need to walk the topology. And having disconnections because routes are not accepted prunes events to reconcile kuadrant policies.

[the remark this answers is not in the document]
guicassolato: > Configuration for "programmed gateways only" option was introduced by #715. […] As you know, currently the topology has [hard-coded](https://github.com/Kuadrant/kuadrant-operator/blob/c8d02d0437956dc66c14f4a04327c435a1b66fa3/pkg/library/gatewayapi/topology.go#L198) linking only accepted routes.

This indeed confirms the PR does not aggravate the problem, and it's enough for me to approve it. Thanks for explaining it and for collecting the evidences in the code to support it, @eguzki .

To elaborate on my main criticism, which I still stand for...

<table>
<tr>
<td>
<b>Disclaimer for all readers:</b> please try to read in between the lines and focus on the generalized form of the problem I'm trying to describe, rather than the individual examples I use.
</td>
</tr>
</table>

I missed #715. I think this is a very good example of how a well-intended change, that we needed indeed and that, looked individually, made our code better than before, but, at the same time, if observed holistically, emerges nonetheless as a change that creates a problem (or aggravates one) we didn't have before (or that simply had a smaller surface before.)

We spotted we were treating non-programmed gateways as if they were ready and thus we were signaling the policies configured on those gateways as enforced. This was clearly a bug and had to be fixed. We have tons of code that implement business logic based on a "view of the world" provided by this topology struct. We made that topology struct to only account for programmed gateways. Done. Problem solved. We fixed all those business logic with one move.

However, unfortunately there are controllers (possibly all of them AFAIK!) that mix business logic based on the topology struct with business logic that does not use the topology struct, where sometimes the status of the Gateways and HTTPRoutes are properly taken into account, sometimes not. (I mentioned the reconciliation of AuthConfigs before, but look at [`ComputeGatewayDiff`](https://github.com/Kuadrant/kuadrant-operator/blob/c8d02d0437956dc66c14f4a04327c435a1b66fa3/pkg/library/reconcilers/gateway_diffs.go#L27) for plenty of more examples.)

With the topology struct now doing what seems to be the right thing, we created two "views of the world" – one that we believe is right and another one that may or may not be right. These two views (as well as the decisions based on them) are often intertwined together, leading to a situation where we can no longer say _"Kuadrant overlooks the state of the gateways. As a user, it's up to you to make sure to bring all your gateways to a programmed state. Until then, do not trust status of Kuadrant policies."_ Unfortunately, neither we can say today, after #715, _"No worries! Kuadrant makes sure to always configure and report status of your policies right, according to the state of the gateways."_ Between the two statements, obviously we want to issue the latter, but if that's not true, then we're probably better off with the former IMO.

I used "programmed gateways" as example, but we can run over again replacing it with "accepted routes", or ["listener status"](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.ListenerStatus), and possibly others that do not occur to me right now. The important part is, while we are not consistent about the views of the world that the code bases its decisions upon, it becomes very hard to debug behaviours such as:
- why a policy with status enforced doesn't seem to be enforced at all
- why an override that seems not have been accepted appears to be taken over my route-level rules (or the other way around)
- etc

To sum up, let's be careful with changes that seem to be improving the code but until they are rolled out completely they may actually put us (as well as our users) in a difficult position. This is not the case of this PR as I'm now convinced, and neither it is a negative criticism to the DAG/topology approach in general. Rather, I'm concerned with how we roll out these changes.

guicassolato (approved):

PR #752, opened 2024-07-09
base reconciler reconciles status

eguzki:
### What

Common status reconciler. The signature

```go
func(ctx context.Context, objKey client.ObjectKey, obj client.Object, mutator StatusMutator) 
```

Then, it can be used as follows:

```go
        newStatus := r.calculateStatus(rlp, specErr)
         r.ReconcileResourceStatus(                                                            
              ctx,                                                                                        
              client.ObjectKeyFromObject(rlp),                                                            
             &kuadrantv1beta2.RateLimitPolicy{},                                                         
             kuadrantv1beta2.RateLimitPolicyStatusMutator(newStatus),                            
       )
```

And the mutator looks like

```go
  func RateLimitPolicyStatusMutator(desiredStatus *RateLimitPolicyStatus) reconcilers.StatusMutatorFunc {
        return func(obj client.Object) (bool, error) {                                                  
              existingRLP, ok := obj.(*RateLimitPolicy)                                                   
              if !ok {                                                                                    
                   return false, fmt.Errorf("unsupported object type %T", obj)                             
              }                                                                                           
                                                                                                                                                                                                                                                                                     
             if cmp.Equal(*desiredStatus, existingRLP.Status, opts) {                                    
                  return false, nil                                                                       
              }                                                                                           
                                                                                                     
              existingRLP.Status = *desiredStatus                                                         
                                                                                                     
              return true, nil
}
```



This is part of a series of PR's that will end with a new controller that will only reconcile limitador limits configuration.


api/v1beta2/ratelimitpolicy_types.go
```diff
@@ -191,12 +200,32 @@ func (s *RateLimitPolicyStatus) Equals(other *RateLimitPolicyStatus, logger logr
...
+var _ reconcilers.ObjectWithStatus = &RateLimitPolicy{}
+
+func (r *RateLimitPolicy) GetStatusObject() reconcilers.StatusObject {
+	return &r.Status
+}
+
+func (r *RateLimitPolicy) SetStatusObject(statusObject reconcilers.StatusObject) {
+	switch status := statusObject.(type) {
+	case *RateLimitPolicyStatus:
+		r.Status = *status
```
Boomatang, line 226: I am not sure about this silently failing if the wrong kind of `StatusObject` is passed in, it seems wrong. At the same time I don't think just logging an error message is enough.

didierofrivia: Maybe we could return `error` if as a `default` case (nothing matched) and we set the status to `nil` ? we could bubble up that error and decide given the setter expectations what to do

eguzki: before considering returning an `error`, any proposal?

Boomatang: My mind is starting to go to extremes and I really don't know how well it suits. I am thinking that raising a panic might be the correct approach. If we return an error the reconcile ends up doing possibly twos things. One logs the error message, and two, reschedule the reconcile to try again. But if the wrong type is being past in, it was most likely an error caused by a developer in the code base where no matter how many times it is reschedule, reconcile, it will never pass. 

I know causing the panic is extreme and not really done in operators, but I don't see how it would ever natural recover. 

I do wonder what this would look like if the function is made more generalized  and taken off the struct its self. But possible would have other issue, along with some of the same issue that I can currently visualize in my head.

eguzki: thanks for the ideas... let me think and try something else. I reckon the type assertion is a bad smell.

eguzki:
@Boomatang 

I have tried a different approach. Iteration #2. 

I do not like the `panic` approach because in production you kill the controller entirely, while returning an error and doing nothing only disables one feature. 

This is not entirely clean approach either and there is still a chance to mistakenly mix types by the developer, so the status updater tries again and again returning error. But it is also true that it is the same mutator approach as the controllers use to reconcile resource specs (not status). It is being used by all the controllers and never hit any issue like that. So I think it should be good enough.

You might be interested in looking at the envoygateway controller implementation of the generic status updater: https://github.com/envoyproxy/gateway/blob/main/internal/provider/kubernetes/status_updater.go#L72-L121

pkg/library/reconcilers/base_reconciler.go
```diff
@@ -34,6 +34,32 @@ import (
...
+
+// StatusMutator is an interface to hold mutator functions for status updates.
+type StatusMutator interface {
+	Mutate(obj client.Object) (bool, error)
+}
+
+// StatusMutatorFunc is a function adaptor for StatusMutators.
+type StatusMutatorFunc func(client.Object) (bool, error)
+
+// StatusMutatorFunc adapts the MutatorFunc to fit through the StatusMutator interface.
```
Boomatang, line 54: Should document the function not the struct. Guessing a copy paste error
```suggestion
// Mutate adapts the MutatorFunc to fit through the StatusMutator interface.
```

Boomatang (changes requested):
There was the one documentation issue, but other than that every thing is working as I expected. This newer method seems far nicer than before.

eguzki:
> There was the one documentation issue, but other than that every thing is working as I expected. This newer method seems far nicer than before.

good catch. Addressed

Boomatang (approved):
Looks good to me

PR #753, opened 2024-07-10
Add Alerts Panels to PE Dashboard

ehearneRedHat:
### What:
4 panels have been added: 

- Total Firing Alerts (Stat Panel)
- Total Pending Alerts (Stat Panel)
- All Pending/Firing Alerts Table
- Alert State Timeline (Firing)

### Verify:

```bash
KUADRANT_REF=733-add-alerts-panels-to-pe-dashboard ./hack/quickstart-setup.sh
kubectl -n monitoring wait --for=condition=Available deployment grafana --timeout=300s # Not necessary, though convenient :D 
kubectl -n monitoring port-forward service/grafana 3000:3000
```

Log into local Grafana instance with username and password `admin/admin` and verify that `Platform Engineer Dashboard` now has `Alerts` row with the panels mentioned above. Wait a few minutes for data to come in from alerts . Also verify that new panels do not affect the other panels .

### DIFF
--- a/examples/dashboards/controller-resources-metrics.json
+++ b/examples/dashboards/controller-resources-metrics.json
@@ -232,7 +232,6 @@
   "style": "dark",
   "tags": [],
   "templating": {
-    
     "list": [
       {
         "current": {
@@ -253,7 +252,7 @@
         "skipUrlSync": false,
         "type": "datasource"
       },
-            {
+      {
         "datasource": {
           "type": "prometheus",
           "uid": "${datasource}"
@@ -287,7 +286,7 @@
         "definition": "label_values(controller_runtime_reconcile_total, namespace)",
         "hide": 0,
         "includeAll": false,
-                "multi": false,
+        "multi": false,
         "name": "namespace",
         "options": [],
         "query": {
--- a/examples/dashboards/controller-runtime-metrics.json
+++ b/examples/dashboards/controller-runtime-metrics.json
@@ -1,5 +1,4 @@
 {
-  
   "__requires": [
     {
       "type": "datasource",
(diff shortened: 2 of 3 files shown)


david-martin (approved):
Approving the addition of new panels focused on alerting.
I have contributed changes as well, so I'll defer to @R-Lawton for another approval

david-martin:
@R-Lawton I'd really appreciate it if you had time to sanity check the dashboard changes in here.
Myself and Evan collaborated on them, so it would be good to get a 3rd party

R-Lawton:
2 minor nit-picky things. 
Can we make the heading of the columns align to the centre to match the rest of the tables? 
Second thing is wdyt about make the active state timeline panel the full width of the dashboard rather then not going under the single stat panels.

david-martin:
> Can we make the heading of the columns align to the centre to match the rest of the tables?

image

These headings? Should be trivial to do that I think.




> wdyt about make the active state timeline panel the full width of the dashboard rather then not going under the single stat panels.

It should be fine, and gives extra viewing space for the timeline and alert names.
I aligned it with the panel above as it was showing a different view of the data, but I don't see a strong enough visual correlation to keep them aligned.

R-Lawton:
> > Can we make the heading of the columns align to the centre to match the rest of the tables?
> 
> image
> These headings? Should be trivial to do that I think.
> 
Yes exactly, setting Column alignment to center should do the trick

ehearneRedHat:
@david-martin @R-Lawton will work to resolve...

R-Lawton:
Changes look good to me 👍

R-Lawton (approved):

Issue #241, opened 2023-08-30
Create label to trigger Quay images


didierofrivia:
Currently we're facing mainly 2 issues:

1. Images overflow in our [quay.io](quay.io/kuadrant/) repo
2. Pull requests coming from forks cannot trigger the build and push of the images

It would be good to have a label or a comment that administrators of Kuadrant org can assign to any PR in order to allow the trigger of the required quay job.

This needs to be applied to every Kuadrant repo.

Something that's been used by k8s that could be of help is https://prow.k8s.io/command-help#ok_to_test.

Boomatang:
I am not sure how much the ok_to_test label that k8s uses will help. K8s uses prow and not github actions. I am not say that github actions can't do the same, its more any prow feature would need to be researched to find the equivalent github action feature. A scan of https://github.com/marketplace/actions/label-actions shows that it is possible to do the triggers on labels. If it is enough for what we want, that is a different question.

PR #768, opened 2024-07-19
Workflow: PR Conditional Build and Push Image to Quay.io Repo

dlaw4608:
Closes: [#241]([https://github.com/Kuadrant/kuadrant-operator/issues/241])

## Summary:

This workflow leverages the pull_request_target event and performs the build and push only if the forked_image_approved label is applied to the PR. This ensures that only approved changes can trigger the image build and push process, maintaining security for secrets.

```
name: PR Conditional Build and Push Image to Quay.io Repo

on:
  pull_request_target:
    types: [labeled, opened, synchronize, reopened]

jobs:
  workflow-build:
    if: ${{ github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'forked_image_approved') }}
    name: Calls build-images-base workflow
    uses: ./.github/workflows/build-images-base.yaml
    secrets: inherit
    with:
      kuadrantOperatorVersion: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }}
      kuadrantOperatorTag: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }} 
```

## Verification Steps

1. Create a Pull Request from a Fork:
+ Fork the repository and make changes that affect the Docker image build.
+ Create a pull request (PR) from the forked repository to the base repository.

2. Apply the forked_image_approved Label:
+ As a maintainer, review the incoming PR.
+ If the changes are approved, apply the forked_image_approved label to the PR.

3. Check Workflow Execution:
+ Verify that the PR Conditional Build and Push Image to Quay.io Repo workflow is triggered upon applying the label.
+ Ensure the workflow runs successfully, building and pushing the Docker images.

4. Verify Image Push:
+ After the workflow completes, check the Quay.io repository.
+ Confirm that the image is built and pushed with the tags corresponding to the PR (e.g., username-PRnumber).

# Should look like this:
(https://github.com/user-attachments/assets/563717f2-9442-47fb-afbe-e0a83333e8cd)

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -194,4 +194,4 @@ jobs:
           tags: |
             ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}-catalog:${{ env.IMG_TAGS }}
       - name: Print Catalog Image URL
-        run: echo "Catalog image pushed to ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}-catalog:${{ env.IMG_TAGS }}"
+        run: echo "Catalog image pushed to ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}-catalog:${{ env.IMG_TAGS }}"        
--- a/.github/workflows/build-images-pr.yaml
+++ b/.github/workflows/build-images-pr.yaml
@@ -1,0 +1,29 @@
+name: PR Conditional Build and Push Image to Quay.io Repo
+
+on:
+  pull_request_target:
+    types: [labeled, opened, synchronize, reopened]
+
+concurrency:
+  group: pr-conditional-build-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  workflow-build:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'forked_image_approved') }}
+    name: Calls build-images-base workflow
+    uses: ./.github/workflows/build-images-base.yaml
+    secrets: inherit
+    with:
+      kuadrantOperatorVersion: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }}
+      kuadrantOperatorTag: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }}
+
+  remove-label:
+    runs-on: ubuntu-latest
+    needs: workflow-build
+    steps:
+      - name: Remove forked_image_approved label after workflow is triggered
+      - uses: actions-ecosystem/action-remove-labels@v1
+        with:
+          labels: forked_image_approved
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-images-pr.yaml
+++ b/.github/workflows/build-images-pr.yaml
@@ -13,18 +13,20 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ contains(github.event.pull_request.labels.*.name, 'forked_image_approved') }}
     name: Calls build-images-base workflow
-    uses: ./.github/workflows/build-images-base.yaml
-    secrets: inherit
-    with:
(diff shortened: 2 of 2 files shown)


eguzki:
This is solving only one part of the #241 

We might want to create another github issue for `Images overflow in our quay.io repo`

.github/workflows/build-images-base.yaml
```diff
@@ -207,4 +207,4 @@ jobs:
           username: ${{ secrets.IMG_REGISTRY_USERNAME }}
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
       - name: Print Image URL
-        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
```
KevFan, line 210: nitpick: lets keep this line break so that the PR doesn't touch this file since it's doesn't change anything other than removing a line break

.github/workflows/build-images-pr.yaml
```diff
@@ -0,0 +1,15 @@
+name: PR Conditional Build and Push Image to Quay.io Repo
+
+on:
+  pull_request_target:
+    types: [labeled, opened, synchronize, reopened]
+
+jobs:
+  workflow-build:
+    if: ${{ github.event_name == 'pull_request_target' && contains(github.event.pull_request.labels.*.name, 'forked_image_approved') }}
```
KevFan, line 9:
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'forked_image_approved')
```

I think we can simplify to this instead since we should only get events from `pull_request_target` defined in the `on` previously already

mikenairn:
@didierofrivia re. https://github.com/Kuadrant/kuadrant-operator/issues/241, why do we want to trigger this job on pull requests?

eguzki:
> rigger this job on pull requ

In other words, why do we want to push images to quay for forked repos? Build make sense as yet another test

.github/workflows/build-images-pr.yaml
```diff
@@ -0,0 +1,15 @@
...
+    name: Calls build-images-base workflow
+    uses: ./.github/workflows/build-images-base.yaml
+    secrets: inherit
+    with:
+      kuadrantOperatorVersion: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }}
+      kuadrantOperatorTag: ${{ github.event.pull_request.user.login }}-${{ github.event.pull_request.number }}
```
KevFan, line 15: I think we should also consider to always remove the `forked_image_approved` label regardless of whether the image build was successfull or not as an additional security measure - https://github.com/actions-ecosystem/action-remove-labels.

And also possibly disallow multiples of this workflow from running concurrently (i.e. I'm thinking someone might first get the label and then quickly push something that could leak our secrets so we should try to prevent/mitigate this from happening) 🤔  - https://docs.github.com/en/actions/using-jobs/using-concurrency

didierofrivia:
@eguzki @mikenairn I'd imagine for some testing during the review... would you reckon it's better to not build at all on PRs?

eguzki:
> @eguzki @mikenairn I'd imagine for some testing during the review... would you reckon it's better to not build at all on PRs?

I like building. It is "cheap" and I see it as another type of testing which needs to succeed. 

I do not think we should push to quay.io kuadrant org the builds from forks. What is the value?

KevFan:
I see this as a way for some maintainers (such as me) that typically make PRs from forks rather than branches  to run the image build since the image build flow is not run these "external" PRs. If QE or anyone want to build the image they can as it can serve as another type of test as @eguzki mentioned. Downside is this does allow for a security risk given the scope that the `pull_request_target` has to secrets etc. Not sure is this worth the risk 🤷 

It can also potentially replace the `build-images-branches.yaml` if we choose to only want to build at the point where the PR is made and a maintainer / QE deems it as necessary through a label. Personally I don't see always building the images on branch on every commit as necessary since I don't think it's used that often in general other than to serve as a build test, unless people actually do pull the branch builds regulary for testing 🤔

KevFan:
Going to close this as it doesn't look like it's a priority and not sure is this worth the security risk given the scope that the `pull_request_target` has to secrets etc. and the mitigations that also need to be added to use this.

Will re-open if we want to explore this again.

PR #771, opened 2024-07-22
Provider agnostic gateway name/namespace

adam-cattermole:
## Changes

~Builds on top of the authpolicy branch which should be merged first~

- Renamed the gateway we deploy to `kuadrant-ingressgateway` for both istio and envoygateway
- Updated the namespace from `istio-system`/`envoy-gateway-system` to be `gateway-system` for both providers
- Updated the docs that refer to the gateway to point to this single gateway

Open to suggestions for the naming above, happy to change.

These changes makes it so that a user can follow the user-guides directly without amendment regardless of which gateway api provider they provision with (`make local-setup GATEWAYAPI_PROVIDER=istio`/`make local-setup GATEWAYAPI_PROVIDER=envoygateway`)

### DIFF
--- a/.github/workflows/verify-dashboards-alerts.yaml
+++ b/.github/workflows/verify-dashboards-alerts.yaml
@@ -32,7 +32,7 @@ jobs:
           kind create cluster
           # Install Istio
           kubectl apply -k config/dependencies/istio/sail
-          kubectl -n gateway-system wait --for=condition=Available deployment istio-operator --timeout=300s
+          kubectl -n istio-system wait --for=condition=Available deployment istio-operator --timeout=300s
           kubectl apply -f config/dependencies/istio/sail/istio.yaml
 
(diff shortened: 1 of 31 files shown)


eguzki:
What about sticking to the gateway providers specific namespace names, i.e. `istio-system` and `envoy-gateway-system` **and** create a new namespace for the default ingress gateway in a provider agnosting namespacedname? Something like `gateway-system/ingressgateway` ?? I have the feeling that changing from `istio-system` to something else might raise issues

maleck13:
@adam-cattermole we keeping this PR?

adam-cattermole:
@maleck13 Yep, it's just a little out of date as it's based on the envoy gateway authpolicy branch which is yet to be merged

eguzki (approved):
Sweet 🍻

PR #772, opened 2024-07-23
Simplify status reconciler

eguzki:
### What

Simplify status reconciler leveraging base reconciler `r.ReconcileResourceStatus` method.

Applied to: 

- [X] kuadrant CRD status
- [X] Authpolicy status
- [x] TLSPolicy
- [x] DNSPolicy
- [x] Rate limit policy enforced status controller

Related work https://github.com/Kuadrant/kuadrant-operator/pull/752


api/v1alpha1/tlspolicy_types.go
```diff
@@ -203,6 +202,28 @@ func (p *TLSPolicy) Validate() error {
...
 }
 
+func TLSPolicyStatusMutator(desiredStatus *TLSPolicyStatus, logger logr.Logger) reconcilers.StatusMutatorFunc {
+	return func(obj client.Object) (bool, error) {
+		existing, ok := obj.(*TLSPolicy)
+		if !ok {
+			return false, fmt.Errorf("unsupported object type %T", obj)
+		}
+
+		if equality.Semantic.DeepEqual(*desiredStatus, existing.Status) {
```
eguzki, line 212: @mikenairn Here I stick to the existing implementation. What do you think about moving to `github.com/google/go-cmp/cmp` compare method with `opts` tuned for status condittions?

```go
opts := cmp.Options{                                                        
    cmpopts.IgnoreFields(metav1.Condition{}, "LastTransitionTime"),         
    cmpopts.IgnoreMapEntries(func(k string, _ any) bool {                   
        return k == "lastTransitionTime"                                    
    }),                                                                     
}                                                                           
                                                                            
if cmp.Equal(*desiredStatus, existing.Status, opts) {                    
    return false, nil                                                       
}                                                                           

```

mikenairn: I don't have a strong opinion either way. If we are using `google/go-cmp/cmp` in other policies then that's probably a good enough reason to change it.

eguzki: done

eguzki:
@mikenairn ready for review

api/v1alpha1/dnspolicy_types.go
```diff
@@ -215,6 +215,35 @@ func (p *DNSPolicy) DirectReferenceAnnotationName() string {
...
+			diff := cmp.Diff(*desiredStatus, existing.Status)
+			logger.V(1).Info("status not equal", "difference", diff)
+		}
+
+		existing.Status = *desiredStatus
+
+		return true, nil
+	}
+}
+
```
mikenairn, line 246: Is this any different from the `v1beta1.KuadrantStatusMutator` and `v1beta2.AuthPolicyStatusMutator`version? Are they likely to ever differ between polices or change between versions of the API?

api/v1alpha1/dnspolicy_types.go
```diff
@@ -134,19 +139,14 @@ type LoadBalancingGeo struct {
 
 // DNSPolicyStatus defines the observed state of DNSPolicy
 type DNSPolicyStatus struct {
+	reconcilers.StatusMeta `json:",inline"`
```
mikenairn, line 142: Should `StatusMeta` not be in the versioned API? What is the purpose of it? Is it intended to be the common expected fields for the status on all policies? If so, should "Conditions" be in there also?

mikenairn (commented):
Couple of questions, but generally i have no issue with trying to make all the polices work in the same way where possible.

api/v1alpha1/dnspolicy_types.go
eguzki: I do not think so. What do you think? Should I make this mutator abstract and reuse for all the types?

api/v1alpha1/dnspolicy_types.go
eguzki: you are right, this is leftovers of a previous try that did not succeed. I will remove it.

eguzki:
I tried to simplify, but in the end, all the policies had to implement the same reconciliation of the status, as it is now... Thus not being a good progress, closing this.

PR #774, opened 2024-07-23
fix flaky test: TestNewGatewayEventMapper

eguzki:


pkg/library/mappers/gateway_test.go
```diff
@@ -90,7 +91,11 @@ func TestNewGatewayEventMapper(t *testing.T) {
 			},
 		}
 		requests := em.MapToPolicy(context.Background(), gateway, &kuadrantv1beta2.AuthPolicy{})
-		expected := []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: "app-ns", Name: "policy-1"}}, {NamespacedName: types.NamespacedName{Namespace: "app-ns", Name: "policy-2"}}}
-		assert.DeepEqual(subT, expected, requests)
+		assert.Assert(subT, utils.Index(requests, func(r reconcile.Request) bool {
```
eguzki, line 94: order does not matter and it was enforced by the test

KevFan (approved):

PR #777, opened 2024-07-23
Object types

eguzki:
### What

* Policy types

```go
type PolicyType interface {
	GetGVK() schema.GroupVersionKind
	GetInstance() client.Object
	GetList(context.Context, client.Client, ...client.ListOption) ([]Policy, error)
	BackReferenceAnnotationName() string
	DirectReferenceAnnotationName() string
}
```

* `Type` interface for all Gateway API types

```go
type Type interface {
	GetGVK() schema.GroupVersionKind
	GetInstance() client.Object
}
```

* Simplified the `Policy` interface: it should not have anything related to kuadrant.

Useful types to write common controller mappers and controllers that can be used by any policy type. 

This is part of a series of PR's that will end up with a new controller that will only reconcile limitador limits configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,9 +302,12 @@ clean-cov: ## Remove coverage reports
 ifdef TEST_NAME
 test-unit: TEST_PATTERN := --run $(TEST_NAME)
 endif
+ifdef VERBOSE
+test-unit: VERBOSE_FLAG = -v
+endif
 test-unit: clean-cov generate fmt vet ## Run Unit tests.
 	mkdir -p $(PROJECT_PATH)/coverage/unit
-	go test $(UNIT_DIRS) -coverprofile $(PROJECT_PATH)/coverage/unit/cover.out -tags unit -v -timeout 0 $(TEST_PATTERN)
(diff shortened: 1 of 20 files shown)


eguzki:
@Boomatang I would appreciate your :eyes: on this PR :bowing_man:

Boomatang (commented):
I see what is being done and kinda why it is being done. What I am concerned about is the adding of yet more interfaces and structs that we need to be aware of. 

My mind is in two places over wide interfaces compared to deep interfaces. What is your view on the abstraction and abstracting to early?

eguzki:
> I see what is being done and kinda why it is being done. What I am concerned about is the adding of yet more interfaces and structs that we need to be aware of.
> 
> My mind is in two places over wide interfaces compared to deep interfaces. What is your view on the abstraction and abstracting to early?

I plan to  write common (reusable) controller mappers and generic (reusable) controllers that can be used by any policy type. This is kind of preparation work for it.

Boomatang (changes requested):
Looks good to me and is a good stepping stone to something better

PR #780, opened 2024-07-24
Update Gateway API to v1.1

### LINKED ISSUE #705: Switch to using LocalPolicyTargetReference struct instead
From gateway api v1.1.0, the `PolicyTargetReference` struct that we were using had been removed but introduced several types of policy target references including `LocalPolicyTargetReference`:

- https://github.com/kubernetes-sigs/gateway-api/blob/690f754646e8326128fd686e3e46117ac479cfdf/apis/v1alpha2/policy_types.go#L34-L48

When we upgrade to use Gateway API v1.1.0, we shoudl consider using `LocalPolicyTargetReference` instead as it does not include the optional `namespace` field as we do not support setting the namespace field to any other the policy namespace.

This will remove the need for the cross namespace validation code in the policies:
- https://github.com/Kuadrant/kuadrant-operator/blob/d2e301b29fbff0b5b39edc487cf0e4d6a763b95e/api/v1alpha1/dnspolicy_types.go#L204-L206
- https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1alpha1/tlspolicy_types.go#L186-L188
- https://github.com/Kuadrant/kuadrant-operator/blob/d2e301b29fbff0b5b39edc487cf0e4d6a763b95e/api/v1beta2/authpolicy_types.go#L275-L277
- https://github.com/Kuadrant/kuadrant-operator/blob/d2e301b29fbff0b5b39edc487cf0e4d6a763b95e/api/v1beta2/ratelimitpolicy_types.go#L220-L222

### LINKED ISSUE #732: Move to 1.1 of Gateway API
**What**

Currently we use v1.0 of Gateway API. The latest version available is 1.1 we should move to using that version in our installs and local development

adam-cattermole:
## Changes

- Gateway API v1.1
- Use of `LocalPolicyTargetReference`
- (update to the pinned sail-operator go module due to conflict)

Resolves #732, resolves #705


maleck13 (approved):
Code changes look good to me

maleck13:
@adam-cattermole I think it would be worth checking the quick start docs and docs repo also to see do we refer to installing a particular version of Gateway API

KevFan (approved):
Changes looks good to me 👍

didierofrivia (approved):
LGTM, after your explanation regarding `LocalPolicyTargetReference` that is the reason for removing the namespace tests. You'll need to rebase with main tho. 🥧

go.mod
```diff
@@ -1,39 +1,39 @@
 module github.com/kuadrant/kuadrant-operator
 
-go 1.22
+go 1.22.0
```
eguzki, line 3: this was intended?

eguzki (changes requested):

go.mod
adam-cattermole: Yes, as of 1.22 the toolchain directive is added if the go version is not specific. I.e. if I run `go mod tidy` with go version 1.22.5 it updates the go.mod to the following:

```
go 1.22.0

toolchain go1.22.5
```
The only way to avoid is adding the full version afaict

eguzki: TIL this. Thanks.

eguzki (approved):

PR #788, opened 2024-07-30
Update default template with latest manifests

didierofrivia:


adam-cattermole (approved):
Changes lgtm!

Maybe we can have the helm manifests auto-generated by the existing manifests make target (for simplicity), or/and have a check on PR to make sure we update these when we make a change to the manifests

PR #790, opened 2024-07-30
Transition to Cross-Compilation for faster Docker Build Times

### LINKED ISSUE #787: Optimise Docker Build Time by Using Cross-Compilation Instead of QEMU for Multi-Architecture Builds
### Issue:
I noticed when working with GitHub Actions for building Docker images that the build times were quite long, often exceeding 20 minutes. This delay is largely due to the use of QEMU for emulation, which, while supporting multi-architecture builds, significantly slows down the process.

### Proposal:
To reduce build times and enhance efficiency, I propose switching from QEMU emulation to cross-compilation. Cross-compilation can significantly speed up the Docker build process by removing the overhead associated with QEMU emulation.

GitHub plans to offer free ARM runners later in the year, making the need for cross-compilation no longer relevant. However, in the meantime, I believe there is significant upside to improving our development workflows, shortening feedback loops, and speeding up deployment cycles.

An example can be found in a merged PR for limitador: [refactor: cross compile for arm64 image build ](https://github.com/Kuadrant/limitador/pull/222url)

dlaw4608:
closes #787 

## This pull request introduces significant improvements to the Docker build process by implementing cross-compilation.

Building arm64 image via QEMU takes over 15 mins on github actions, making it slow to build for development.

image

Building the image via cross compile dropped this build time to 2 min 30

image

## Changes Made 

1. **Dockerfile Updates**:
    - Introduced cross-compilation support by using `ARG TARGETARCH`.
    - Utilized `--platform=$BUILDPLATFORM` for multi-platform builds with Docker Buildx.
    
2. **Workflow File**:
    - Updated the workflow to leverage Docker Buildx for building and pushing images to multiple platforms (linux/amd64, linux/arm64).
    - Removed reliance on QEMU emulation, significantly speeding up the build process.

## Benefits

- **Faster Build Times**: The new build process reduces the time from approximately 16 minutes to just 2 minutes and 30 seconds, representing an almost 8x improvement.
- **Optimised Build Efficiency**: Reduces overhead and simplifies the build process by eliminating the need for emulation.

### DIFF
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -85,31 +85,27 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v3
-      - name: Install qemu dependency
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y qemu-user-static
-      - name: Build Image
-        id: build-image
-        uses: redhat-actions/buildah-build@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to container registry
+        uses: docker/login-action@v2
         with:
-          image: ${{ env.OPERATOR_NAME }}
-          tags: ${{ env.IMG_TAGS }}
-          platforms: linux/amd64,linux/arm64
-          dockerfiles: |
-            ./Dockerfile
-      - name: Push Image
-        if: ${{ !env.ACT }}
-        id: push-to-quay
-        uses: redhat-actions/push-to-registry@v2
-        with:
-          image: ${{ steps.build-image.outputs.image }}
-          tags: ${{ steps.build-image.outputs.tags }}
-          registry: ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}
           username: ${{ secrets.IMG_REGISTRY_USERNAME }}
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Build and Push Image
+        id: build-image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          provenance: false
+          tags: |
+            ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}:${{ env.IMG_TAGS }}
       - name: Print Image URL
-        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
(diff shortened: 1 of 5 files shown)


.github/workflows/build-images-base.yaml
```diff
@@ -85,31 +85,26 @@ jobs:
...
           password: ${{ secrets.IMG_REGISTRY_TOKEN }}
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Build and Push Image
+        id: build-image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
```
KevFan, line 103:
```suggestion
          push: true
          provenance: false
```

There looks to be an issue with producing unknown manifest when building by cross compile
- https://github.com/docker/build-push-action/issues/820

image

We should probably set `provenance: false` here and other places to where an image is built to prevent this

.github/workflows/build-images-base.yaml
```diff
@@ -138,27 +129,26 @@ jobs:
...
-        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Build and Push Bundle Image
+        id: build-bundle-image
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: ./bundle.Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
```
KevFan, line 147:
```suggestion
          push: true
          provenance: false
```

.github/workflows/build-images-base.yaml
```diff
@@ -182,29 +172,23 @@ jobs:
...
-        run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"
+          registry: ${{ env.IMG_REGISTRY_HOST }}
+      - name: Build and Push Catalog Image
+        id: build-catalog-image
+        uses: docker/build-push-action@v5
+        with:
+          context: ./catalog
+          file: ./catalog/kuadrant-operator-catalog.Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
```
KevFan, line 190:
```suggestion
          push: true
          provenance: false
```

.github/workflows/build-images-base.yaml
dlaw4608: Yeah that's a strange one. adding  provenance and setting it to false seemed to have fixed the issue,

controllers/dnspolicy_status.go
KevFan, line 1: Rebase issue? I dont think this file and the files changes in `tests/common/dnspolicy` should be part of this PR 🤔

dlaw4608: Ah I must have squashed @makslion's commit into my own, will fix.

PR #795, opened 2024-08-01
Remove reconcile of subcomponent configuration

Boomatang:
With the move to GA and a stable API this work is being reverted so to allow for a more action based approach to be taken in the future without introducing breaking changes to the API.


KevFan (approved):
Looks good to me 👍

PR #796, opened 2024-08-01
Release button

didierofrivia:
This PR introduces a new workflow for crafting a release, following the pattern used in [Authorino Operator](https://github.com/Kuadrant/authorino-operator/blob/main/.github/workflows/release.yaml). It makes it convenient to trigger the workflow manually without the need to go through the process of creating a branch and manually triggering the make targets, committing and tagging.

It includes the building of both OLM manifests and Helm Charts.


guicassolato (approved):
LGTM

PR #803, opened 2024-08-14
Fix release workflow

didierofrivia:
Fixing the workflow adding the missing `prerelease` input and the correct version


adam-cattermole (approved):
lgtm

Issue #806, opened 2024-08-15
State-of-the-world reconciler


guicassolato:
Implement a state-of-the-world reconciler based on Policy Machinery’s [`controller`](https://github.com/Kuadrant/policy-machinery/tree/main/controller) package.

Scope:
- Define watchers for:
  - Kuadrant CRs
  - Kuadrant policy kinds: DNSPolicy, TLSPolicy, AuthPolicy, RateLimitPolicy
  - Gateway API kinds: GatewayClass, Gateway, HTTPRoute
  - Gateway provider (internal) resources: Istio WasmPlugin
  - Other object kinds: Secrets
- Define custom link functions for internal/other generic object kinds (i.e. non-targetables, non-policies)
- Base workflow reconciler that simply logs the event initially (placeholder for next steps)
- Options for:
  - Metrics server
  - Webhook server
  - Health check and readiness probe
  - Leader election

Requires:
- https://github.com/Kuadrant/policy-machinery/issues/15

Supersedes:
- #203
- #483

Issue #831, opened 2024-08-16
[state-of-the-world reconciler] Topology ConfigMap


guicassolato:
Reconciliation of the topology graph represented in DOT language stored into a read-only ConfigMap resource.

guicassolato:
@jasonmadigan, can you please confirm if a ConfigMap is OK for storing the topology graph given the applications in sight for it?

We had discussed off-line about defining a new CRD for storing the topology graph. In the last community call, it was proposed to use a ConfigMap instead, to avoid the burden of managing yet another CRD.

jasonmadigan:
@guicassolato config-map works for me, would be just as accessible

PR #838, opened 2024-08-26
State of the world reconciler

Boomatang:
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/806
Partial: https://github.com/Kuadrant/kuadrant-operator/issues/831

This introduces the [policy-machinery](https://github.com/Kuadrant/policy-machinery) package and its concept of workflows. The workflow is very basic and only logs events in the `precondition`. For `tasks` the workflows creates a configmap in the same namespace as the Kuadrant CR with the topology of the cluster saved as dot format. 

All existing controllers still work the same and there should be no change to the existing functionality of the operator.

Generate a local dot file from the cluster:
```sh
watch "kubectl get cm topology -n kuadrant-system -o jsonpath='{.data.topology}' > topology.dot"
```

Run gist to get sample cluster set up
```sh
kubectl apply -f https://gist.githubusercontent.com/Boomatang/15ae1b31f25898553cb0fee5838b5e86/raw/51282b238dffe7573bf52ef55b04eded8347134d/state_of_the_world.yaml
```


controllers/state_of_the_world.go
```diff
@@ -0,0 +1,170 @@
...
+			Name:      "topology",
+			Namespace: "kuadrant-system",
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+
+	configMapRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
+	unstructuredMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(cm)
```
guicassolato, line 141: Can you use [`controller.Destruct`](https://github.com/Kuadrant/policy-machinery/blob/e963418c9566a5e61acdd766c6109d3f0c147222/controller/runnable.go#L213) here maybe?

```suggestion
	unstructuredMap, _ := controller.Destruct(cm)
```

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,170 @@
...
+		controller.WithRunnable("kuadrant watcher", buildWatcher(&kuadrantv1beta1.Kuadrant{}, kuadrantv1beta1.KuadrantResource, metav1.NamespaceAll)),
+		controller.WithRunnable("gatewayclass watcher", buildWatcher(&gwapiv1.GatewayClass{}, controller.GatewayClassesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("gateway watcher", buildWatcher(&gwapiv1.Gateway{}, controller.GatewaysResource, metav1.NamespaceAll)),
+		controller.WithRunnable("httproute watcher", buildWatcher(&gwapiv1.HTTPRoute{}, controller.HTTPRoutesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("dnspolicy watcher", buildWatcher(&kuadrantv1alpha1.DNSPolicy{}, kuadrantv1alpha1.DNSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("tlspolicy watcher", buildWatcher(&kuadrantv1alpha1.TLSPolicy{}, kuadrantv1alpha1.TLSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("authpolicy watcher", buildWatcher(&kuadrantv1beta2.AuthPolicy{}, kuadrantv1beta2.AuthPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("ratelimitpolicy watcher", buildWatcher(&kuadrantv1beta2.RateLimitPolicy{}, kuadrantv1beta2.RateLimitPoliciesResource, metav1.NamespaceAll)),
```
guicassolato, line 41: Add the ConfigMaps here, maybe with a label predicate to focus on the Topology resource?

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		panic(err.Error())
+	}
+	unstructuredCM := &unstructured.Unstructured{
+		Object: unstructuredMap,
+	}
+
+	_, err = r.Client.Resource(configMapRes).Namespace(cm.Namespace).Update(context.TODO(), unstructuredCM, metav1.UpdateOptions{})
+	if errors.IsNotFound(err) {
+		_, err = r.Client.Resource(configMapRes).Namespace(cm.Namespace).Create(context.TODO(), unstructuredCM, metav1.CreateOptions{})
```
guicassolato, line 151: I see here a bit of the pattern we want to avoid with the topology, I guess. Rather, I think the controller could be watching ConfigMaps as well, making sure the `kuadrant-system/topology` ConfigMap object is represented in the topology struct (linked to the Kuadrant CR?) Then, calling `Create` vs `Update` would depend on the presence of the object in the topology. WDYT?

main.go
```diff
@@ -279,18 +280,24 @@ func main() {
...
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err = mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
 
-	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
+	client, err := dynamic.NewForConfig(ctrl.GetConfigOrDie())
```
KevFan, line 292: Don't think it matters, but not sure do we want to get the `rest.Config` from the `mgr` instead here since it's available through it
```suggestion
	client, err := dynamic.NewForConfig(mgr.GetConfig())
```

Boomatang: Sure no problem

controllers/state_of_the_world.go
Boomatang: Okay I have done a bit of a refactor here. I think now it lines up better with what you were expecting.

controllers/state_of_the_world.go
Boomatang: I have done this and dropped the checking of the error. I am not sure how I feel about not doing something with the errors that may come back.

api/v1beta1/topology.go
```diff
@@ -0,0 +1,56 @@
...
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[*Kuadrant])
+
+	return machinery.LinkFunc{
+		From: KuadrantKind,
+		To:   schema.GroupKind{Group: corev1.SchemeGroupVersion.Group, Kind: "ConfigMap"},
+		Func: func(child machinery.Object) []machinery.Object {
+			o := child.(*controller.RuntimeObject)
+			cm := o.Object.(*corev1.ConfigMap)
+			return lo.FilterMap(kuadrants, func(kuadrant *Kuadrant, _ int) (machinery.Object, bool) {
+				if cm.Name == "topology" {
```
guicassolato, line 49: I guess we don't need to filter, right? Maybe simply return the slice of Kuadrants when the ConfigMap is the `topology` one and `nil` otherwise?

I'm also understanding we you want one single `topology` CM, rather than one (sliced) for each Kuadrant CR?

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,132 @@
...
+	"k8s.io/client-go/dynamic"
+	ctrlruntime "sigs.k8s.io/controller-runtime"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+)
+
+var kuadrant_namespace = "kuadrant-system"
```
Boomatang, line 21: I don't like the idea of  a hardcoded namespace, but I am not sure how we can best get the namespace the operator is installed into at the early stage when setting up the runables

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,132 @@
...
+		controller.WithRunnable("gatewayclass watcher", buildWatcher(&gwapiv1.GatewayClass{}, controller.GatewayClassesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("gateway watcher", buildWatcher(&gwapiv1.Gateway{}, controller.GatewaysResource, metav1.NamespaceAll)),
+		controller.WithRunnable("httproute watcher", buildWatcher(&gwapiv1.HTTPRoute{}, controller.HTTPRoutesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("dnspolicy watcher", buildWatcher(&kuadrantv1alpha1.DNSPolicy{}, kuadrantv1alpha1.DNSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("tlspolicy watcher", buildWatcher(&kuadrantv1alpha1.TLSPolicy{}, kuadrantv1alpha1.TLSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("authpolicy watcher", buildWatcher(&kuadrantv1beta2.AuthPolicy{}, kuadrantv1beta2.AuthPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("ratelimitpolicy watcher", buildWatcher(&kuadrantv1beta2.RateLimitPolicy{}, kuadrantv1beta2.RateLimitPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("configmap watcher", buildWatcher(&corev1.ConfigMap{}, controller.ConfigMapsResource, kuadrant_namespace)),
```
guicassolato, line 39: If the only ConfigMap we care about now is the `topology` CM, how about labeling the resource and using a [label filter](https://github.com/Kuadrant/policy-machinery/blob/e963418c9566a5e61acdd766c6109d3f0c147222/controller/runnable.go#L43) predicate here?

controllers/state_of_the_world.go
guicassolato, line 21:
```suggestion
var kuadrantNamespace = "kuadrant-system"
```

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,132 @@
...
+
+func buildWatcher[T controller.Object](obj T, resource schema.GroupVersionResource, namespace string, options ...controller.RunnableBuilderOption[T]) controller.RunnableBuilder {
+	return controller.Watch(obj, resource, namespace, options...)
+}
+
+func buildReconciler(client *dynamic.DynamicClient) controller.ReconcileFunc {
+	topologyFileReconciler := TopologyFileReconciler{Client: client}
+
+	reconciler := &controller.Workflow{
+		Precondition: func(ctx context.Context, resourceEvents []controller.ResourceEvent, _ *machinery.Topology) {
```
guicassolato, line 67: To clean this a little, maybe define an `eventLogger`?

guicassolato: E.g.:

```go
	return (&controller.Workflow{
		Precondition: (&EventLogger{}).Log,
			(&TopologyFileReconciler{Client: client}).Reconcile,
		},
	}).Run
```

main.go
```diff
@@ -279,18 +280,24 @@ func main() {
...
-		setupLog.Error(err, "problem running manager")
+	client, err := dynamic.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		setupLog.Error(err, "unable to create client")
+		os.Exit(1)
+	}
+
+	sateOfTheWorld := controllers.SetupWithManagerA(mgr, client)
+	if err = sateOfTheWorld.Start(ctrl.SetupSignalHandler()); err != nil {
+		setupLog.Error(err, "unable to start sateOfTheWorld controller")
```
KevFan, line 300:
```suggestion
	stateOfTheWorld := controllers.SetupWithManager(mgr, client)
	if err = stateOfTheWorld.Start(ctrl.SetupSignalHandler()); err != nil {
		setupLog.Error(err, "unable to start stateOfTheWorld controller")
```

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,132 @@
...
+	})
+
+	if len(targets) == 0 {
+		_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Create(context.TODO(), unstructuredCM, metav1.CreateOptions{})
+		if err != nil {
+			logger.Error(err, "failed to write topology configmap")
+		}
+		return
+	}
+	_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Update(context.TODO(), unstructuredCM, metav1.UpdateOptions{})
```
KevFan, line 128: Let's use the passed in context in the client calls rather than a new context
```suggestion
	logger := controller.LoggerFromContext(ctx).WithName("topology file")
	cm := &corev1.ConfigMap{
		ObjectMeta: metav1.ObjectMeta{
			Name:      "topology",
			Namespace: kuadrantNamespace,
		},
		Data: map[string]string{
			"topology": topology.ToDot(),
		},
	}

	configMapRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
	unstructuredCM, _ := controller.Destruct(cm)

	targets := topology.Objects().Items(func(object machinery.Object) bool {
		if object.GetName() == cm.GetName() && object.GetNamespace() == cm.GetNamespace() {
			return true
		}
		return false
	})

	if len(targets) == 0 {
		_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Create(ctx, unstructuredCM, metav1.CreateOptions{})
		if err != nil {
			logger.Error(err, "failed to write topology configmap")
		}
		return
	}
	_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Update(ctx, unstructuredCM, metav1.UpdateOptions{})
```

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,132 @@
...
+	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+)
+
+var kuadrantNamespace = "kuadrant-system"
+
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gatewayclasses,verbs=list;watch
+
+func SetupWithManagerA(manager ctrlruntime.Manager, client *dynamic.DynamicClient) *controller.Controller {
```
KevFan, line 25:
```suggestion
func SetupWithManager(manager ctrlruntime.Manager, client *dynamic.DynamicClient) *controller.Controller {
```

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+
+	configMapRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
+	unstructuredCM, _ := controller.Destruct(cm)
+
+	targets := topology.Objects().Items(func(object machinery.Object) bool {
```
eguzki, line 94: this lookup method does not seem very efficient. Does the topology have a method to find object by ID?

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,139 @@
...
+	})
+
+	if len(targets) == 0 {
+		_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Create(ctx, unstructuredCM, metav1.CreateOptions{})
+		if err != nil {
+			logger.Error(err, "failed to write topology configmap")
+		}
+		return
+	}
+	_, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Update(ctx, unstructuredCM, metav1.UpdateOptions{})
```
eguzki, line 108: no checking if the topology has been changed? update regardless?

eguzki (commented):
looking good

main.go
```diff
@@ -279,18 +280,24 @@ func main() {
...
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err = mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
 
-	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-		setupLog.Error(err, "problem running manager")
+	client, err := dynamic.NewForConfig(mgr.GetConfig())
```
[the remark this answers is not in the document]
guicassolato: I don't think there's anything missing in controller-runtime's. Rather, it's more about what it adds.

Right now, given how the Policy Machinery was designed, it doesn't add a lot (though also not nothing.) Generally the same operations are available; the dynamic client is also a cache client for read operations; and the structuring of objects is already built-in and thought for all read operations post-construction of the topology to first target the topology object.

IOW, no hard reason to depend on controller-runtime here IMO.

Maybe a couple diffs worth mentioning are:
| controller-runtime | kube/client-go |
|-|-|
| read/write separation | single client, though with reading operations discouraged as a pattern |
| implicit restructuring of resources from the concrete the Go types | uses declared [resources](https://github.com/Kuadrant/policy-machinery/blob/main/controller/resources.go) and explicit un/marshalling of `Unstructured`, though already handled built-in for the topology and [helper](https://github.com/Kuadrant/policy-machinery/blob/c4c99ba6f9e97306b4c3c1f3fcbb2825f561305d/controller/runnable.go#L213) provided for write operations |

controllers/state_of_the_world.go
guicassolato: It should check before.

controllers/state_of_the_world.go
guicassolato: Sounds like a good addition to have in [Kuadrant/policy-machinery](https://github.com/Kuadrant/policy-machinery) to me.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,139 @@
...
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+)
+
+var kuadrantNamespace = "kuadrant-system"
+
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gatewayclasses,verbs=list;watch
+
+func SetupWithManager(manager ctrlruntime.Manager, client *dynamic.DynamicClient) *controller.Controller {
+	logger := controller.CreateAndSetLogger()
```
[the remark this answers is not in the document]
guicassolato: Yep. Logger should be passed in.

main.go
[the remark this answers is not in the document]
eguzki: The code using this client seems a lot more confusing and cumbersome. Just a personal opinion. Additionally no need for a explicit un/marshalling. 

What do others think about this? cc @Boomatang @KevFan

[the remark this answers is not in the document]
guicassolato: To be clear, I'm not against using controller-runtime's client, but if we do (off the top of my head):
1. The change to expect a controller-runtime client associated with the controller (instead of client-go's <a href="https://pkg.go.dev/k8s.io/client-
go@v0.30.2/dynamic#DynamicClient"><code>dynamic.DynamicClient</code></a>) needs to be done in the Policy Machinery first.

2. We cannot use [`mgr.GetClient()`](https://github.com/Kuadrant/kuadrant-operator/blob/b7a58cab263e38b105fcde5dc9670bfe73a73e05/main.go#L146C3-L146C18) as this returns a controller-runtime's [`cluster.Client`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.16.3/pkg/cluster#Cluster.GetClient). Instead, we'd need a new client created using [`NewWithWatch`](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#NewWithWatch).

3. In the Policy Machinery, we will no longer be able to instantiate runnables (informers/reconcilers) only from the singular type; we'd need to pass along in all the calls to [`controller.Watch[T Object](obj T, …)`](https://github.com/Kuadrant/policy-machinery/blob/c4c99ba6f9e97306b4c3c1f3fcbb2825f561305d/controller/runnable.go#L61) also an instance of a List type – i.e. sth like `controller.Watch[O Object, L ctrlruntimeclient.ObjectList](obj O, listObj L, …)`. This means a fairly heavy refactoring of https://github.com/Kuadrant/policy-machinery/blob/main/controller/runnable.go

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,139 @@
...
+	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+)
+
+var kuadrantNamespace = "kuadrant-system"
+
+//+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gatewayclasses,verbs=list;watch
+
+func SetupWithManager(manager ctrlruntime.Manager, client *dynamic.DynamicClient) *controller.Controller {
```
guicassolato, line 25: I got tricked by the name of this function. WDYT about renaming it to one of the following?
- `NewPolicyController`
- `NewPolicyMachineryController`
- `NewStateOfTheWorldReconciler`

.gitignore
```diff
@@ -26,10 +26,11 @@ testbin/*
 *~
 /kuadrant-operator
 tmp
+topology.dot
```
guicassolato, line 29: Probably no longer needed since the topology is saved to a ConfigMap?

main.go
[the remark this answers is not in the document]
eguzki: I am happy to do that heavy refactor in the policy machinery lib if people do not like doing

```go
cm := &corev1.ConfigMap{ ... }                                                                                                                          
                                                                                                                           
configMapRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}                              
unstructuredCM, _ := controller.Destruct(cm)                                                                               
 _, err := r.Client.Resource(configMapRes).Namespace(cm.Namespace).Create(ctx, unstructuredCM, metav1.CreateOptions{})  
```

I am here to serve people command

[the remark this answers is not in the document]
Boomatang: I do think using the `DynamicClient` is cumbersome to a point but I also know this is the first time I have had to use it so it is also different. I don't know know if the lack of practice with the client leads to it feeling cumbersome. We always know there was going to be a learning curve with doing these changes so I am happy to let it cook. 

There is all my natural dislike of the controller-runtime. I believe it is a dependency that we can live without and using the `DynamicClient` reduces that dependency which can only be a good thing.

controllers/state_of_the_world.go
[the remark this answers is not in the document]
Boomatang: Working on this

controllers/state_of_the_world.go
Boomatang: Working on this.

controllers/state_of_the_world.go
Boomatang: Ya I wasn't overly happy with the name my self. The `NewPolicyMachineryController` seem to be the best fit when looking at it here.

.gitignore
Boomatang: sure can do.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,149 @@
...
+		controller.WithRunnable("gateway watcher", controller.Watch(&gwapiv1.Gateway{}, controller.GatewaysResource, metav1.NamespaceAll)),
+		controller.WithRunnable("httproute watcher", controller.Watch(&gwapiv1.HTTPRoute{}, controller.HTTPRoutesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("dnspolicy watcher", controller.Watch(&kuadrantv1alpha1.DNSPolicy{}, kuadrantv1alpha1.DNSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("tlspolicy watcher", controller.Watch(&kuadrantv1alpha1.TLSPolicy{}, kuadrantv1alpha1.TLSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("authpolicy watcher", controller.Watch(&kuadrantv1beta2.AuthPolicy{}, kuadrantv1beta2.AuthPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("ratelimitpolicy watcher", controller.Watch(&kuadrantv1beta2.RateLimitPolicy{}, kuadrantv1beta2.RateLimitPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("configmap watcher", controller.Watch(&corev1.ConfigMap{}, controller.ConfigMapsResource, kuadrantNamespace, controller.FilterResourcesByLabel[*corev1.ConfigMap]("kuadrant.io/topology=true"))),
```
guicassolato, line 40: Could store the name of the name label in a constant, to avoid mismatching where the resource is created/reconciled.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	configMapRes := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
+	unstructuredCM, _ := controller.Destruct(cm)
+
+	targets := topology.Objects().Items(func(object machinery.Object) bool {
+		if object.GetName() == cm.GetName() && object.GetNamespace() == cm.GetNamespace() {
+			return true
+		}
+		return false
```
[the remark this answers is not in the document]
guicassolato: You probably want to filter by kind too:

```suggestion
		_, ok := o.(*corev1.ConfigMap)
		return ok && object.GetName() == cm.GetName() && object.GetNamespace() == cm.GetNamespace()

```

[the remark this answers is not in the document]
guicassolato: Or yet the following.

Say you name the desired topology ConfiMap `topologyConfigMap` instead of `cm`

```go
topologyConfigMap := &corev1.ConfigMap{
	ObjectMeta: metav1.ObjectMeta{
		Name:      "topology",
		Namespace: kuadrantNamespace,
		Labels:    map[string]string{"kuadrant.io/topology": "true"},
	},
	Data: map[string]string{
		"topology": topology.ToDot(),
	},
}

[…]

topologyConfigMaps := lo.FilterMap(objs, func(obj machinery.Object, _ int) (*corev1.ConfigMap, bool) {
	cm, ok := obj.(*corev1.ConfigMap)
	return cm, ok && cm.GetNamespace() == topologyConfigMap.GetNamespace() && cm.GetName() == topologyConfigMap.GetName()
})

if len(topologyConfigMaps) == 0 {
	// …
}
…
```

I guess this is the option I like the most. WDYT?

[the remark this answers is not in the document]
guicassolato: To be even more accurate in the search if you want, given the topology ConfigMap must be linked to the Kuadrant object, first find the Kuadrant object in the topology, then...

```go
existingTopologyConfigMaps := lo.FilterMap(topology.Objects().Children(kuadrantObj), …)
```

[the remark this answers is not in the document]
guicassolato: Finally, just a friendly reminder that the Kuadrant object is a root object in the topology. So...

```go
kuadrantObject := topology.Objects().Root()[0]
```

🙂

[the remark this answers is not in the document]
Boomatang: A couple of things here. Firstly the refactor to have to remove the if block and have a direct return is of course something that I can do.

For the rest of the suggestions I have issue with them.

The checking if the object is in fact a kind of configmap makes sense but when I try the type assertion I am getting the following error: `Impossible type assertion: '*corev1.ConfigMap' does not implement 'machinery.Object'`. It is quite possible I am doing something wrong but I would assume `_, ok := object.(*corev1.ConfigMap)` would be the correct way.

If the preferred filter method is use `lo.FitlerMap(topology.Objects().Items(), func()....` do we have a design flaw in the policy-machinery by allowing the user to pass in a filter function into `topology.Objects().Items()`?

The last point is on `kuadrantObject := topology.Objects().Root()[0]`, this may not be true. The topology allows for multiple kuadrant CRs. While currently today the operator only support one instance of the Kuadrant CR, I am not sure how long this will hold true. I know there has being talks about supporting multiple gateways by using the different Kuadrant CRs for that. I am well aware that this will not be happening for a long time, but we would be wrong to make the assumption that there will only ever be one Kuadrant CR.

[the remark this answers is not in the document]
guicassolato: > The checking if the object is in fact a kind of configmap makes sense but when I try the type assertion I am getting the following error: `Impossible type assertion: '*corev1.ConfigMap' does not implement 'machinery.Object'`. It is quite possible I am doing something wrong but I would assume `_, ok := object.(*corev1.ConfigMap)` would be the correct way.

Actually, the `Impossible type assertion` error makes sense. This is because the object is actually a [`controller.RuntimeObject`](https://github.com/Kuadrant/policy-machinery/blob/85837a3f6a5ae4b84594de2124c5338f678a57f3/controller/object.go#L17) which wraps a `*corev1.ConfigMap`. It cannot be a `*corev1.ConfigMap` indeed. So my previous suggestion was misleading. Instead, you'd need something like this:

```go
ro, ok := obj.(*RuntimeObject)
if !ok
	return nil, false
end
cm, ok := ro.Object.(*corev1.ConfigMap)
return cm, ok && cm.GetNamespace() == topologyConfigMap.GetNamespace() && cm.GetName() == topologyConfigMap.GetName()
```

Another option is checking the `GroupKind`, tho you'd still need to make the conversion to a ConfigMap in a separate step:

```go
obj.GroupVersionKind().GroupKind() == ConfigMapGroupKind
```

<br/>

> If the preferred filter method is use `lo.FitlerMap(topology.Objects().Items(), func()....` do we have a design flaw in the policy-machinery by allowing the user to pass in a filter function into `topology.Objects().Items()`?

I don't see it as a design flaw. If all you want is filtering, then you can pass a predicate to the `Items()` func. I proposed using `lo.FilterMap` instead because in this case not only we want to filter, but also we want the objects converted to `*corev1.ConfigMap` if possible.

That said, if this is something we believe we'll be doing a lot, it may be a good addition to the Policy Machinery.

<br/>

> The last point is on `kuadrantObject := topology.Objects().Root()[0]`, this may not be true. The topology allows for multiple kuadrant CRs. While currently today the operator only support one instance of the Kuadrant CR, I am not sure how long this will hold true. I know there has being talks about supporting multiple gateways by using the different Kuadrant CRs for that. I am well aware that this will not be happening for a long time, but we would be wrong to make the assumption that there will only ever be one Kuadrant CR.

This is the hardest one to unpack IMO.

First, we need to decide if we want to allow more than one Kuadrant object per cluster or not and what to do in case of not allowing it.

Allowing multiple Kuadrant _objects_ does not necessarily mean allowing multiple Kuadrant _instances_. We can let users create as many Kuadrant CRs they want, but commit to reconcile only the first that gets created, leaving all the others stale (maybe just updating their status). These objects would be loose roots to which nothing would be linked.

Another option is not to allow multiple Kuadrant objects. Be harsh by perhaps making the reconciliation workflow to force delete any second CR of this kind and on, keeping only one present at a time. Not very user-friendly, but a possibility nonetheless.

Either way, the reason to structure the topology as a DAG is so we can navigate it from the root, "making assumptions" about the structure.

If we have to list and filter all resources then not only this could be computationally costly, but also we might very well have simply listed the resources in a map.

I guess what I'm trying to say is that we want to read resources from the topology by navigating the DAG, not in the traditional way that the API server would otherwise let us. Each time we use a func such as `Parent` or `Children`, there's a semantic embedded into it, whose counterpart is the linking functions. We want to use those functions and thus ensure the objects handled in the process are where they are supposed to be.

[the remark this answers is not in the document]
eguzki: we have already some code that assumes there is only one kuadrant CR. If that simplifies your work, feel free to assume it. Drop a comment and we can easily spot that later if the assumption happens be no longer true in the future.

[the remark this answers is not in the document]
guicassolato: I agree with @eguzki. Let users create as many Kuadrant CRs as they want, have a well-commented func that filters amongst all roots of the topology and returns the oldest of the Kuadrant CRs only, use this func to link the GatewayClasses to this one Kuadrant CR only and ignore all the others, use the func to retrieve the one and only root of the DAG whenever we want to navigate it (to read other resources).

main.go
```diff
@@ -279,18 +280,24 @@ func main() {
...
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err = mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
 
-	setupLog.Info("starting manager")
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
```
eguzki, line 292: Why remove this `mgr.Start(ctrl.SetupSignalHandler())` ? It's not the responsibility of a controller to do that.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,149 @@
...
+	"k8s.io/client-go/dynamic"
+	ctrlruntime "sigs.k8s.io/controller-runtime"
+	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	kuadrantv1beta2 "github.com/kuadrant/kuadrant-operator/api/v1beta2"
+)
+
+var kuadrantNamespace = "kuadrant-system"
```
eguzki, line 23: The kuadrant namespace is not hardcoded, it could be any.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,149 @@
...
+		return
+	}
+
+	if len(targets) > 1 {
+		logger.Info("multiple topology configmaps found, continuing but unexpected behaviour may occur")
+	}
+	target := targets[0].(controller.Object).(*controller.RuntimeObject)
+	cmTarget := target.Object.(*corev1.ConfigMap)
+
+	if d, found := cmTarget.Data["topology"]; !found || strings.Compare(d, cm.Data["topology"]) != 0 {
```
eguzki, line 116: the string comparing is usually very weak and prone to false positives (yielding positive when actually is not). 

Maybe not for now, but a good improvement is to parse the topology (maybe `topology.FromDot(...)` method needed first) and then compare topologies with another method like `parsedTopology.EqualsTo(currentTopology)`.

main.go
guicassolato: Because now the `stateOfheWorld` controller will start the manager: https://github.com/Kuadrant/policy-machinery/blob/85837a3f6a5ae4b84594de2124c5338f678a57f3/controller/controller.go#L176

> It's not the responsibility of a controller to do that.

The the responsibility of a Policy Machinery controller to do that.

controllers/state_of_the_world.go
Boomatang: I agree, and raised this concern two weeks ago with out any feedback. The reason I have done this is for the topology configmap.  I am not happy watching every namespace on the cluster for configmaps, it feels like an over reach of access. But at the same time at initialization time of the controller I do not know how to get the namespace it is deployed into. If you have  suggestions on how to get the namespace dynamically during start up I will be more than happy to redo this.

controllers/state_of_the_world.go
Boomatang: I agree. As the first pass I think this is acceptable. I think the `EqualsTo` function should be part of the Policy-machinery package.

controllers/state_of_the_world.go
guicassolato: If we're filtering the ConfigMaps by label, watching all namespaces should be fine. Not ideal, I agree, but good enough for now.

main.go
eguzki: I do not think it is the responsibility of a policy machinery controller to do that. What if I am a that library user that wants to use that kuadrant's policy machinery together with another controller or  what if I want to manage the start of the manager myself?

controllers/state_of_the_world.go
eguzki: You can easily get the kuadrant namespace reading kuadrant CRs, pick one with a valid status (the remaining ones should not exist or they are irrelevant) and read the namespace.

api/v1beta2/topology.go
```diff
@@ -0,0 +1,65 @@
+package v1beta2
```
guicassolato, line 1: For v1alpha1 (DNSPolicy and TLSPolicy) and v1beta2 (AuthPolicy and RateLimitPolicy) I think this is fine. For v1alpha2/v1beta3 (and therefore v1 as well), let's make sure these methods are defined along with each API type and not in a separate file as a separate concern, please. Reasoning behind is that the `machinery.Policy` interface is what makes the type a policy, not something we need to do just to comply with the Policy Machinery reqs.

controllers/state_of_the_world.go
guicassolato: A topology may exist before the first Kuadrant CR is created. The "topology" ConfigMap still needs to be reconciled nonetheless.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+
+func NewTopologyFileReconciler(client *dynamic.DynamicClient) *TopologyFileReconciler {
+	return &TopologyFileReconciler{Client: client}
+}
+
+func (r *TopologyFileReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+	logger := controller.LoggerFromContext(ctx).WithName("topology file")
+
+	if len(topology.Objects().Roots()) == 0 {
+		logger.Info("no Kuadrant CR found, can create topology configmap")
```
eguzki, line 83: `can` or `can't`?

eguzki (changes requested):

main.go
guicassolato: It's by design @eguzki. If you want to use a Policy Machinery controller, the manager associated with this controller will be started when you start the controller.

Keep in mind that the proposed pattern here is of the Policy Machinery controller being the one and only controller. One who feels the need for more controllers aside is probably not making enough use of Policy Machinery's reconcile funcs and patterns. Using a `Workflow` with more tasks instead maybe a good idea, for example.

But if you are "that user" who wants to break the pattern by having other controllers on the side, decoupling the starting of the manager, then I'm guessing your two sets of controllers (one set being actually a set of Policy Machinery reconciliation "tasks" and the other set being the collection other kinds of custom controllers not based on the Policy Machinery) are for completely independent "reconciliation realms". In this case, I would recommend creating two completely separate managers then. You can share the same context returned by `ctrl.SetupSignalHandler()` in both if you want. Start the Policy Machinery-based controller in a goroutine and then the other manager blocking the process.

IMHO, thinking in terms of "reconciliation realms" (as opposed to all things are interconnected) has been the core problem we want to fix with this work in the Kuadrant Operator.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+		controller.WithRunnable("gateway watcher", controller.Watch(&gwapiv1.Gateway{}, controller.GatewaysResource, metav1.NamespaceAll)),
+		controller.WithRunnable("httproute watcher", controller.Watch(&gwapiv1.HTTPRoute{}, controller.HTTPRoutesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("dnspolicy watcher", controller.Watch(&kuadrantv1alpha1.DNSPolicy{}, kuadrantv1alpha1.DNSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("tlspolicy watcher", controller.Watch(&kuadrantv1alpha1.TLSPolicy{}, kuadrantv1alpha1.TLSPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("authpolicy watcher", controller.Watch(&kuadrantv1beta2.AuthPolicy{}, kuadrantv1beta2.AuthPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("ratelimitpolicy watcher", controller.Watch(&kuadrantv1beta2.RateLimitPolicy{}, kuadrantv1beta2.RateLimitPoliciesResource, metav1.NamespaceAll)),
+		controller.WithRunnable("configmap watcher", controller.Watch(&corev1.ConfigMap{}, controller.ConfigMapsResource, metav1.NamespaceAll, controller.FilterResourcesByLabel[*corev1.ConfigMap](fmt.Sprintf("%s=true", kuadrant.TopologyAnnotation)))),
```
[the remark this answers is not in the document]
guicassolato: Everything that we watch will be added to the topology graph. We can add watchers for these types now or gradually while we move the logic more directly related to them into the workflow. The sooner we do, the safer we can write code that navigates the topology for implementing any reconciliation task.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (r *TopologyFileReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+	logger := controller.LoggerFromContext(ctx).WithName("topology file")
+
+	if len(topology.Objects().Roots()) == 0 {
+		logger.Info("no Kuadrant CR found, can create topology configmap")
+		return
+	}
+
+	kuadrantCR := topology.Objects().Roots()[0]
```
guicassolato, line 87: Both, the `Kuadrant` kind and the `ConfigMap` kind, are generic `Object`s in the topology (as opposed to `Targetable` and `Policy` kinds). If  the Kuadrant CR does not exist, but the ConfigMap does, then the ConfigMap _is_ the root object of the topology.

This particular instance may not be a problem because `kuadrantCR` here is only used to infer the namespace. However, technically speaking, this root object may not be the Kuadrant CR, but the ConfigMap itself (or any other generic object once added a watcher for it).

If you want to ensure the root object is indeed the Kuadrant CR, you need to add type-checking.

A more concerning issue here though, I guess, is the possibility of the topology having no root `Object`s at all – i.e., only `Targetable`s and `Policy`s were created. In this case, I think line 91 will raise a null pointer exception.

guicassolato (changes requested):

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+	configMapRes := controller.ConfigMapsResource
+	unstructuredCM, _ := controller.Destruct(cm)
+
+	targets := topology.Objects().Items(func(object machinery.Object) bool {
+		return object.GetName() == cm.GetName() && object.GetNamespace() == cm.GetNamespace() && object.GroupVersionKind().Kind == cm.GroupVersionKind().Kind
```
guicassolato, line 102: Does `cm.GroupVersionKind().Kind` work here? I remember having issues with trying to get `GroupVersionKind` from structs having the `metav1.TypeMeta` unset (null pointer).

We're used to trusting this method because resources read from the cluster using an API client are typically returned with the TypeMeta fields properly filled. I don't think this is true if you build a minimum representation of the resource in memory, like in line 88.

Because type-checking is so common, and now more even so due to reading resources from the topology instead of using an API client, having the GroupKind of the most used resources stored in variables can be useful. `schema.GroupKind{Group: corev1.GroupName, Kind: "ConfigMap"})` is mentioned in line 49 too, for example.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+			Labels:    map[string]string{kuadrant.TopologyAnnotation: "true"},
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+	configMapRes := controller.ConfigMapsResource
+	unstructuredCM, _ := controller.Destruct(cm)
+
+	targets := topology.Objects().Items(func(object machinery.Object) bool {
```
guicassolato, line 101: Do you mind renaming `targets` here as sth like `existingTopologyConfigMaps` or `knownTopologyConfigMap` please?

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,155 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "topology",
+			Namespace: kuadrantCR.GetNamespace(),
+			Labels:    map[string]string{kuadrant.TopologyAnnotation: "true"},
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+	configMapRes := controller.ConfigMapsResource
```
guicassolato, line 98: nit: If you want, you can get rid of this variable as use `controller.ConfigMapsResource` directly.

controllers/state_of_the_world.go
Boomatang: Should be can't

controllers/state_of_the_world.go
Boomatang: Fair point, I will add some checks. 

This seems like some thing that would need to be done often and doing the checks is can be verbose. Is this something that we should consider adding as a feature to the policy-machinery? A way to define a resources is the root of the graph and give an easy way to get the resources at the root expected root of the graph.

controllers/state_of_the_world.go
Boomatang: I wasn't getting any issue but I will look into it more and make sure there is not unexpected behavior happening.

controllers/state_of_the_world.go
Boomatang: That is fair.

api/v1beta2/topology.go
Boomatang: I am not sure I follow what you are saying here. I will address the other points and then catch up  with you on this point.

controllers/state_of_the_world.go
Boomatang: After some unexpected behavior with using the kuadrant CR the name space for the topology configmap is now taken from an environment variable. It was also agreed that the topology configmap should also not be linked to the kuadrant CR. It is an independent node in the graph.

guicassolato: We can revisit this decision about linking the node to the Kuadrant CR(s) in the future if needed. I agree that, for now, this is the best move. After all, a topology exists independently of Kuadrant CRs and a topology that depicts this relationship otherwise seems a bit artificial. Maybe one day we'll support multiple Kuadrant instances in a cluster, with distinct topology graphs per instance.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func NewTopologyFileReconciler(client *dynamic.DynamicClient) *TopologyFileReconciler {
+	return &TopologyFileReconciler{Client: client}
+}
+
+func (r *TopologyFileReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+	logger := controller.LoggerFromContext(ctx).WithName("topology file")
+
+	if topologyNamespace == "" {
```
guicassolato, line 97: nit:

Can we please make `TopologyNamespace` a field of the `TopologyFileReconciler` struct (filled from the `topologyNamespace` package-wide variable when the reconciler is initialised)?

I'm thinking on the (de)coupling of this type from the rest, maybe so it can be moved to its own file or even package at some point.

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,167 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "topology",
+			Namespace: topologyNamespace,
+			Labels:    map[string]string{kuadrant.TopologyAnnotation: "true"},
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+	unstructuredCM, _ := controller.Destruct(cm)
```
guicassolato, line 111: Indeed I would expect no error here, but just in case, if you want to, we could not discard the error returned by `Destruct` and at least log it. WDYT?

config/manager/manager.yaml
```diff
@@ -30,6 +30,8 @@ spec:
           env:
             - name: RELATED_IMAGE_WASMSHIM
               value: "oci://quay.io/kuadrant/wasm-shim:latest"
+            - name: TOPOLOGY_NAMESPACE
+              value: "kuadrant-system"
```
guicassolato, line 34:
```suggestion
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```

WDYT?

Boomatang: This is the more robust approach, we do loss the ability to switch of the reconcile of the topology configmap easily with using the make targets. It would mean editing the deployment to change or remove that entry. Which I think is okay for now.

controllers/state_of_the_world.go
Boomatang: Ya I can log the message.

controllers/state_of_the_world.go
Boomatang: Should the `TopologyFileReconciler` raise an error if there is a empty string passed into to the constructor. If so, how would you suggest handling the error in the [buildReconciler](https://github.com/Kuadrant/kuadrant-operator/blob/c18da761d52b2f4bf4fdf98a43f3a0c2f6350c11/controllers/state_of_the_world.go#L75-L84) function?

controllers/state_of_the_world.go
```diff
@@ -0,0 +1,159 @@
...
+			Namespace: r.Namespace,
+			Labels:    map[string]string{kuadrant.TopologyLabel: "true"},
+		},
+		Data: map[string]string{
+			"topology": topology.ToDot(),
+		},
+	}
+	unstructuredCM, err := controller.Destruct(cm)
+	if err != nil {
+		logger.Error(err, "failed to destruct topology configmap")
```
guicassolato, line 102: This error should never happen, so it's good if that we log it.

I wonder if we should return if it happens.

guicassolato (approved):
One minor comment left only, but in general excellent work @Boomatang!

I've tried the changes locally by pushing the content of the `topology` ConfigMap to a file while rendering the graph on screen (using a [Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=tintinweb.graphviz-interactive-preview)).

```sh
make local-setup
```

Screenshot 2024-09-13 at 11 12 30

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

Screenshot 2024-09-13 at 11 13 23

```sh

kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/cars"
    - method: GET
      path:
        type: PathPrefix
        value: "/dolls"
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

Screenshot 2024-09-13 at 11 14 06


```sh
kubectl apply -f - <<EOF
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-authn":
        apiKey:
          selector: {}
        credentials:
          authorizationHeader:
            prefix: APIKEY
    authorization:
      "only-admins":
        opa:
          rego: |
            groups := split(object.get(input.auth.identity.metadata.annotations, "kuadrant.io/groups", ""), ",")
            allow { groups[_] == "admins" }
        routeSelectors:
        - matches:
          - path:
              type: PathPrefix
              value: "/admin"
EOF
```

Screenshot 2024-09-13 at 11 16 29


```sh
kubectl apply -f -<<EOF
kind: HTTPRoute
metadata:
  name: other
spec:
  parentRefs:
  - name: istio-ingressgateway
    namespace: istio-system
  hostnames:
  - "*.other-apps.com"
EOF
````

Screenshot 2024-09-13 at 11 16 59


```sh
kind: AuthPolicy
metadata:
  name: gw-auth
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  rules:
    authorization:
      deny-all:
        opa:
          rego: "allow = false"
    response:
      unauthorized:
        headers:
          "content-type":
            value: application/json
        body:
          value: |
            {
              "error": "Forbidden",
            }
EOF
```

Screenshot 2024-09-13 at 11 28 17

```sh
kind: RateLimitPolicy
metadata:
  name: gw-rlp
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  limits:
    "global":
      rates:
      - limit: 5
        duration: 10
        unit: second
EOF
```
Screenshot 2024-09-13 at 11 32 18

```sh
kubectl -n istio-system apply -f -<<EOF
kind: DNSPolicy
metadata:
  name: ingress
spec:
  targetRef:
    kind: Gateway
    name: istio-ingressgateway
  routingStrategy: simple
  providerRefs:
  - name: aws
EOF
```

Screenshot 2024-09-13 at 11 37 44


(After this last command, the operator panicked at [this line](https://github.com/Kuadrant/kuadrant-operator/blob/024a68cf537827f03538de7abd233f5dc65fcdb1/controllers/dnspolicy_dnsrecords.go#L65). I don't think it's related to this PR though.)

eguzki (approved):
Great job @Boomatang :medal_military:

PR #844, opened 2024-08-30
Update auth config json test

adam-cattermole:
One of the integration tests is failing https://github.com/Kuadrant/kuadrant-operator/actions/runs/10625397804/job/29455286217#step:7:122

I think this is related to the changes to store as v1beta2. I can't quite pinpoint exactly why but before the `credentials` field had the `authorizationHeader` present (even though it's `omitempty`), whereas now it does not - I'm guessing some logic in the webhook was incorrect and was creating an empty `Prefixed` so the pointer was not null - I've updated all 3 occurrences in the json string.

Before:
```
"credentials":{"authorizationHeader":{}}
```
After:
```
"credentials":{}
```

 I believe the behaviour now looks correct but looking to @guicassolato or @alexsnaps to confirm.


adam-cattermole:
My only guess is something to do with this [kubebuilder validation](https://github.com/Kuadrant/authorino/blob/main/api/v1beta1/auth_config_types.go#L187-L188) was making `Credentials_In` not empty which meant that [building the credentials](https://github.com/Kuadrant/authorino/blob/1eaead9212cbecd750bfae638b3025a8b9327bd6/api/v1beta2/auth_config_conversion.go#L337-L343) in the webhook made an empty `Prefixed` instead of null?

guicassolato (approved):
IDK why exactly we were getting `{"authorizationHeader":{}}` before, but it could be due to a difference in the procedure between mashalling the content before versus just reading it from storage now.

Either way, the fix seems correct to me.

Thanks, @adam-cattermole!

PR #848, opened 2024-09-02
Envoygateway wasm controller

eguzki:
# What

Controller to configure kuadrant wasm-shim when EnvoyGateway is the gateway provider.

It enables kuadrant rate limiting


controllers/envoygateway_wasm_controller.go
```diff
@@ -0,0 +1,222 @@
...
+	if err := r.Client().Get(ctx, req.NamespacedName, kObj); err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("no kuadrant object found")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "failed to get kuadrant object")
+		return ctrl.Result{}, err
+	}
+
+	raw_topology, err := kuadranttools.TopologyForPolicies(ctx, r.Client(), kuadrantv1beta2.NewRateLimitPolicyType())
```
guicassolato, line 58: nit:

```suggestion
	rawTopology, err := kuadranttools.TopologyForPolicies(ctx, r.Client(), kuadrantv1beta2.NewRateLimitPolicyType())
```

pkg/rlptools/wasm/utils.go
```diff
@@ -287,3 +297,130 @@ func dataFromLimit(limitIdentifier string, limit *kuadrantv1beta2.Limit) (data [
...
+	}
+
+	route := t.GetPolicyHTTPRoute(rlp)
+
+	if route == nil {
+		// The policy is targeting a gateway
+		// This gateway policy will be enforced into all HTTPRoutes that do not have a policy attached to it
+
+		// Build imaginary route with all the routes not having a RLP targeting it
+		untargetedRoutes := t.GetUntargetedRoutes(gw)
```
guicassolato, line 314: I was banging my head against the wall for 20 min or so, trying to find where in the code we "fixed" the imaginary route thing for RLPs. It turns out we haven't. 🙄

Leaving this comment here then only so we remember to fix the issue here as well whenever it is addressed with #688.

IOW, this LOC is wrong, but consistently wrong with all the other places where we made the same mistake, making it kind of right, if it makes any sense 😵‍💫

eguzki: good point. Thanks for that. That will ease the process of rebasing when we fix that in `main`. With a little bit of luck, we can merge this into `main` before we fix it and then no rebase conflicts.

eguzki:
@adam-cattermole ready for review

make/envoy-gateway.mk
```diff
@@ -39,10 +39,19 @@ $(EGCTL):
 .PHONY: egctl
 egctl: $(EGCTL) ## Download egctl locally if necessary.
 
+envoy-gateway-enable-envoypatchpolicy: $(YQ)
```
adam-cattermole, line 42: This is interesting, I guess if we document support for envoygateway in the repo / on the website we should call out this requirement to enable the envoy patch policy API

adam-cattermole (approved):
Nice job here, lgtm

make/envoy-gateway.mk
eguzki: Sure!

Issue #794, opened 2024-08-01
Changing gateway listener with DNSPolicy attached results in orphaned DNS records


azgabur:
It is the same issue: https://github.com/Kuadrant/multicluster-gateway-controller/issues/601

After changing Gateway listener from domain specified in Managed zone to a domain specified in different Managed zone the previous DNS record wont get cleaned up and stays in DNS provider.

## Steps to reproduce
1. Have two Managed zones.
2. Create correctly Gateway, HttpRoute and DnsPolicy.
3. Edit the Gateway listener so the domain is now under different managed zone.
4. The DnsPolicy should have status: 
```
'DNSPolicy has encountered some issues: policy is not enforced on any dns record: no routes attached for listeners'
``` 
HttpRoute has status: 
```
no hostnames matched parent hostname
```
**DNSRecord CR is deleted and DNS record on provider is NOT deleted.**
5. Edit the HttpRoute so it matches the Gateway new listener.
6. Everything reconciles successfully and new hostname works. New DNSRecord CR gets created.
7. Doing DNS lookup on the old hostname will still work.

azgabur:
I found simpler way to reproduce this with just one ManagedZone.
1. Create Gateway and HttpPolicy to have same hostname (no wildcards)
2. Create DNSPolicy
3. Edit Gateway listener to different hostname
4. Observe the errors in HttpPolicy and DNSPolicy (same as in first post). Observe the DNSRecord gets deleted.
5. Edit HttpPolicy to the new hostname
6. Everything reconciles successfuly
7. DNS query for old hostname still resturns a result.

mikenairn:
This should be fixed as part of the managed zone removal https://github.com/Kuadrant/dns-operator/pull/203

mikenairn:
> This should be fixed as part of the managed zone removal [Kuadrant/dns-operator#203](https://github.com/Kuadrant/dns-operator/pull/203)

Will be looked into after [Kuadrant/dns-operator#203](https://github.com/Kuadrant/dns-operator/pull/203) as we will probably mark the `rootHost ` filed as immutable and delete and re-create the DNSRecord resource in the kuadrant operator instead of handling `rootHost` changes in the dnsrecord controller.

azgabur:
Yes I can confirm the managed zone removal did not fix this bug.

PR #850, opened 2024-09-03
Delete dns records when listener hostname changes

mikenairn:
fixes: #794 

The DNSRecord resource does not handle changes made to the `rootHost` which can cause things to be left around in the provider if it's modified after initial record creation. This change modifies the logic detecting if existing records should be removed to take into account possible hostname changes, removing any that no longer match (qw.listener.hostname != record.spec.rootHost).

related: https://github.com/Kuadrant/dns-operator/pull/219

### DIFF
--- a/controllers/dns_helper.go
+++ b/controllers/dns_helper.go
@@ -254,13 +254,15 @@ func (dh *dnsHelper) removeDNSForDeletedListeners(ctx context.Context, upstreamG
 
 	for i, dnsRecord := range dnsList.Items {
 		listenerExists := false
+		rootHostMatches := false
 		for _, listener := range upstreamGateway.Spec.Listeners {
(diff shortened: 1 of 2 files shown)


makslion (approved):
Looks good. 
Would it be a good idea to test for the wildcard records as well? Currently `rootHost` (dns-record) is identical to `hostname` (gateway) always but I'm worried that it might change

mikenairn:
> Looks good. Would it be a good idea to test for the wildcard records as well? Currently `rootHost` (dns-record) is identical to `hostname` (gateway) always but I'm worried that it might change

I'm not really sure why it would change, but that said there is no harm in testing it. Test updated now to ensure that other listeners DNSRecords are not modified due to updates to another listener, and also that the wildcard listener DNSRecord can be updated in the same way.

Issue #769, opened 2024-07-22
Manage Image Overflow in our Quay.io Repository


dlaw4608:
(Follow-up from Issue #241)

The original issue #241 addressed two main concerns: image overflow in our Quay.io repository and pull requests from forks not triggering builds and pushes of images. While the second issue has been resolved, we still need to address the first issue regarding image overflow.

## Problem
# Image Overflow
The Quay.io repository is accumulating a large number of image tags over time.
It would be a good idea to filter out older tags no longer relevant and delete them.
Though some tags remain relevant and should not be deleted, even if they are old.

## Proposed Solution
+ Automate Cleanup: Automatically identify and delete image tags older than six months (or however old is deemed no longer relevant).
+ Preserve Relevant Tags: Ensure the relevant tags are not deleted. Ex. (version-release)

Boomatang:
@KevFan is this relative do you know? If not I think we should close it.

PR #851, opened 2024-09-05
feat: set quay image expiry to prevent overflow of images

KevFan:
# Description
Part of: https://github.com/Kuadrant/kuadrant-operator/issues/769
Associated with: https://github.com/Kuadrant/kuadrant-operator/pull/843

Currently, all built Quay images never expire and are consuming unnecessary quota on Quay.io. Quay provides a way to [set tag expiration from a Dockerfile](https://docs.redhat.com/en/documentation/red_hat_quay/3.4/html/use_red_hat_quay/working_with_tags#setting_tag_expiration_from_a_dockerfile) to prevent this issue for images that don't need to be kept indefinitely, such as development branches and nightly builds. This can help prevent the problem from escalating over time.

# Verification
* You can check that the `expire_images` tag has an expiry at the following quay repos which was build by the CI for this branch:
  * https://quay.io/repository/kuadrant/kuadrant-operator?tab=tags
  * https://quay.io/repository/kuadrant/kuadrant-operator-bundle?tab=tags
  * https://quay.io/repository/kuadrant/kuadrant-operator-catalog?tab=tags
* If you want to verify manually locally that the image label is set:
  *  Remove local image
     ```
     docker rmi quay.io/kuadrant/kuadrant-operator:latest
     docker rmi quay.io/kuadrant/kuadrant-operator-bundle:latest
     docker rmi quay.io/kuadrant/kuadrant-operator-catalog:latest
     ```
  * Build image
    ```
    make docker-build bundle-build catalog-dockerfile catalog-build
    ```
  * Inspect image labels 
    ```
    docker inspect quay.io/kuadrant/kuadrant-operator:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/kuadrant-operator-bundle:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/kuadrant-operator-catalog:latest --format='{{json .Config.Labels}}'
    ```
  * The `quay.expires-after` label should be set to `never`
     
    image

  * Remove local image again
     ```
     docker rmi quay.io/kuadrant/kuadrant-operator:latest
     docker rmi quay.io/kuadrant/kuadrant-operator-bundle:latest
     docker rmi quay.io/kuadrant/kuadrant-operator-catalog:latest
     ```
  * Build image with a set expiry
    ```
     # Delete the generated catalog Dockerfile as it won't regenerate again with the correct label if it's already present
     rm catalog/kuadrant-operator-catalog.Dockerfile 
     QUAY_IMAGE_EXPIRY=3d make docker-build bundle-build catalog-dockerfile catalog-build
    ```
  * Inspect image labels 
    ```
    docker inspect quay.io/kuadrant/kuadrant-operator:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/kuadrant-operator-bundle:latest --format='{{json .Config.Labels}}'
    docker inspect quay.io/kuadrant/kuadrant-operator-catalog:latest --format='{{json .Config.Labels}}'
    ```
  * The `quay.expires-after` label should be set to `3d`
    
    image

### DIFF
--- a/.github/workflows/build-images-branches.yaml
+++ b/.github/workflows/build-images-branches.yaml
@@ -13,3 +13,5 @@ jobs:
     with:
       kuadrantOperatorVersion: ${{ github.ref_name }}
       kuadrantOperatorTag: ${{ github.ref_name }}
+      # Conditionally set quayImageExpiry to expire quay images only for non-release branches
+      quayImageExpiry: ${{ startsWith(github.ref_name, 'release') && 'never' || '1w' }}
--- a/.github/workflows/build-images-nightly.yaml
+++ b/.github/workflows/build-images-nightly.yaml
@@ -25,3 +25,4 @@ jobs:
       limitadorOperatorVersion: ${{ vars.LIMITADOR_OPERATOR_SHA }}
       dnsOperatorVersion: ${{ vars.DNS_OPERATOR_SHA }}
       wasmShimVersion: ${{ vars.WASM_SHIM_SHA }}
+      quayImageExpiry: 2w
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -7,9 +7,22 @@ CATALOG_IMG ?= $(IMAGE_TAG_BASE)-catalog:$(IMAGE_TAG)
 CATALOG_FILE = $(PROJECT_PATH)/catalog/kuadrant-operator-catalog/operator.yaml
 CATALOG_DOCKERFILE = $(PROJECT_PATH)/catalog/kuadrant-operator-catalog.Dockerfile
 
+# Quay image default expiry
+QUAY_IMAGE_EXPIRY ?= never
+
+# A LABEL that can be appended to a generated Dockerfile to set the Quay image expiration through Docker arguments.
+define QUAY_DOCKERFILE_LABEL
+
+# Quay image expiry
+ARG QUAY_IMAGE_EXPIRY
+ENV QUAY_IMAGE_EXPIRY=$${QUAY_IMAGE_EXPIRY:-never}
+LABEL quay.expires-after=$${QUAY_IMAGE_EXPIRY}
+endef
+export QUAY_DOCKERFILE_LABEL
+
 $(CATALOG_DOCKERFILE): $(OPM)
 	-mkdir -p $(PROJECT_PATH)/catalog/kuadrant-operator-catalog
-	cd $(PROJECT_PATH)/catalog && $(OPM) generate dockerfile kuadrant-operator-catalog
+	cd $(PROJECT_PATH)/catalog && $(OPM) generate dockerfile kuadrant-operator-catalog -l quay.expires-after=$(QUAY_IMAGE_EXPIRY)
 catalog-dockerfile: $(CATALOG_DOCKERFILE) ## Generate catalog dockerfile.
 
 CHANNELS ?= preview
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -11,14 +11,14 @@ CATALOG_DOCKERFILE = $(PROJECT_PATH)/catalog/kuadrant-operator-catalog.Dockerfil
 QUAY_IMAGE_EXPIRY ?= never
 
 # A LABEL that can be appended to a generated Dockerfile to set the Quay image expiration through Docker arguments.
-define QUAY_DOCKERFILE_LABEL
+define QUAY_EXPIRY_TIME_LABEL
 
 # Quay image expiry
 ARG QUAY_IMAGE_EXPIRY
 ENV QUAY_IMAGE_EXPIRY=$${QUAY_IMAGE_EXPIRY:-never}
 LABEL quay.expires-after=$${QUAY_IMAGE_EXPIRY}
 endef
-export QUAY_DOCKERFILE_LABEL
+export QUAY_EXPIRY_TIME_LABEL
 
 $(CATALOG_DOCKERFILE): $(OPM)
 	-mkdir -p $(PROJECT_PATH)/catalog/kuadrant-operator-catalog
--- a/.github/workflows/build-images-base.yaml
+++ b/.github/workflows/build-images-base.yaml
@@ -35,6 +35,10 @@ on:
         description: Bundle and catalog channels, comma separated
         default: preview
         type: string
+      quayImageExpiry:
+        description: When to expire the built quay images. The time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively, from the time the image is built.
+        default: never
+        type: string
   workflow_dispatch:
     inputs:
       kuadrantOperatorTag:
@@ -69,6 +73,10 @@ on:
         description: Bundle and catalog channels, comma separated
         default: preview
         type: string
+      quayImageExpiry:
+        description: When to expire the built quay images. The time values could be something like 1h, 2d, 3w for hours, days, and weeks, respectively, from the time the image is built.
+        default: never
+        type: string
 
 env:
   IMG_TAGS: ${{ inputs.kuadrantOperatorTag }}
@@ -77,6 +85,7 @@ env:
   IMG_REGISTRY_ORG: kuadrant
   MAIN_BRANCH_NAME: main
   OPERATOR_NAME: kuadrant-operator
+  QUAY_IMAGE_EXPIRY: ${{ inputs.quayImageExpiry }}
 
 jobs:
   build:
@@ -107,6 +116,7 @@ jobs:
           provenance: false
           tags: |
             ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}:${{ env.IMG_TAGS }}
+          build-args: QUAY_IMAGE_EXPIRY=${{ inputs.quayImageExpiry }}
       - name: Print Image URL
         run: echo "Image pushed to ${{ env.IMG_REGISTRY_HOST }}/${{ env.IMG_REGISTRY_ORG }}/${{ env.OPERATOR_NAME }}:${{ env.IMG_TAGS }}"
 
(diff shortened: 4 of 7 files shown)


eguzki (changes requested):
TIL [setting tag expiration from a dockerfile](https://docs.redhat.com/en/documentation/red_hat_quay/3.4/html/use_red_hat_quay/working_with_tags#setting_tag_expiration_from_a_dockerfile) :+1: 

operator image and bundle image covered.

Missing the catalog image. In https://github.com/Kuadrant/kuadrant-operator/blob/main/make/catalog.mk#L10-L12 `opm generate dockerfile` allows adding extra labels

```sh
bin/opm generate dockerfile kuadrant-operator-catalog
Error: stat kuadrant-operator-catalog: no such file or directory
Usage:
  opm generate dockerfile <dcRootDir> [flags]

Flags:
  -i, --binary-image string    Image in which to build catalog. (default "quay.io/operator-framework/opm:latest")
  -l, --extra-labels strings   Extra labels to include in the generated Dockerfile. Labels should be of the form 'key=value'.
  -h, --help                   help for dockerfile

Global Flags:
      --skip-tls-verify   skip TLS certificate verification for container image registries while pulling bundles
      --use-http          use plain HTTP for container image registries while pulling bundles
```

make/catalog.mk
```diff
@@ -7,9 +7,22 @@ CATALOG_IMG ?= $(IMAGE_TAG_BASE)-catalog:$(IMAGE_TAG)
...
+ENV QUAY_IMAGE_EXPIRY=$${QUAY_IMAGE_EXPIRY:-never}
+LABEL quay.expires-after=$${QUAY_IMAGE_EXPIRY}
+endef
+export QUAY_DOCKERFILE_LABEL
+
 $(CATALOG_DOCKERFILE): $(OPM)
 	-mkdir -p $(PROJECT_PATH)/catalog/kuadrant-operator-catalog
-	cd $(PROJECT_PATH)/catalog && $(OPM) generate dockerfile kuadrant-operator-catalog
+	cd $(PROJECT_PATH)/catalog && $(OPM) generate dockerfile kuadrant-operator-catalog -l quay.expires-after=$(QUAY_IMAGE_EXPIRY)
 catalog-dockerfile: $(CATALOG_DOCKERFILE) ## Generate catalog dockerfile.
```
KevFan, line 26: We can also allow setting the Quay image expiry by following the same pattern to how it was done for the bundle image. If we choose to implement this, the Docker label will be consistently set via build arguments. However, I'm not sure if it's worth the extra step just for the sake of consistency. If anyone has any thoughts on this, feel free to share, otherwise achieving this through `opm` labeling is simpler. 
```suggestion
	cd $(PROJECT_PATH)/catalog && $(OPM) generate dockerfile kuadrant-operator-catalog 
	echo "$$QUAY_DOCKERFILE_LABEL" >> $(CATALOG_DOCKERFILE)
catalog-dockerfile: $(CATALOG_DOCKERFILE) ## Generate catalog dockerfile.

# Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
# Ref https://olm.operatorframework.io/docs/tasks/creating-a-catalog/#catalog-creation-with-raw-file-based-catalogs
.PHONY: catalog-build
catalog-build: ## Build a catalog image.
	# Build the Catalog
	docker build --build-arg QUAY_IMAGE_EXPIRY=$(QUAY_IMAGE_EXPIRY)  $(PROJECT_PATH)/catalog -f $(PROJECT_PATH)/catalog/kuadrant-operator-catalog.Dockerfile -t $(CATALOG_IMG)
```

.github/workflows/build-images-branches.yaml
```diff
@@ -13,3 +13,4 @@ jobs:
     with:
       kuadrantOperatorVersion: ${{ github.ref_name }}
       kuadrantOperatorTag: ${{ github.ref_name }}
+      quayImageExpiry: 1w
```
KevFan, line 16: Abitary value of when to expire branch images. If anyone has a better suited value, I can update it to this value

.github/workflows/build-images-nightly.yaml
```diff
@@ -25,3 +25,4 @@ jobs:
       limitadorOperatorVersion: ${{ vars.LIMITADOR_OPERATOR_SHA }}
       dnsOperatorVersion: ${{ vars.DNS_OPERATOR_SHA }}
       wasmShimVersion: ${{ vars.WASM_SHIM_SHA }}
+      quayImageExpiry: 2w
```
KevFan, line 28: Abitary value of when to expire nighty images. If anyone has a better suited value, I can update it to this value

.github/workflows/build-images-branches.yaml
KevFan: This will also affect releases branches currently which is probably something we don't want 🤔

.github/workflows/build-images-branches.yaml
```diff
@@ -13,3 +13,5 @@ jobs:
     with:
       kuadrantOperatorVersion: ${{ github.ref_name }}
       kuadrantOperatorTag: ${{ github.ref_name }}
+      # Conditionally set quayImageExpiry to expire quay images only for non-release branches
+      quayImageExpiry: ${{ startsWith(github.ref_name, 'release') && 'never' || '1w' }}
```
KevFan, line 17: Abitary value of when to expire branch images. If anyone has a better suited value, I can update it to this value.

I'm assuming we **only** want to expire dev branch images and **not** release branches. Or do we care about release branch images? Seems the actual tagged v.X.Y.Z(-RC) images are run manually using the base images workflow where the image should be set to never expire 🤔

.github/workflows/build-images-nightly.yaml
KevFan: Maybe @jsmolar (QE) can recommend a value that would work well

didierofrivia (approved):
LGTM!, regarding the default expiry time, would trust QA input.

🔑

.github/workflows/build-images-nightly.yaml
jsmolar: 1 or 2 weeks for nightlies is reasonable. I don't see any reason why we would need to keep them longer.

jsmolar (approved):

make/catalog.mk
eguzki: no hard opinion on this matter. For me what it matters is that the label should be there. 

I do have a "soft opinion", though. I would do the labeling using the `opm` command. Mainly because we do not have a committed version of the dockerfile, it is autogenerated. So, the expiry time is not going to be committed in the repo even though it is hardcoded in the (ephemeral) dockerfile.

eguzki: I also do not like 

```
echo "$$QUAY_DOCKERFILE_LABEL" >> $(CATALOG_DOCKERFILE)
```
if we need to do that, we do that. But modifying dynamically the dockerfile is somewhat unwanted. Can we ensure the value of `$QUAY_DOCKERFILE_LABEL` is syntactically correct?

KevFan: > if we need to do that, we do that. But modifying dynamically the dockerfile is somewhat unwanted. 

@eguzki  Yes, we don't need to do this for the **catalog** dockerfile 👍  But I think we need to do this for the **bundle** Dockerfile since this is also autogenerated, is overwritten every time and committed. 
https://github.com/Kuadrant/kuadrant-operator/blob/107947ed5b7478c523c13cb66e373394db5167f8/Makefile#L393
https://github.com/Kuadrant/kuadrant-operator/blob/107947ed5b7478c523c13cb66e373394db5167f8/bundle.Dockerfile#L23-L26

I'm not aware of a way to add extra labels to the generated bundle dockerfile via operator-sdk 🤔  

> Can we ensure the value of $QUAY_DOCKERFILE_LABEL is syntactically correct?

It should be correct, but not sure how we can ensure this? It's functioning at least since the label was set correctly for the bundle image built 🤔

eguzki: right, I ran into the same issue when wanted to add openshift specific labels to the bundle image. 

I am afraid, I do not know how to do this cleanly, either. 

We decided to have git commited the bundle.Dockerfile (and `bundle` dir) for a reason. Reason being having traceability. Updating the file on-the-fly sort of defeats that purpose. 

I propose to do what you suggest. A sacrifice for the greater good. Do no git update `bundle.Dockerfile`, change it on the fly instead.

eguzki (changes requested):
following https://github.com/Kuadrant/kuadrant-operator/pull/851/files#r1749898448

`bundle.Dockerfile` cannot be changed, it is autogenerated.

make/catalog.mk
KevFan: > I propose to do what you suggest. A sacrifice for the greater good. Do no git update bundle.Dockerfile and commit it, change it on the fly instead and revert changes when the image build is done.

@eguzki  I'm thinking of reverting the [last commit](https://github.com/Kuadrant/kuadrant-operator/pull/851/commits/e4f3ac3208d10223839c0a074987ce0b7f88c28b) for this. Reason is the CI build for the **bundle** image needs to be updated also to account for this ([why it's not working on the CI build currently](https://quay.io/repository/kuadrant/kuadrant-operator-bundle?tab=tags)) since it does not use the makefile commands to build the image, and likely be a duplicated effort to also do this in the CI.
https://github.com/Kuadrant/kuadrant-operator/blob/e4f3ac3208d10223839c0a074987ce0b7f88c28b/.github/workflows/build-images-base.yaml#L138-L168

This, to me, is not as clean as the previous approach with just always appending the neccessay `QUAY_DOCKERFILE_LABEL` in the `make bundle` command after bundle image dockerfile has been generated, which will be the actual dockerfile used when building the bundle image rather than hiding this detail 🤔 
Since we commit this Dockerfile, it's apparent from the Dockerfile that it can accept a build argument to allow expirying quay images if passed. Otherwise this detail is hidden and harder to debug in general I feel if an issue occurs.

What do you think?

eguzki: So, when you run `bundle` makefile target, `$(OPERATOR_SDK) generate bundle ... ` will overwrite the bundle.Dockerfile and then we add the label again, right? something like

```
bundle:
    .... 
    $(OPERATOR_SDK) generate bundle -q --overwrite
    .....
    @echo "$$QUAY_DOCKERFILE_LABEL" >> bundle.Dockerfile
```

If this is what you suggest, I am ok with it. And I propose:

`s/QUAY_DOCKERFILE_LABEL/QUAY_EXPIRY_TIME_LABEL/`

Or something similar.

KevFan: > So, when you run `bundle` makefile target, `$(OPERATOR_SDK) generate bundle ... ` will overwrite the bundle.Dockerfile and then we add the label again, right? If this is what you suggest, I am ok with it.

Yes, exactly 👍 

> . And I propose: `s/QUAY_DOCKERFILE_LABEL/QUAY_EXPIRY_TIME_LABEL/`. Or something similar.

Sounds good, updated to `QUAY_EXPIRY_TIME_LABEL` 👍

eguzki (approved):

PR #858, opened 2024-09-13
chore: Simplify DNSPolicy API

### LINKED ISSUE #855: Simplify geo and weighted options in DNSPolicy
**What**

In the past we worked with OCM which brought the idea of a hub cluster and multiple views of the same gateway. We no longer support OCM but some of that legacy has remained in our API we should simplify this to reflect a policy can only target one gateway and even if it could target multiple they would be in the same cluster.

Currently we have 

```
  loadBalancing:
    weighted:
      defaultWeight: 120
      custom: # <--- New Custom Weights being added
        - weight: 255
          selector:
            matchLabels:
              kuadrant.io/lb-attribute-custom-weight: AWS
    geo: 
      defaultGeo: US #catch all geo         

```

we then have to set attributes on the gateway in the form of annotations to set these custom weights. This no longer makes sense. 

Proposed change is to have the values set directly on the policy and indicated if geo is a default or not

```
  loadBalancing: # each of these values would be required.
    weight: 100
    geo: U.S
    defaultGEO: true #indicates whether to use this as a the wilcard / catchall geo
```

mikenairn:
closes #855 

* Remove RoutingStrategy field. LoadBalanced vs Simple can be inferred from the value of the LoadBalancingSpec on the policy.
* Simplify LoadBalancingSpec:
  - No longer uses label selectors to get values for geo or weight from a metav1.Object, but the values themselves should be set directly in the policy.
  - LoadBalancingSpec.DefaultGeo specifies if the current geo is the default one or not and signals to the endpoint generation whether to create the wildcard catch all endpoint.


api/v1alpha1/dnspolicy_types.go
```diff
@@ -70,47 +64,32 @@ type DNSPolicySpec struct {
...
 	//
-	// Route53: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-weighted.html
-	DefaultWeight Weight `json:"defaultWeight"`
-
-	// custom list of custom weight selectors.
-	// +optional
-	Custom []*CustomWeight `json:"custom,omitempty"`
+	// Route53: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-geo.html
+	// Google: https://cloud.google.com/compute/docs/regions-zones
+	// Azure: ToDo (mnairn) get link for this
```
philbrookes, line 83: https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-geographic-regions

api/v1alpha1/dnspolicy_types.go
```diff
@@ -69,45 +63,34 @@ type DNSPolicySpec struct {
...
-	// defaultWeight is the record weight to use when no other can be determined for a dns target cluster.
+	// weight value to apply to weighted endpoints.
 	//
 	// The maximum value accepted is determined by the target dns provider, please refer to the appropriate docs below.
 	//
 	// Route53: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/routing-policy-weighted.html
-	DefaultWeight Weight `json:"defaultWeight"`
+	// Google: https://cloud.google.com/dns/docs/overview/
+	// Azure: https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-routing-methods#weighted-traffic-routing-method
+	Weight int `json:"weight"`
```
maleck13, line 80: I think making this optional with a preset default is reasonable rather than forcing it to be set

tests/common/dnspolicy/dnspolicy_controller_multi_cluster_test.go
```diff
@@ -1,714 +0,0 @@
-//go:build integration
-
-package dnspolicy
```
maleck13, line 3: why did we end up removing this as part of this work?

maleck13:
The changes look reasonable to me just not sure why we removed the test file and also perhaps we should go with optional  on weight

tests/common/dnspolicy/dnspolicy_controller_multi_cluster_test.go
mikenairn: These tests were specifically for the multi cluster status on the gateway (ACM). Since we are removing that support I'd rather just remove the tests than update them and then remove them later. 

Note this change is also partially dropping support for multicluster since there will be no longer any way to set geos/weights for individual clusters via labels on the ACM gateway so a lot of the tests are redundant now anyway.

api/v1alpha1/dnspolicy_types.go
mikenairn: I have no real issue with this ether way tbh. It was made mandatory [here](https://github.com/Kuadrant/kuadrant-operator/pull/578), reason for that as far as i remember was to make it more explicit to users that LoadBalanced is geo weighted, these are provider specific values that you as the user should be aware of, and since you have to set a geo you might as well tell us the weight as well so it's not a surprise where it's coming from.

Since all current providers will support a non negative value and will default to a round robin, equal weighted, when that value whatever it is is the same, setting a default value that will work with all three seems fine to me. "120" seem ok?

mikenairn: This is still required in the spec, but now has a default set so you don't have to explicitly set it if you are happy enough using the default value (120)

api/v1alpha1/dnspolicy_types.go
```diff
@@ -69,45 +64,35 @@ type DNSPolicySpec struct {
...
+	// The values accepted are determined by the target dns provider, please refer to the appropriate docs below.
+	//
+	// Route53: https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-values-geo.html
+	// Google: https://cloud.google.com/compute/docs/regions-zones
+	// Azure: https://learn.microsoft.com/en-us/azure/traffic-manager/traffic-manager-geographic-regions
+	// +kubebuilder:validation:MinLength=2
+	Geo string `json:"geo"`
+
+	// defaultGeo specifies if this is the default geo for providers that support setting a default catch all geo endpoint such as Route53.
+	DefaultGeo bool `json:"defaultGeo"`
```
mikenairn, line 95: @maleck13 This could also have a default value of false if we really wanted? Minimum loadBalancing spec a user would need to specify would be:
```
loadBalancing:
  geo: EU
```
Gets you weight: 120, geo: EU, defaultGeo false in the created resources.

api/v1alpha1/dnspolicy_types.go
```diff
@@ -56,7 +51,7 @@ const (
 )
 
 // DNSPolicySpec defines the desired state of DNSPolicy
-// +kubebuilder:validation:XValidation:rule="!(self.routingStrategy == 'loadbalanced' && !has(self.loadBalancing))",message="spec.loadBalancing is a required field when spec.routingStrategy == 'loadbalanced'"
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.loadBalancing) || has(self.loadBalancing)) && (has(oldSelf.loadBalancing) || !has(self.loadBalancing))", message="loadBalancing is immutable"
```
mikenairn, line 54: Find these validations a bit confusing to read tbh, but this just means you can't change the filed  `loadBalancing` from set to unset and vice versa after creation. Doing it on the field itself would mean you couldn't change the value of  `loadBalancing` and therefore any fields in that struct.

Tests to check all this were added.

maksymvavilov (approved):
Looks good.

Issue #274, opened 2023-10-16
Use sharding for Kuadrants Authorino deployment


pehala:
Currently, the Authorino deployed by Kuadrant is configured as cluster-wide and without sharding, which means it will reconcile all AuthConfigs on the cluster, which means you cannot use standalone authorino deployments on the same cluster as everything will be also reconciled by the Kuadrant Authorino.

Solution would be to deploy Kuadrant Authorino with sharding to some label like `kuadrant.io/managed=true` and add the same label to AuthConfigs generated from AuthPolicy

PR #865, opened 2024-09-20
Authorino CR reconcile moved to state of the world reconciler.

Boomatang:
closes: https://github.com/Kuadrant/kuadrant-operator/issues/812

Move the reconcile of the Authorino CR to the state of the world reconcile. There should be no functional change.

### DIFF
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -2,6 +2,7 @@ package v1beta1
 
 import (
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
+	authorinov1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
 	"github.com/samber/lo"
@@ -10,9 +11,11 @@ import (
 )
 
 var (
+	AuthorinoKind     = schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"}
 	KuadrantKind  = schema.GroupKind{Group: GroupVersion.Group, Kind: "Kuadrant"}
 	LimitadorKind = schema.GroupKind{Group: limitadorv1alpha1.GroupVersion.Group, Kind: "Limitador"}
 
+	AuthorinoResource = authorinov1beta1.GroupVersion.WithResource("authorinos")
 	KuadrantResource  = GroupVersion.WithResource("kuadrants")
 	LimitadorResource = limitadorv1alpha1.GroupVersion.WithResource("limitadors")
 )
@@ -49,6 +52,24 @@ func LinkKuadrantToLimitador(objs controller.Store) machinery.LinkFunc {
 			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
 				return kuadrant.GetNamespace() == child.GetNamespace() && child.GetName() == "limitador"
 			})
+		},
+	}
+}
+
+func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[*Kuadrant])
+
+	return machinery.LinkFunc{
+		From: KuadrantKind,
+		To:   schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"},
+		Func: func(child machinery.Object) []machinery.Object {
+			parents := make([]machinery.Object, len(kuadrants))
+			for _, parent := range kuadrants {
+				if parent.Namespace == child.GetNamespace() {
+					parents = append(parents, parent)
+				}
+			}
+			return parents
 		},
 	}
 }
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -64,7 +64,7 @@ func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
 		To:   AuthorinoKind,
 		Func: func(child machinery.Object) []machinery.Object {
 			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
-				return kuadrant.GetNamespace() == child.GetNamespace()
+				return kuadrant.GetNamespace() == child.GetNamespace() && child.GetName() == "authorino"
 			})
 		},
 	}
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -1,8 +1,8 @@
 package v1beta1
 
 import (
-	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
 	authorinov1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
+	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
 	"github.com/samber/lo"
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	AuthorinoKind     = schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"}
+	AuthorinoKind = schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"}
 	KuadrantKind  = schema.GroupKind{Group: GroupVersion.Group, Kind: "Kuadrant"}
 	LimitadorKind = schema.GroupKind{Group: limitadorv1alpha1.GroupVersion.Group, Kind: "Limitador"}
 
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -57,19 +57,15 @@ func LinkKuadrantToLimitador(objs controller.Store) machinery.LinkFunc {
 }
 
 func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
-	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[*Kuadrant])
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[machinery.Object])
 
 	return machinery.LinkFunc{
 		From: KuadrantKind,
-		To:   schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"},
+		To:   AuthorinoKind,
 		Func: func(child machinery.Object) []machinery.Object {
-			parents := make([]machinery.Object, len(kuadrants))
-			for _, parent := range kuadrants {
-				if parent.Namespace == child.GetNamespace() {
-					parents = append(parents, parent)
-				}
-			}
-			return parents
+			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
+				return kuadrant.GetNamespace() == child.GetNamespace()
+			})
 		},
 	}
 }
--- a/controllers/authorino_task.go
+++ b/controllers/authorino_task.go
@@ -1,0 +1,116 @@
+package controllers
+
+import (
+	"context"
+	v1beta2 "github.com/kuadrant/authorino-operator/api/v1beta1"
+	"github.com/kuadrant/kuadrant-operator/api/v1beta1"
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/ptr"
+)
+
+type AuthorinoCrReconciler struct {
+	Client *dynamic.DynamicClient
+}
+
+func NewAuthorinoCrReconciler(client *dynamic.DynamicClient) *AuthorinoCrReconciler {
+	return &AuthorinoCrReconciler{Client: client}
+}
+
+func (r *AuthorinoCrReconciler) Subscription() *controller.Subscription {
+	return &controller.Subscription{
+		ReconcileFunc: r.Reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: ptr.To(v1beta1.KuadrantKind), EventType: ptr.To(controller.CreateEvent)},
+			{Kind: ptr.To(v1beta1.AuthorinoKind), EventType: ptr.To(controller.DeleteEvent)},
+		},
+	}
+}
+
+func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+	logger := controller.LoggerFromContext(ctx).WithName("AuthorinoCrReconciler")
+	logger.Info("reconciling authorino resource", "status", "started")
+	defer logger.Info("reconciling authorino resource", "status", "completed")
+
+	kobjs := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (*v1beta1.Kuadrant, bool) {
+		if item.GroupVersionKind().Kind == v1beta1.KuadrantKind.Kind {
+			return item.(*v1beta1.Kuadrant), true
+		}
+		return nil, false
+	})
+
+	kobj, err := GetOldestKuadrant(kobjs)
+	if err != nil {
+		logger.Error(err, "cannot find Kuadrant resource", "status", "error")
+	}
+
+	if kobj.GetDeletionTimestamp() != nil {
+		logger.Info("root kuadrant marked for deletion", "status", "skipping")
+		return
+	}
+
+	aobjs := lo.FilterMap(topology.Objects().Objects().Items(), func(item machinery.Object, _ int) (machinery.Object, bool) {
+		if item.GroupVersionKind().Kind == v1beta1.AuthorinoKind.Kind {
+			return item, true
+		}
+		return nil, false
+	})
+
+	if len(aobjs) > 0 {
+		logger.Info("authorino resource already exists, no need to create", "status", "skipping")
+		return
+	}
+
+	authorino := &v1beta2.Authorino{
+		TypeMeta: v1.TypeMeta{
+			Kind:       "Authorino",
+			APIVersion: "operator.authorino.kuadrant.io/v1beta1",
+		},
+		ObjectMeta: v1.ObjectMeta{
+			Name:      "authorino",
+			Namespace: kobj.Namespace,
+			OwnerReferences: []v1.OwnerReference{
+				{
+					APIVersion:         kobj.GroupVersionKind().GroupVersion().String(),
+					Kind:               kobj.GroupVersionKind().Kind,
+					Name:               kobj.Name,
+					UID:                kobj.UID,
+					BlockOwnerDeletion: ptr.To(true),
+					Controller:         ptr.To(true),
+				},
+			},
+		},
+		Spec: v1beta2.AuthorinoSpec{
+			ClusterWide:            true,
+			SupersedingHostSubsets: true,
+			Listener: v1beta2.Listener{
+				Tls: v1beta2.Tls{
+					Enabled: ptr.To(false),
+				},
+			},
+			OIDCServer: v1beta2.OIDCServer{
+				Tls: v1beta2.Tls{
+					Enabled: ptr.To(false),
+				},
+			},
+		},
+	}
+
+	unstructuredAuthorino, err := controller.Destruct(authorino)
+	if err != nil {
+		logger.Error(err, "failed to destruct authorino", "status", "error")
+	}
+	logger.Info("creating authorino resource", "status", "processing")
+	_, err = r.Client.Resource(v1beta1.AuthorinoResource).Namespace(authorino.Namespace).Create(ctx, unstructuredAuthorino, v1.CreateOptions{})
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			logger.Info("already created authorino resource", "status", "acceptable")
+		} else {
+			logger.Error(err, "failed to create authorino resource", "status", "error")
+		}
+	}
+}
--- a/controllers/authorino_task.go
+++ b/controllers/authorino_task.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"strings"
+	"sync"
 
 	v1beta2 "github.com/kuadrant/authorino-operator/api/v1beta1"
 	"github.com/kuadrant/policy-machinery/controller"
@@ -34,7 +35,7 @@ func (r *AuthorinoCrReconciler) Subscription() *controller.Subscription {
 	}
 }
 
-func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, _ *sync.Map) error {
 	logger := controller.LoggerFromContext(ctx).WithName("AuthorinoCrReconciler")
 	logger.Info("reconciling authorino resource", "status", "started")
 	defer logger.Info("reconciling authorino resource", "status", "completed")
@@ -50,10 +51,10 @@ func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	if err != nil {
 		if strings.Contains(err.Error(), "empty list passed") {
 			logger.Info("kuadrant resource not found, ignoring", "status", "skipping")
-			return
+			return err
 		}
 		logger.Error(err, "cannot find Kuadrant resource", "status", "error")
-		return
+		return err
 	}
 
 	aobjs := lo.FilterMap(topology.Objects().Objects().Children(kobj), func(item machinery.Object, _ int) (machinery.Object, bool) {
@@ -65,7 +66,7 @@ func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, _ []controller.Re
 
 	if len(aobjs) > 0 {
 		logger.Info("authorino resource already exists, no need to create", "status", "skipping")
-		return
+		return nil
 	}
 
 	authorino := &v1beta2.Authorino{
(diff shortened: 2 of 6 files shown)


api/v1beta1/topology.go
```diff
@@ -34,3 +37,21 @@ func LinkKuadrantToGatewayClasses(objs controller.Store) machinery.LinkFunc {
 		},
 	}
 }
+
+func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[*Kuadrant])
+
+	return machinery.LinkFunc{
+		From: KuadrantKind,
+		To:   schema.GroupKind{Group: authorinov1beta1.GroupVersion.Group, Kind: "Authorino"},
```
guicassolato, line 46:
```suggestion
		To:   AuthorinoKind,
```

api/v1beta1/topology.go
```diff
@@ -34,3 +37,21 @@ func LinkKuadrantToGatewayClasses(objs controller.Store) machinery.LinkFunc {
...
+		Func: func(child machinery.Object) []machinery.Object {
+			parents := make([]machinery.Object, len(kuadrants))
+			for _, parent := range kuadrants {
+				if parent.Namespace == child.GetNamespace() {
+					parents = append(parents, parent)
+				}
+			}
+			return parents
```
guicassolato, line 54: Nothing wrong with the current approach, but just a minor suggestion leveraging `github.com/samber/lo`:

```suggestion
			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) {
				kuadrant.Namespace == child.GetNamespace()
			})
```

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
 }
 
 func buildReconciler(client *dynamic.DynamicClient) controller.ReconcileFunc {
+	authorinoCrSubscription := &controller.Subscription{
+		ReconcileFunc: NewAuthorinoCrReconciler(client).Reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: ptr.To(kuadrantv1beta1.KuadrantKind), EventType: ptr.To(controller.CreateEvent)},
+			{Kind: ptr.To(kuadrantv1beta1.AuthorinoKind), EventType: ptr.To(controller.DeleteEvent)},
+		},
+	}
```
guicassolato, line 80: Could all this be part of `NewAuthorinoReconciler()`?

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
 		Precondition: NewEventLogger().Log,
 		Tasks: []controller.ReconcileFunc{
 			NewTopologyFileReconciler(client, operatorNamespace).Reconcile,
+			authorinoCrSubscription.Reconcile,
 		},
 	}
 	return reconciler.Run
 }
 
+type AuthorinoCrReconciler struct {
```
guicassolato, line 91: I wonder if we should start moving these type to their separate files. Otherwise very soon this one will become too long and hard to navigate. WDYT?

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+			{Kind: ptr.To(kuadrantv1beta1.AuthorinoKind), EventType: ptr.To(controller.DeleteEvent)},
+		},
+	}
 	reconciler := &controller.Workflow{
 		Precondition: NewEventLogger().Log,
 		Tasks: []controller.ReconcileFunc{
 			NewTopologyFileReconciler(client, operatorNamespace).Reconcile,
+			authorinoCrSubscription.Reconcile,
 		},
 	}
```
guicassolato, line 87: This is fine for now. Probably what we'll want soon is:

```suggestion
	reconciler := &controller.Workflow{ // main
		Precondition: &controller.Workflow{ // logging, topology, and error handling
			Precondition: NewEventLogger().Log,
			Tasks: []controller.ReconcileFunc{
				NewTopologyFileReconciler(client, operatorNamespace).Reconcile,
			},
		},
		Tasks: []controller.ReconcileFunc{
			authorinoCrSubscription.Reconcile,
		},
	}
```

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
 }
 
 func buildReconciler(client *dynamic.DynamicClient) controller.ReconcileFunc {
+	authorinoCrSubscription := &controller.Subscription{
+		ReconcileFunc: NewAuthorinoCrReconciler(client).Reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: ptr.To(kuadrantv1beta1.KuadrantKind), EventType: ptr.To(controller.CreateEvent)},
```
guicassolato, line 77: Why Kuadrant CREATE only? What about when it's updated (with changes affecting the Authorino CR) or deleted?

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+			{Kind: ptr.To(kuadrantv1beta1.AuthorinoKind), EventType: ptr.To(controller.DeleteEvent)},
```
guicassolato, line 78: I guess we also want to reconcile the Authorino CR when its updated, no? Or are we going to allow modifications to this CR after it was created by the operator? (I think we should not BTW.)

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+				logger.Error(fmt.Errorf("muiltply kuadrant CRs found"), "unexpected behaviour may happen")
+			}
+			kobj = kobjs[0]
+			if kobj.GetDeletionTimestamp() != nil {
+				logger.Info("kuadrant CR marked for deletion, wont create Authorino CR.")
+				return
+			}
+			break
+		}
+	}
```
guicassolato, line 139: I had pictured this a bit different in my head:
- Find a `Kuadrant` object that is the root of the topology
- If there is none,
  - delete all `Authorino` objects that exist in the topology
  - return
- Otherwise
  - build the desired `Authorino` object for the `Kuadrant` root
  - create it if it doesn't exist as a child of the `Kuadrant` root
  - update it otherwise

controllers/state_of_the_world.go
```diff
@@ -63,15 +71,122 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+	unstructuredAuthorino, err := controller.Destruct(authorino)
+	if err != nil {
+		logger.Error(err, "failed to destruct authorino")
+	}
+	_, err = r.Client.Resource(kuadrantv1beta1.AuthorinoResource).Namespace(authorino.Namespace).Create(ctx, unstructuredAuthorino, metav1.CreateOptions{})
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			logger.Info("already created authorino Cr")
+		} else {
+			logger.Error(err, "failed to create authorino Cr")
```
guicassolato, line 185: This is an interesting one. I wonder if the Kuadrant CR should reflect in its status that no corresponding Authorino CR exists and then use this update to the status to retry.

controllers/state_of_the_world.go
Boomatang: Ya, I was thinking that also. Currently I am unsure of how we want to structure the packages.

controllers/state_of_the_world.go
Boomatang: This was something I was going to do after this was merged. I guess I could do it as part of this PR.

controllers/state_of_the_world.go
Boomatang: So currently the status controller will set the status in the kuadrant CR if the authorino CR is missing or if authorino is not in a ready state. That should be creating the retry events. Not sure if with the status controller outside the state of the world workflow if that can cause a timing issue.

controllers/state_of_the_world.go
Boomatang: The lack of delete is the easy one. Because of the owner reference when the kuadrant CR is fully remove from the cluster kubernetes will delete the Authorino CR. Delete is an event that we don't beed to handle.

The kuadrant CR update events also don't matter. There is no spec setable in the kuadrant CR that will affect the authorino CR. That feature was removed from the kuadrant spec for V1 as it is going to be redone to suit what is outlined in https://github.com/Kuadrant/architecture/pull/97 , so update events also don't matter.

Only the create events matter.

controllers/state_of_the_world.go
Boomatang: I think want is being done is mostly what you listed above. The things that are not being done are:
* deleting the Authorino CRs when the there is not kuadrant root. The cluster is handling this by the owner reference
* update if otherwise. The current implantation also does not do this. I don't think it would be right to introduction that change now.

controllers/state_of_the_world.go
Boomatang: Not sure. Would need to play with it to see what that would look like. I will look into it.

controllers/state_of_the_world.go
guicassolato: Fair enough. Let's stick with the current behaviour then and keep the discussion about leaking abstractions vs ownership fully granted to the Kuadrant CR outside of this PR.

controllers/state_of_the_world.go
guicassolato: Good point. I missed this reconciliation task that updates the status of the Kuadrant CR represented in the workflow then.

To cover for that, I think Authorino and Limitador reconciliation could be wrapped as two concurrent tasks of a "Data plane components reconciliation" workflow, with a postcondition taking care of the status update on the Kuadrant CR.

The task introduced by this PR then, focused on the Authorino CR only, may need to inject some hint of the status of Authorino (using the `sync.Map`) for the status updater to pick up – e.g., when it just created the CR or in case of failures like such as in this line.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,137 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+	for _, event := range events {
+		if event.Kind == kuadrantv1beta1.KuadrantKind && event.EventType == controller.CreateEvent {
+			kobjs := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (*kuadrantv1beta1.Kuadrant, bool) {
+				if item.GetName() == event.NewObject.GetName() && item.GetNamespace() == event.NewObject.GetNamespace() && item.GroupVersionKind().Kind == event.NewObject.GetObjectKind().GroupVersionKind().Kind {
+					return item.(*kuadrantv1beta1.Kuadrant), true
+				}
+				return nil, false
+			})
+			if len(kobjs) != 1 {
+				logger.Error(fmt.Errorf("muiltply kuadrant CRs found"), "unexpected behaviour may happen")
```
guicassolato, line 120:
```suggestion
				logger.Error(fmt.Errorf("multiple Kuadrant resources found"), "cannot select root Kuadrant resource")
```

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,137 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+	}
+}
+
+func (r *AuthorinoCrReconciler) Reconcile(ctx context.Context, events []controller.ResourceEvent, topology *machinery.Topology, _ error) {
+	logger := controller.LoggerFromContext(ctx).WithName("AuthorinoCrReconciler")
+	logger.Info("Reconciling Authorino Cr")
+
+	kobj := &kuadrantv1beta1.Kuadrant{}
+	for _, event := range events {
+		if event.Kind == kuadrantv1beta1.KuadrantKind && event.EventType == controller.CreateEvent {
```
guicassolato, line 112: Sorry @Boomatang. I'm not convinced about this approach of matching the event type again, to decide how to search for the Kuadrant object in the topology.

IMO, the first root that matches the Kuadrant kind is the Kuadrant object we are looking for. If there are more than one, then only the first Kuadrant should get an Authorino instance created for it. If there is none, than no Authorino object should be created and all existing Authorino instances should be deleted.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,137 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+	}
+
+	authorino := &authorinov1beta1.Authorino{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Authorino",
+			APIVersion: "operator.authorino.kuadrant.io/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "authorino",
+			Namespace: kobj.Namespace,
```
guicassolato, line 155: If I understand the selection of the Kuadrant object above correctly, if a second Kuadrant CR is created (in another namespace), another Authorino instance will be created for it too. Is that what we want? I think the only way to support this is by also supporting sharded instances (#274).

If a second Kuadrant CR is created in the same namespace where another exists, then we know we won't create another instance of Authorino in that same namespace (with static name `authorino`), right? So why even continue if don't reconcile existing Authorino CRs? And if we do decide to reconcile those, as soon as it starts having any field set directly from the Kuadrant CR, then the current code could trigger an infinite loop of reconciliation events with 2+ Kuadrant CRs.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,137 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+	kobj := &kuadrantv1beta1.Kuadrant{}
+	for _, event := range events {
+		if event.Kind == kuadrantv1beta1.KuadrantKind && event.EventType == controller.CreateEvent {
+			kobjs := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (*kuadrantv1beta1.Kuadrant, bool) {
+				if item.GetName() == event.NewObject.GetName() && item.GetNamespace() == event.NewObject.GetNamespace() && item.GroupVersionKind().Kind == event.NewObject.GetObjectKind().GroupVersionKind().Kind {
+					return item.(*kuadrantv1beta1.Kuadrant), true
+				}
+				return nil, false
+			})
+			if len(kobjs) != 1 {
```
guicassolato, line 119: The only way for this to be true is if the count is zero, right?

`len(kobjs) > 1` cannot ever be true, given the way we're filtering Kuadrant objects fro the topology here, I reckon.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,137 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+				logger.Error(fmt.Errorf("muiltply kuadrant CRs found"), "unexpected behaviour may happen")
+			}
+			kobj = kobjs[0]
+			break
+		} else if event.Kind == kuadrantv1beta1.AuthorinoKind && event.EventType == controller.DeleteEvent {
+			kobjs := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (*kuadrantv1beta1.Kuadrant, bool) {
+				if item.GetNamespace() == event.OldObject.GetNamespace() && item.GroupVersionKind().Kind == kuadrantv1beta1.KuadrantKind.Kind {
+					return item.(*kuadrantv1beta1.Kuadrant), true
+				}
```
guicassolato, line 128: It may not be obvious why this has to be the Kuadrant CR previously linked to the deleted Authorino object. We know this is true because in line 155, the Authorino object is built to exist in the same namespace as `kobj.Namespace`. If we continue with this approach for selecting the associated Kuadrant root, maybe add comment here?

controllers/state_of_the_world.go
guicassolato: To clarify the suggestion here... It's two-fold actually:
1. `Logger.Error` will log this as `level=error` and whatever is in the second argument as the main error message, while the first argument (the error) becomes a nested detail (a value) of the log entry. So, in my head, it makes sense that the top-level error message is "cannot select root Kuadrant resource", and the reason (detail) is because "multiple Kuadrant resources [were] found". (Not that I do not agree that "unexpected behaviour may happen" 🙂)
2. nitpick: Avoid using acronyms (e.g. "CR") in the log messages.

controllers/state_of_the_world.go
Boomatang: Ya looking at it again you are right there was no need to act depending on the event kind/type. That will only matter when we allow multiple Kuadrant resources. Refactored to be some what cleaner I think.

controllers/state_of_the_world.go
Boomatang: `len(kobjs) >1` can be true, but it would be the user doing something really wrong. Hence the logging. If I could I would have logged this as a warning but we only seem to have info and error level logs.

Odd you mention the count being zero, that caused a panic which affected the test. Is fixed now.

controllers/state_of_the_world.go
Boomatang: Avoiding acronyms, check

Ya, your version of the log reads better in the actual logs, happy to make the change.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,124 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+
+	authorino := &authorinov1beta1.Authorino{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Authorino",
+			APIVersion: "operator.authorino.kuadrant.io/v1beta1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "authorino",
+			Namespace: kobj.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
```
eguzki, line 151: There is a method to add ownerrefs. Up to you.

api/v1beta1/topology.go
```diff
@@ -34,3 +37,17 @@ func LinkKuadrantToGatewayClasses(objs controller.Store) machinery.LinkFunc {
...
+
+func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantKind), controller.ObjectAs[machinery.Object])
+
+	return machinery.LinkFunc{
+		From: KuadrantKind,
+		To:   AuthorinoKind,
+		Func: func(child machinery.Object) []machinery.Object {
+			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
+				return kuadrant.GetNamespace() == child.GetNamespace()
```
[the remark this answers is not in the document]
guicassolato: Agreed. If we name the Authorino CR after the Kuadrant CR, then we can have just that instance linked and any other one not linked and assumed to have been created by the user manually.

controllers/kuadrant_controller.go
```diff
@@ -427,41 +423,6 @@ func (r *KuadrantReconciler) reconcileLimitador(ctx context.Context, kObj *kuadr
...
-	}
-
-	err := r.SetOwnerReference(kObj, authorino)
-	if err != nil {
-		return err
-	}
-
-	return r.ReconcileResource(ctx, &authorinov1beta1.Authorino{}, authorino, reconcilers.CreateOnlyMutator)
-}
-
```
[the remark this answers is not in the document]
guicassolato: Correct. It should be removed from line 440 inside of `SetupWithManager`.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,124 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+func NewAuthorinoCrReconciler(client *dynamic.DynamicClient) *AuthorinoCrReconciler {
+	return &AuthorinoCrReconciler{Client: client}
+}
+
+func (r *AuthorinoCrReconciler) Subscription() *controller.Subscription {
+	return &controller.Subscription{
+		ReconcileFunc: r.Reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: ptr.To(kuadrantv1beta1.KuadrantKind), EventType: ptr.To(controller.CreateEvent)},
+			{Kind: ptr.To(kuadrantv1beta1.AuthorinoKind), EventType: ptr.To(controller.DeleteEvent)},
```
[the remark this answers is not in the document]
guicassolato: The only thing more specific than this the subscriber allows us to go is the adding name of the resource. I'm afraid the only way for us to know that is if it's statically named. If it's dynamically named after the Kuadrant CR (as I [suggested](https://github.com/Kuadrant/kuadrant-operator/pull/865/files#r1772974162)), then it won't work.

An alternative is labeling the Authorino resources we create and adding the label selector to the watcher. This will the Kuadrant Operator will completely ignore any other instance manually created by users, not even adding them to the topology.

controllers/state_of_the_world.go
```diff
@@ -64,14 +72,124 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+		return nil, false
+	})
+	if len(kobjs) == 0 {
+		logger.Info("no kuadrant resources found", "status", "skipping")
+		return
+	}
+	if len(kobjs) > 1 {
+		logger.Error(fmt.Errorf("multiple Kuadrant resources found"), "cannot select root Kuadrant resource", "status", "error")
+	}
+	kobj := kobjs[0]
```
guicassolato, line 124: We may think about the order these objects are fetched.

controllers/state_of_the_world.go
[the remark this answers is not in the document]
Boomatang: I would like to be more filtered on what events we get in general. I like the idea of using the labels to do this filtering and I am going to suggest that we create a common one called `kuadrant.io/resource` as this is not the only resource we should be filtering on. The same steps should be taken for the Limitador CR, as the step is mostly the very same, https://github.com/Kuadrant/kuadrant-operator/pull/871.

However I have come across a possible issue with the approach and you can create the issue today with the topology configmap. When you filter by labels and a user removes said labels, you **do not** get that event. This causes a number of issues. Operator restarts will try recreate the resource but fail, which is normally okay as you would also get the create event for that resource if it was labeled. 

There is a problem here, and I am not to sure how much of an edge case it is, or even how easy it maybe to solve. I have one idea but it is not compatible with the controller-runtime as far as I am aware.

controllers/kuadrant_controller.go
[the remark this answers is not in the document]
Boomatang: Good catch, slipped my mind to event check that.

controllers/state_of_the_world.go
Boomatang: Ya, I looked at that and the ownerrefs method I found at least expects a schema to also be passed in, which we do have access to at this stage. So I went with manually adding the reference.

controllers/state_of_the_world.go
Boomatang: In what sense are you say we should think about the order the resources are fetched. Is this how I get the the kuadrant CR's before the Authorino CR's or related to the `kobjs[0]` may not be the oldest on cluster and therefore the wrong one? The age of the CR is something I had though about so possible should fix.

controllers/state_of_the_world.go
Boomatang: I am not sure if this is still valid.

api/v1beta1/topology.go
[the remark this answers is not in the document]
Boomatang: Today I don't think we can rename the Authorino CR after the Kuadrant CR. I have tried to do this and it fails because of the status reconciler. Once the the status reconciler is move to the state of the world controller we can do a dynamic name based on the Kuadrant CR.

For now I can narrow this down to the hard coded name for the Authorino CR.

controllers/state_of_the_world.go
```diff
@@ -164,3 +180,30 @@ func (e *EventLogger) Log(ctx context.Context, resourceEvents []controller.Resou
 
 	return nil
 }
+
+// GetOldestKuadrant returns the oldest kuadrant resource from a list of kuadrant resources that is not marked for deletion.
+func GetOldestKuadrant(kuadrants []*kuadrantv1beta1.Kuadrant) (*kuadrantv1beta1.Kuadrant, error) {
```
guicassolato, line 185: We could probably pass the topology here. This is a pattern that likely will repeat a lot across tasks.

Boomatang: That's a good idea, but I am not going to do it in this PR. There was a topic I wish to discuss about unit testing where the input is a topology. I have a place holder issue, https://github.com/Kuadrant/policy-machinery/issues/33. But for now I would like to leave this as is with its unit tests until there is a good way to generate the a topology for a unit test.

controllers/state_of_the_world.go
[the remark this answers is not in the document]
Boomatang: I am going to ask if adding the filtering by label or by field selector (needs research) can be added in a follow up PR. I am conscious of time and don't want to see this getting held up over bike sheding. For now we would get some more events than we really want.

controllers/authorino_task.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		if strings.Contains(err.Error(), "empty list passed") {
+			logger.Info("kuadrant resource not found, ignoring", "status", "skipping")
+			return err
+		}
+		logger.Error(err, "cannot find Kuadrant resource", "status", "error")
+		return err
+	}
+
+	aobjs := lo.FilterMap(topology.Objects().Objects().Children(kobj), func(item machinery.Object, _ int) (machinery.Object, bool) {
```
KevFan, line 60: nit: redundant consecutive `.Objects()` call? 🤔  
```suggestion
	aobjs := lo.FilterMap(topology.Objects().Children(kobj), func(item machinery.Object, _ int) (machinery.Object, bool) {
```

controllers/state_of_the_world.go
```diff
@@ -26,7 +28,7 @@ import (
 
 var (
 	ConfigMapGroupKind = schema.GroupKind{Group: corev1.GroupName, Kind: "ConfigMap"}
-	operatorNamespace  = env.GetString("OPERATOR_NAMESPACE", "")
+	operatorNamespace  = env.GetString("OPERATOR_NAMESPACE", "kuadrant-system")
```
KevFan, line 31: This would make the panic at the following redundant, do we want to remove the panic now?
https://github.com/Kuadrant/kuadrant-operator/blob/abc2e19a3610e0857d30047570b24441ad90c372/controllers/state_of_the_world.go#L152-L154

Or we can remove setting a default ns env value and set it across all the makefile integration test commands that expect this env value to be set if we want to keep the panic?

KevFan (approved):
Minor nits, but this looks good to me 👍

controllers/state_of_the_world.go
guicassolato: Can we please add a TODO to refactor this ASAP. I do not want to risk people passing different lists of Kuadrant objects to this function. I want to make sure it is going to be always the same Kuadrant object that will be returned to all tasks that depend on it to navigate the topology from the common root.

Issue #811, opened 2024-08-16
[state-of-the-world reconciler] Dependency check


guicassolato:
Reconciliation function for checking top-level dependencies without which reconciliation of policies cannot proceed (e.g. Gateway API CRDs, gateway providers, etc)

PR #870, opened 2024-09-24
feat: check dependency for state of the world reconciler

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/811

Checks for dependencies for whether specific watchers and object kinds should be added to the controller opts of the state of the world reconciler

### DIFF
--- a/controllers/state_of_the_world.go
+++ b/controllers/state_of_the_world.go
@@ -5,10 +5,15 @@ import (
 	"fmt"
 	"strings"
 
+	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	egv1alpha1 "github.com/envoyproxy/gateway/api/v1alpha1"
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
+	istioclientgoextensionv1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
+	istioclientnetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
+	istioclientgosecurityv1beta1 "istio.io/client-go/pkg/apis/security/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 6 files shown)


controllers/state_of_the_world.go
```diff
@@ -59,6 +64,72 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+			controller.WithRunnable("envoyextensionpolicy watcher", controller.Watch(&egv1alpha1.EnvoyExtensionPolicy{}, envoygateway.EnvoyExtensionPoliciesResource, metav1.NamespaceAll)),
+			controller.WithRunnable("envoysecuritypolicy watcher", controller.Watch(&egv1alpha1.SecurityPolicy{}, envoygateway.SecurityPoliciesResource, metav1.NamespaceAll)),
+			controller.WithObjectKinds(
+				envoygateway.EnvoyPatchPolicyGroupKind,
+				envoygateway.EnvoyExtensionPolicyGroupKind,
+				envoygateway.SecurityPolicyGroupKind,
+			),
+			// TODO: add object links
+		)
+		// TODO: add specific tasks to workflow
```
KevFan, line 93: Future TODOs that would be completed by other issues as functionality is migrated to the workflow

controllers/state_of_the_world.go
```diff
@@ -59,6 +64,71 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
 		controller.WithReconcile(buildReconciler(client)),
 	}
 
+	ok, err := kuadrantgatewayapi.IsGatewayAPIInstalled(manager.GetRESTMapper())
```
Boomatang, line 67: What should happen if there is no gateway currently installed that we can use? We would need at least one, I am asking the same question of the cert manger if that is not installed should we try reconcile. I am aware that the user may install these resources later, but that would mean restarting the operator (which I am okay with). 

I guess my question is; should we even start the operator if there is no gateway that we can use installed?

Boomatang (commented):
Have yet to run it but wanted to ask about what happens if there is not gateways first. 

Also I know you have the to do's marked in. So I guess the linking functions would be the next step.

controllers/state_of_the_world.go
KevFan: Yeah, not sure what's the best approach but I've went with the approach of the existing controllers which gets disabled rather than crash when a dependency CRD is not detected (which we have existing integration tests for) 🤔 

Personally, I think it's better not to crash and have this reported in the Policy Status also since a user may not initially look through logs on why things are not working as expected.

eguzki: I agree, the operator should not crash. Personally, for now, disable watchers and maybe steps in the workflow. There should be a task responsible for the kuadrant CR status that watches for changes on Kuadrant CRDs (which we can assume they exist). Then, task checks for GatewayAPI being installed and reports in the status if not. Same for policies.

eguzki:
LGTM

I leave approval to @Boomatang

controllers/state_of_the_world.go
Boomatang: After the call I am happy that we have a plan to get the errors these missing CRDs and resources can cause into the status of kuadrant and / or the related policies. Would be good to create issue for any follow up work this may need.

Boomatang (approved):

PR #875, opened 2024-09-25
ratelimitpolicy v1beta3

eguzki:


api/v1beta3/ratelimitpolicy_types.go
```diff
@@ -141,7 +141,7 @@ type RateLimitPolicySpec struct {
 	// TargetRef identifies an API object to apply policy to.
 	// +kubebuilder:validation:XValidation:rule="self.group == 'gateway.networking.k8s.io'",message="Invalid targetRef.group. The only supported value is 'gateway.networking.k8s.io'"
 	// +kubebuilder:validation:XValidation:rule="self.kind == 'HTTPRoute' || self.kind == 'Gateway'",message="Invalid targetRef.kind. The only supported values are 'HTTPRoute' and 'Gateway'"
-	TargetRef gatewayapiv1alpha2.LocalPolicyTargetReference `json:"targetRef"`
+	TargetRef gatewayapiv1alpha2.LocalPolicyTargetReferenceWithSectionName `json:"targetRef"`
```
guicassolato, line 144: Do you mind keeping this one commit out if this PR please? It's done in https://github.com/Kuadrant/kuadrant-operator/pull/893 where it makes more sense IMO, because the functionality for it is also implemented there.

eguzki: I need that to implement the removal of routeSelectors. It's part of my assignment on https://github.com/Kuadrant/kuadrant-operator/issues/810

guicassolato: Please. Don't do that either. It's also removed in #893, along with the meaning of that.

eguzki: I am confused. I thought we agreed on doing the version bumping first together with the route selector removal. That's a change big enough to deserve it's own PR. Then in another PR, the refactor on the state of the world controller. 

I think it is reasonable and we should stick to the plan.

eguzki: > adding the sectionName has no meaning without the task the computes the effect RLPs.

Completely disagree. Adding the sectionName and removing the routeSelectors is itself the point of this PR. And it can be done without computing the effective policies and without refactor to to the state of the world controller

guicassolato: @eguzki, I see your point, but we are doing the same work twice. The work in #893 is done already, and that branch targets this one in case you want to merge it in one go.

Redoing the same thing in this PR is a waste of your time IMO, first because it's done in there and it will all be part of the same change anyway, but more importantly to me because here it would be a change to the API without the counterpart function that uses this field.

That said, if you really want to work on some part of this particular change still in this PR, the only thing I ask is please do not add `sectionName` just yet. You can remove `routeSelectors` and all the logic related to it. I'm fine with rebasing mine. Np.

eguzki:
With all due respect, @guicassolato, do not change the scope of the issues that we had agree upon in the planning stage on your own. That's not nice. You could be doing some work that someone else had already done because it was enclosed on another issue. That's wasting someones time.

I will remove all the changes that you already did and open PR with only the version bump.

guicassolato:
@eguzki,

I see where you are coming from. Indeed "Add `sectionName` field to targetRef" is listed in #810.

In the end, many of these tasks are convoluted I'm afraid. There's very little left to merging policies together without the sectionName and there is no reason to introduce sectionName other than for changing how we let policies target sections of the resources and consequently how we merge policy specs that add definitions only for those sections. IOW, sectionName and effective policies with D&O and merge strategies in context are all mixed together into one.

Unfortunately, the issues in GitHub do not do a great job at setting clearly the boundaries of what's one thing and what's another, especially in this case where lines are in fact blurred. We need more human communication to sync on this kind of stuff IMO.

Apologies if you have perceived as a change of plans. It was never my intention. I'm only trying to be proactive here. Last week we agreed on working on this part of the workflow together. You were kind enough to share a draft. I wanted added the next bit.

eguzki:
ready for rewiew @guicassolato

guicassolato (approved):
Just need to sort out those couple unit tests failing due to now seeing the hostnames in the conditions to call the RL service, but the general change LGTM. In fact, very thorough job on removing the `routeSelectors`. Thank you, @eguzki!

guicassolato (approved):

guicassolato (approved):
Neat.

eguzki:
heads up @trepel @azgabur 

This is breaking backward compat. More changes coming in!

PR #879, opened 2024-09-26
ConsolePlugin resource

jasonmadigan:
WIP

Ref: https://github.com/Kuadrant/console-plugin/issues/46

On OCP/OKD, apply a `ConsolePlugin` resource and associated deployment, to deploy https://github.com/Kuadrant/console-plugin alongside the rest of our components.

### DIFF
--- a/bundle/manifests/kuadrant-console-plugin.yaml
+++ b/bundle/manifests/kuadrant-console-plugin.yaml
@@ -1,0 +1,133 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kuadrant-console
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kuadrant-console
+  namespace: kuadrant-console
+  labels:
+    app: kuadrant-console
+    app.kubernetes.io/component: kuadrant-console
+    app.kubernetes.io/instance: kuadrant-console
+    app.kubernetes.io/name: kuadrant-console
+    app.kubernetes.io/part-of: kuadrant-console
+    app.openshift.io/runtime-namespace: kuadrant-console
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: kuadrant-console
+  template:
+    metadata:
+      labels:
+        app: kuadrant-console
+        app.kubernetes.io/component: kuadrant-console
+        app.kubernetes.io/instance: kuadrant-console
+        app.kubernetes.io/name: kuadrant-console
+        app.kubernetes.io/part-of: kuadrant-console
+    spec:
+      containers:
+      - name: kuadrant-console
+        image: quay.io/kuadrant/console-plugin:latest
+        ports:
+        - containerPort: 9443
+          protocol: TCP
+        imagePullPolicy: Always
+        volumeMounts:
+        - name: plugin-serving-cert
+          readOnly: true
+          mountPath: /var/serving-cert
+        - name: nginx-conf
+          readOnly: true
+          mountPath: /etc/nginx/nginx.conf
+          subPath: nginx.conf
+      volumes:
+      - name: plugin-serving-cert
+        secret:
+          secretName: plugin-serving-cert
+          defaultMode: 420
+      - name: nginx-conf
+        configMap:
+          name: nginx-conf
+          defaultMode: 420
+      restartPolicy: Always
+      dnsPolicy: ClusterFirst
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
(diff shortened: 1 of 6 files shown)


bundle/manifests/kuadrant-console-plugin.yaml
```diff
@@ -0,0 +1,133 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: kuadrant-console
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kuadrant-console
+  namespace: kuadrant-console
```
eguzki, line 10: this namespace is quite arbitrary. I would try to make it configurable at operator build time (or runtime.. let's discuss about this). One candidate for the default value would be the namespace where the operator is installed. But it is ok as well to be in `kuadrant-console`.

controllers/kuadrant_controller.go
```diff
@@ -167,6 +179,17 @@ func (r *KuadrantReconciler) Reconcile(eventCtx context.Context, req ctrl.Reques
 		return statusResult, nil
 	}
 
+	if r.isConsolePluginAvailable() {
+		logger.Info("ConsolePlugin API is available, proceeding to apply manifest")
+		if err := r.applyManifest(ctx, "bundle/manifests/kuadrant-console-plugin.yaml", r.Client()); err != nil {
```
eguzki, line 184: I am afraid, this is not gonna work 

I see four approaches:
* The kuadrant operator binary has embedded the manifests. (we use this approach for the grafana dashboards on the 3scale operator)
* The kuadrant operator reads from a configmap
* The kuadrant operator has the manifests available in the docker image, so the container can read the file from local filesystem as you try there. 
* The kuadrant operator creates deployment, service, nginx configmap... So all those resources are encoded in golang code. Pretty much like the limitador operator does with limitador. 

Each approach has pro's & con's. Maybe a combination of some of them. IDK.  I would go first with the fourth approach. But let's discuss.

controllers/kuadrant_controller.go
```diff
@@ -479,3 +502,58 @@ func (r *KuadrantReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&authorinov1beta1.Authorino{}).
 		Complete(r)
 }
+
+func (r *KuadrantReconciler) isConsolePluginAvailable() bool {
```
eguzki, line 506: It is already implemented in `kuadrantgatewayapi.IsCRDInstalled`
I patched it and the outcome is:
```diff
diff --git a/controllers/kuadrant_controller.go b/controllers/kuadrant_controller.go
index b7c89dd..7d7589f 100644
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -177,7 +177,13 @@ func (r *KuadrantReconciler) Reconcile(eventCtx context.Context, req ctrl.Reques
                return statusResult, nil
        }
 
-       if r.isConsolePluginAvailable() {
+       consolePluginGVK := consolev1.GroupVersion.WithKind("ConsolePlugin")
+       isConsolePluginAvailable, err := kuadrantgatewayapi.IsCRDInstalled(r.RestMapper, consolePluginGVK.Group, consolePluginGVK.Kind, consolePluginGVK.Version)
+       if err != nil {
+               return ctrl.Result{}, err
+       }
+
+       if isConsolePluginAvailable {
                logger.Info("ConsolePlugin API is available, proceeding to apply manifest")
                if err := r.applyManifest(ctx, "bundle/manifests/kuadrant-console-plugin.yaml", r.Client()); err != nil {
                        logger.Error(err, "failed to apply manifest")
@@ -501,17 +507,6 @@ func (r *KuadrantReconciler) SetupWithManager(mgr ctrl.Manager) error {
                Complete(r)
 }
 
-func (r *KuadrantReconciler) isConsolePluginAvailable() bool {
-       gvk := consolev1.GroupVersion.WithKind("ConsolePlugin")
-       mapping, err := r.RestMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
-       if err != nil || mapping == nil {
-               log.Log.Info("ConsolePlugin API is not available")
-               return false
-       }
-       log.Log.Info("ConsolePlugin API is available")
-       return true
-}
-
 func (r *KuadrantReconciler) applyManifest(ctx context.Context, filePath string, k8sClient client.Client) error {
        logger := logr.FromContextOrDiscard(ctx)
```

eguzki (commented):
Sweet, very cool stuff.

Left some comments we can use for further discussion

jasonmadigan:
superseded by #884

PR #799, opened 2024-08-09
[wip] use API to generate EPs and remove OCM

makslion:
Use a new API from `dns-operator` and remove all multicluster stuff

PR #882, opened 2024-09-27
[WIP] update integration tests

maksymvavilov:
Since we are no longer generating endpoints here in this PR I am:
1. Removing validation of the endpoints (checking only for their presence) 
2. Merging two test files reducing total amount of tests run against the dnspolicy controller 

It is a follow up on the #799


maksymvavilov:
won't do. We want those tests kept

PR #883, opened 2024-09-27
ADD: Watch limitador CR on cluster

Boomatang:
Adds a watch for the limitador CR on the cluster. This watches all limitador CRDs across the cluster. Later this will be filtered down to be more focused based of labels or owner reference. Owner reference filtering is still to be explored.

Linking function adds the only expected limitador CR into the topology based on the hard-coded name for that CR.

### DIFF
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -1,6 +1,7 @@
 package v1beta1
 
 import (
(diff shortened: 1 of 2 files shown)


KevFan (approved):
Looks good to me

PR #884, opened 2024-09-27
consoleplugin reconciler

eguzki:
### What

Deployment of Kuadrant's Openshift dynamic plugin.

ConsolePlugin v1.0 available on OCP >= 4.12 https://docs.openshift.com/container-platform/4.12/rest_api/console_apis/consoleplugin-console-openshift-io-v1.html 

The operator deploys the consoleplugin resources together with the service,deployment and configuration configmap in the operator's namespace

The lifecycle  of the consoleplugin resource, together with the related deployment resources, is linked to the lifecycle of the *topology* configmap. When the topology configmap is created, so is the consoleplugin (and related resources). Same when deleting the topology configmap. 

The consoleplugin will be deployed only on clusters where the CRDs are present, namely Openshift. 

Fixes https://github.com/Kuadrant/console-plugin/issues/46

TODO:
- [x] Prepare release task to add `RELATED_IMAGE_CONSOLEPLUGIN`
- [x] Update helm charts
- [x] Tests
- [x] Doc
- [ ] ~Lifecycle management: delete the plugin when the operator is uninstalled~ -> Left out for future implementation https://github.com/Kuadrant/kuadrant-operator/issues/896

### Verification Steps

Verified on OpenShift v4.15

* Install Gateway API

```shell
make gateway-api-install
```

* Install Istio

Follow [user guide](https://docs.kuadrant.io/0.10.0/kuadrant-operator/doc/install/install-openshift/#step-4-install-and-configure-istio-with-the-sail-operator)

Once istio is installed, we can deploy gateways and routes to be shown in the console plugin 🎨

* Deploy the kuadrant operator

Create `kuadrant-system` namespace. Currently it has to be `kuadrant-system` as it is hardcoded in the consoleplugin service.

```
kubectl create ns kuadrant-system
```

To install the Kuadrant Operator, enter the following command to deploy catalog from this `46-consoleplugin` branch:

```yaml
kubectl apply -f - <<EOF
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  namespace: kuadrant-system
spec:
  sourceType: grpc
  image: quay.io/kuadrant/kuadrant-operator-catalog:46-consoleplugin
  publisher: grpc
  updateStrategy:
    registryPoll:
      interval: 45m
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: kuadrant-operator
  namespace: kuadrant-system
spec:
  channel: preview
  installPlanApproval: Automatic
  name: kuadrant-operator
  source: kuadrant-operator-catalog
  sourceNamespace: kuadrant-system
---
kind: OperatorGroup
apiVersion: operators.coreos.com/v1
metadata:
  name: kuadrant
  namespace: kuadrant-system
spec:
  upgradeStrategy: Default
EOF
```
Wait for the Kuadrant Operators to be installed as follows:

```
kubectl get installplan -n kuadrant-system -o=jsonpath='{.items[0].status.phase}'
```

After some time, this command should return `complete`.

Installation done :rocket: Let's try it out 

* Enable Kuadrant's dynamic plugin

It may take some time to load the plugin. The first time it will be **Disabled**. It needs to be enabled.

In `Home -> Overview `, the **status** box has `Dynamic Plugins`. Click on it and then click on **view all**. Enable **kuadrant-console** plugin on the form as below:

Screenshot 2024-10-01 at 11-45-05 cluster · Console · Console plugins · Red Hat OpenShift

Once enabled, wait until the **status** says it is loaded, like in the snapshot below

Screenshot 2024-10-01 at 11-47-06 cluster · Console · Console plugins · Red Hat OpenShift

Refresh the openshift's console and a new entry `Kuadrant` in the left menu will show up

Screenshot 2024-10-01 at 11-51-39 cluster · Console · Console plugins · Red Hat OpenShift

If you check **Overview** and **Policies** it should be all empty (provided there is no gateway, httproute and kuadrant policy installed. Let's populate the cluster with some interesting resources.

* Deploy resources

Apply the Kuadrant custom resource

```yaml
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec: {}
EOF
```

Add one gateway

```yaml
---
kind: Gateway
metadata:
  labels:
    istio: ingressgateway
  name: kuadrant-ingressgateway
spec:
  gatewayClassName: istio
  listeners:
  - name: http
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
EOF
```

Add one HTTPRoute
```yaml
---
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/toys"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

Add one RateLimitPolicy

```yaml
---
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    "create-toy":
      rates:
      - limit: 5
        duration: 10
        unit: second
EOF
```

After that, the topology graph looks like this:

Screenshot 2024-10-01 at 12-07-33 Policy Topology

* Check console plugin resource (cluster wide resource):
```
kubectl get consoleplugin kuadrant-console -o yaml
```

You should get something like:

```yaml
apiVersion: console.openshift.io/v1
kind: ConsolePlugin
metadata:
  creationTimestamp: "2024-10-01T09:39:54Z"
  generation: 1
  labels:
    app: kuadrant-console
    app.kubernetes.io/component: kuadrant-console
    app.kubernetes.io/instance: kuadrant-console
    app.kubernetes.io/name: kuadrant-console
    app.kubernetes.io/part-of: kuadrant-console
  name: kuadrant-console
  resourceVersion: "130839645"
  uid: 5318cd76-993b-4597-b681-fe89257eadff
spec:
  backend:
    service:
      basePath: /
      name: kuadrant-console
      namespace: kuadrant-system
      port: 9443
    type: Service
  displayName: Kuadrant Console Plugin
  i18n:
    loadType: ""
```

### Cleanup

```
kubectl delete consoleplugin kuadrant-console
```

```
kubectl delete namespace kuadrant-system
```


controllers/state_of_the_world.go
```diff
@@ -130,16 +132,40 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
 		// TODO: add tls policy specific tasks to workflow
 	}
 
+	ok, err = openshift.IsConsolePluginInstalled(manager.GetRESTMapper())
+	if err != nil || !ok {
+		logger.Info("console plugin is not installed, skipping related watches and reconcilers", "err", err)
+	} else {
+		controllerOpts = append(controllerOpts,
+			controller.WithRunnable("consoleplugin watcher", controller.Watch(&consolev1.ConsolePlugin{}, openshift.ConsolePluginsResource, metav1.NamespaceAll)),
```
guicassolato, line 140: Should we use label predicates here?

```suggestion
			controller.WithRunnable("consoleplugin watcher", controller.Watch(&consolev1.ConsolePlugin{}, openshift.ConsolePluginsResource, metav1.NamespaceAll, controller.FilterResourcesByLabel[*consolev1.ConsolePlugin](fmt.Sprintf("app=%s", consoleplugin.KUADRANT_CONSOLE)))),
```

.github/workflows/build-images-base.yaml
```diff
@@ -27,6 +27,10 @@ on:
         description: WASM Shim version
         default: latest
         type: string
+      consolePluginImageURL:
```
eguzki, line 30: @didierofrivia  @maleck13 trying here complete URL instead of some "version". What are your thoughts about this?

controllers/state_of_the_world.go
eguzki: done

controllers/state_of_the_world.go
```diff
@@ -130,16 +133,42 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
 		// TODO: add tls policy specific tasks to workflow
 	}
 
+	ok, err = openshift.IsConsolePluginInstalled(manager.GetRESTMapper())
+	if err != nil || !ok {
+		logger.Info("console plugin is not installed, skipping related watches and reconcilers", "err", err)
+	} else {
+		controllerOpts = append(controllerOpts,
```
jasonmadigan, line 140: think it makes sense to show the kuadrant `ConsolePlugin`, but noticed from the topology screenshot it's showing _all_ `ConsolePlugin` resources, which we probably don't want (e.g. the monitoring-plugin, which isn't related to Kuadrant).

eguzki: Right, I will implement two "filters" in that regards:
* the consoleplugin watcher now contains a predicate with a label selector
* Only watch on the expected namespace, that the operator knows at this stage.

jasonmadigan: perfect

jasonmadigan:
Merged the docs

jasonmadigan:
LGTM

jasonmadigan (approved):

Makefile
```diff
@@ -340,7 +340,7 @@ build: generate fmt vet ## Build manager binary.
 
 run: export LOG_LEVEL = debug
 run: export LOG_MODE = development
-run: export OPERATOR_NAMESPACE = kuadrant-system
+run: export OPERATOR_NAMESPACE = $(shell kubectl config view --minify -o jsonpath='{..namespace}')
```
KevFan, line 343: This gives an empty value for me for a kind and crc cluster, so it will panic for most people using `make run` here unless it's just me 🤔 
https://github.com/Kuadrant/kuadrant-operator/blob/128ea20995e1b81fb4f26c9b63e8c175ee68e17e/controllers/state_of_the_world.go#L158

```json
{
    "kind": "Config",
    "apiVersion": "v1",
    "preferences": {},
    "clusters": [
        {
            "name": "api-crc-testing:6443",
            "cluster": {
                "server": "https://api.crc.testing:6443",
                "certificate-authority-data": "DATA+OMITTED"
            }
        }
    ],
    "users": [
        {
            "name": "developer/api-crc-testing:6443",
            "user": {
                "token": "REDACTED"
            }
        }
    ],
    "contexts": [
        {
            "name": "/api-crc-testing:6443/developer",
            "context": {
                "cluster": "api-crc-testing:6443",
                "user": "developer/api-crc-testing:6443"
            }
        }
    ],
    "current-context": "/api-crc-testing:6443/developer"
}
```

controllers/consoleplugin_reconciler.go
```diff
@@ -0,0 +1,128 @@
...
+	})
+
+	topologyExists := len(existingTopologyConfigMaps) > 0
+
+	// Service
+	service := consoleplugin.Service(r.namespace)
+	if !topologyExists {
+		utils.TagObjectToDelete(service)
+	}
+	err := r.ReconcileResource(ctx, &corev1.Service{}, service, reconcilers.CreateOnlyMutator)
```
KevFan, line 85: I guess if we want to align with the conventions agreed, we do not want to use this `r.ReconcileResource` function, since it performs a `Get` from the api server. Instead we want to get this resource from the topology 🤔 
Although not sure do we want to block this PR for this

Makefile
eguzki: The command line `kubectl config view --minify -o jsonpath='{..namespace}'` is pretty common way to parse the current context's namespace. I do not think there is anything wrong with it. 

Your kubeconfig does not have namespace set on the current context. This is a portion of my kubeconfig

```yaml
contexts:
- context:                                                                 
    cluster: mycluster:443                   
    namespace: kuadrant-system                                             
    user: eguzki/mycluster:443               
  name: kuadrant-system/mycluster:443/eguzki 
current-context: kuadrant-system/mycluster:443/eguzki 
```

Maybe `oc project kuadrant-system` will help??

I will try to enhance makefile command to handle error when the namespace is not available.

controllers/consoleplugin_reconciler.go
eguzki: `ReconcileResource` has some functionality we still does not have in the new approach. Do you mind letting this implementation go through? In exchange, I will open issue to address that as tech debt

BTW, I realize that the old issue of endless reconciliation loop for authconfig https://github.com/Kuadrant/kuadrant-operator/issues/354 we fixed with the dry run https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/library/reconcilers/base_reconciler.go#L156

```go
desired.SetResourceVersion(obj.GetResourceVersion())
	if err := b.Client().Update(ctx, desired, client.DryRunAll); err != nil {
		return err
	}
```
can now become an issue again with the topology being the source of truth.

eguzki: added

Makefile
eguzki: added

KevFan: Hmm, yeah adding `oc project kuadrant-system` will add the `namespace` to the project, but without it there no namespace in the context

controllers/consoleplugin_reconciler.go
KevFan: Yeah, I'm fine with leaving it to be addressed later 👍

Makefile
KevFan: Yeah, I don't know about this change tbh. I'm getting the following whenever I run `make run` on a fresh **kind** cluster made using `make local-env-setup`, so this will affect the `make run` command once merged to main.
```
🚨 namespace could not be parsed from kubeconfig 💥
```
This does get set on a **CRC** cluster as I was interfacting with the cluster by following the verification steps. So I assume this gets set typically when interacting with an openshift type cluster unless I'm missing something 🤔 

TBH the env var has a default now also so we probably don't need this set in the makefile also unless people want to explicitly change it
https://github.com/Kuadrant/kuadrant-operator/blob/c194aa4144349fc4f5be61cb549fa2042ca3e49c/controllers/state_of_the_world.go#L35

KevFan:
Verified steps on a local CRC cluster on Openshift version 4.16.7

Makefile
eguzki: Sorry, I wanted to read the namespace from the kubeconfig context, but turns out that in local env (kind), there is no namespace for the current context. 

Reverting to previous state with the addition of adding `OPERATOR_NAMESPACE` makefile variable to override default value (`kuadrant-system`).

eguzki: Actually, I think that for `make run`, the `OPERATOR_NAMESPACE` should be:
* When `OPERATOR_NAMESPACE` makefile variable is set, it takes that value. Else
* namespace from kubeconfig current context. If empty, then
* default from `KUADRANT_NAMESPACE` makefile variable, i.e. `kuadrant-system`.

But we can leave this for other PR.

KevFan (approved):
Looks good to me!

.github/workflows/build-images-base.yaml
eguzki: sorry guys, accepted by (answer) omission. 

We can always change that later

Makefile
eguzki: cc @Boomatang ^^

controllers/consoleplugin_reconciler.go
eguzki: Created issue https://github.com/Kuadrant/kuadrant-operator/issues/916

Issue #824, opened 2024-08-16
[state-of-the-world reconciler] TLSPolicy status


guicassolato:
Move the business logic of TLSPolicy status reconciliation to a workflow state-of-the-world reconcile func.

PR #885, opened 2024-09-27
refactor: tls policy status to state of the world tasks

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/824

* Refactor the TLSPolicy status code to workflow tasks
* Added a new `MissingDependency` status of TLS Policy for if CertManager is not installed

# Verification
* Passing integration tests should be sufficient to ensure current functionality around TLSPolicy Status still works as intended
* To verify new `MissingDependency` TLSPolicy status
  *  Create cluster without CertManager and run operator
  ```
  make local-env-setup
  # Delete CertManager CRDs to minic cert manager is uninstalled
  kubectl delete crd certificaterequests.cert-manager.io 
  kubectl delete crd certificates.cert-manager.io
  kubectl delete crd challenges.acme.cert-manager.io
  kubectl delete crd clusterissuers.cert-manager.io
  kubectl delete crd issuers.cert-manager.io
  kubectl delete crd orders.acme.cert-manager.io
  # Run operator
  make run
  ``` 
  * Create kuadrant-system namespace and tls policiy
  ```sh
  # Create TLSPolicy
  kubectl apply -f - <<EOF
  kind: TLSPolicy
  metadata:
    name: gw-tls
  spec:
    targetRef:
      kind: Gateway
      name: istio-ingressgateway
    issuerRef:
      group: cert-manager.io
      kind: ClusterIssuer
      name: selfsigned-issuer
  EOF
  ```
  * Get TLSPolicy status 
  ```sh
  kubectl get tlspolicy -A -o yaml | yq '.items[].status'
  ```
  * TLSPolicy status should be the following:
  ```yaml
  conditions:
  - lastTransitionTime: "2024-10-02T13:34:59Z"
    message: Cert Manager is not installed, please restart pod once dependency has been installed
    reason: MissingDependency
    status: "False"
    type: Accepted
  observedGeneration: 1
  ```


controllers/state_of_the_world.go
```diff
@@ -181,6 +191,11 @@ func (r *TopologyFileReconciler) Reconcile(ctx context.Context, _ []controller.R
 	if len(existingTopologyConfigMaps) == 0 {
 		_, err := r.Client.Resource(controller.ConfigMapsResource).Namespace(cm.Namespace).Create(ctx, unstructuredCM, metav1.CreateOptions{})
 		if err != nil {
+			if errors.IsAlreadyExists(err) {
+				// This error can happen when the operator is starting, and the create event for the topology has not being processed.
```
guicassolato, line 195: Hmmm... We may expect more of these, and not only with this particular resource.

This is because, as a pattern, we check the presence of the resource in the topology, and not with the API server. But more importantly, because, even though it's "state of the world" reconciliation and therefore the controller already knows about all resources that exist (because it lists them all), the library watching for changes to the cache writes to the subscription channel [resource by resource](https://github.com/Kuadrant/policy-machinery/blob/585f10911abdbb919512b3317c9db073eccf99a3/controller/cache.go#L100).

I think we should consider fixing this in the Policy Machinery instead, because now it may be defeating the purpose of "state of the world". We may be able to make the call to [`cache.Replace`](https://github.com/Kuadrant/policy-machinery/blob/585f10911abdbb919512b3317c9db073eccf99a3/controller/controller.go#L220C2-L220C24) truly atomic – assuming it won't break with the granularity of multiple event types. I have a couple of ideas.

guicassolato: I believe https://github.com/Kuadrant/policy-machinery/pull/35 may have fixed this issue.

KevFan: This still happens with the same error:

```
{"level":"error","ts":"2024-10-02T08:56:41+01:00","logger":"kuadrant-operator","msg":"reconciliation error","error":"configmaps \"topology\" already exists","stacktrace":"github.com/kuadrant/policy-machinery/controller.(*Controller).propagate\n\t/Users/chfan/dev/kuadrant-operator/vendor/github.com/kuadrant/policy-machinery/controller/controller.go:262\ngithub.com/kuadrant/policy-machinery/controller.(*Controller).subscribe.func2\n\t/Users/chfan/dev/kuadrant-operator/vendor/github.com/kuadrant/policy-machinery/controller/controller.go:312"}
{"level":"info","ts":"2024-10-02T08:56:41+01:00","logger":"kuadrant-operator.controller-runtime.metrics","msg":"Starting metrics server"}
{"level":"info","ts":"2024-10-02T08:56:41+01:00","logger":"kuadrant-operator","msg":"starting server","name":"health probe","addr":"[::]:8081"}
```

What looks to happens from debugging is that this code is ran on boot even when there has been no events and when the topology has been loaded with the cluster data yet 🤔 

My guess is that since this is a task with no event matchers, the initial creation of the empty cache store is an event that causes these task to run 🤔

controllers/state_of_the_world.go
```diff
@@ -234,3 +369,25 @@ func (e *EventLogger) Log(ctx context.Context, resourceEvents []controller.Resou
 
 	return nil
 }
+
+func NewTLSPolicyWorkflow(client *dynamic.DynamicClient) *controller.Workflow {
+	return &controller.Workflow{
+		Precondition:  NewValidateTLSPolicyTask().Reconcile,
+		Postcondition: NewTLSPolicyStatusTask(client).Reconcile,
+	}
+}
```
guicassolato, line 378: Care to move this func to `tlspolicy_tasks.go` please?

controllers/state_of_the_world.go
```diff
@@ -234,3 +371,18 @@ func (e *EventLogger) Log(ctx context.Context, resourceEvents []controller.Resou
 
 	return nil
 }
+
+func isObjectOwnedByKind(o client.Object, ownerGroup, ownerKind string) bool {
```
guicassolato, line 375: Could receive a `schema.GroupKind` instead of 2 strings, and thus help with consistently defining and reusing all those XGroupKind variables, without any weird stuff.

controllers/state_of_the_world.go
```diff
@@ -116,33 +194,92 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+				UpdateFunc: func(e event.TypedUpdateEvent[*certmanagerv1.Certificate]) bool {
+					return isCertificateOwnedByTLSPolicy(e.ObjectNew)
+				},
+				DeleteFunc: func(e event.TypedDeleteEvent[*certmanagerv1.Certificate]) bool {
+					return isCertificateOwnedByTLSPolicy(e.Object)
+				},
+				GenericFunc: func(e event.TypedGenericEvent[*certmanagerv1.Certificate]) bool {
+					return isCertificateOwnedByTLSPolicy(e.Object)
+				},
+			})),
```
guicassolato, line 244: Interesting approach. I wonder how it performs compared to labels.

controllers/state_of_the_world.go
KevFan: That's a good idea, I'll update 👍

controllers/state_of_the_world.go
KevFan: Yeah, I went with ownerReference instead since we already set the ownerRef to point to `TLSPolicy` since we want the associated `Certificates` deleted when the `TLSPolicy` is deleted.

Also we don't currently set a common label across all created `Certificate`'s
https://github.com/Kuadrant/kuadrant-operator/blob/4793dc4d2ad522a96f696c68f448e8b4518a6147/controllers/tlspolicy_certmanager_certificates.go#L161-L184

controllers/state_of_the_world.go
```diff
@@ -62,17 +96,31 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+				&gwapiv1.GatewayClass{},
+				controller.GatewayClassesResource,
+				metav1.NamespaceAll,
+				controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*gwapiv1.GatewayClass]{})),
+			),
+			controller.WithRunnable("gateway watcher", controller.Watch(
+				&gwapiv1.Gateway{},
+				controller.GatewaysResource,
+				metav1.NamespaceAll,
+				controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*gwapiv1.Gateway]{})),
```
guicassolato, line 116: We probably want to watch gateways, as well as routes, when their statuses change.

controllers/state_of_the_world.go
```diff
@@ -45,13 +50,45 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+		controller.WithRunnable("authpolicy watcher", controller.Watch(
+			&kuadrantv1beta2.AuthPolicy{},
+			kuadrantv1beta2.AuthPoliciesResource,
+			metav1.NamespaceAll,
+			controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*kuadrantv1beta2.AuthPolicy]{})),
+		),
+		controller.WithRunnable("ratelimitpolicy watcher", controller.Watch(
+			&kuadrantv1beta2.RateLimitPolicy{},
+			kuadrantv1beta2.RateLimitPoliciesResource,
+			metav1.NamespaceAll),
```
guicassolato, line 80: Any reason why we wouldn't have `TypedGenerationChangedPredicate` for RateLimitPolicy too?

controllers/state_of_the_world.go
```diff
@@ -67,17 +104,30 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
 	} else {
 		controllerOpts = append(controllerOpts,
-			controller.WithRunnable("gatewayclass watcher", controller.Watch(&gwapiv1.GatewayClass{}, controller.GatewayClassesResource, metav1.NamespaceAll)),
-			controller.WithRunnable("gateway watcher", controller.Watch(&gwapiv1.Gateway{}, controller.GatewaysResource, metav1.NamespaceAll)),
-			controller.WithRunnable("httproute watcher", controller.Watch(&gwapiv1.HTTPRoute{}, controller.HTTPRoutesResource, metav1.NamespaceAll)),
+			controller.WithRunnable("gatewayclass watcher", controller.Watch(
+				&gwapiv1.GatewayClass{},
+				controller.GatewayClassesResource,
+				metav1.NamespaceAll,
+				controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*gwapiv1.GatewayClass]{})),
```
guicassolato, line 119: I'm afraid GatewayClasses can change status to `Accepted: false` and `Unknown` too.

controllers/state_of_the_world.go
```diff
@@ -86,9 +136,24 @@ func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.D
...
+				envoygateway.EnvoyExtensionPoliciesResource,
+				metav1.NamespaceAll,
+				controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*egv1alpha1.EnvoyExtensionPolicy]{})),
+			),
+			controller.WithRunnable("envoysecuritypolicy watcher", controller.Watch(
+				&egv1alpha1.SecurityPolicy{},
+				envoygateway.SecurityPoliciesResource,
+				metav1.NamespaceAll,
+				controller.WithPredicates(&ctrlruntimepredicate.TypedGenerationChangedPredicate[*egv1alpha1.SecurityPolicy]{})),
+			),
```
guicassolato, line 156: Same https://github.com/Kuadrant/kuadrant-operator/pull/885/files#r1787506477

controllers/state_of_the_world.go
KevFan: Hmm, no, I think I just missed this, will update 👍

controllers/state_of_the_world.go
KevFan: Oh cool, I'll remove this predicate 👍

controllers/state_of_the_world.go
KevFan: Will remove this predicate, and also for this isio CRDs since I assume we want status changes for those also

controllers/tlspolicies_validator.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	for _, policy := range policies {
+		if policy.DeletionTimestamp != nil {
+			logger.V(1).Info("tls policy is marked for deletion, skipping", "name", policy.Name, "namespace", policy.Namespace)
+			continue
+		}
+
+		if !isCertManagerInstalled {
+			s.Store(TLSPolicyAcceptedKey(policy.GetUID()), kuadrant.NewErrDependencyNotInstalled("Cert Manager"))
```
guicassolato, line 68: Interesting. You opted for one entry in the state map per validated TLSPolicy. In #893, I'm going with one `map[string]error` for all RateLimitPolicies.

In the key, I'm using `GetLocator()`. I guess you went with `GetUID()` because you had cast the policies to the concrete type already, right?

controllers/tlspolicies_validator.go
```diff
@@ -0,0 +1,96 @@
...
+			// Can't find gateway target ref
+			if !ok {
+				logger.V(1).Info("tls policy cannot find target ref", "name", policy.Name, "namespace", policy.Namespace)
+				s.Store(TLSPolicyAcceptedKey(policy.GetUID()), kuadrant.NewErrTargetNotFound(policy.Kind(), policy.GetTargetRef(), apierrors.NewNotFound(kuadrantv1alpha1.TLSPoliciesResource.GroupResource(), policy.GetName())))
+				continue
+			}
+
+			logger.V(1).Info("tls policy found target ref", "name", policy.Name, "namespace", policy.Namespace)
+			s.Store(TLSPolicyAcceptedKey(policy.GetUID()), nil)
+		}
```
guicassolato, line 92: Interestingly, we may not need to find the object the policy points to.

Because policies are linked to their targets already, a policy `p` where `len(p.GetTargetRefs()) > 0 && len(topology.Targetables().Children(p)) == 0` cannot be valid.

guicassolato: Obviously, what I'm suggesting won't give enough details to report in the message which target cannot be found.

controllers/tlspolicy_status_updater.go
```diff
@@ -0,0 +1,233 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func TLSPolicyAcceptedKey(uid types.UID) string {
+	return fmt.Sprintf("TLSPolicyValid:%s", uid)
+}
```
guicassolato, line 233: Perhaps move this func to `tls_workflow.go`. It's used by the validator and the status updater, right?

controllers/tlspolicy_status_updater.go
```diff
@@ -0,0 +1,233 @@
...
+
+func (t *TLSPolicyStatusUpdaterReconciler) enforcedCondition(ctx context.Context, tlsPolicy *kuadrantv1alpha1.TLSPolicy, topology *machinery.Topology) *metav1.Condition {
+	if err := t.isIssuerReady(ctx, tlsPolicy, topology); err != nil {
+		return kuadrant.EnforcedCondition(tlsPolicy, kuadrant.NewErrUnknown(tlsPolicy.Kind(), err), false)
+	}
+
+	if err := t.isCertificatesReady(ctx, tlsPolicy, topology); err != nil {
+		return kuadrant.EnforcedCondition(tlsPolicy, kuadrant.NewErrUnknown(tlsPolicy.Kind(), err), false)
+	}
+
```
guicassolato, line 117: TLS is an interesting case. If the gateway is not Programmed, it doesn't mean we cannot provision the certificates for its listeners. Is that why you didn't include the status of the gateway here?

controllers/tls_workflow.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return machinery.LinkFunc{
+		From: machinery.GatewayGroupKind,
+		To:   CertManagerCertificateKind,
+		Func: func(child machinery.Object) []machinery.Object {
+			o := child.(*controller.RuntimeObject)
+			cert := o.Object.(*certmanagerv1.Certificate)
+
+			gateway, ok := lo.Find(gateways, func(g *gwapiv1.Gateway) bool {
+				for _, l := range g.Spec.Listeners {
```
guicassolato, line 44: I wonder if these objects should be liked to the gateway listeners instead of the gateways as a whole. If not now, in the near future.

With `sectionName` introduced to the TLSPolicy, the policies themselves may be linked to the entire gateway or to a specific listener only. I suppose that could make a difference in the decision of the number of CSRs that are created for each policy resource – one with multiple CNs or one per listener.

Either way, the case of a certificate linked to a individual listener obviously exists. It's actually the default today as we currently do not create CSRs with multiple CNs IIRC. As I can tell from this loop, the listener is definitely the driver to decide if the object is linked or not anyway.

WDYT?

Interested in your thoughts too @maleck13.

maleck13: Yes I think having TLSPolicy work against individual listeners is the way to go. I think we can think of it similar to DNS 
- Each listener is evaluated and the DNSPolicy reconciled for that listener
- Each listener has its status updated in the gateway with enforced  / accpeted conditions
- Probably remove the overall gateway status. 

^ some of this may be out of scope for now

controllers/tlspolicy_status_updater.go
KevFan: Exactly, a gateway can transition to a `Programmed` state from the certs created by the current tls policy controller.

controllers/tlspolicies_validator.go
KevFan: Oh cool, yeah that's a nicer way of knowing whether a policy can't find their targetRef, I'll update 👍

KevFan: `len(p.GetTargetRefs()) == len(topology.Targetables().Children(p))` is probably what we expect instead? 🤔 
I think this is fine as we currently don't support multiple target refs but this may not work if we want to report which is not found as you've said, and also not sure what the policy status should be in the case where at least one target ref in the list is not found  🤔

controllers/state_of_the_world.go
guicassolato: > the initial creation of the empty cache store is an event that causes these task to run

Oh! This makes total sense and explains why it's triggering before the controller finishes starting. By simply creating the cache we may be generating events already. Nice catch, @KevFan!

guicassolato: I'm guessing that somewhere [here](https://github.com/Kuadrant/policy-machinery/blob/d4be292ed9e72de1cd8e04cd57c46415339336b2/controller/controller.go#L273), when the snapshot contains zero updates, then we just `continue`?

Another option is, `if len(events) == 0 { continue }` [here](https://github.com/Kuadrant/policy-machinery/blob/d4be292ed9e72de1cd8e04cd57c46415339336b2/controller/controller.go#L312).

controllers/tls_workflow.go
KevFan: Sounds good, I'll see can I do the linking to certificates instead as part of this PR. If not, I'll do that change with the refactor of main reconcile logic 👍

controllers/cert_manager_installed.go
KevFan, line 1: Not sure about this task. Technically this value can be passed from the inital boot check in `state_of_the_world.go` for whether the watchers should be registered. I suppose with this being a constant check, reconciling functions that depend of this boolean can still happen if the CRDs gets installed **after** the operator starts. It's only that the operator will not have watchers/receives events from these CRDs until the pod is restarted 🤔

controllers/tls_workflow.go
```diff
@@ -1,9 +1,156 @@
...
+	CertManagerIssuerKind        = schema.GroupKind{Group: certmanager.GroupName, Kind: certmanagerv1.IssuerKind}
+	CertManagerClusterIssuerKind = schema.GroupKind{Group: certmanager.GroupName, Kind: certmanagerv1.ClusterIssuerKind}
 )
 
-func NewTLSWorkflow() *controller.Workflow {
-	return &controller.Workflow{}
+func NewTLSWorkflow(client *dynamic.DynamicClient) *controller.Workflow {
+	return &controller.Workflow{
+		Precondition:  NewValidateTLSPoliciesValidatorReconciler().Validate,
+		Postcondition: NewTLSPolicyStatusUpdaterReconciler(client).UpdateStatus,
```
guicassolato, line 33: Are we going to subscribe to TLS-specific events?

```suggestion
		Precondition:  NewValidateTLSPoliciesValidatorReconciler().Subscription().Reconcile,
		Postcondition: NewTLSPolicyStatusUpdaterReconciler(client).Subscription().Reconcile,
```

controllers/tlspolicies_validator.go
```diff
@@ -0,0 +1,82 @@
...
+			logger.V(1).Info("tls policy is marked for deletion, skipping", "name", p.Name, "namespace", p.Namespace)
+			continue
+		}
+
+		if !isCertManagerInstalled {
+			isPolicyValidErrorMap[p.GetLocator()] = kuadrant.NewErrDependencyNotInstalled("Cert Manager")
+			continue
+		}
+
+		// TODO: What should happen if multiple target refs is supported in the future in terms of reporting in log and policy status?
```
guicassolato, line 68: We're probably going to have to report status by target, i.e. split the definition of status stanza of the policy so we can have one set of status conditions per targetRef.

controllers/cert_manager_installed.go
guicassolato: > reconciling functions that depend of this boolean can still happen if the CRDs gets installed **after** the operator starts. It's only that the operator will not have watchers/receives events from these CRDs until the pod is restarted

What I do not like about this is that it may feel like the operator is missing events. In fact, it _will_ miss events.

Say the tasks subscribe to `Gateway`, `TLSPolicy`, `Certificate`, `Issuer` and `ClusterIssuer`. At boot-time, cert-manager CRDs are not available, so watchers for `Certificate`, `Issuer` and `ClusterIssuer` are not registered. Then, these CRDs become available and reconciliation starts working perfectly fine for every event related to `Gateway` and `TLSPolicy`. However, because the controller is not watching for `Certificate`, `Issuer` and `ClusterIssuer`, events related to these other custom resources will be ignored, including when the status of the `Certificate` changes.

Only from that we can tell that the policies cannot ever move to `Enforced: true`.

This may become hard to debug for a user. Everything seems to be working perfectly fine, except that the policies report `Enforced: false`. Nevertheless, they _are_ being enforced, because the CSRs _were_ created.

On one hand, we want to update the status of the policies regarding missing a dependency installed in the cluster if that's the case. On the other hand, we do not want to put the user in such a situation where it's hard to explain the actual state of things.

Perhaps a better approach would be sticking with the info we have at boot time regarding the state of the deps all along until the operator is redeployed. Instead of having this task, we'd just store that info once, in a field of the state-of-the-world reconciler. Status update would continue to run for `Gateway` and `TLSPolicy` related events, but it would continuously think cert-manager is not installed even after it is. Therefore, CSRs would not be created, the policies would not be enforced, but at least that would be consistent with what it reports in the status.

WDYT?

controllers/tls_workflow.go
KevFan: Yes, missed this. Good catch 👍

controllers/tlspolicies_validator.go
KevFan: I think that makes sense, but I think i'll leave that as a seperate PR as this would be an enhancement

controllers/cert_manager_installed.go
KevFan: Yeah, agreed, this makes sense. I remove this task and just use the value at boot time 👍

KevFan: Updated

controllers/state_of_the_world.go
```diff
@@ -39,9 +42,17 @@ var (
...
+	isEnvoyGatewayInstalled  bool
+	isIstioInstalled         bool
+	isCertManagerInstalled   bool
+	isConsolePluginInstalled bool
+}
+
 //+kubebuilder:rbac:groups=gateway.networking.k8s.io,resources=gatewayclasses,verbs=list;watch
 
-func NewPolicyMachineryController(manager ctrlruntime.Manager, client *dynamic.DynamicClient, logger logr.Logger) *controller.Controller {
+func (c *PolicyMachineryController) Controller(manager ctrlruntime.Manager, client *dynamic.DynamicClient, logger logr.Logger) *controller.Controller {
```
guicassolato, line 55: I would probably keep this as a direct constructor that returns a `*controller.Controller`, instead of a method of a struct that now needs to be known at `main.go`.

Additionally, I would make `PolicyMachineryController` into `policyMachineryControllerOptionsBuilder`. Make it an internal thing used by `NewPolicyMachineryController`. Name is probably too long BTW.

The builder could implement methods such as `getGatewayAPIOptions() ([]controller.ControllerOption, error)`, `getEnvoyGatewayoptions() ([]controller.ControllerOption, error)` and `buildReconciler() controller.ReconcileFunc`.

I think this can help breaking down the constructor into smaller steps and avoid any temptation of doing `&PolicyMachineryController{}` and getting something that actually is not a Policy Machinery Controller yet.

WDYT?

KevFan: Yeah, this sounds a lot nicer, I'll push a change for this 👍

controllers/tlspolicy_certmanager_certificates.go
```diff
@@ -130,12 +131,36 @@ func (r *TLSPolicyReconciler) expectedCertificatesForGateway(ctx context.Context
...
+			Name: string(certRef.Name),
+		}
+		if certRef.Namespace != nil {
+			secretRef.Namespace = string(*certRef.Namespace)
+		} else {
+			secretRef.Namespace = l.GetNamespace()
+		}
+		// Gateway API hostname explicitly disallows IP addresses, so this
+		// should be OK.
+		tlsHosts[secretRef] = append(tlsHosts[secretRef], string(*l.Hostname))
```
guicassolato, line 153: I believe if the listener has no hostname, this line will throw a nil pointer dereference exception.

guicassolato: A workaround (backed by the [docs](https://pkg.go.dev/sigs.k8s.io/gateway-api@v1.1.0/apis/v1#Listener.Hostname)) is considering `*` when `l.Hostname == nil`.

controllers/tlspolicy_status_updater.go
```diff
@@ -0,0 +1,240 @@
...
+	logger := controller.LoggerFromContext(ctx).WithName("TLSPolicyStatusUpdaterReconciler").WithName("Reconcile")
+
+	policies := lo.FilterMap(topology.Policies().Items(), func(item machinery.Policy, index int) (*kuadrantv1alpha1.TLSPolicy, bool) {
+		p, ok := item.(*kuadrantv1alpha1.TLSPolicy)
+		return p, ok
+	})
+
+	store, ok := s.Load(TLSPolicyAcceptedKey)
+	if !ok {
+		logger.Error(errors.New("TLSPolicyAcceptedKey not found, skipping update of tls policy statuses"), "sync map error")
```
guicassolato, line 59: `ok == false` should never happen, given both reconcilers subscribe to the exact same event matchers, unless the validator panics half-way. Good call logging the error.

controllers/tlspolicy_status_updater.go
```diff
@@ -0,0 +1,240 @@
...
+		// Do not set enforced condition if Accepted condition is false
+		if meta.IsStatusConditionFalse(newStatus.Conditions, string(gatewayapiv1alpha2.PolicyReasonAccepted)) {
+			meta.RemoveStatusCondition(&newStatus.Conditions, string(kuadrant.PolicyConditionEnforced))
+		} else {
+			enforcedCond := t.enforcedCondition(ctx, policy, topology)
+			meta.SetStatusCondition(&newStatus.Conditions, *enforcedCond)
+		}
+
+		// Nothing to do
+		equalStatus := equality.Semantic.DeepEqual(newStatus, policy.Status)
```
guicassolato, line 89: We need to be careful to never allow unexported fields in the [`TLSPolicyStatus`](https://github.com/Kuadrant/kuadrant-operator/blob/b8f0bf6b9a4a13bb0defcad29b553ab88cd810ec/api/v1alpha1/tlspolicy_types.go#L117) type, otherwise this line will start panicking.

controllers/tlspolicy_status_updater.go
```diff
@@ -0,0 +1,240 @@
...
+		resource := t.Client.Resource(kuadrantv1alpha1.TLSPoliciesResource).Namespace(policy.GetNamespace())
+		un, err := controller.Destruct(policy)
+		if err != nil {
+			logger.Error(err, "unable to destruct policy")
+			continue
+		}
+
+		_, err = resource.UpdateStatus(ctx, un, metav1.UpdateOptions{})
+		if err != nil {
+			logger.Error(err, "unable to update status for TLSPolicy", "name", policy.GetUID(), "namespace", policy.GetNamespace(), "uid", policy.GetUID())
```
guicassolato, line 106: The log message at line 67 uses the policy name instead of the object's UID, which I personally think it's easier for debugging. Any particular reason not to do the same here?

controllers/tlspolicy_status_updater.go
KevFan: Yeah, this should never happen. This can also happen if there is a difference in the subscriptions between the validator and status updater in the future

controllers/tlspolicy_status_updater.go
KevFan: Nope, I think it's meant to be the name. I'll update 👍

controllers/tlspolicy_certmanager_certificates.go
```diff
@@ -130,12 +131,40 @@ func (r *TLSPolicyReconciler) expectedCertificatesForGateway(ctx context.Context
...
+			secretRef.Namespace = string(*certRef.Namespace)
+		} else {
+			secretRef.Namespace = l.GetNamespace()
+		}
+		// Gateway API hostname explicitly disallows IP addresses, so this
+		// should be OK.
+		hostname := "*"
+		if l.Hostname != nil {
+			hostname = string(*l.Hostname)
+		}
```
guicassolato, line 156: nit: this could be done outside the loop I guess, right?

controllers/tlspolicy_status_updater.go
guicassolato: Here's how I'm solving for RateLimitPolicy the issue of updating the status after having previously validated the policy vs. when the policy validator is skipped but the status updater still has to run (e.g. due to different event matchers).

guicassolato (approved):
I haven't tested it, but the code LGTM.

I've left a comment about a pattern for picking up state from previous tasks vs falling back to the topology state, but I'd recommend considering another PR for it.

PR #887, opened 2024-09-30
Sotw add limitador reconcile

Boomatang:
Introduces the reconcile function for the limitador resource.


Boomatang:
@KevFan Yes, I can do that bit of the rename. The point about the removing the logic from the kuadrant controller being part of this PR is not need. There is no updating of the resources only creating. If both methods are in place, one will get the update first while the other will get a version out of date error. Which they both can handle. The PR for removing the work is already opened, #888. Once this would be merged that can be merged, (would require a rebase to pass test)

controllers/limitador_reconciler.go
```diff
@@ -14,5 +27,85 @@ func NewLimitadorReconciler(client *dynamic.DynamicClient) *LimitadorReconciler
...
+		if item.GroupVersionKind().Kind == v1beta1.LimitadorGroupKind.Kind {
+			return item, true
+		}
+		return nil, false
+	})
+
+	if len(aobjs) > 0 {
+		logger.Info("limitador resource already exists, no need to create", "status", "skipping")
+		return nil
+	}
```
KevFan, line 71: nit: `aobjs` being authorino objects? So this should be `lobjs` instead? 🤔 
```suggestion
	lobjs := lo.FilterMap(topology.Objects().Objects().Items(), func(item machinery.Object, _ int) (machinery.Object, bool) {
		if item.GroupVersionKind().Kind == v1beta1.LimitadorGroupKind.Kind {
			return item, true
		}
		return nil, false
	})

	if len(lobjs) > 0 {
		logger.Info("limitador resource already exists, no need to create", "status", "skipping")
		return nil
	}
```

KevFan (approved):

PR #888, opened 2024-09-30
REMOVE: kuadrant controller management of limitador

Boomatang:
With the move to the state of the world reconcile the kuadrant controller no longer needs to reconcile the creation of the limitador resource.

This change is independent but will introduce a breaking change if merged before https://github.com/Kuadrant/kuadrant-operator/pull/883 and https://github.com/Kuadrant/kuadrant-operator/pull/887

Required PRs:
- [x] #883 
- [x] #887

### DIFF
--- a/controllers/kuadrant_controller.go
+++ b/controllers/kuadrant_controller.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


KevFan (approved):
Looks good to me 👍

Boomatang:
@KevFan So you are aware I need to add back in the watches for the limitador and authorino as these are need to trigger the status update in kuadrant.

PR #894, opened 2024-10-01
ADD: Resources to Topology for External Authorizer (Istio)

Boomatang:
Add the resources used in the setting up of the external authorizer to the topology.


guicassolato (commented):
I don't think we'll need this with auth into wasm, no?

Boomatang:
Not need

PR #895, opened 2024-10-01
ADD: Resources to Topology for External Authorizer (OSSM)

Boomatang:
Add the resources used in the setting up of external authorizer to the topology for using OSSM.


Boomatang:
Not need

PR #903, opened 2024-10-03
update health check propagation

philbrookes:


maleck13 (commented):
changes look fine. Should there also be an update to DNSPolicy reference doc?

PR #909, opened 2024-10-04
refactor: sotw conventions

KevFan:
# Description
* Align with agreed conventions on GroupKind and Resource naming
* Align with agreed conventions on naming of tasks & workflows
* Placeholders for workflows and some tasks


guicassolato (approved):
Gold!

PR #910, opened 2024-10-04
Update: Improve get kuadrant resource

Boomatang:
Updates the getting of the kuadrant CR to use the topology as the input data structure. This adds consistency to how others will get the kuadrant CR.

### DIFF
--- a/controllers/state_of_the_world.go
+++ b/controllers/state_of_the_world.go
@@ -404,29 +404,28 @@ func finalStepsWorkflow(client *dynamic.DynamicClient, isIstioInstalled, isEnvoy
 	return workflow
 }
 
-// GetOldestKuadrant returns the oldest kuadrant resource from a list of kuadrant resources that is not marked for deletion.
-func GetOldestKuadrant(kuadrants []*kuadrantv1beta1.Kuadrant) (*kuadrantv1beta1.Kuadrant, error) {
-	if len(kuadrants) == 1 {
-		return kuadrants[0], nil
-	}
-	if len(kuadrants) == 0 {
-		return nil, fmt.Errorf("empty list passed")
-	}
-	oldest := kuadrants[0]
-	for _, k := range kuadrants[1:] {
-		if k == nil || k.DeletionTimestamp != nil {
-			continue
+var ErrNoKandrantResource = fmt.Errorf("no kuadrant resources in topology")
+
+// GetKuadrant returns the oldest Kuadrant from the root objects in the topology
+func GetKuadrant(topology *machinery.Topology) (*kuadrantv1beta1.Kuadrant, error) {
+	kuadrantList := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (*kuadrantv1beta1.Kuadrant, bool) {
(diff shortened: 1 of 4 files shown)


KevFan (approved):
Looks good to me!

controllers/state_of_the_world.go
```diff
@@ -183,29 +185,28 @@ func finalStepsWorkflow(client *dynamic.DynamicClient, isIstioInstalled, isEnvoy
...
-		if k == nil || k.DeletionTimestamp != nil {
-			continue
-		}
-		if oldest == nil {
-			oldest = k
-			continue
-		}
+	oldest := kuadrantList[0]
+	for _, k := range kuadrantList[1:] {
 		if k.CreationTimestamp.Before(&oldest.CreationTimestamp) {
```
guicassolato, line 207: By bumping Policy Machinery to v0.4.0, we use `controller.ObjectsByCreationTimestamp` to sort the objects. E.g.: https://github.com/Kuadrant/kuadrant-operator/blob/f6d36aad28b7f5b6fd9308698e7899e19fb62963/controllers/state_of_the_world.go#L283

controllers/state_of_the_world_test.go
```diff
@@ -5,14 +5,51 @@ import (
 	"testing"
 	"time"
 
+	"github.com/kuadrant/policy-machinery/machinery"
```
eguzki, line 8: Missing `//go:build unit`

eguzki (changes requested):

eguzki (approved):

controllers/state_of_the_world.go
```diff
@@ -184,29 +187,24 @@ func finalStepsWorkflow(client *dynamic.DynamicClient, isIstioInstalled, isEnvoy
...
+func GetKuadrant(topology *machinery.Topology) (*kuadrantv1beta1.Kuadrant, error) {
+	kuadrantList := lo.FilterMap(topology.Objects().Roots(), func(item machinery.Object, _ int) (controller.Object, bool) {
+		k, ok := item.(controller.Object)
+		if ok && k.GetObjectKind().GroupVersionKind().GroupKind() == kuadrantv1beta1.KuadrantGroupKind && k.GetDeletionTimestamp() == nil {
+			return k, true
 		}
+		return nil, false
+	})
+	if len(kuadrantList) == 1 {
+		return kuadrantList[0].(*kuadrantv1beta1.Kuadrant), nil
```
guicassolato, line 202: This optimisation is probably not needed. If the number of elements in the slice is 1, then sorting is trivial.

Boomatang: The way I see this is we have to do the zero length check, so if the length check has to be done we should short circuit sooner. With that said possible I should have added this to a switch statement and not use the if statements.

guicassolato: The zero length is correct. I'm saying we don't need the check `if len(kuadrantList) == 1`.

Boomatang: I am happy to remove it. To me it seemed more logical to test that the list is how we expect it to be 99% of the time and if not in that state then go do the sorting.

PR #917, opened 2024-10-08
Script

R-Lawton:
The quickstart script was failing as a namespace was missing. The namespace no longer gets installed with the preferred documented method of installing istio via sail.  The script should finish properly now and the observability piece should install without issues

KUADRANT_REF=script ./hack/quickstart-setup.sh

### DIFF
--- a/config/observability/openshift/telemetry.yaml
+++ b/config/observability/openshift/telemetry.yaml
@@ -2,7 +2,7 @@ apiVersion: telemetry.istio.io/v1alpha1
 kind: Telemetry
 metadata:
   name: namespace-metrics
-  namespace: gateway-system
+  namespace: istio-system
 spec:
   metrics:
   - providers:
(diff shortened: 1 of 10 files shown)


R-Lawton:
@maleck13 Looking at the different types of observability pieces in didn't make sense to just overlay the namespace as envoy and istio have different ports and selectors etc.  Split out what was needed by istio into its own separate kustomise that the quickstart deploys and the same for envoy that standalone observability doc installs.  Both also still deploy the obvserve pieces that oringally were there require.

config/observability/openshift/telemetry.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: telemetry.istio.io/v1alpha1
 kind: Telemetry
 metadata:
   name: namespace-metrics
-  namespace: gateway-system
+  namespace: istio-system
```
R-Lawton, line 5: Telemetry resource is a istio thing not a envoy resource. Envoy has there own version but that can be post v1

maleck13 (approved):
worked for me.

PR #920, opened 2024-10-08
topology CM lookup, inject into console-plugin deployment

jasonmadigan:
### What

Load the location of the `topology` ConfigMap dynamically. An update to the kuadrant-operator now sets some env-vars on the ConsolePlugin deployment, which the plugin uses to discover the topology resource. 

Resolves: https://github.com/Kuadrant/kuadrant-console-plugin/issues/48

### Verification

Verified on OpenShift v4.15

* Install Gateway API

```shell
make gateway-api-install
```

* Install Istio

Follow [user guide](https://docs.kuadrant.io/0.11.0/kuadrant-operator/doc/install/install-openshift/#step-4-install-and-configure-istio-with-the-sail-operator)

* Deploy the kuadrant operator to a NS other than `kuadrant-system`

```
kubectl create ns quadrant-system
```

Install the operator from this branch (`dynamic-topology`):

```yaml
kubectl apply -f - <<EOF
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  namespace: quadrant-system
spec:
  sourceType: grpc
  image: quay.io/kuadrant/kuadrant-operator-catalog:dynamic-topology
  publisher: grpc
  updateStrategy:
    registryPoll:
      interval: 45m
---
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: kuadrant-operator
  namespace: quadrant-system
spec:
  channel: preview
  installPlanApproval: Automatic
  name: kuadrant-operator
  source: kuadrant-operator-catalog
  sourceNamespace: quadrant-system
---
kind: OperatorGroup
apiVersion: operators.coreos.com/v1
metadata:
  name: kuadrant
  namespace: quadrant-system
spec:
  upgradeStrategy: Default
EOF
```
Wait for the Kuadrant Operators to be installed as follows:

```
kubectl get installplan -n quadrant-system -o=jsonpath='{.items[0].status.phase}'
```

After some time, this command should return `complete`.

* Enable Kuadrant's dynamic plugin: https://docs.kuadrant.io/dev/kuadrant-operator/doc/install/install-openshift/#step-10-configure-the-kuadrant-console-plugin

* Deploy resources

Apply the Kuadrant custom resource

```yaml
kubectl -n quadrant-system apply -f - <<EOF
---
kind: Kuadrant
metadata:
  name: quadrant-sample
spec: {}
EOF
```

Note: a patched version of the `kuadrant-console-plugin` has already been published as `:latest`, because I like to live dangerously.

Ensure the topology is loaded OK on the Kuadrant > Policy Topology view

Verify the two env vars are set in the kuadrant-console deployment:

```bash
kubectl get deployment kuadrant-console -n quadrant-system -o jsonpath='{.spec.template.spec.containers}' | yq e -P
```
```yaml
- env:
    - name: TOPOLOGY_CONFIGMAP_NAME
      value: topology
    - name: TOPOLOGY_CONFIGMAP_NAMESPACE
      value: quadrant-system
  image: quay.io/kuadrant/console-plugin:latest
  imagePullPolicy: Always
  name: kuadrant-console
  ports:
    - containerPort: 9443
      protocol: TCP
  resources: {}
  terminationMessagePath: /dev/termination-log
  terminationMessagePolicy: File
  volumeMounts:
    - mountPath: /var/serving-cert
      name: plugin-serving-cert
      readOnly: true
    - mountPath: /etc/nginx/nginx.conf
      name: nginx-conf
      readOnly: true
      subPath: nginx.conf
```

### DIFF
--- a/pkg/openshift/consoleplugin/nginx_configmap.go
+++ b/pkg/openshift/consoleplugin/nginx_configmap.go
@@ -34,6 +34,9 @@ http {
 		location / {
 			root                /usr/share/nginx/html;
 		}
+		location /config.js {
+			root /tmp;
+		}
 	}
 }
 `,
--- a/controllers/consoleplugin_reconciler.go
+++ b/controllers/consoleplugin_reconciler.go
@@ -89,7 +89,7 @@ func (r *ConsolePluginReconciler) Run(eventCtx context.Context, _ []controller.R
 	}
 
 	// Deployment
-	deployment := consoleplugin.Deployment(r.namespace, ConsolePluginImageURL)
+	deployment := consoleplugin.Deployment(r.namespace, ConsolePluginImageURL, TopologyConfigMapName)
 	deploymentMutators := make([]reconcilers.DeploymentMutateFn, 0)
 	deploymentMutators = append(deploymentMutators, reconcilers.DeploymentImageMutator)
 	if !topologyExists {
--- a/pkg/openshift/consoleplugin/deployment.go
+++ b/pkg/openshift/consoleplugin/deployment.go
@@ -7,7 +7,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/env"
 	"k8s.io/utils/ptr"
 )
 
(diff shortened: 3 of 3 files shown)


eguzki:
Thanks for the verification steps. Working like a charm!!

pkg/openshift/consoleplugin/nginx_configmap.go
```diff
@@ -34,6 +34,9 @@ http {
 		location / {
 			root                /usr/share/nginx/html;
 		}
+		location /config.js {
```
eguzki, line 37: out of curiosity. What is this doing?

jasonmadigan: Basically:

- there's a new custom entrypoint script in the plugin image
- that script injects the two env vars into config.js
- the topology view in plugin loads this config.js, with the correct pointers to the ns/name of the topology configmap

eguzki: Thanks for the explanation.

PR #924, opened 2024-10-09
Authpolicy/v1beta3

Boomatang:
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/809

This change introduces the AuthPolicy v1beta3.

The RouteSelector has being removed from the API. There is still work required to bring in the Section Name to the API.

In the Docs only the reference to the RouteSelector has being removed from the reference sections. The guides have not being updated to use v1betat3. I expect this to happen with or shortly after the adding of the Section Name.

A number of integration test have been removed which depend on or directly tested the behavior around the RouteSelector.


guicassolato (commented):
Can we also delete `api/v1beta2/route_selectors.go`?

api/v1beta3/authpolicy_types.go
```diff
@@ -1,4 +1,4 @@
-package v1beta2
+package v1beta3
```
guicassolato, line 1: I think we can also remove all `+kubebuilder:validation:MaxProperties=10` markers. They were added to keep the CRD within the complexity budget when we first introduced reoute selectors.

api/v1beta3/authpolicy_types.go
```diff
@@ -293,47 +242,11 @@ func (ap *AuthPolicy) GetWrappedNamespace() gatewayapiv1.Namespace {
...
-	}
-
-	appendRuleHosts(ap.Spec.CommonSpec())
-	appendCommonSpecRuleHosts(ap.Spec.CommonSpec())
-
-	return
+// GetRulesHostnames
+// in v1beta2 this returned the list of route selectors
+// in v1beta3 this should work with section name, once implemented.
+func (ap *AuthPolicy) GetRulesHostnames() []string {
```
guicassolato, line 248: Is there anything left that still depends on this? Maybe it's required to fully implement some interface by the actual method is no longer being used? (I didn't check.)

KevFan (changes requested):
Looks good generally but I think there's some changes missing 🧑‍💻 

Can we also update the api version of the guides where an `AuthPolicy` is used as it's now using the wrong api version? 
- For example: https://github.com/Boomatang/kuadrant-operator/blob/authpolicy/v1beta3/doc/user-guides/auth-for-app-devs-and-platform-engineers.md#-protect-the-toy-store-application-persona-app-developer

Can we rename and update the authpolicy v1beta2 sample to v1beta3 so that the example is still shown in the generated csv?
- https://github.com/Boomatang/kuadrant-operator/blob/authpolicy/v1beta3/config/samples/kuadrant_v1beta2_authpolicy.yaml

Boomatang:
@KevFan The sample authpolicy that I can update no problem. 

I don't feel updating the guides right now is the best thing to do right now. These guides will need to be again updated when the section name is introduced. I had the intentional decision not to update guides till we had the section name in place.

Boomatang:
> Can we also delete `api/v1beta2/route_selectors.go`?

@guicassolato yes, I had it remove and reintroduced it when trying to fix test. I most have forgotten to remove it afterwards. It was never meant to make it into this PR.

api/v1beta3/authpolicy_types.go
Boomatang: Yap can update this. I expect this will should also effect some tests also.

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -49,34 +49,6 @@ metadata:
...
-                }
-              }
-            },
-            "targetRef": {
-              "group": "gateway.networking.k8s.io",
-              "kind": "HTTPRoute",
-              "name": "toystore"
-            }
-          }
-        },
```
KevFan, line 79: Let's rename https://github.com/Kuadrant/kuadrant-operator/blob/1ac1104edae00a64cc3eae224f07fee46aadc190/config/samples/kuadrant_v1beta2_authpolicy.yaml to `kuadrant_v1beta3_authpolicy.yaml ` and add it to the samples path:
https://github.com/Kuadrant/kuadrant-operator/blob/1ac1104edae00a64cc3eae224f07fee46aadc190/config/samples/kustomization.yaml#L4
Since I feel like it's going to be missed if done later 🤔

controllers/orphaned_dnsrecord_reconciler.go
KevFan, line 1: I'm assuming this file is not meant to be in this PR?

KevFan (commented):
Looks good overall 👍  Left 2 comments 🧑‍💻

controllers/orphaned_dnsrecord_reconciler.go
Boomatang: Correct, when you break form your normal workflow things like this happen. I blame `git add . ` for that.

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
Boomatang: Ya sure, it is outside the doc guides and would be missed.

guicassolato:
> The guides have not being updated to use v1betat3. I expect this to happen with or shortly after the adding of the Section Name.

Maybe I'm being naive here, but with this PR (unlike #875), we should be able to "just" `sed s/v1beta2/v1beta3/g` across all md files and docs would be done, no? Personally, I wouldn't wait for `sectionName` thing to do that if it turns out to be indeed that easy. This way, we have functioning docs from day 1, and section name can be only about adding stuff.

Boomatang:
@guicassolato using sed would be the fastest way to do this, but there is reference to the route selectors in the guides. So it is not just a matter of doing a find and replace. I will work on that refactor now. That said I think we can merge the current changes without the docs update.

guicassolato:
> @guicassolato using sed would be the fastest way to do this, but there is reference to the route selectors in the guides. So it is not just a matter of doing a find and replace. I will work on that refactor now. That said I think we can merge the current changes without the docs update.

No problem, @Boomatang. If you prefer this way, I'm OK with it.

Let's just make sure we don't cut a release of the Operator before fixing those references in the docs. Especially if the route selectors are mentioned, then IMO it's imperative that we remove them as soon as possible.

cc @maleck13

guicassolato (approved):

Issue #819, opened 2024-08-16
[state-of-the-world reconciler] Effective policy TLS


guicassolato:
Reconciliation of effective policies for TLS. Creation of TLS certificate requests for each applicable path in the topology graph (TBD: Gateway or Gateway → Gateway Listener).

mikenairn:
Closed by https://github.com/Kuadrant/kuadrant-operator/pull/927

PR #927, opened 2024-10-09
refactor: effective tls policies reconciler

KevFan:
# Description
Part of https://github.com/Kuadrant/kuadrant-operator/issues/819 https://github.com/Kuadrant/kuadrant-operator/issues/824

* Refactor the logic for the management of TLSPolicy Certificates into a workflow task
* Re-add validation for TLS Policy to check for if Issuer kind is correct and if the issuer is present on cluster, with the addition of integrations tests to test for this

# Verification
* Passing integration tests should be enough to test that nothing is broken from this change


controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,289 @@
...
+		listeners := lo.FilterMap(topology.Targetables().Items(), func(t machinery.Targetable, index int) (*machinery.Listener, bool) {
+			l, ok := t.(*machinery.Listener)
+			return l, ok && lo.Contains(l.Gateway.Policies(), p)
+		})
+
+		// Policy is deleted
+		if policy.DeletionTimestamp != nil {
+			logger.V(1).Info("policy is marked for deletion, nothing to do", "name", policy.Name, "namespace", policy.Namespace, "uid", policy.GetUID())
+			continue
+		}
```
KevFan, line 76: No need for finalizer since associated `Certificate`'s will also be deleted by owner ref

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,289 @@
...
+//+kubebuilder:rbac:groups="cert-manager.io",resources=issuers,verbs=get;list;watch;
+//+kubebuilder:rbac:groups="cert-manager.io",resources=clusterissuers,verbs=get;list;watch;
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="cert-manager.io",resources=certificates,verbs=get;list;watch;create;update;patch;delete
+
+func (t *EffectiveTLSPoliciesReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, s *sync.Map) error {
+	logger := controller.LoggerFromContext(ctx).WithName("EffectiveTLSPoliciesReconciler").WithName("Reconcile")
+
+	// Get all TLS Policies
+	policies := lo.Filter(topology.Policies().Items(), func(item machinery.Policy, index int) bool {
```
maleck13, line 57: So are we reconciling every policy on the cluster with every change? Not quite understanding how this is working

KevFan: That's fair. There's is a way to do reconcile only affected tls policies instead by using the events received. I've pushed a commit for this so this should be addressed now https://github.com/Kuadrant/kuadrant-operator/pull/927/commits/cd35907ece1813cb9f7ed7c70acbe092ebcbf52f

controllers/tls_workflow.go
```diff
@@ -133,24 +126,124 @@ func LinkGatewayToClusterIssuerFunc(objs controller.Store) machinery.LinkFunc {
...
 
-			// Can infer linked gateways through the policy
-			linkedGateways := lo.Filter(gateways, func(g *gwapiv1.Gateway, index int) bool {
-				for _, l := range linkedPolicies {
-					if string(l.Spec.TargetRef.Name) == g.GetName() && g.GetNamespace() == l.GetNamespace() {
+			g := machinery.Gateway{Gateway: ob.(*gwapiv1.Gateway)}
+
+			affectedPolicies = append(affectedPolicies, lo.Filter(policies, func(item machinery.Policy, index int) bool {
+				for _, tg := range item.GetTargetRefs() {
+					if g.GetLocator() == tg.GetLocator() {
```
guicassolato, line 189: Have you considered using [`g.Policies()`](https://github.com/Kuadrant/policy-machinery/blob/72b15c675f8d03489ae926288b3fd2d426174fe6/machinery/types.go#L39C2-L39C12)?

controllers/tls_workflow.go
```diff
@@ -133,24 +126,124 @@ func LinkGatewayToClusterIssuerFunc(objs controller.Store) machinery.LinkFunc {
...
-		},
+		if event.Kind == kuadrantv1alpha1.TLSPolicyGroupKind {
+			ob := event.NewObject
+			if ob == nil {
+				ob = event.OldObject
+			}
+
+			affectedPolicies = append(affectedPolicies, lo.Filter(policies, func(item machinery.Policy, index int) bool {
+				return item.GetName() == ob.GetName() && item.GetNamespace() == ob.GetNamespace()
+			})...)
```
guicassolato, line 205: I understand this will filter out all TLSPolicies that have been deleted. Unless that's what indeed you want, could we simply cast `ob` maybe?

```suggestion
			p, _ := ob.(*TLSPolicy)
			affectedPolicies = append(affectedPolicies, p)
```

controllers/tls_workflow.go
```diff
@@ -133,24 +126,124 @@ func LinkGatewayToClusterIssuerFunc(objs controller.Store) machinery.LinkFunc {
...
+		if event.Kind == CertManagerCertificateKind {
+			ob := event.NewObject
+			if ob == nil {
+				ob = event.OldObject
+			}
+
+			affectedPolicies = append(affectedPolicies, lo.Filter(policies, func(item machinery.Policy, index int) bool {
+				p := item.(*kuadrantv1alpha1.TLSPolicy)
+				return utils.IsOwnedBy(ob, p)
+			})...)
```
guicassolato, line 217: I believe you could leverage the topology links here too:

```suggestion
			c, _ := ob.(*controller.RuntimeObject).Object.(*certmanagerv1.Certificate)
			listeners := topology.Targetables().Parents(machinery.LocatorFromObject(c))
			affectedPolicies = append(affectedPolicies, lo.FlatMap(listeners, func(l machinery.Targetable) ([]machinery.Policy, bool) {
				return l.(*machinery.Listener).Gateway.Policies()
			})...)
```

In this particular case, not sure if it's better or worse. I'm just thinking about avoiding filtering the same list of policies so many times.

controllers/tls_workflow.go
```diff
@@ -133,24 +126,124 @@ func LinkGatewayToClusterIssuerFunc(objs controller.Store) machinery.LinkFunc {
...
+
+			affectedPolicies = append(affectedPolicies, lo.Filter(policies, func(item machinery.Policy, index int) bool {
+				p := item.(*kuadrantv1alpha1.TLSPolicy)
+				return ob.GetName() == p.Spec.IssuerRef.Name && p.Spec.IssuerRef.Kind == certmanagerv1.ClusterIssuerKind
+			})...)
+		}
 	}
+
+	// Return only unique policies as there can be duplicates from multiple events
+	return lo.Uniq(affectedPolicies)
```
guicassolato, line 248: Hmm. Maybe specify a uniqueness key and thus avoid possible problems with pointer comparison:

```suggestion
	return lo.UniqBy(affectedPolicies, func(p machinery.Policy) string { return p.GetLocator() } )
```

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,390 @@
...
+	var expectedCerts []*certmanv1.Certificate
+
+	for _, p := range policies {
+		policy := p.(*kuadrantv1alpha1.TLSPolicy)
+
+		// Get all listeners where the gateway contains this policy
+		// TODO: Update when targeting by section name is allowed, the listener will contain the policy rather than the gateway
+		listeners := lo.FilterMap(topology.Targetables().Items(), func(t machinery.Targetable, index int) (*machinery.Listener, bool) {
+			l, ok := t.(*machinery.Listener)
+			return l, ok && lo.Contains(l.Gateway.Policies(), p)
```
guicassolato, line 88: I'm actually surprised equality (`==`) between two policy works.

controllers/tls_workflow.go
KevFan: Hmm, dont think i'm able to since `g` is a new `machinery.Gateway` instance, so the `policies` will be `nil` unless you're saying to get the `machinery.Gateway` instance elsewhere? 🤔

controllers/tls_workflow.go
KevFan: Hmm, I didn't intend to do this, but that is simpler indeed. Why it's working as is due to there's no work to be done on tlspolicies marked for deletion 🤔

controllers/tls_workflow.go
KevFan: Hmm, that's true. If you feel strongly on this I can update, but using the owner refernence looks a lot more readable unless filtering the list of policies should be avoided in general? 🤔

controllers/effective_tls_policies_reconciler.go
KevFan: Yeah, I think it compares the pointers between the two policies

controllers/tls_workflow.go
guicassolato: True. I was thinking on getting the gateway from the topology. Indeed `g := machinery.Gateway{Gateway: ob.(*gwapiv1.Gateway)}` is no good.

controllers/tls_workflow.go
guicassolato: I would not say we should avoid filtering policies. Rather, I would say we should avoid writing new logic for things otherwise represented in the topology.

I believe what we want here is get to the list of policies. There must always be a path in the topology between every object directly or indirectly related to a policy and the policy object itself. Getting to the policies some other way suggests to me that we may have failed to represent those links between objects in the topology, that we are underusing the links (and therefore the helper functions such as `Parents` and `Children`, thought to impose a given structure to the topology), or yet that we're using the wrong relationship between objects as topological links (e.g. the object is linked to a resource _and_ it's owned by the resource).

If ownership in this case defines the link between objects, then we should use the link. (Using `IsOwnedBy` would be repeating ourselves.) If the ownership is not the basis for the link, but added only as a mean to facilitate cleaning up resources, then let's not use it as the strong link tying the objects together.

controllers/tls_workflow.go
```diff
@@ -133,24 +126,126 @@ func LinkGatewayToClusterIssuerFunc(objs controller.Store) machinery.LinkFunc {
...
+
+func GetTLSPoliciesByEvents(topology *machinery.Topology, events []controller.ResourceEvent) []machinery.Policy {
+	policies := lo.Filter(topology.Policies().Items(), func(item machinery.Policy, index int) bool {
+		_, ok := item.(*kuadrantv1alpha1.TLSPolicy)
+		return ok
+	})
+
+	var affectedPolicies []machinery.Policy
+	for _, event := range events {
+		if event.Kind == machinery.GatewayGroupKind {
```
Boomatang, line 179: I am not sure how I feel about looping over the events like this. However, would it not make more sense to use a  switch statement over this individual if blocks,

controllers/effective_tls_policies_reconciler.go
KevFan: I've reverted it since it looks to give more concerns that needed. Let's have a look can this be optimized in another PR or later point

controllers/tls_workflow.go
KevFan: Reverted this filtering

controllers/tls_workflow.go
KevFan: Yeah, that's fair, I've reverted this part where this is done, but I'll have a look in the other code changes to see can I better leverage the links 👍

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,394 @@
...
+				resource := t.client.Resource(CertManagerCertificatesResource).Namespace(cert.GetNamespace())
+
+				// Check is cert already in topology
+				objs := topology.Objects().Children(l)
+				obj, ok := lo.Find(objs, func(o machinery.Object) bool {
+					return o.GroupVersionKind().GroupKind() == CertManagerCertificateKind && o.GetNamespace() == cert.GetNamespace() && o.GetName() == cert.GetName()
+				})
+
+				// Create
+				if !ok {
```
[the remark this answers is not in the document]
KevFan: Sure but whether to do `!ok` or `ok` first depends also on whether we `create` or `update` first, so if we want to be more positive in our check here `if ok`, this would mean `update` is first 🤔 

Not sure which action do we expect more of. Maybe @mikenairn can answer on this

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,394 @@
...
+
+		// Only want certs owned by this policy
+		return c, utils.IsOwnedBy(c, p)
+	})
+
+	for _, cert := range certs {
+		resource := t.client.Resource(CertManagerCertificatesResource).Namespace(cert.GetNamespace())
+
+		if err := resource.Delete(ctx, cert.Name, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+			return err
```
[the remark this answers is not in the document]
KevFan: True, I'll log the error here instead 👍

controllers/tls_workflow.go
KevFan: I've reverted this logic so maybe we can revisit this again at a later date if we want to re-introduce it

controllers/tlspolicy_status_updater.go
```diff
@@ -46,53 +47,43 @@ func (t *TLSPolicyStatusUpdaterReconciler) Subscription() *controller.Subscripti
...
-	}
-
-	isPolicyValidErrorMap := store.(map[string]error)
+	policies := GetTLSPoliciesByEvents(topology, events)
 
 	for _, policy := range policies {
-		if policy.DeletionTimestamp != nil {
-			logger.V(1).Info("tls policy is marked for deletion, skipping", "name", policy.GetName(), "namespace", policy.GetNamespace(), "uid", policy.GetUID())
+		p := policy.(*kuadrantv1alpha1.TLSPolicy)
+		if p.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
KevFan: With the revert, this should be a valid check again

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+				_, err = resource.Update(ctx, un, metav1.UpdateOptions{})
+				if err != nil {
+					logger.Error(err, "unable to update certificate", "policy", policy.Name)
+				}
+			}
+		}
+	}
+
+	// Clean up orphaned certs
```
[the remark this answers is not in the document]
KevFan: I believe this is because there is a indirect link between policies and the subresouce (certificates in this case) and so gives the expected state of the cluster. Though true, this probably can be done as a separate task, that I can look into if you feel strongly about this. 

A cert can be orphaned when a gateway listener is removed, or is the target ref of a tls policy is changed to another gateway. In both of these cases the cert(s) is created and will be orphaned since the policy still exists and still valid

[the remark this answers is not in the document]
guicassolato: Indeed we need to cleanup certs that remained orphan due to, e.g., a listener being removed from a gateway _ceteris paribus_.

Let's keep this as-is now since it does the job. Maybe in the future we can look for this pattern across the reconcilers (not only this one) and try to improve them all by having "janitors" as separate tasks that can run in parallel to other tasks.

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+
+	var expectedCerts []*certmanv1.Certificate
+
+	for _, listener := range listeners {
+		l := listener.(*machinery.Listener)
+
+		policies := l.Policies()
+		if len(policies) == 0 {
+			policies = l.Gateway.Policies()
+		}
```
guicassolato, line 83: Nice work, @KevFan. This will easily adapt to when we decide to support `sectionName` and effectively inheritance on TLSPolicies.

guicassolato (approved):

controllers/effective_tls_policies_reconciler.go
mikenairn: Not sure this will work if you have more than one different policy on the listener (DNSPolicy and TLSPolicy)

controllers/effective_tls_policies_reconciler.go
```diff
@@ -0,0 +1,388 @@
...
+
+		for _, p := range policies {
+			policy := p.(*kuadrantv1alpha1.TLSPolicy)
```
mikenairn, line 86: Could policy here be a different type? Is the policies list filtered by type anywhere?

guicassolato: @mikenairn is right. `.Policies()` will return policies of all kinds.

controllers/effective_tls_policies_reconciler.go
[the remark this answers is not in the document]
mikenairn: I wouldn't expect too many updates of a Certificate resource after creation. I don't have strong opinions on this, but for what its worth i did `if ok` in the dnspolicy equivalent task, and deal with updates first.

controllers/effective_tls_policies_reconciler.go
mikenairn: Is it good enough to just assume if you no policies use the gateway ones? Should it not check for a policy of the expected type being attached to the listener and if none fallback? Will this not also reconcile a gateway policy multiple times, once per listener?

controllers/effective_tls_policies_reconciler.go
KevFan: Oh yes, good catch 👍  This might be why integrations tests are failing now 😅

controllers/effective_tls_policies_reconciler.go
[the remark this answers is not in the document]
guicassolato: `if ok { update }` / `if !ok { create }`. Does it matter? The number of jump instructions in the final compiled code is literally the same I think.

If `!ok` feels like being negative (it doesn't to me BTW), then perhaps we could rename the variable to `missing`. Is it better? I.e. `if missing { create }`.

IMO, if it doesn't have any obvious downside in performance, the important thing is that the code is readable. In my head (and it could be only me), I read this in terms of chronology and progressive use cases.

The first thing that happens is having nothing, then:
1. user creates the network resources
2. user creates the policies
3. the operator creates the internal resources (for the first time after seeing the policies)
4. user adjusts the policies and/or network resources
5. operator updates the internal resources accordingly.

controllers/effective_tls_policies_reconciler.go
KevFan: Yes, that's true, this should check for policies of specific type before failing back also 👍 
Yes, unfortunately, this would reconcile a gateway policy multiple times. Not quite sure how to approach this one 🤔

controllers/effective_tls_policies_reconciler.go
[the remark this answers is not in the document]
KevFan: Thanks guys 👍  I don't think it matters in fairness unless people really feel it hinders readability. My preference is to keep it as is tbh as I also prefer dealing with the `create` first followed by `update` action. If we feel strongly about this, maybe this can be done in a separate where this can be dealt with in the entire codebase

controllers/effective_tls_policies_reconciler.go
KevFan: Updated to filter for tls policies. I'm aware of the multiple reconciles for gateway policies but I think I'll look to address this as I look into the reconcile for section name support in another PR.

guicassolato (approved):

controllers/effective_tls_policies_reconciler.go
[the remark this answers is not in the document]
Boomatang: I am happy with leaving this for now. It is some thing that can be done in the future.

Boomatang (approved):
I am happy to approve this. Any comment I had was based around styling which can be done later.

PR #931, opened 2024-10-10
feat: introduce tls policy v1alpha2

KevFan:
# Description
Part of https://github.com/Kuadrant/kuadrant-operator/issues/808

Currently just a straight copy of v1alpha1 in preparation for more api changes in the future


KevFan:
Might not be needed since v1apha1 probably wont have any major api changes, so if we decide to add section name support, this can be done to v1alpha1 instead since there's not too much value in adding v1alpha2 if they are broadly exactly the same

KevFan:
Decided that there's no need for v1apha2 for TLSPolicy. Closing

PR #935, opened 2024-10-11
Consoleplugin mutator

eguzki:
### What

Reconcile ConsolePugin configuration when the operator changes namespace (for whatever reason).


KevFan (approved):
Looks good to me 💯

PR #936, opened 2024-10-11
extend health check spec

maksymvavilov:
Make sure to transfer healthcheck spec from policy to the record properly


controllers/dnspolicy_dnsrecords.go
```diff
@@ -137,11 +137,13 @@ func (r *DNSPolicyReconciler) desiredDNSRecord(gateway *gatewayapiv1.Gateway, cl
...
-			Protocol:         dnsPolicy.Spec.HealthCheck.Protocol,
-			FailureThreshold: dnsPolicy.Spec.HealthCheck.FailureThreshold,
-			Interval:         dnsPolicy.Spec.HealthCheck.Interval,
+			Path:                     dnsPolicy.Spec.HealthCheck.Path,
+			Port:                     dnsPolicy.Spec.HealthCheck.Port,
+			Protocol:                 dnsPolicy.Spec.HealthCheck.Protocol,
+			FailureThreshold:         dnsPolicy.Spec.HealthCheck.FailureThreshold,
+			Interval:                 dnsPolicy.Spec.HealthCheck.Interval,
+			AdditionalHeadersRef:     dnsPolicy.Spec.HealthCheck.AdditionalHeadersRef,
+			AllowInsecureCertificate: dnsPolicy.Spec.HealthCheck.AllowInsecureCertificate,
```
maleck13, line 146: I am kinda wondering does this make sense in the API. Or would it be better to make it an explicit option when starting the DNS operator? `--insecure-health-checks` that just always set that flag for health checks. It being in the API feels like it could accidently get set for one policy and not noticed?

maksymvavilov: Removed insecure checks from code. On the dns-operator side will also remove it from API (the probe-worker PR) We could wait with this PR and I will use it to update dnspolicy API and introduce flags for local dev to allow probes and insecure certs

maleck13 (approved):

Issue #823, opened 2024-08-16
[state-of-the-world reconciler] DNS status


guicassolato:
Move the business logic of DNSPolicy status reconciliation to a workflow state-of-the-world reconcile func.

mikenairn:
Closed by https://github.com/Kuadrant/kuadrant-operator/pull/937

PR #937, opened 2024-10-14
sotw: dnspolicy

### LINKED ISSUE #818: [state-of-the-world reconciler] Effective policy DNS
Reconciliation of effective policies for DNS. Creation of DNSRecords for each applicable path in the topology graph (TBD: Gateway or Gateway → Gateway Listener).

mikenairn:
Adds state of the world tasks for DNSPolicy.

- [x] Init
- [x] DNSPolicy Validation (DNSPoliciesValidator)
- [x] Policy Reconciliation (EffectiveDNSPoliciesReconciler)
  - [x] Reconcile resources
  - [x] Orphan record deletion
- [x] Status Updater (DNSPolicyStatusUpdater)
- [x] ~~Cleanup old controller files~~  (Left as is for easier reviews, will do this in a separate PR after this one with no logic changes)

closes #818 #823


controllers/dns_workflow.go
```diff
@@ -1,7 +1,60 @@
...
+	DNSRecordGroupKind = schema.GroupKind{Group: kuadrantdnsv1alpha1.GroupVersion.Group, Kind: "DNSRecord"}
+)
+
+//+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies,verbs=get;list;watch;update;patch;delete
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnspolicies/finalizers,verbs=update
+
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnsrecords,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnsrecords/status,verbs=get
```
mikenairn, line 28: Moved all these to here for now, but will need to review these permissions again. I'm not sure why we need the namespace permissions for instance.

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+				GVK: schema.GroupVersionKind{
+					Group: gk.Group,
+					Kind:  gk.Kind,
+				},
+			}
+			break
+		}
+	}
+	return ownerRef
+}
```
mikenairn, line 133: This should all go somewhere else, state_of_the_world.go maybe? This also, if we agree it seems like a reasonable way to link an OwnerReference to a Policy could be reused by at least the TLSPolicy.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/kuadrant/authorino-operator v0.11.1
 	github.com/kuadrant/dns-operator v0.0.0-20241002074817-d0cab9eecbdb
 	github.com/kuadrant/limitador-operator v0.9.0
-	github.com/kuadrant/policy-machinery v0.5.0
+	github.com/kuadrant/policy-machinery v0.5.1-0.20241014081934-0c6af68f7cfe
```
mikenairn, line 17: Requires https://github.com/Kuadrant/policy-machinery/pull/42

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, _ *sync.Map) error {
+	//ToDo Implement DNSRecord reconcile
+	return r.deleteOrphanDNSPolicyRecords(ctx, topology)
+}
+
+// deleteOrphanDNSPolicyRecords deletes any DNSRecord resources created by a DNSPolicy but no longer have a valid path in the topology to that policy.
+func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSPolicyRecords(ctx context.Context, topology *machinery.Topology) error {
+	logger := controller.LoggerFromContext(ctx).WithName("effectiveDNSPoliciesReconciler")
```
mikenairn, line 48: Should probably agree on how we want logging to work in all these, do we want a named logger per task?

controllers/dns_workflow.go
```diff
@@ -1,7 +1,60 @@
...
+
+func NewDNSWorkflow(client *dynamic.DynamicClient) *controller.Workflow {
+	return &controller.Workflow{
+		Precondition: NewDNSPoliciesValidator().Subscription().Reconcile,
```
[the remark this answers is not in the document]
guicassolato: `.Validate` would read better indeed.

I'm afraid this is a conflict in the conventions we agreed though:
1. Naming tasks after what they do.
2. Abstract the task event subscription within a `Subscription()` method of the task.
3. Name the reconcile func of the task in a meaningful way.

I guess we cannot really keep 2 and 3, unless the name chosen at 1 is `<Something>Reconciler`.

controllers/dns_workflow.go
```diff
@@ -1,7 +1,60 @@
...
+	listeners := lo.FlatMap(lo.Map(gateways, func(g *gwapiv1.Gateway, _ int) *machinery.Gateway {
+		return &machinery.Gateway{Gateway: g}
+	}), machinery.ListenersFromGatewayFunc)
+
+	return machinery.LinkFunc{
+		From: machinery.ListenerGroupKind,
+		To:   DNSRecordGroupKind,
+		Func: func(child machinery.Object) []machinery.Object {
+			return lo.FilterMap(listeners, func(l *machinery.Listener, _ int) (machinery.Object, bool) {
+				o := child.(*controller.RuntimeObject)
```
maleck13, line 53: suggestion: any reason not to just cast this to `DNSRecord` ? might also make this more readable? and name the var `dnsRecord` took me a while reading this to figure out which objects were which naming and using `dnsRecord` might help others

controllers/dnspolicies_validator.go
```diff
@@ -0,0 +1,44 @@
...
+		ReconcileFunc: r.validate,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: &machinery.GatewayGroupKind},
+			{Kind: &kuadrantv1alpha1.DNSPolicyGroupKind},
+		},
+	}
+}
+
+func (r *DNSPoliciesValidator) validate(_ context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
+	policies := topology.Policies().Items(func(o machinery.Object) bool {
```
maleck13, line 35: night be a nice option in the future to have a `topology.Policies().ForGVK(...)` ?

controllers/dns_workflow.go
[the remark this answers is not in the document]
maleck13: sure its not a big deal really just an observation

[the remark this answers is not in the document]
mikenairn: Done this way for consistency with other workflows [see](https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/tls_workflow.go#L32), which was discussed [here](https://github.com/Kuadrant/kuadrant-operator/pull/885#discussion_r1789871802)

I think `.Validate` would read better also, but I'd prefer keeping everything consistent more so.

controllers/dnspolicies_validator.go
```diff
@@ -0,0 +1,44 @@
...
+		return o.GroupVersionKind().GroupKind() == kuadrantv1alpha1.DNSPolicyGroupKind
+	})
+
+	state.Store(StateDNSPolicyValid, lo.SliceToMap(policies, func(policy machinery.Policy) (string, bool) {
+		return policy.GetLocator(), len(policy.GetTargetRefs()) == 0 || len(topology.Targetables().Parents(policy)) > 0
+	}))
```
mikenairn, line 41: Update this to work in the same way as TLS `state.Store(DNSPolicyAcceptedKey, s.Store(TLSPolicyAcceptedKey, isPolicyValidErrorMap))`

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+			// Any DNSRecord that does not have a link in the topology back to its owner DNSPolicy should be removed
+			if len(topology.All().Paths(policyOwnerRef, item)) == 0 {
+				logger.Info(fmt.Sprintf("dnsrecord object is no longer linked to it's policy owner, dnsrecord: %v, policy: %v", item.GetLocator(), policyOwnerRef.GetLocator()))
+				return item, true
+			}
+		}
+		return nil, false
+	})
+
+	for i := range orphanRecords {
```
maleck13, line 69: open question. If we are using `lo` should we not end up using it for slices also? ```https://github.com/samber/lo?tab=readme-ov-file#foreach``` @guicassolato ?

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	return machinery.LocatorFromObject(o)
+}
+
+var _ machinery.PolicyTargetReference = &PolicyOwnerReference{}
+
+func getObjectPolicyOwnerReference(item machinery.Object, gk schema.GroupKind) *PolicyOwnerReference {
+	var ownerRef *PolicyOwnerReference
+	for _, o := range item.(*controller.RuntimeObject).GetOwnerReferences() {
+		oGV, err := schema.ParseGroupVersion(o.APIVersion)
+		if err != nil {
```
maleck13, line 116: guessing we skip the error here as it would be none kuadrant owner ref that would cause this error?

controllers/state_of_the_world.go
```diff
@@ -291,13 +293,28 @@ func (b *BootOptionsBuilder) getConsolePluginOptions() []controller.ControllerOp
...
+		),
+	)
+
+	return opts
+}
+
 func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
 	mainWorkflow := &controller.Workflow{
 		Precondition: initWorkflow(b.client).Run,
 		Tasks: []controller.ReconcileFunc{
```
maleck13, line 314: are each of these dispatched in their own routine?

maleck13 (commented):
These changes all look pretty good to me. Couple of minor comments but  nothing major.

controllers/dns_workflow.go
```diff
@@ -1,7 +1,59 @@
...
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnsrecords,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=kuadrant.io,resources=dnsrecords/status,verbs=get
+
+func NewDNSWorkflow(client *dynamic.DynamicClient) *controller.Workflow {
+	return &controller.Workflow{
+		Precondition: NewDNSPoliciesValidator().Subscription().Reconcile,
+		Tasks: []controller.ReconcileFunc{(&controller.Workflow{
+			Tasks: []controller.ReconcileFunc{
+				NewEffectiveDNSPoliciesReconciler(client).Subscription().Reconcile,
+			},
```
Boomatang, line 34: Is there plans to put a precondition or post-condition in the Workflow defined here. If there is no plan does it not make more sense to add the reconcile directly to the outer tasks list.

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+// deleteOrphanDNSPolicyRecords deletes any DNSRecord resources created by a DNSPolicy but no longer have a valid path in the topology to that policy.
+func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSPolicyRecords(ctx context.Context, topology *machinery.Topology) error {
+	logger := controller.LoggerFromContext(ctx).WithName("effectiveDNSPoliciesReconciler")
+	logger.Info("deleting orphan policy DNS records")
+
+	orphanRecords := lo.FilterMap(topology.Objects().Items(), func(item machinery.Object, _ int) (machinery.Object, bool) {
+		if item.GroupVersionKind().GroupKind() == DNSRecordGroupKind {
+			policyOwnerRef := getObjectPolicyOwnerReference(item, kuadrantv1alpha1.DNSPolicyGroupKind)
+
+			// Ignore all DNSRecords that weren't created by a DNSPolicy
```
Boomatang, line 55: Why is there DNSRecords in the topology that were not created by a DNSPolicy. There many need to be improvements done to the DNSRecords watchable filters.

Boomatang, line 55: Why would there be DNSRecords in the topology that are not linked to a DNSPoicy if they are just going to be ignored. There may need to be improvements done to the DNSRecords watchable filters.

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// deleteOrphanDNSPolicyRecords deletes any DNSRecord resources created by a DNSPolicy but no longer have a valid path in the topology to that policy.
+func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSPolicyRecords(ctx context.Context, topology *machinery.Topology) error {
+	logger := controller.LoggerFromContext(ctx).WithName("effectiveDNSPoliciesReconciler")
+	logger.Info("deleting orphan policy DNS records")
+
+	orphanRecords := lo.FilterMap(topology.Objects().Items(), func(item machinery.Object, _ int) (machinery.Object, bool) {
+		if item.GroupVersionKind().GroupKind() == DNSRecordGroupKind {
+			policyOwnerRef := getObjectPolicyOwnerReference(item, kuadrantv1alpha1.DNSPolicyGroupKind)
```
Boomatang, line 53: After looking at the implementation for `getObjectPolicyOwnerReference`, I am wondering if the function is need at all.

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	//ToDo Implement DNSRecord reconcile
+	return r.deleteOrphanDNSPolicyRecords(ctx, topology)
+}
+
+// deleteOrphanDNSPolicyRecords deletes any DNSRecord resources created by a DNSPolicy but no longer have a valid path in the topology to that policy.
+func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSPolicyRecords(ctx context.Context, topology *machinery.Topology) error {
+	logger := controller.LoggerFromContext(ctx).WithName("effectiveDNSPoliciesReconciler")
+	logger.Info("deleting orphan policy DNS records")
+
+	orphanRecords := lo.FilterMap(topology.Objects().Items(), func(item machinery.Object, _ int) (machinery.Object, bool) {
```
Boomatang, line 51: Is this filtering being done wrongly, only records that are created by a DNSPolicy should ever end up in the topology. With how the linking function is set up an orphaned record would become a root node in the topology. So then the filter by filter the root objects for any type of DNSRecord would return all the orphaned records. This relates back to my earlier comment that possible the DNSRecord watchable filters need to be updated as this method would not work if there are DNSRecords that were not created by kuadrant in the topology.

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,133 @@
...
+	return o.OwnerReference.Name
+}
+
+func (o *PolicyOwnerReference) GetLocator() string {
+	return machinery.LocatorFromObject(o)
+}
+
+var _ machinery.PolicyTargetReference = &PolicyOwnerReference{}
+
+func getObjectPolicyOwnerReference(item machinery.Object, gk schema.GroupKind) *PolicyOwnerReference {
```
Boomatang, line 112: I question if this function is needed

controllers/state_of_the_world.go
```diff
@@ -294,13 +296,28 @@ func (b *BootOptionsBuilder) getConsolePluginOptions() []controller.ControllerOp
 	return opts
 }
 
+func (b *BootOptionsBuilder) getDNSOperatorOptions() []controller.ControllerOption {
+	var opts []controller.ControllerOption
+	opts = append(opts,
+		controller.WithRunnable("dnsrecord watcher", controller.Watch(&kuadrantdnsv1alpha1.DNSRecord{}, DNSRecordResource, metav1.NamespaceAll)),
```
Boomatang, line 302: I think this needs address as it brings in results that are not created by kuadrant and kuadrant has no right to know about.

Boomatang (commented):
Sorry for not posting all comments as one review. I pressed the wrong button to many times.

I have a concern that there are DNSRecords being added to the topology that should not be there. This is then affecting the logic in the reconciles as it needs to account for these resources. If the filtering is done at the api server level, our logic becomes easier and safer.

controllers/common.go
```diff
@@ -0,0 +1,21 @@
...
+func CommonLabels() map[string]string {
+	return map[string]string{
+		AppLabelKey:                    AppLabelValue,
+		"app.kubernetes.io/component":  KuadrantAppName,
+		"app.kubernetes.io/managed-by": "kuadrant-operator",
+		"app.kubernetes.io/instance":   KuadrantAppName,
+		"app.kubernetes.io/name":       KuadrantAppName,
+		"app.kubernetes.io/part-of":    KuadrantAppName,
+	}
+}
```
mikenairn, line 21: Copied the layout and labels from https://github.com/Kuadrant/kuadrant-operator/blob/main/pkg/openshift/consoleplugin/common.go

controllers/state_of_the_world.go
```diff
@@ -294,13 +296,35 @@ func (b *BootOptionsBuilder) getConsolePluginOptions() []controller.ControllerOp
...
+	isDNSRecordOwnedByDNSPolicy := func(c *kuadrantdnsv1alpha1.DNSRecord) bool {
+		return true
+	}
+
+	var opts []controller.ControllerOption
+	opts = append(opts,
+		controller.WithRunnable("dnsrecord watcher", controller.Watch(
+			&kuadrantdnsv1alpha1.DNSRecord{}, DNSRecordResource, metav1.NamespaceAll,
+			controller.FilterResourcesByLabel[*kuadrantdnsv1alpha1.DNSRecord](fmt.Sprintf("%s=%s", AppLabelKey, AppLabelValue)),
+			controller.WithPredicates(ctrlruntimepredicate.NewTypedPredicateFuncs(isDNSRecordOwnedByDNSPolicy)))),
```
mikenairn, line 309: Added this predicate based on what had been added for TLS but I'm unsure of its usefulness when we are using a label filter. The label filter gets added as a predicate internally anyway. I'm thinking we can probably just drop it entirely and rely solely on the label filter.

controllers/state_of_the_world.go
mikenairn: Yes, so i actually wasn't aware until you commented and i looked more closely at the controller code that there was a way to filter out the objects that get added to the topology. This is definitely something i think we should do as we don't want to interfere with resources that weren't created by a policy as you correctly pointed out.

The only way i can see we can really do that with the current implementation of the controller is by using a label filter, and ensuring that all resources we create on behalf of the kuadrant operator are consistently labelled. I had a go at that here https://github.com/Kuadrant/kuadrant-operator/pull/937/commits/8e1afd24629047c0e2c11c6767b6eff3dbbb9d8b.

If this is the approach we want to take, i think we should do it for all resources that we create on behalf of any policy. I noticed for instance TLS Certificates and Issuers that are not created by a TLSPolicy are also added to the topoloy.

@guicassolato @KevFan @Boomatang wdyt about common labelling for all created resources?

controllers/effective_dnspolicies_reconciler.go
mikenairn: Related to https://github.com/Kuadrant/kuadrant-operator/pull/937#discussion_r1805111455, can keep the conversation about this there.

controllers/effective_dnspolicies_reconciler.go
mikenairn: removed

api/v1alpha1/dnspolicy_types.go
```diff
@@ -70,8 +73,23 @@ type DNSPolicySpec struct {
...
+		//Only a CIDR will have  / in the address so attempt to parse fail if not valid
+		if strings.Contains(exclude, "/") {
+			_, _, err := net.ParseCIDR(exclude)
+			if err != nil {
+				return fmt.Errorf("could not parse the CIDR from the excludeAddresses field %w", err)
+			}
+		}
+	}
+	return nil
 }
```
mikenairn, line 93: @maleck13 fyi, these are the changes to the v1alpha api file i mentioned this morning. Since this is really a validation issue, adding the validation directly to resource seemed to make sense rather than relying on it failing during reconciliation. Ideally we would be able to do this with CEL validations and prevent the policy ever getting created/updated with invalid values, but now at least we can call this in the validation task.

controllers/dnspolicy_status_updater.go
```diff
@@ -0,0 +1,124 @@
...
+			meta.SetStatusCondition(&newStatus.Conditions, *enforcedCond)
+
+			//ToDo: Deal with messages, these should probably be retrieved from state after the reconciliation task
+			// add some additional user friendly context
+			//if errors.Is(specErr, ErrNoAddresses) && !strings.Contains(eCond.Message, ErrNoAddresses.Error()) {
+			//	eCond.Message = fmt.Sprintf("%s : %s", eCond.Message, ErrNoAddresses.Error())
+			//}
+			//if errors.Is(specErr, ErrNoRoutes) && !strings.Contains(eCond.Message, ErrNoRoutes.Error()) {
+			//	eCond.Message = fmt.Sprintf("%s : %s", eCond.Message, ErrNoRoutes)
+			//}
```
mikenairn, line 94: @maleck13 These more specific error messages are not currently being put into the status of the policy. I have an idea of how to do it, but i don't think this needs to block this PR, i can create a follow on task to fix this after merge.

go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/kuadrant/authorino-operator v0.11.1
 	github.com/kuadrant/dns-operator v0.0.0-20241018131559-f2ce8b6aaaef
 	github.com/kuadrant/limitador-operator v0.9.0
-	github.com/kuadrant/policy-machinery v0.5.0
+	github.com/kuadrant/policy-machinery v0.5.1-0.20241020183426-2dc665c31409
```
mikenairn, line 17: Merge this PR only after main of policy-machinery is tagged and kuadrant operator is updated in a separate PR that @KevFan is creating.

mikenairn: https://github.com/Kuadrant/kuadrant-operator/pull/950

controllers/dns_workflow.go
mikenairn: The namespace permission is needed when we get the cluster id since we base it on the `kube-system` ns uid.

Boomatang (commented):
I have looked at the changes but I do not have access to Route53 so I didn't test to ensure it works correctly. I have looked it at looking and the linking functions seems to be working correctly. Error messages are getting propagated to the policies in a clear manner. 

There is a few comments, nothing big and no reason to block this work.

controllers/state_of_the_world.go
KevFan: Yeah, I'd say drop it since it's currentlt always returning `true` anyway 👍

controllers/dnspolicies_validator.go
```diff
@@ -0,0 +1,58 @@
...
+
+	logger.V(1).Info("finished validating dns policies")
+
+	return nil
+}
+
+func (r *DNSPoliciesValidator) policyValid(p *kuadrantv1alpha1.DNSPolicy) error {
+	return p.Spec.ExcludeAddresses.Validate()
+}
```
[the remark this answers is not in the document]
mikenairn: Changed this by adding a Validate method to the policy itself.

pkg/library/utils/k8s_utils.go
```diff
@@ -154,17 +159,16 @@ func GetLabel(obj metav1.Object, key string) string {
...
-func GetClusterUID(ctx context.Context, c client.Client) (string, error) {
+func GetClusterUID(ctx context.Context, c dynamic.Interface) (string, error) {
 	//Already calculated? return it
 	if clusterUID != "" {
 		return clusterUID, nil
 	}
 
-	ns := &corev1.Namespace{}
-	err := c.Get(ctx, client.ObjectKey{Name: clusterIDNamespace}, ns)
+	un, err := c.Resource(corev1.SchemeGroupVersion.WithResource("namespaces")).Get(ctx, clusterIDNamespace, metav1.GetOptions{})
```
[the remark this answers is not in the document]
mikenairn: Fair point, but i don't really think we want to add this to the topology. 

I was actually wondering when i was having to update this, if we even need it any more. It was required when we supported ACM that we had a consistent unique id per cluster and using this namespace id was determined to be the best way to do that. I'm not really sure if this is really required any longer though since ownership of records(in the provider) now comes down to the gateway/listener/dnsrecord combination and really is nothing to do with clusters (You could have many on the same cluster contributing to the same record set).

I don't want to change any of this as part of this PR, this is how it works currently, but a follow on task to look into the need for this any longer might be worthwhile.

@maleck13 fyi

KevFan (commented):
First look through. Looks good overall 👍 Left some comments / minor nits (that do not have to be addressed in this PR unless you deem it necessary) 🧑‍💻

controllers/dnspolicy_status_updater.go
mikenairn: This is updated now

controllers/state_of_the_world.go
mikenairn: Removed the predicates

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,291 @@
...
+			if len(gateway.Status.Addresses) > 0 {
+				gatewayHasAddresses = true
+			}
+
+			hasAttachedRoute := false
+			for _, statusListener := range gateway.Status.Listeners {
+				if string(listener.Name) == string(statusListener.Name) {
+					hasAttachedRoute = statusListener.AttachedRoutes > 0
+				}
+			}
```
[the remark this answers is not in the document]
mikenairn: This is existing code that has been around for a while https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/dnspolicy_dnsrecords.go#L83-L87. I would be surprised if you can have more than one status with the same section name, would kind of make it hard to know which one is related to what. It should possibly break when it finds it (assumes only one), or i guess do what you suggested and take the first positive one if there happened to be more.

pkg/library/utils/k8s_utils.go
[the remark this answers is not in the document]
Boomatang: Ya that is fair. I think the less we change how it previously worked the better. Should there be a ticket created to address this and possible other things that were required because of the ACM support?

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,290 @@
...
+					l, lok := t.(*machinery.Listener)
+					lPolicies := lo.FilterMap(l.Policies(), policyTypeFilterFunc)
+					return l, lok && len(lPolicies) == 0
+				})
+				return listeners, true
+			}
+		}
+		return nil, false
+	}))
+}
```
KevFan, line 231: Think we can use `topology.Targetables().Children(policy)` here to get the `Targetables` of a `Policy` which would make this a bit simpler 🤔 
```suggestion
func (r *EffectiveDNSPoliciesReconciler) listenersForPolicy(_ context.Context, topology *machinery.Topology, policy machinery.Policy, policyTypeFilterFunc dnsPolicyTypeFilter) []*machinery.Listener {
	return lo.Flatten(lo.FilterMap(topology.Targetables().Children(policy), func(t machinery.Targetable, _ int) ([]*machinery.Listener, bool) {
		if l, ok := t.(*machinery.Listener); ok {
			return []*machinery.Listener{l}, true
		}
		if g, ok := t.(*machinery.Gateway); ok {
			listeners := lo.FilterMap(topology.Targetables().Children(g), func(t machinery.Targetable, _ int) (*machinery.Listener, bool) {
				l, lok := t.(*machinery.Listener)
				lPolicies := lo.FilterMap(l.Policies(), policyTypeFilterFunc)
				return l, lok && len(lPolicies) == 0
			})
			return listeners, true
		}
		
		return nil, false
	}))
```

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,290 @@
...
+			pLogger.V(1).Info("reconciling policy for gateway listeners", "listeners", listenerLocators)
+		}
+
+		var gatewayHasAttachedRoutes = false
+		var gatewayHasAddresses = false
+
+		clusterID, err := utils.GetClusterUID(ctx, r.client)
+		if err != nil {
+			return fmt.Errorf("failed to generate cluster ID: %w", err)
+		}
```
KevFan, line 88: Can we move this outside the loop since we I'm assuming we only need to get this once overall? 🤔

controllers/effective_dnspolicies_reconciler.go
mikenairn: Makes sense :+1:

controllers/effective_dnspolicies_reconciler.go
mikenairn: yes we can

controllers/effective_dnspolicies_reconciler.go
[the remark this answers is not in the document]
KevFan: Yeah, this is a `status` per `Listener` defined in the `Gateway` so I guess it's possible to have multiple `Listeners` and only a `Route` attached to just one of the `Listener` using the route `sectionName` in the `parentRef` 🤔 

For example:
```yaml
      listeners:
        - attachedRoutes: 1
          conditions:
            - lastTransitionTime: "2024-10-23T08:54:09Z"
              message: No errors found
              observedGeneration: 1
              reason: Accepted
              status: "True"
              type: Accepted
            - lastTransitionTime: "2024-10-23T08:54:09Z"
              message: No errors found
              observedGeneration: 1
              reason: NoConflicts
              status: "False"
              type: Conflicted
            - lastTransitionTime: "2024-10-23T09:17:54Z"
              message: No errors found
              observedGeneration: 1
              reason: Programmed
              status: "True"
              type: Programmed
            - lastTransitionTime: "2024-10-23T09:17:54Z"
              message: No errors found
              observedGeneration: 1
              reason: ResolvedRefs
              status: "True"
              type: ResolvedRefs
          name: http
          supportedKinds:
              kind: HTTPRoute
              kind: GRPCRoute
        - attachedRoutes: 0
          conditions:
            - lastTransitionTime: "2024-10-23T08:54:09Z"
              message: No errors found
              observedGeneration: 1
              reason: Accepted
              status: "True"
              type: Accepted
            - lastTransitionTime: "2024-10-23T08:54:09Z"
              message: No errors found
              observedGeneration: 1
              reason: NoConflicts
              status: "False"
              type: Conflicted
            - lastTransitionTime: "2024-10-23T09:17:54Z"
              message: No errors found
              observedGeneration: 1
              reason: Programmed
              status: "True"
              type: Programmed
            - lastTransitionTime: "2024-10-23T09:17:54Z"
              message: No errors found
              observedGeneration: 1
              reason: ResolvedRefs
              status: "True"
              type: ResolvedRefs
          name: http2
          supportedKinds:
              kind: HTTPRoute
              kind: GRPCRoute

```
Since the `hasAttachedRoute` will always use the last `Listener` in the status block, this can be incorrect since this bool value also help to set `gatewayHasAttachedRoutes ` 🤔 

Since it's existing logic, it's probably a seperate issue (if it is) in a separate PR, but looks a bit odd to me at the moment anyway

api/v1alpha1/dnspolicy_types.go
maleck13: I can copy this change over if it is the only API change we expect

controllers/dns_workflow.go
```diff
@@ -1,7 +1,141 @@
...
+			return err == nil, err
+		}
+		return dnsPolicyAcceptedStatus(policy)
+	}
+}
+
+func dnsPolicyAcceptedStatus(policy machinery.Policy) (accepted bool, err error) {
+	p, ok := policy.(*v1alpha1.DNSPolicy)
+	if !ok {
+		return
```
maleck13, line 111: is it expected we will get non DNSPolicies? or should there be an error returned?

controllers/dnspolicies_validator.go
```diff
@@ -0,0 +1,54 @@
...
+
+type DNSPoliciesValidator struct{}
+
+func (r *DNSPoliciesValidator) Subscription() controller.Subscription {
+	return controller.Subscription{
+		ReconcileFunc: r.validate,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: &machinery.GatewayGroupKind},
+			{Kind: &kuadrantv1alpha1.DNSPolicyGroupKind},
+		},
```
maleck13, line 30: Do we also want events for DNSRecord? As DNSPolicy owns DNSRecord and if it is changed manually, we would want to set it back. Or is DNSPolicy enough to trigger that behaviour?

controllers/dnspolicies_validator.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func (r *DNSPoliciesValidator) validate(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
+	logger := controller.LoggerFromContext(ctx).WithName("DNSPoliciesValidator")
+
+	policyTypeFilterFunc := dnsPolicyTypeFilterFunc()
+
+	policies := lo.FilterMap(topology.Policies().Items(), policyTypeFilterFunc)
```
maleck13, line 39: nit: any reason  not to just pass the function straight instead of assigning it here?

controllers/dnspolicy_controller.go
```diff
@@ -17,184 +17,11 @@ limitations under the License.
...
-	)
-
-	r.dnsHelper = dnsHelper{Client: r.Client()}
-	ctrlr := ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.DNSPolicy{}).
-		Owns(&kuadrantdnsv1alpha1.DNSRecord{}).
-		Watches(&gatewayapiv1.Gateway{}, handler.EnqueueRequestsFromMapFunc(gatewayEventMapper.Map))
-	return ctrlr.Complete(r)
-}
-
```
maleck13, line 197: doesn't seem to be much left in here, do we want to get rid entirely?

controllers/dns_workflow.go
```diff
@@ -1,7 +1,141 @@
...
+	}
+	if condition := meta.FindStatusCondition(p.Status.Conditions, string(gatewayapiv1alpha2.PolicyConditionAccepted)); condition != nil {
```
maksymvavilov, line 113: why not `meta.IsStatusConditionTrue()?`

controllers/dnspolicy_status_updater.go
```diff
@@ -0,0 +1,121 @@
...
+
+		equalStatus := equality.Semantic.DeepEqual(newStatus, policy.Status)
+		if equalStatus && policy.Generation == policy.Status.ObservedGeneration {
+			pLogger.V(1).Info("policy status unchanged, skipping update")
+			continue
+		}
+		newStatus.ObservedGeneration = policy.Generation
+		policy.Status = *newStatus
+
+		obj, err := controller.Destruct(policy)
```
maleck13, line 106: does this remove it from the topology and then it gets re-added when we update it?

controllers/dnspolicy_status_updater.go
```diff
@@ -0,0 +1,121 @@
...
+
+	policyTypeFilterFunc := dnsPolicyTypeFilterFunc()
+	policyAcceptedFunc := dnsPolicyAcceptedStatusFunc(state)
+	policyErrorFunc := dnsPolicyErrorFunc(state)
+
+	policies := lo.FilterMap(topology.Policies().Items(), policyTypeFilterFunc)
+
+	logger.V(1).Info("updating dns policy statuses", "policies", len(policies))
+
+	for _, policy := range policies {
```
maksymvavilov, line 54: how do we end up only with policies that target a DNS record under reconciliation? and should there be more than one policy?

controllers/dnspolicy_status.go
```diff
@@ -41,37 +34,7 @@ import (
 
 var NegativePolarityConditions []string
```
maksymvavilov, line 35: What `NegativePolirity` meats for the record condition?

pkg/library/utils/k8s_utils.go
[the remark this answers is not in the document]
maksymvavilov: We use this ID to ensure records in the DNSProvider don't conflict with each other in the multicluster context. I'm wondering if we can make topologies unique when creating them (we do client requests at that point as far as I understand) to booth be in a line with the rest policies and make sure we can differentiate from other clusters in providers :thinking:

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+			gateway := listener.Gateway
+			if listener.Hostname == nil || *listener.Hostname == "" {
+				lLogger.Info("skipping listener no hostname assigned")
+				continue
+			}
+
+			hasAttachedRoute := false
+			for _, statusListener := range gateway.Status.Listeners {
+				if string(listener.Name) == string(statusListener.Name) {
+					hasAttachedRoute = statusListener.AttachedRoutes > 0
```
maleck13, line 106: yeah I think we could break here once found?

maleck13: minor thing

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+				continue
+			}
+			obj, err := controller.Destruct(desiredRecord)
+			if err != nil {
+				lLogger.Error(err, "unable to destruct dns record") // should never happen
+				continue
+			}
+			resource := r.client.Resource(DNSRecordResource).Namespace(desiredRecord.GetNamespace())
+
+			existingRecordObj, ok := lo.Find(topology.Objects().Children(listener), func(o machinery.Object) bool {
```
maleck13, line 129:
```suggestion
			existingRecordObj, recordExists := lo.Find(topology.Objects().Children(listener), func(o machinery.Object) bool {
```

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+				_, ok := o.(*controller.RuntimeObject).Object.(*kuadrantdnsv1alpha1.DNSRecord)
+				return ok && o.GetNamespace() == listener.GetNamespace() && o.GetName() == dnsRecordName(listener.Gateway.Name, string(listener.Name))
+			})
+
+			//Update
+			if ok {
```
maleck13, line 139:
```suggestion
			if recordExists {
```

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+				existingRecord := existingRecordObj.(*controller.RuntimeObject).Object.(*kuadrantdnsv1alpha1.DNSRecord)
+
+				//Deal with the potential deletion of a record first
+				if !hasAttachedRoute || len(desiredRecord.Spec.Endpoints) == 0 {
+					if logger.V(1).Enabled() {
```
maleck13, line 146: do we need this? I see it being used elsewhere without this check

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+	logger := controller.LoggerFromContext(ctx).WithName("EffectiveDNSPoliciesReconciler")
+
+	policyTypeFilterFunc := dnsPolicyTypeFilterFunc()
+	policyAcceptedFunc := dnsPolicyAcceptedStatusFunc(state)
+
+	policies := lo.FilterMap(topology.Policies().Items(), policyTypeFilterFunc)
+
+	logger.V(1).Info("updating dns policies", "policies", len(policies))
+
+	for _, policy := range policies {
```
maleck13, line 65: this is one massive for loop :) wondering should we (in a follow up) split it into a set of function calls if possible and reduce the congnitive overhead (at least in reviews)

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+
+	return r.deleteOrphanDNSRecords(controller.LoggerIntoContext(ctx, logger), topology)
+}
+
+// listenersForPolicy returns an array of listeners that are targeted by the given policy.
+// If the target is a Listener a single element array containing that listener is returned.
+// If the target is a Gateway all listeners that do not have a DNS policy explicitly attached are returned.
+func (r *EffectiveDNSPoliciesReconciler) listenersForPolicy(_ context.Context, topology *machinery.Topology, policy machinery.Policy, policyTypeFilterFunc dnsPolicyTypeFilter) []*machinery.Listener {
```
maleck13, line 211: so with this do we get section name support then?

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -0,0 +1,268 @@
...
+				})
+				pTargetablesLocs := lo.Map(pTargettables, func(item machinery.Targetable, _ int) string {
+					return item.GetLocator()
+				})
+				logger.V(1).Info("dns record parents", "record", item.GetLocator(), "targetables", pTargetablesLocs, "polices", pPoliciesLocs)
+			}
+
+			//Target removed from topology
+			if len(pTargettables) == 0 {
+				return true
```
maleck13, line 250: worth info logging here (orphan record being removed as no targets remain)?

maleck13 (approved):
few minor things. But generally changes look ok and approachable. Will try it out locally but happy to approve

maleck13:
@mikenairn I did a bunch of addiitonal manual tests with multiple gateways, multiple httproutes, deleting httproutes, deleting gateways, no credentials etc all worked as expected

👍 
/lgtm

maleck13 (approved):

controllers/dns_workflow.go
mikenairn: If you passed on in you would, but where its being used it would never not be. That said, we don't want to return an error in this case as the errors returned from this function are supposed to be related to the dnspolicy status.

Same code as RLP for consistency https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimit_workflow.go#L267-L280

controllers/dnspolicies_validator.go
mikenairn: This even this on the validation task of the policy so i don't believe this is required here. No change to the record will cause the policy to be invalidated.

controllers/dnspolicy_controller.go
mikenairn: In this PR i am deliberately not moving code that i don't need to so i don't have to deal with a review of existing code. I will be following this PR up with another to remove some of the existing controller files, this will have no functional changes, only moving stuff about to fit better with the new way of working.

controllers/dns_workflow.go
mikenairn: Yes it probably could be.

Same code as RLP for consistency https://github.com/Kuadrant/kuadrant-operator/blob/main/controllers/ratelimit_workflow.go#L267-L280

controllers/dnspolicy_status_updater.go
mikenairn: If you are referring to `controller.Destruct(policy)`, this is just converting it to a `unstructured.Unstructured` type so it works with the client https://github.com/Kuadrant/policy-machinery/blob/main/controller/runnable.go#L231-L237

I found this method name confusing when i saw it being used also.

controllers/dnspolicy_status.go
mikenairn: I'll defer that question to @philbrookes who added it.

I'm only removing code from this file in this PR, this already exists.

controllers/dnspolicy_status_updater.go
mikenairn: You don't really, this iterates around all polices ensuring there status is correct. Majority of times this will do nothing, but if any changes are detected in the topology related to a policy, records removed or in a different state etc.., the status will be updated accordingly.

controllers/effective_dnspolicies_reconciler.go
mikenairn: sure, it's not really much bigger than what we had previously, but it could probably benefit from a refactor of sorts. I would do that after though.

controllers/effective_dnspolicies_reconciler.go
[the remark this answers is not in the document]
mikenairn: > Since the `hasAttachedRoute` will always use the last `Listener` in the status block, this can be incorrect since this bool value also help to set `gatewayHasAttachedRoutes ` 🤔

I'm still not really seeing the issue. Your status there has two listener status with different names, one with attached routes one without. The expected value of `gatewayHasAttachedRoutes` in this case will be `true` as we only set it to `true` when a listener [has attached routes](https://github.com/Kuadrant/kuadrant-operator/pull/937/files#diff-b6f99dd8a13a65bd856d8867f97c855b436c973a324732b8bc0e8a91f7663545R109-R111)  , default value is [false](https://github.com/Kuadrant/kuadrant-operator/pull/937/files#diff-b6f99dd8a13a65bd856d8867f97c855b436c973a324732b8bc0e8a91f7663545R87) 

`hasAttachedRoute` is set to true only when we find the listener with the desired name and [defaults to false](https://github.com/Kuadrant/kuadrant-operator/pull/937/files#diff-b6f99dd8a13a65bd856d8867f97c855b436c973a324732b8bc0e8a91f7663545R103).

I'd say create an issue if you think there is still an issue, there may well be one, but i can't see it currently.

I will however add a break.

controllers/effective_dnspolicies_reconciler.go
mikenairn: possibly not here since its not actually calculating anything additional to add into the log statements (probably was at one point and i removed it)

controllers/effective_dnspolicies_reconciler.go
mikenairn: Yes, this should allow section name to be added without new changes to this task. I have already got this working and will create a separate PR for that after this one is merged.

api/v1alpha1/dnspolicy_types.go
mikenairn: There will be additional changes to the API if/when we add section name, but it will be minimal.

controllers/effective_dnspolicies_reconciler.go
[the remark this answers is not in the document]
KevFan: > hasAttachedRoute is set to true only when we find the listener with the desired name and [defaults to false](https://github.com/Kuadrant/kuadrant-operator/pull/937/files#diff-b6f99dd8a13a65bd856d8867f97c855b436c973a324732b8bc0e8a91f7663545R103).

Oh, I see now. I completely missed the fact that it does a match of listener status name with the desired name 🤦 In that case, it's not an issue. Ignore this 😅

controllers/dnspolicy_status_updater.go
maksymvavilov: I understand now, thanks! This sotw thing is weird

KevFan (approved):
Looks good to me 👍

KevFan (approved):

controllers/common.go
guicassolato: I've been adding a few common vars and helper funcs for things like labeling internal resources we create, etc, including one to identify the resource as "managed-by" kuadrant, here: https://github.com/Kuadrant/kuadrant-operator/blob/b6494ca686716a4fb90f03ae8360fb25a1ab7745/controllers/state_of_the_world.go#L43

(and also at the top of specific workflow files, when the var is not just as common as these ones.)

Happy to move those here once this is merged, if you believe it would help having them all in one place rather than separated by area.

pkg/library/utils/k8s_utils.go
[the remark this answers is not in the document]
mikenairn: https://github.com/Kuadrant/kuadrant-operator/issues/959

PR #943, opened 2024-10-16
ci: Add kuadrant status check step to test jobs

mikenairn:
Adds a step to the integration test jobs to display the pods in the kuadrant-system namespace. More info could be added here, but for now this is enough to tell you if the pods were restarting during the test run for some reason (panic in the operator itself).

Add workflow dispatch to "Test" workflow.


KevFan (approved):
Looks good to me

Issue #846, opened 2024-09-02
Consistent labeling for Operator metric Services


jsmolar:
### Current labeling for Operator Metric Services
- **kuadrant-operator-metrics** - `control-plane=controller-manager`
- **authorino-operator-metrics** - `control-plane=authorino-operator`
- **limitador-operator-metrics** - `control-plane=controller-manager`
- **dns-operator-controller-manager-metrics-service** -` control-plane=dns-operator-controller-manager`

### Proposed Change
To achieve consistency in the labeling of Operator metric services, I propose aligning the labels according to the pattern implemented by the authorino-operator, or by a new common pattern.

Two potential approaches:

1. Change label value to reflect correct service
    **kuadrant-operator-metrics** - `control-plane=kuadrant-operator`
    **authorino-operator-metrics** - `control-plane=authorino-operator`
    **limitador-operator-metrics** - `control-plane=limitador-operator`
    **dns-operator-metrics** - `control-plane=dns-operator`
   

2. Change both key and value 
    **kuadrant-operator-metrics** - `component=kuadrant-operator`
    **authorino-operator-metrics** - `component=authorino-operator`
    **limitador-operator-metrics** - `component=limitador-operator`
    **dns-operator-metrics** - `component=dns-operator`

david-martin:
When labels are changed, please also update the ServiceMonitors as needed to ensure they target the correct Services.
See https://github.com/Kuadrant/kuadrant-operator/pull/953#discussion_r1814655693 for related discussion on this.

Boomatang:
@david-martin Now that V1 is out. How much issues can be caused by changing the labels for metrics? Would it be regarded as a breaking change?

david-martin:
> Would it be regarded as a breaking change

Strictly, yes.

Less strictly, depends on if they are being used by anyone/anything yet. Even then, the surface area may be sufficiently small that the beneift of the change indicated in this issue outweighs the work required.

PR #953, opened 2024-10-23
Update steps for setting up metrics on openshift, focusing on single …

david-martin:
…cluster

- combine service monitors for operators for easier deploy
- update openshift install steps to reference this file and explain the setup better
- remove reference to multi cluster thanos setup to avoid confusion about user workload thanos-querier and custom thanos
- add missing metrics service proxy for scraping important metrics directly from the gateway pod


doc/install/install-openshift.md
```diff
@@ -172,36 +172,42 @@ Wait for Envoy Gateway to become available::
...
 ```bash
 kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/main/config/observability/openshift/kube-state-metrics.yaml
 kubectl apply -k https://github.com/Kuadrant/gateway-api-state-metrics?ref=main
 ```
 
-To enable request metrics in Istio, you must create a `telemetry` resource as follows:
+To enable request metrics in Istio and scrape them, create the following resources:
 
 ```bash
 kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/main/config/observability/openshift/telemetry.yaml
```
roivaz, line 192: This is not strictly necessary to have request metrics in Istio, those are enabled by default. This Telemetry configuration adds the request path as a label to the request metrics, which is not always desirable as it is a high cardinality label that can flood your prometheus instance if you have a big API. For example each resource in an API would generate a different prometheus time-series. We probably should warn about this at least.

doc/user-guides/secure-protect-connect-single-multi-cluster.md
```diff
@@ -162,6 +162,46 @@ kubectl get gateway ${gatewayName} -n ${gatewayNS} -o=jsonpath='{.status.listene
...
+spec:
+  selector:
+    istio.io/gateway-name: ${gatewayName}
+  ports:
+  - name: metrics
+    protocol: TCP
+    port: 15020
+    targetPort: 15020    
+EOF
+```
```
roivaz, line 203: This can be done using a single PodMonitor that targets all your Gateways because Istio annotates the gateway pods with the port where the metrics are served. It's slightly convoluted though, but has the advantage of targeting all gateways in the namespace with a single PodMonitor:

```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: istio-proxies-monitor
spec:
  selector:
    matchExpressions:
      - key: istio-prometheus-ignore
        operator: DoesNotExist
  podMetricsEndpoints:
    - path: /stats/prometheus
      interval: 30s
      relabelings:
        - action: keep
          sourceLabels: ["__meta_kubernetes_pod_container_name"]
          regex: "istio-proxy"
        - action: keep
          sourceLabels:
            ["__meta_kubernetes_pod_annotationpresent_prometheus_io_scrape"]
        - action: replace
          regex: (\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})
          replacement: "[$2]:$1"
          sourceLabels:
            [
              "__meta_kubernetes_pod_annotation_prometheus_io_port",
              "__meta_kubernetes_pod_ip",
            ]
          targetLabel: "__address__"
        - action: replace
          regex: (\d+);((([0-9]+?)(\.|$)){4})
          replacement: "$2:$1"
          sourceLabels:
            [
              "__meta_kubernetes_pod_annotation_prometheus_io_port",
              "__meta_kubernetes_pod_ip",
            ]
          targetLabel: "__address__"
        - action: labeldrop
          regex: "__meta_kubernetes_pod_label_(.+)"
        - sourceLabels: ["__meta_kubernetes_namespace"]
          action: replace
          targetLabel: namespace
        - sourceLabels: ["__meta_kubernetes_pod_name"]
          action: replace
          targetLabel: pod_name
```

I recall getting this from the Istio documentation, but I can't find it now ...

doc/install/install-openshift.md
david-martin: Good point.
I'll split this out and explain better with a warning.

roivaz (commented):
Just a couple of comments, looks good overall.

config/observability/prometheus/monitors/operators.yaml
```diff
@@ -0,0 +1,73 @@
...
+  name: limitador-operator-metrics
+  namespace: kuadrant-system
+spec:
+  endpoints:
+    - path: /metrics
+      port: metrics
+      scheme: http
+  selector:
+    matchLabels:
+      control-plane: controller-manager
```
jsmolar, line 71: this `matchLabel` is the same as for kuadrant. This can be a problem for monitor to target correct service. I created an [issue for this](https://github.com/Kuadrant/kuadrant-operator/issues/846), but do we want to address this here in doc?

doc/user-guides/secure-protect-connect-single-multi-cluster.md
roivaz: Ah, here is the source https://github.com/istio-ecosystem/sail-operator/blob/main/docs/README.md#observability-integrations
Sail documentation, not Istio's.

config/observability/prometheus/monitors/operators.yaml
```diff
@@ -0,0 +1,73 @@
...
+      scheme: http
+  selector:
+    matchLabels:
+      control-plane: authorino-operator
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    control-plane: controller-manager
```
jsmolar, line 22: Is it required to use an extended amount of labels for the DNS monitor?

doc/user-guides/secure-protect-connect-single-multi-cluster.md
david-martin: I have a similar looking 'additionalScrapeConfig' [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/prometheus/additional-scrape-configs.yaml) but it doesn't work with user workload monitoring on Openshift due to restrictions on what can be configured.

If this single PodMonitor approach works with UWM, I think that would be more robust than the Service/ServiceMonitor approach.

doc/install/install-openshift.md
```diff
@@ -172,36 +172,42 @@ Wait for Envoy Gateway to become available::
...
+```bash
+kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/refs/heads/main/config/observability/prometheus/monitors/operators.yaml
 ```
 
-If you have Grafana installed in your cluster, you can import the [example dashboards and alerts](https://docs.kuadrant.io/latest/kuadrant-operator/doc/observability/examples).
+!!! note
+
+    There is 1 more metrics configuration that needs to be applied so that all relevant metrics are being scraped.
+    That configuration depends on where you deploy your Gateway.
+    The steps to configure that are detailed in the follow on 'Secure, protect, and connect' guide.
```
jsmolar, line 206: Maybe we should provide link to `Secure, protect, and connect`

doc/install/install-openshift.md
```diff
@@ -172,36 +172,42 @@ Wait for Envoy Gateway to become available::
...
 
-For example installation details, see [installing Grafana on OpenShift](https://cloud.redhat.com/experts/o11y/ocp-grafana/). When installed, you must add your Thanos instance as a data source to Grafana. Alternatively, if you are using only the user workload monitoring stack in your OpenShift cluster, and not writing metrics to an external Thanos instance, you can [set up a data source to the thanos-querier route in the OpenShift cluster](https://docs.openshift.com/container-platform/4.15/observability/monitoring/accessing-third-party-monitoring-apis.html#accessing-metrics-from-outside-cluster_accessing-monitoring-apis-by-using-the-cli).
+The [example Grafana dashboards and alerts](https://docs.kuadrant.io/latest/kuadrant-operator/doc/observability/examples/) for observing Kuadrant functionality use low-level CPU metrics and network metrics available from the user monitoring stack in OpenShift. They also use resource state metrics from Gateway API and Kuadrant resources.
```
jsmolar, line 208: The HTTPRoute must include the service and deployment labels, which are marked as "Important" in the dashboard. I believe we should mention this here as well. If the user follows this guide and imports the dashboards, they will need to return to OCP to add the labels before any data is visible on the dashboard. My suggestion is to include this step in the guide, allowing the user to complete the entire setup in one go.

doc/install/install-openshift.md
```diff
@@ -172,36 +172,42 @@ Wait for Envoy Gateway to become available::
...
 
+For Grafana installation details, see [installing Grafana on OpenShift](https://cloud.redhat.com/experts/o11y/ocp-grafana/). When installed, you must [set up a data source to the thanos-querier route in the OpenShift cluster](https://docs.openshift.com/container-platform/4.15/observability/monitoring/accessing-third-party-monitoring-apis.html#accessing-metrics-from-outside-cluster_accessing-monitoring-apis-by-using-the-cli).
```
jsmolar, line 210: Thanos datasource setup is also described in `install Grafana on OpenShifte guide.

doc/user-guides/secure-protect-connect-single-multi-cluster.md
roivaz: Yes, I have tested this in OCP with UWM and works as expected.

config/observability/prometheus/monitors/operators.yaml
david-martin: These look generated by kubebuilder/kustomize.
I don't see any harm in having them

config/observability/prometheus/monitors/operators.yaml
david-martin: Consistency is important to avoid confusion, so +1 for that issue.
I believe the monitor targets both services, so metrics would all be scraped (so the suggested monitors should work).
I'll link back to this comment in that issue so the matchLabels config is updated accordingly when labels are updated.

doc/install/install-openshift.md
david-martin: Good catch.
I think i'll call that out here, but keep the 2nd link as well as it's a more 'more details' and official way of accessing thanos-querier.

doc/install/install-openshift.md
david-martin: Sure, i'll mention it here.

roivaz (approved):

jsmolar (approved):

PR #954, opened 2024-10-23
Istio installed using Sail operator added to integration tests

eguzki:
Additionally, istio default intallation tool in dev environments now it is Sail Operator


eguzki:
tests are expected to fail because the base branch `install-sail-tp` needs to be rebased. The base branch needs to pick the fix of the authorino-operator  deployment

adam-cattermole (approved):
I've just rebased the base branch so this will need one too, but changes here lgtm!

PR #961, opened 2024-10-25
dnspolicy section name support

mikenairn:
Add section name support to DNSPolicy.

image


controllers/effective_dnspolicies_reconciler.go
```diff
@@ -300,3 +297,34 @@ func (r *EffectiveDNSPoliciesReconciler) deleteRecord(ctx context.Context, obj m
...
+	// DNSRecord doesn't currently support record type changes due to a limitation of the dns operator
+	// https://github.com/Kuadrant/dns-operator/issues/287
+	canUpdate := true
+	for _, curEp := range current.Spec.Endpoints {
+		for _, desEp := range desired.Spec.Endpoints {
+			if curEp.DNSName == desEp.DNSName {
+				if curEp.RecordType != desEp.RecordType {
+					logger.V(1).Info("record type for existing endpoint has changed",
+						"dnsName", curEp.DNSName, "current", curEp.RecordType, "desired", desEp.RecordType)
+					canUpdate = false
```
guicassolato, line 320: It looks like once this turns into `false`, it won't ever get back to `true` anymore. Could we simply `return false` maybe?

mikenairn: Yes, no point in continuing. Updated.

maleck13:
**Verifiation**

- Created Gateway
- Created multiple listeners
- Created section name target dnspolicy and none section name dnspolicy
- Verified DNS records created as expected
- Deleted each policy individually and all together 
- verified dns records are removed and all related objects are removed
- Changed a listener name
- verified dnsrecords removed as expected.

👍

maleck13 (approved):
/lgtm

controllers/effective_dnspolicies_reconciler.go
```diff
@@ -300,3 +297,30 @@ func (r *EffectiveDNSPoliciesReconciler) deleteRecord(ctx context.Context, obj m
...
+					logger.V(1).Info("record type for existing endpoint has changed",
+						"dnsName", curEp.DNSName, "current", curEp.RecordType, "desired", desEp.RecordType)
+					return false
+				}
+			}
+		}
+	}
+
+	return true
+}
```
mikenairn, line 326: @maleck13 fyi, because i had to do this to deal with possible changes in record type when attaching a policy to a listener that already has existing records via policy attached to a gateway, it means we could probably remove this now https://github.com/Kuadrant/kuadrant-operator/blob/main/api/v1alpha1/dnspolicy_types.go#L56 and allow switching from simple to loadbalanced without the need for manual intervention.

maleck13: nice ok. Lets go for it

mikenairn: I'll do it in a different PR, will need some tests

mikenairn: https://github.com/Kuadrant/kuadrant-operator/pull/968

PR #966, opened 2024-10-30
'Update dependencies image versions' workflow_dispatch

guicassolato:
Adds `workflow_dispatch` option for the 'Update dependencies image versions' workflow


didierofrivia (approved):
🙏🏼

PR #968, opened 2024-10-31
fix: Remove dnspolicy lb immutable validation

mikenairn:
Removes the immutable validation on the loadBalancing field of the dnspolicy spec. 

Updates made as part of [section name support ](https://github.com/Kuadrant/kuadrant-operator/pull/961) allow the controller to handle changes to endpoint record types,  so it's now possible to switch from what we used to refer to as simple policy types (loadBalancing field not set) to loadbalanced (loadBalancing field set) and vice versa without deleting and re-creating the policy.


maleck13 (approved):
/lgtm

PR #971, opened 2024-11-01
Integrate `AuthConfig`'s `v1beta3` changes in `AuthPolicy`

adam-cattermole:
### Verification

```sh
make local-setup
```

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

```sh

kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
    namespace: gateway-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/cars"
    - method: GET
      path:
        type: PathPrefix
        value: "/dolls"
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

```sh
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1beta3
kind: AuthPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  when:
    - predicate: request.headers['my-header'] == 'test'
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
EOF
```

```sh
kubectl apply -f - <<EOF
apiVersion: v1
kind: Secret
metadata:
  name: bob-key
  labels:
    app: toystore
  annotations:
stringData:
  api_key: IAMBOB
type: Opaque
---
apiVersion: v1
kind: Secret
metadata:
  name: alice-key
  labels:
    app: toystore
  annotations:
stringData:
  api_key: IAMALICE
type: Opaque
EOF
```

```sh
```

```sh
curl -H 'Host: api.toystore.com' http://$GATEWAY_URL/cars -i
# HTTP/1.1 200 OK
```

```sh
curl -H 'Host: api.toystore.com' -H 'my-header: test' http://$GATEWAY_URL/cars -i
# HTTP/1.1 401 Unauthorized
```

```sh
curl -H 'Host: api.toystore.com' -H 'my-header: test' -H 'Authorization: APIKEY IAMALICE' http://$GATEWAY_URL/cars -i
# HTTP/1.1 200 OK
```

```sh
kubectl get -n gateway-system wasmplugin/kuadrant-kuadrant-ingressgateway -o yaml
```
```yaml
pluginConfig:
  actionSets:
  - actions:
    - predicates:
      - request.headers['my-header'] == 'test'
      scope: 2c627402626ee2c07410deeaccf34c5adef98942f7b3ad15ac9cd044aba42aa1
      service: auth-service
    name: 3333a094e3493a905c35f12ffe3fa372e02a38c9700389a532f722e00d28d898
    routeRuleConditions:
      hostnames:
      - api.toystore.com
      predicates:
      - request.url_path.startsWith('/admin')
  - actions:
    - predicates:
      - request.headers['my-header'] == 'test'
      scope: e2db39952dd3bc72e152330a2eb15abbd9675c7ac6b54a1a292f07f25f09f138
      service: auth-service
    name: 5d3152cd9b75ba0b71f4c631dac258c905748ba6cce17a9b887b8bebe43a77f0
    routeRuleConditions:
      hostnames:
      - api.toystore.com
      predicates:
      - request.method == 'GET'
      - request.url_path.startsWith('/dolls')
  - actions:
    - predicates:
      - request.headers['my-header'] == 'test'
      scope: e2db39952dd3bc72e152330a2eb15abbd9675c7ac6b54a1a292f07f25f09f138
      service: auth-service
    name: 221b3f3322ef9c2a2a833343073f3d5997ad81354f67d8b4c652cbbd054d692c
    routeRuleConditions:
      hostnames:
      - api.toystore.com
      predicates:
      - request.method == 'GET'
      - request.url_path.startsWith('/cars')
  services:
    auth-service:
      endpoint: kuadrant-auth-service
      failureMode: deny
      type: auth
    ratelimit-service:
      endpoint: kuadrant-ratelimit-service
      failureMode: allow
      type: ratelimit
```


pkg/wasm/types.go
```diff
@@ -131,16 +131,16 @@ func (r *RouteRuleConditions) EqualTo(other RouteRuleConditions) bool {
...
+		if r.Predicates[i] != other.Predicates[i] {
 			return false
 		}
 	}
 
 	return true
 }
 
-type Predicate struct {
+type Condition struct {
```
adam-cattermole, line 143: Somewhat confusingly we called the [wasm plugin `Condition`](https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration/action.rs#L16) (i.e. `PatternExpression`) `Predicate`.. but now we have a [`Predicates` field in wasm](https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration/action.rs#L18) and so to retain both I've renamed `Predicate` -> `Condition` and introduced `Predicates` (`[]string`) in routeRuleConditions..

api/v1beta3/ratelimitpolicy_types.go
```diff
@@ -32,8 +32,8 @@ import (
 const (
 	EqualOperator      WhenConditionOperator = "eq"
 	NotEqualOperator   WhenConditionOperator = "neq"
-	StartsWithOperator WhenConditionOperator = "startswith"
-	EndsWithOperator   WhenConditionOperator = "endswith"
+	StartsWithOperator WhenConditionOperator = "startsWith"
+	EndsWithOperator   WhenConditionOperator = "endsWith"
```
adam-cattermole, line 36: I'm not 100% about changing these - they're not used elsewhere but it's a little confusing using some within CEL predicates and others for `PatternExpression`s

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/goccy/go-yaml v1.12.0
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
-	github.com/kuadrant/authorino v0.18.0
+	github.com/kuadrant/authorino v0.18.1-0.20241030162053-82d7619acc40
```
adam-cattermole, line 13: For now importing the version from main until we have a `v0.19.0` release

adam-cattermole: Updated to use the [v0.19.0 release](https://github.com/Kuadrant/authorino/releases/tag/v0.19.0)

pkg/wasm/types.go
```diff
@@ -116,12 +115,12 @@ func (s *ActionSet) EqualTo(other ActionSet) bool {
 }
 
 type RouteRuleConditions struct {
-	Hostnames []string    `json:"hostnames"`
-	Matches   []Predicate `json:"matches,omitempty"`
+	Hostnames  []string `json:"hostnames"`
+	Predicates []string `json:"predicates,omitempty"`
```
eguzki, line 119: Check out https://github.com/Kuadrant/kuadrant-operator/pull/976/files#diff-509b896bd679d72634dfb60a3d6b4bed29d15972232cc2068d0d22efce1c1518R110-R116

I tried to describe the existing Wasm configuration. However, only using CEL expressions from control plane.

controllers/ratelimit_workflow_helpers.go
```diff
@@ -154,7 +154,7 @@ func wasmActionFromLimit(limit *kuadrantv1beta3.Limit, limitIdentifier, scope st
 		ServiceName: wasm.RateLimitServiceName,
 		Scope:       scope,
 	}
-	if conditions := wasm.PredicatesFromWhenConditions(limit.When...); len(conditions) > 0 {
+	if conditions := wasm.ConditionsFromWhenConditions(limit.When...); len(conditions) > 0 {
```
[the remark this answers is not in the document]
adam-cattermole: The new thing is `Predicates`, but it still supports `Conditions` - this is mapping existing PatternExpression Conditions -> wasm Conditions

api/v1beta3/ratelimitpolicy_types.go
adam-cattermole: Unfortunately we can't remove the `WhenCondition` because we import the `AuthConfig` which still has selector expressions present.. so I've left it so it can support both.. but if we're not using these types I could remove them

adam-cattermole: Unless if a user sets them we just ignore them?

pkg/wasm/types.go
adam-cattermole: Sure this is similar to what I did, but I thought we could avoid `Matches` entirely given the routeRuleConditions will always be generated by Kuadrant and replaced it with `Predicates`

alexsnaps: Can we report in the status when people use them and _NOT_ apply the Policy? @guicassolato wdyt?

alexsnaps: That's for Auth btw, I think they should :fire: in RLP

guicassolato: I think we should burn them in Auth as well actually. However, I'd do it as @adam-cattermole proposed it, i.e., by keeping the old user-defined `when` conditions (based on `selector, operator, value`) alongside with CEL as an addition only in a first iteration, generating the route conditions internally in CEL only (because why would we do it otherwise?), and then, in a separate PR, remove the old `when` conditions from both kinds of policy at once.

guicassolato: Reporting in the status the use of old `when` conditions as invalid doesn't make sense to me. Why would we have them in the API if the user cannot use them?

guicassolato: BTW, the amount of work to burn them in Auth in non-negligible. It's a lot of Authorino types to refine inside the Kuadrant Operator.

alexsnaps: Right... the amount of work is what annoys me 😮‍💨

api/v1beta3/ratelimitpolicy_types.go
eguzki: Up to you. I still believe that the operator can generate CEL expressions out of `namePatterns` instead of `WhenConditions` and we can :fire:  `WhenConditions`. 

But can be done later on.

controllers/ratelimit_workflow_helpers.go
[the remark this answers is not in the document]
eguzki: I was hoping we could :fire: some "deprecated" configuration at the wasm side. Moving to CEL for all use cases.

Fair enough for this iteration.

eguzki (approved):

Issue #759, opened 2024-07-12
Move to v1 for DNSPolicy


maleck13:
Move DNSPolicy to `kuadarant.io/v1`

Update docs to refer to v1 instead of v1alpha1

maleck13:
https://github.com/Kuadrant/kuadrant-operator/pull/948

PR #915, opened 2024-10-07
sotw: event predicates

guicassolato:
Filter ~topology configmap~ events with generation-changed predicate type.

PR #973, opened 2024-11-04
Move DNS and TLS Policies to v1

### LINKED ISSUE #762: Move to v1 of TLSPolicy

mikenairn:
closes #762 #759


api/v1/merge_strategies.go
```diff
@@ -32,6 +32,7 @@ const (
 	PolicyRuleMergeStrategy = "merge"
 )
 
+// +kubebuilder:object:generate=false
```
mikenairn, line 35: @guicassolato Had to exclude this due to this error in generate:
```
merge_strategies.go:36:2: invalid field type: any
Error: not all generators ran successfully
```

I don't think it needs to be included anyway, let me know if you think otherwise.

config/manifests/bases/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -32,26 +32,16 @@ spec:
...
-        name: dnshealthcheckprobes.kuadrant.io
-        version: v1alpha1
       - description: DNSPolicy configures how North-South based traffic should be balanced and reach the gateways
         displayName: DNSPolicy
         kind: DNSPolicy
         name: dnspolicies.kuadrant.io
-        version: v1alpha1
-      - description: DNSRecord represents DNS endpoints for gateway network services
-        displayName: DNSRecord
-        kind: DNSRecord
```
azgabur, line 47: Does this mean the `DNSRecord` and `DNSHealthCheckProbe` CRs will be removed? Or it is defined elsewhere?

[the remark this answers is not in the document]
maleck13: I don't think so. Also the tests would fail in that case.

maleck13 (approved):
changes look good to me

mikenairn:
From what i can tell the code coverage report is broken because we have been squashing commits onto main, creating a new commit that has never had a report done on it, and codecov is using the last common commit it knows about when generating its PR reports, which is currently [this one](https://github.com/Kuadrant/kuadrant-operator/commit/63f1d28a9ee06bbc3d001c81bf19b0b2cd35f897)  from [this](https://github.com/Kuadrant/kuadrant-operator/pull/915) PR.

[This PR](https://github.com/Kuadrant/kuadrant-operator/pull/952) looks to have decreased it enough to the point of it failing all the PR builds.

Going to merge this, creating a merge commit, so any new branches should report from this commit on. Might be an idea to disable the squash option.

Issue #962, opened 2024-10-27
A potential risk in Kuadrant that could lead to takeover of the cluster 


HouqiyuA:
Dear Team Members:

Greetings! Our team is very interested in your project and we recently identified a potential RBAC security risk while doing a security assessment of your project. Therefore, we would like to report it to you and provide you with the relevant details so that you can fix and improve it accordingly.

Details:

In this Kubernetes project, there is a ClusterRole named ’ manager-role‘ that has high-risk privileges to list, get, and watch on secrets resources. （https://github.com/Kuadrant/kuadrant-operator/blob/main/config/rbac/role.yaml). This ClusterRole has the privilege to directly list the secrets of the entire cluster. If an attacker steals a service account with this privilege, he can elevate the privilege and further take over the entire cluster.

Please confirm the purpose of the assignment of this permission and consider using a more granular permission non-assignment rule.
Best wishes.

HouqiyuA

maleck13:
Thanks for raising this, it was the case that Kuadrant Operator would read secrets for enforcing TLSPolicy. I believe this may have changed now. We will review the permissions

PR #974, opened 2024-11-04
remove secret perm as not using it

maleck13:
TLSPolicy no longer interacts with the secret created. So this permission is no longer needed

#962 

rh-pre-commit.version: 2.2.0


mikenairn (approved):
/lgtm

@KevFan You might want to cast your eye over this being the last person to go through the TLS flow, but fairly sure this was needed for previous iterations of TLS support and if the tests all pass we should be good.

KevFan (approved):

KevFan (approved):

PR #976, opened 2024-11-05
951 finalize ratelimitpolicy v1beta3

eguzki:
### What

Finalize ratelimitpolicy v1beta3. Basically, exposing [CEL](https://cel.dev/) at RateLimitPolicy level to express predicates at top level and limit levels. 

Updates applied

- [X] Top level CEL `predicate`'s
- [X] Limit level CEL `predicate`'s
- [X] Rates `window` in [GEP-2257: Gateway API Duration Format](https://gateway-api.sigs.k8s.io/geps/gep-2257/?h=duration)
- [X] Counters CEL `expression`

TODO:
- [x] unittests
- [x] e2e tests
- [x] doc
- [x] Test user guides with new RLP API

```yaml
apiVersion: kuadrant.io/v1beta3
kind: RateLimitPolicy
metadata:
  name: toy
  namespace: gateway-system
spec:
  targetRef:
    kind: Gateway
    name: kuadrant-ingressgateway
  when:
    - predicate: "auth.identity.group == 'admin'"   # a list of "top-level" `Predicate`s
  limits:
    "expensive-operation":
      rates:
        - limit: 2
          window: 30s   # a duration as we now express them elsewhere and in GwAPI
      when:
        - predicate: "request.method == 'POST'"  # a list of per-limit `Predicate`s as were our `Condition`s, but CEL

    "limit-per-ip":
      rates:
        - limit: 5
          window: 30s
      counters:
        - expression: source.remote_address  # this maps more or less to what we do in authorino wrt `properties` et al 
```

### Verification

```sh
make local-setup
```

```sh
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

```sh

kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
    namespace: gateway-system
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - method: GET
      path:
        type: PathPrefix
        value: "/cars"
    - method: GET
      path:
        type: PathPrefix
        value: "/dolls"
    backendRefs:
    - name: toystore
      port: 80
  - matches:
    - path:
        type: PathPrefix
        value: "/admin"
    backendRefs:
    - name: toystore
      port: 80
EOF
```

```sh
```

Verify the route works:

```sh
curl -H 'Host: api.toystore.com' http://$GATEWAY_URL/cars -i
# HTTP/1.1 200 OK
```

Enforce rate limiting on requests to the Toy Store API

```sh
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1beta3
kind: RateLimitPolicy
metadata:
  name: toystore
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  limits:
    "create-toy":
      rates:
      - limit: 5
        window: 10s
      when:
      - predicate: "request.url_path != '/admin'"
EOF
```

> **Note:** It may take a couple of minutes for the RateLimitPolicy to be applied depending on your cluster.

<br/>

Verify the rate limiting works by sending requests in a loop.

Up to 5 successful (`200 OK`) requests every 10 seconds to `GET /cars`, then `429 Too Many Requests`:

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: api.toystore.com' http://$GATEWAY_URL/cars  | grep -E --color "\b(429)\b|$"; sleep 1; done
```

Unlimited successful (`200 OK`) to `GET /admin`:

```sh
while :; do curl --write-out '%{http_code}\n' --silent --output /dev/null -H 'Host: api.toystore.com' http://$GATEWAY_URL/admin | grep -E --color "\b(429)\b|$"; sleep 1; done
```

Check wasm plugin

```sh
kubectl get -n gateway-system wasmplugin/kuadrant-kuadrant-ingressgateway -o yaml
```
```yaml
apiVersion: extensions.istio.io/v1alpha1
kind: WasmPlugin
metadata:
  creationTimestamp: "2024-11-06T14:13:41Z"
  generation: 2
  labels:
    kuadrant.io/managed: "true"
  name: kuadrant-kuadrant-ingressgateway
  namespace: gateway-system
  ownerReferences:
  - apiVersion: gateway.networking.k8s.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Gateway
    name: kuadrant-ingressgateway
    uid: d0f27e1d-2cb8-4dc3-b1f6-94204d70aaef
  resourceVersion: "6002"
  uid: 3e506fd8-0517-4ab0-9307-c822ce3b0032
spec:
  phase: STATS
  pluginConfig:
    actionSets:
    - actions:
      - data:
        - expression:
            key: limit.create_toy__fb82c51a
            value: "1"
        predicates:
        - request.url_path != '/admin'
        scope: default/toystore
        service: ratelimit-service
      name: 3333a094e3493a905c35f12ffe3fa372e02a38c9700389a532f722e00d28d898
      routeRuleConditions:
        hostnames:
        - api.toystore.com
        predicates:
        - request.url_path.startsWith('/admin')
    - actions:
      - data:
        - expression:
            key: limit.create_toy__fb82c51a
            value: "1"
        predicates:
        - request.url_path != '/admin'
        scope: default/toystore
        service: ratelimit-service
      name: 5d3152cd9b75ba0b71f4c631dac258c905748ba6cce17a9b887b8bebe43a77f0
      routeRuleConditions:
        hostnames:
        - api.toystore.com
        predicates:
        - request.method == 'GET'
        - request.url_path.startsWith('/dolls')
    - actions:
      - data:
        - expression:
            key: limit.create_toy__fb82c51a
            value: "1"
        predicates:
        - request.url_path != '/admin'
        scope: default/toystore
        service: ratelimit-service
      name: 221b3f3322ef9c2a2a833343073f3d5997ad81354f67d8b4c652cbbd054d692c
      routeRuleConditions:
        hostnames:
        - api.toystore.com
        predicates:
        - request.method == 'GET'
        - request.url_path.startsWith('/cars')
    services:
      auth-service:
        endpoint: kuadrant-auth-service
        failureMode: deny
        type: auth
      ratelimit-service:
        endpoint: kuadrant-ratelimit-service
        failureMode: allow
        type: ratelimit
  targetRefs:
    kind: Gateway
    name: kuadrant-ingressgateway
  url: oci://quay.io/kuadrant/wasm-shim:latest
```


eguzki:
:warning: RateLimitPolicy API changing here @azgabur @jsmolar @trepel

alexsnaps (approved):

maleck13:
Nice work @eguzki love how the API is looking in those examples.

api/v1beta3/ratelimitpolicy_types.go
```diff
@@ -46,6 +41,9 @@ const (
 var (
 	RateLimitPolicyGroupKind  = schema.GroupKind{Group: SchemeGroupVersion.Group, Kind: "RateLimitPolicy"}
 	RateLimitPoliciesResource = SchemeGroupVersion.WithResource("ratelimitpolicies")
+	// Top level predicate rules key starting with # to prevent conflict with limit names
+	// TODO(eastizle): this coupling between limit names and rule IDs is a bad smell. Merging implementation should be enhanced.
```
guicassolato, line 45: We can ID the rules however we want, as long as each rule has a unique ID. That was the case of limits as the only kind of rule before this PR. It made sense IDing the rule after the unique limit name rather than coming up with something different and possibly more opaque for the rule ID.

Now, we want to differentiate these two kinds of rules (conditions and limits) within a RateLimitPolicy spec. Similarly to how we handle this in the AuthPolicy (where there are 10 different kinds of rules), we can define different prefixes for each kind. E.g. conditions could be `conditions#<condition-index>` and limits `limits#<limit-name>`.

Sure, the new rule ID limits prefix would ultimately show up in the Limitador CR and wasm config as `limit.limits_<limit-name>__<hash:rlp_ns/rlp_name>`. To avoida that, a trick that I think would work is setting the prefix in `Rules()`/`SetRules()` to `limit.` and changing [this line](https://github.com/Kuadrant/kuadrant-operator/blob/3b8e313d552090c52d8aadca95f6952f42a03192/controllers/ratelimit_workflow_helpers.go#L69) to `identifier := ""`.

guicassolato (approved):
I've chatted with @eguzki offline about https://github.com/Kuadrant/kuadrant-operator/pull/976#discussion_r1832407288.

We made a call together that the approach proposed by this PR is preferable than the behaviour implemented in the AuthPolicy. Merging conditions based on index position feels like dark magic and better avoiding it.

I will unblock this PR with the promise of the work on the AuthPolicy to copy the one here (i.e. the entire set of top-level `when` conditions always treated as one single rule) will follow this one ASAP.

PR #977, opened 2024-11-05
add on-cluster healtcheck test and prevent enforced false on not healhy probes

maksymvavilov:


pkg/library/kuadrant/apimachinery_status_conditions.go
```diff
@@ -78,3 +81,27 @@ func EnforcedCondition(policy Policy, err PolicyError, fully bool) *metav1.Condi
 
 	return cond
 }
+
+func HealthyCondition(policy Policy, err PolicyError) *metav1.Condition {
```
maleck13, line 85: not 100% sure if this should be here or not @mikenairn ?

tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -1244,6 +1244,102 @@ var _ = Describe("DNSPolicy controller", func() {
...
+							"Status":  Equal(metav1.ConditionTrue),
+							"Reason":  Equal(string(kuadrant.PolicyReasonEnforced)),
+							"Message": Equal("DNSPolicy has been successfully enforced"),
+						}),
+						MatchFields(IgnoreExtras, Fields{
+							"Type":   Equal(string(kuadrant.PolicyConditionSubResourcesHealthy)),
+							"Status": Equal(metav1.ConditionFalse),
+							"Reason": Equal(string(kuadrant.PolicyReasonUnknown)),
+							"Message": And(
+								ContainSubstring("DNSPolicy has encountered some issues: not all sub-resources of policy are healthy. Not healthy DNSRecords are:"),
```
maleck13, line 1333: should this message be 
` not all sub-resources of policy are passing the policy defined health check?` ? might be a bit more obvious then?

maleck13 (changes requested):
We need these changes against v1 now. Also not sure if we want the condition stuff under library will wait for @mikenairn  on that one.  Otherwise it looks good

pkg/library/kuadrant/apimachinery_status_conditions.go
mikenairn: Are you expecting this to ever be used by anything other than DNSPolicy? Seems unlikely, I'd probably move it all into the dns_workflow

tests/common/dnspolicy/dnspolicy_controller_test.go
maksymvavilov: changed the message

pkg/library/kuadrant/apimachinery_status_conditions.go
maksymvavilov: It is strange to have a condition defined in the workflow and not alongside other conditions :thinking:  don't have a strong opinion on this, however - moved

api/v1alpha1/dnspolicy_types.go
```diff
@@ -289,9 +289,9 @@ func (p *DNSPolicy) WithTargetGatewayListener(gwName string, lName string) *DNSP
 func (p *DNSPolicy) WithHealthCheckFor(endpoint string, port int, protocol string, failureThreshold int) *DNSPolicy {
 	return p.WithHealthCheck(dnsv1alpha1.HealthCheckSpec{
 		Path:             endpoint,
-		Port:             &port,
+		Port:             port,
```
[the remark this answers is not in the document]
maksymvavilov: rebased. It is v1 now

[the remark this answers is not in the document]
maleck13: no as in it needs go into the types in the v1 folde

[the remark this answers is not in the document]
maksymvavilov: I have a bad habit of commenting before the push. It is, I promise :laughing:

tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -1244,6 +1245,143 @@ var _ = Describe("DNSPolicy controller", func() {
...
+								"Type":    Equal(string(kuadrantdnsv1alpha1.ConditionTypeReady)),
+								"Status":  Equal(metav1.ConditionTrue),
+								"Reason":  Equal(string(kuadrantdnsv1alpha1.ConditionReasonProviderSuccess)),
+								"Message": Equal("Provider ensured the dns record"),
+							}),
+						),
+						Not(ContainElement(
+							MatchFields(IgnoreExtras, Fields{
+								"Type": Equal(string(kuadrantdnsv1alpha1.ConditionTypeHealthy)),
+							}),
```
maksymvavilov, line 1345: This one is redundant as well. If we are ready then we either don't have healthy cond or it is true. For this suite we never have `healthy - true` but I still would like to have a check for this

tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -1244,6 +1245,143 @@ var _ = Describe("DNSPolicy controller", func() {
...
+							"Message": Equal("DNSPolicy has been accepted"),
+						}),
+						MatchFields(IgnoreExtras, Fields{
+							"Type":   Equal(string(kuadrant.PolicyConditionEnforced)),
+							"Status": Equal(metav1.ConditionTrue),
+							"Reason": Equal(string(kuadrant.PolicyReasonEnforced)),
+							"Message": And(
+								ContainSubstring("DNSPolicy has been partially enforced. Not ready DNSRecords are:"),
+								ContainSubstring(recordName),
+								Not(ContainSubstring(wildcardRecordName))),
```
maksymvavilov, line 1368: If we are partially enforced with only two records it is sufficient to only check for the presence of one of the records. Leaving it here to be extra cautious.

PR #981, opened 2024-11-05
Support HTTPQueryParamMatch as CEL routeRuleConditions

alexsnaps:


pkg/wasm/utils.go
```diff
@@ -144,9 +144,14 @@ func PredicatesFromHTTPRouteMatch(match gatewayapiv1.HTTPRouteMatch) []string {
 		predicates = append(predicates, predicateFromHeader(headerMatch))
 	}
 
-	// TODO(eguzki): query params. Investigate integration with wasm regarding Envoy params
-	// from https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes
-	// request.query -> string : The query portion of the URL in the format of “name1=value1&name2=value2”.
+	// query param, only consider the first in case of repetition, as per spec
+	queryParams := make(map[gatewayapiv1.HTTPHeaderName]bool)
```
alexsnaps, line 148: Dunno if that's idiomatic Go...

maleck13: its fine, I think you would find most use `queryParams := map[gatewaypiv1.HTTPHeaderName]bool{}` but I don't think it is that important

maleck13 (commented):
code looks ok, but will leave it to someone more familar with the impact of this to approve

alexsnaps:
So I don't know how this makes math... but there is _no way_ this introduces a regression in test coverage as it tests _my additions_ as well _as code that wasn't test before_... 

## Go home codecov, You're drunk!

guicassolato (approved):
🏅

eguzki:
Depends on https://github.com/Kuadrant/wasm-shim/pull/126

PR #983, opened 2024-11-06
feat: tls section name support

KevFan:
# Description
Close: https://github.com/Kuadrant/kuadrant-operator/issues/985 https://github.com/Kuadrant/kuadrant-operator/issues/984

Adding section name support for TLS Policy:
* refactor to one cert per listener
* update to latest version of certmanager (v1.16.1)
* With one cert per listener, we will no longer allow listeners sharing secrets. Sharing secrets (i.e. referencing the same secret name by multiple certs) is **not** expected by from cert manager 1.14+ https://github.com/cert-manager/cert-manager/pull/6406 so I believe we should follow cert manager as unexpected behaviour. In this case, the Cert CR with the duplicate secret would have following status:
```yaml
status:
  conditions:
  - lastTransitionTime: "2024-11-05T16:05:15Z"
    message: Secret was issued for "istio-ingressgateway-http". If this message is
      not transient, you might have two conflicting Certificates pointing to the same
      secret.
    observedGeneration: 1
    reason: IncorrectCertificate
    status: "False"
    type: Ready
  notAfter: "2025-02-03T16:05:15Z"
  notBefore: "2024-11-05T16:05:15Z"
  renewalTime: "2025-01-04T16:05:15Z" 
```
The policy will be `Enforced: false` in this case notifying the user to use unique certificate references in the gateway listeners
* Re-add conflicting reason in status for policies that target an already targeted policy
# Verification
* Deploy
```
make local-setup
```
* Create Gateway with listeners referncing the same certificate ref
```
kind: Gateway             
metadata:         
  name: istio-ingressgateway
spec:                
  gatewayClassName: istio 
  listeners:                        
  - name: http   
    hostname: '*.toys.io'     
    port: 443
    protocol: HTTPS       
    tls:                            
      mode: Terminate      
      certificateRefs:
      - name: toys  
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All             
  - name: http2    
    hostname: '*.soys.io'
    port: 599            
    protocol: HTTPS
    tls:        
      mode: Terminate
      certificateRefs:    
      - name: toys
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All
---              
kind: ClusterIssuer
metadata:                
  name: selfsigned-issuer
spec:              
  selfSigned: {}
EOF
```
* Create policy targeting whole gateway
```
apiVersion: kuadrant.io/v1
kind: TLSPolicy   
metadata:           
  name: gw-tls    
spec:            
  targetRef:     
    kind: Gateway             
    name: istio-ingressgateway
  issuerRef:
    group: cert-manager.io
    kind: ClusterIssuer
    name: selfsigned-issuer
EOF
```
* Verify that 2 certs are created (1 per listener) but only 1 is ready
```
kubectl get certificates -A
```
* Verify the unready cert is due to `IncorrectCertificate`
```
kubectl get certificates istio-ingressgateway-http2 -n istio-system -o yaml | yq '.status'
```
* Verify TLSPolicy is `Enforced: false`
```
 kubectl get tlspolicy gw-tls -n istio-system -o yaml | yq '.status'
```
* Update gateway to have different cert refs the listeners
```
kind: Gateway             
metadata:         
  name: istio-ingressgateway
spec:                
  gatewayClassName: istio 
  listeners:                        
  - name: http   
    hostname: '*.toys.io'     
    port: 443
    protocol: HTTPS       
    tls:                            
      mode: Terminate      
      certificateRefs:
      - name: toys  
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All             
  - name: http2    
    hostname: '*.soys.io'
    port: 599            
    protocol: HTTPS
    tls:        
      mode: Terminate
      certificateRefs:    
      - name: soys
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All
EOF
```
* Verify that both certs are now ready and policy is enforced
```
kubectl get certificates -A
kubectl get tlspolicy gw-tls -n istio-system -o yaml | yq '.status'
```
* Create a second TLSPolicy with section name
```
apiVersion: kuadrant.io/v1
kind: TLSPolicy   
metadata:           
  name: section-tls
spec:            
  targetRef:     
    kind: Gateway             
    name: istio-ingressgateway
    sectionName: http2
  issuerRef:
    group: cert-manager.io
    kind: ClusterIssuer
    name: selfsigned-issuer
  commonName: soys.io
EOF
```
* Verify both policies are enforced and certs are still ready
```
kubectl get certificates -A
kubectl get tlspolicy -n istio-system -o wide
```
* Verify that section name cert is being enforced by checking the cert defines the correct common name
```
kubectl get certificates istio-ingressgateway-http2 -n istio-system -o yaml | yq '.spec'
```
* Delete gateway policy
```
kubectl delete tlspolicy gw-tls -n istio-system
```
* Verify section name policy is still enforced
```
kubectl get certificates -A
kubectl get tlspolicy -n istio-system -o wide
```

image


maleck13:
@KevFan I am assuming this is not going to make it by today?

KevFan:
Rebased with main to reolve conflict

Makefile
```diff
@@ -200,7 +200,7 @@ operator-sdk: $(OPERATOR_SDK) ## Download operator-sdk locally if necessary.
 
 CONTROLLER_GEN = $(PROJECT_PATH)/bin/controller-gen
 $(CONTROLLER_GEN):
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.5)
```
KevFan, line 203: Using latest released version of cert manager bumped k8s dependencies which required bumping controller-gen version to work correctly

controllers/limitador_status_to_rlp_gateway_event_handler.go
KevFan, line 1: Using newer k8s dependencies caused the interface implemented in this file to be out of date. Opted to delete this file instead since it's dead code now with the refactor to SOTW in general

controllers/tlspolicies_validator.go
```diff
@@ -99,6 +104,22 @@ func (t *TLSPoliciesValidator) isTargetRefsFound(topology *machinery.Topology, p
...
+			reflect.DeepEqual(conflictTLSPolicy.GetTargetRefs(), p.GetTargetRefs())
+	})
+
+	if ok {
+		return kuadrant.NewErrConflict(kuadrantv1.TLSPolicyGroupKind.Kind, client.ObjectKeyFromObject(conflictingP.(*kuadrantv1.TLSPolicy)).String(), errors.New("conflicting policy"))
+	}
+
+	return nil
+}
+
```
KevFan, line 122: @mikenairn This is what I'd done if you want to re-introduce `conflict` for DNSPolicy also 🤔

controllers/tlspolicies_validator.go
```diff
@@ -99,6 +104,22 @@ func (t *TLSPoliciesValidator) isTargetRefsFound(topology *machinery.Topology, p
 	return nil
 }
 
+// isConflict Validates if there's already an older policy with the same target ref
+func (t *TLSPoliciesValidator) isConflict(policies []machinery.Policy, p *kuadrantv1.TLSPolicy) error {
+	conflictingP, ok := lo.Find(policies, func(item machinery.Policy) bool {
+		conflictTLSPolicy := item.(*kuadrantv1.TLSPolicy)
+		return p != conflictTLSPolicy && conflictTLSPolicy.DeletionTimestamp == nil &&
```
maleck13, line 111: is this correct? Looks like we are checking two pointers? Would they have the same address?

controllers/tlspolicies_validator.go
```diff
@@ -99,6 +104,22 @@ func (t *TLSPoliciesValidator) isTargetRefsFound(topology *machinery.Topology, p
...
+			conflictTLSPolicy.CreationTimestamp.Before(&p.CreationTimestamp) &&
+			reflect.DeepEqual(conflictTLSPolicy.GetTargetRefs(), p.GetTargetRefs())
```
maleck13, line 113: ok this I expected. would it make sense above just to check the names are not equal as that couldn't happen right?

maleck13 (approved):
Verified
Sanity check on code
Minor question around a equality check
We may want to enhance the tests with a test to prove TLSPolicy will not overwrite an existing secret is did not create (this is to allow for a user adding their own certs not managed by cert-manager). However I tested this manually and looks good so happy for this to merge and the test be improved as a follow on,

@KevFan I will approve but feel free to ask for a different review given I am not as familar with the SOTW code

Makefile
mikenairn: hmm, maybe this could be done in it's own PR? One that only does this?

controllers/tlspolicies_validator.go
KevFan: Yes, we don't want the conflicting policy to be itself

controllers/tlspolicies_validator.go
KevFan: Hmm, I went with deep equal to also cover the case of same gateway but different section. This would be allowed. It's only when the target is exactly the same that we report that it's conflcted by another policy 🤔

Makefile
KevFan: Yeah, I can break it out if you want but it'll possibly prevent this PR from merging sooner, since it'll need a rebase on top once that's merged 🤔

controllers/tlspolicies_validator.go
maleck13: I think I wasn't clear. This is fine

mikenairn: I _think_ this is fine, but i was wondering if already accepted policies should be explicitly checked, and not just the timestamps. One policy per target should only ever be accepted, if one is already accepted for this target, all others should conflict regardless of when they were created.

Makefile
mikenairn: I guess if time is an issue we can go with it as is, but generally these things would be better done in there own PRs

PR #991, opened 2024-11-07
REFACTOR: Kuadtrant Status

Boomatang:
Move the old controller logic for reconciling the kuadrant status to the state of the world workflow.

This works the same as before with one main change that ensures status of the kuadrant CR is updated when resources are created out of order. For example if a user wanted to deploy limitador that was redis backed they can now create the the limitador resources a head of time.


controllers/state_of_the_world.go
```diff
@@ -50,15 +50,40 @@ var (
...
+// core, apps, coordination.k8s,io permissions
+//+kubebuilder:rbac:groups=core,resources=serviceaccounts;configmaps;services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=configmaps;leases,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups="",resources=leases,verbs=get;list;watch;create;update;patch;delete
+
+// maistra.io permissions
+// +kubebuilder:rbac:groups=maistra.io,resources=servicemeshcontrolplanes,verbs=get;list;watch;update;use;patch
+// +kubebuilder:rbac:groups=maistra.io,resources=servicemeshmembers,verbs=get;list;watch;create;update;delete;patch
```
guicassolato, line 82: These are not needed anymore.

```suggestion
```

guicassolato (approved):
Left one minor comment, but I'm happy to address it in #992 if we want to get this one merged first.

PR #995, opened 2024-11-08
fix: Add policy conflict to dnspolicy validation

mikenairn:
Updates the dnspolicies validator task to check for polices with conflicting target refs, only one policy can currently target a specific gateway or listener. Uses the same logic as TLS, first policy created is given preference over any created later.

Some other small changes to align DNSPolicy and TLSPolicy.

### DIFF
--- a/api/v1/tlspolicy_types.go
+++ b/api/v1/tlspolicy_types.go
@@ -175,6 +175,7 @@ func (p *TLSPolicy) Kind() string {
 	return TLSPolicyGroupKind.Kind
 }
 
+// Deprecated: Use GetTargetRefs instead
 func (p *TLSPolicy) GetTargetRef() gatewayapiv1alpha2.LocalPolicyTargetReference {
 	return p.Spec.TargetRef.LocalPolicyTargetReference
 }
--- a/controllers/dnspolicies_validator.go
+++ b/controllers/dnspolicies_validator.go
@@ -2,11 +2,14 @@ package controllers
 
 import (
 	"context"
+	"errors"
+	"reflect"
 	"sync"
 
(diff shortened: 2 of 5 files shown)


api/v1/tlspolicy_types.go
```diff
@@ -191,7 +191,7 @@ func (p *TLSPolicy) GetRulesHostnames() []string {
 	return make([]string, 0)
 }
 
-// DEPRECATED: Use GetTargetRefs instead
+// Deprecated: kuadrant.Policy.
```
mikenairn, line 194: "Deprecated:" seems to be the correct way to this in Go https://go.dev/wiki/Deprecated, linter will fail if this is used anywhere.

We can likely remove all these soon anyway, if not already.

controllers/dnspolicies_validator.go
```diff
@@ -34,19 +37,54 @@ func (r *DNSPoliciesValidator) Subscription() controller.Subscription {
...
+			policy.CreationTimestamp.Before(&p.CreationTimestamp) &&
+			reflect.DeepEqual(policy.GetTargetRefs(), p.GetTargetRefs())
+	})
+
+	if ok {
+		return kuadrant.NewErrConflict(kuadrantv1.DNSPolicyGroupKind.Kind, client.ObjectKeyFromObject(conflictingP.(*kuadrantv1.DNSPolicy)).String(), errors.New("conflicting policy"))
+	}
+
+	return nil
+}
```
mikenairn, line 90: These are pretty much copied from the TLSPolicy, we should probably look at making generic functions that can be used by both.

maleck13 (approved):

PR #1000, opened 2024-11-11
fix: incorrect kuadrant status if limitador/authorino is not found

KevFan:
# Description
* fix: incorrect kuadrant status if limitador/authorino is not found
  * Kuadrant status would incorrectly report as `Ready` if limitador or authorino was not found within the topology
* test: kuadrant status on bare k8s 
  * Kuadrant status test on bare k8s was incorrect and should report `GatewayAPIProviderNotFound` instead

# Verification
* Passing integration tests should be enough

Otherwise to manually verify:
* watch Kuadrant CR
```
watch -n 0.5 "kubectl get kuadrant -A -o yaml"
```
* Deploy Kuadrant CR
```
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```
* Verify Kuadrant CR status eventually becomes ready as Limitador or Authorino recomes ready

### DIFF
--- a/controllers/kuadrant_status_updater.go
+++ b/controllers/kuadrant_status_updater.go
@@ -8,8 +8,6 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
-	authorinov1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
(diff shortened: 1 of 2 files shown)


Boomatang (approved):
Changes look good to me. Its hard to see the error for when the resources are not in the topology.

PR #992, opened 2024-11-08
remove unused code

guicassolato:
**This PR should not affect behaviour.**

Here are a few steps to help test a few cases, to ensure everything works exactly as before:

Setup the environment:

```sh
make local-setup
```
```sh
make envoy-gateway-install

# Restart the Kuadrant Operator so it can acknowledge the presence of Envoy Gateway
kubectl rollout restart deployment/kuadrant-operator-controller-manager -n kuadrant-system
```
```sh
kubectl apply -f -<<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: envoygateway
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
EOF

kubectl apply -n gateway-system -f -<<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: kuadrant-envoygateway
spec:
  gatewayClassName: envoygateway
  listeners:
  - name: https
    hostname: "*.eg.apps.io"
    protocol: HTTPS
    port: 443
    tls:
      mode: Terminate
      certificateRefs:
      - name: kuadrant-envoygateway-cert
        kind: Secret
    allowedRoutes:
      namespaces:
        from: All
EOF
```

Configure TLS on the Envoy Gateway-provided gateway:

```sh
kubectl apply -n gateway-system -f -<<EOF
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: kuadrant-ca
spec:
  selfSigned: {}
---
apiVersion: kuadrant.io/v1alpha1
kind: TLSPolicy
metadata:
  name: kuadrant-envoygateway-tls
  namespace: gateway-system
spec:
  targetRef:
    group: gateway.networking.k8s.io
    kind: Gateway
    name: kuadrant-envoygateway
  issuerRef:
    group: cert-manager.io
    kind: ClusterIssuer
    name: kuadrant-ca
EOF
```

Deploy an application:

```sh
kubectl apply -f examples/toystore/toystore.yaml

kubectl apply -f - <<EOF
apiVersion: gateway.networking.k8s.io/v1
kind: HTTPRoute
metadata:
  name: toystore
spec:
  parentRefs:
  - name: kuadrant-ingressgateway
    namespace: gateway-system
  - name: kuadrant-envoygateway
    namespace: gateway-system
  rules:
  - backendRefs:
    - name: toystore
      port: 80
    matches:
    - method: GET
  - backendRefs:
[...]

PR #1003, opened 2024-11-11
🪛 permissions to update kuadrant finalizer 🔙

eguzki:
### What

On Openshift, when the Kuadrant CR is created, limitador and authorino are not deployed. On trying to create limitador and authorino resources, the operator logs the error:

```
{"level":"error","ts":"2024-11-11T16:57:46Z","logger":"kuadrant-operator.AuthorinoReconciler","msg":"failed to create authorino resource","status":"error","error":"authorinos.operator.authorino.kuadrant.io \"authorino\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>","stacktrace":"github.com/kuadrant/kuadrant-operator/controllers.(*AuthorinoReconciler).Reconcile\n\t/remote-source/app/controllers/authorino_reconciler.go:112\ngithub.com/kuadrant/policy-machinery/controller.Subscription.Reconcile\n\t/remote-source/deps/gomod/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/subscriber.go:34\ngithub.com/kuadrant/policy-machinery/controller.(*Workflow).Run.func1\n\t/remote-source/deps/gomod/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/workflow.go:42\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/remote-source/deps/gomod/pkg/mod/golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78"}
```

Important bits:

```
authorinos.operator.authorino.kuadrant.io \"authorino\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on
```

Openshift has admission controller that rejects adding ownerefs with `blockOwnerDeletion: true` when the controller does not have permission to add finalizer on the owner object. 

In the PR https://github.com/Kuadrant/kuadrant-operator/pull/992 a 🐛 🪲 was introduced, creating a regression, which removed permissions to add finalizers to Kuadrant CR's. Thus, the operator cannot create resources with ownerrefs to Kuadrant CR like the Limtador CR and Authorino CR managed by the operator. 

This PR adds the permissions to update finalizers on the Kuadrant CR.

Additionally, the Kuadrant CR status it does not reported the error (the creation of limitador and authorino CR's fail) and reports "READY". It is left as TODO to fix the error reporting, catching error on creating limitador and authorino resources and report back on the status.

### DIFF
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -105,7 +105,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2024-11-08T16:43:37Z"
+    createdAt: "2024-11-11T17:42:02Z"
     description: A Kubernetes Operator to manage the lifecycle of the Kuadrant system
     operators.operatorframework.io/builder: operator-sdk-v1.32.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
@@ -328,6 +328,7 @@ spec:
           resources:
           - authpolicies/finalizers
           - dnspolicies/finalizers
+          - kuadrants/finalizers
           - ratelimitpolicies/finalizers
           - tlspolicies/finalizers
           verbs:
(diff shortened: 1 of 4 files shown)


guicassolato:
Nice catch, @eguzki. Thank you!

Maybe `make generate manifests bundle helm-build`

eguzki:
I always forget helm charts :facepalm:

eguzki:
Happy that I added a test to blame me

Issue #1004, opened 2024-11-12
Policy Status not updated when kuadrant is remove


Boomatang:
# What
When the case where kuadrant is up and running with policies created. This applies more to Ratelimiting and Auth as they require kuadrant to manage resources.  

If the kuadrant CR is removed all the resources are cleaned up but the status of the policies are never updated to reflect that the policy is still not enforced.

For a better user experiences the policies status should be updated.

# Steps to reproduce
Run the guide [RateLimitPolicy for Platform Engineers](https://docs.kuadrant.io/dev/kuadrant-operator/doc/user-guides/gateway-rl-for-cluster-operators/), but don't do the last clean up step. 

Check the status of the ratelimit policy 
```
kubectl get ratelimitpolicy gw-rlp -n gateway-system -o yaml | yq
```

Remove the kuadrant CR.
```
kubectl delete kuadrant kuadrant -n default
```

If you rerun the curl commands from the guide you will see the routes are no longer protect, which is expected.
However checking the ratelimit policy status again show that it is `Accepted` and `Enforced`

PR #1012, opened 2024-11-12
FIX: Context being cancelled and better Policy status

Boomatang:
The GetKuadrantFromTopology was somehow cancelling the context of functions that called it.

This improves the policy status of ratelimit policies when there is no kuadrant CR on cluster, or it was removed.

Closes: #1004 
Closes: #1011 

## Reproduce error
Install kuadrant from main. 
Create the a working ratelimit example. [RateLimitPolicy for Platform Engineers](https://docs.kuadrant.io/dev/kuadrant-operator/doc/user-guides/gateway-rl-for-cluster-operators/)

Delete the kuadrant CR. Then check the state of the wasm plugin. It would be expected that this would be removed. However it will not. Restarting the kuadrant deployment will fix the issue and clean the resources up. Also the status in the ratelimiting policy will say it is still accepted.

Install the operator from this branch and get the cluster back into the state before the deletion of the kuadrant CR. This time when the Kuadrant CR is remove the wasm plugin will be removed and the enforced status will be false.

### DIFF
--- a/controllers/ratelimit_policy_status_updater.go
+++ b/controllers/ratelimit_policy_status_updater.go
@@ -113,6 +113,10 @@ func (r *RateLimitPolicyStatusUpdater) UpdateStatus(ctx context.Context, _ []con
 }
 
 func (r *RateLimitPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.RateLimitPolicy, topology *machinery.Topology, state *sync.Map) *metav1.Condition {
+	kObj := GetKuadrantFromTopology(topology)
+	if kObj == nil {
+		return kuadrant.EnforcedCondition(policy, kuadrant.NewErrSystemResource("kuadrant"), false)
+	}
 	policyKind := kuadrantv1.RateLimitPolicyGroupKind.Kind
 
 	effectivePolicies, ok := state.Load(StateEffectiveRateLimitPolicies)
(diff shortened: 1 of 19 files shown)


controllers/ratelimit_policy_status_updater.go
```diff
@@ -113,6 +113,10 @@ func (r *RateLimitPolicyStatusUpdater) UpdateStatus(ctx context.Context, _ []con
 }
 
 func (r *RateLimitPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.RateLimitPolicy, topology *machinery.Topology, state *sync.Map) *metav1.Condition {
+	kObj := GetKuadrantFromTopology(topology)
+	if kObj == nil {
+		return kuadrant.EnforcedCondition(policy, kuadrant.NewErrSystemResource("kuadrant"), false)
+	}
```
KevFan, line 119: Should this also be done for `AuthPolicy` ? The current condition for `AuthPolicy` without Kuadrant installed is:

```yaml
  status:
    conditions:
    - lastTransitionTime: "2024-11-13T08:50:35Z"
      message: AuthPolicy has been accepted
      reason: Accepted
      status: "True"
      type: Accepted
    - lastTransitionTime: "2024-11-13T08:50:54Z"
      message: 'AuthPolicy has encountered some issues: missing auth effective policies
        stored in the reconciliation state'
      reason: Unknown
      status: "False"
      type: Enforced
```

Boomatang: Yes this should have being done.

KevFan (approved):
Verified! Looks good to me 👍

PR #1017, opened 2024-11-13
Case insensitive header matching and safe lookup

alexsnaps:
As discussed this makes for safe Map lookups in CEL. 
Also, as per the spec if I read it correctly, header names should be case insensitive... now they are


maleck13 (approved):
changes look good to me
/lgtm

Issue #990, opened 2024-11-07
Crash after adding listener to Gateway


azgabur:
I noticed a crash loop of kuadrant-operator pod after adding new listener to Gateway before adding it to associated HTTProute.

Logs:
```
{"level":"error","ts":"2024-11-07T17:24:14Z","logger":"kuadrant-operator.IstioExtensionReconciler.buildWasmConfigs","msg":"failed to build wasm policies for path","pathID":"istio|kuadrant/gw-azgabur--85d|kuadrant/gw-azgabur--85d#api-second|kuadrant/route-azgabur--efth|kuadrant/route-azgabur--efth#rule-1","error":"invalid path: http route does not belong to the listener","stacktrace":"github.com/kuadrant/kuadrant-operator/controllers.(*IstioExtensionReconciler).buildWasmConfigs\n\t/workspace/controllers/istio_extension_reconciler.go:198\ngithub.com/kuadrant/kuadrant-operator/controllers.(*IstioExtensionReconciler).Reconcile\n\t/workspace/controllers/istio_extension_reconciler.go:58\ngithub.com/kuadrant/policy-machinery/controller.Subscription.Reconcile\n\t/go/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/subscriber.go:34\ngithub.com/kuadrant/policy-machinery/controller.(*Workflow).Run.func1\n\t/go/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/workflow.go:42\ngolang.org/x/sy...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1ab5ffb]
goroutine 4568 [running]:
github.com/kuadrant/kuadrant-operator/controllers.(*AuthConfigsReconciler).Reconcile(0xc0007901e8, {0x23ad5b0, 0xc001da44b0}, {0xc0030623f0?, 0x1?, 0x2?}, 0xc002f907a0, {0x0?, 0x0?}, 0xc002f907c0)
/workspace/controllers/authconfigs_reconciler.go:73 +0x57b
[...]

PR #1018, opened 2024-11-13
fix: nil pointer for invalid path for auth

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/990

Fixes a `nil` pointer due to attempting to create  an`AuthConfig` for an invalid path.

In the described scenario in the issue, `HTTPRouteRule` is returned `nil` due to invalid path as the `HTTPRoute` only matches only one of the Listener hostname from:
https://github.com/Kuadrant/kuadrant-operator/blob/d5f0184506ee4a71b397b8686add3e348e26d87f/pkg/policymachinery/utils.go#L62-L87

# Verification
* Passing integration tests should be enough as it now includes an integration test for this

If you want to verify manually:
* Checkout main
* Create env
```sh
make local-env-setup
```
* Create resources
```
kubectl apply -f - <<EOF
kind: Kuadrant  
metadata:
  name: kuadrant
spec: {} 
---
kind: Gateway
metadata:
  name: gw
spec:
  gatewayClassName: istio
  listeners:
    - allowedRoutes:
        namespaces:
          from: All
      hostname: prefix1.example.com
      name: api
      port: 80
      protocol: HTTP
    - allowedRoutes:
        namespaces:
          from: All
      hostname: prefix2.example.com
      name: api-second
      port: 81
      protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: route
spec:
  hostnames:
    - prefix1.example.com
  parentRefs:
      kind: Gateway
      name: gw
  rules:
    - backendRefs:
        - group: ''
          kind: Service
          name: httpbin
          namespace: kuadrant
          port: 8080
          weight: 1
      matches:
        - path:
            type: PathPrefix
            value: /
---                               
apiVersion: kuadrant.io/v1
kind: AuthPolicy
metadata:
  name: authz
spec:
  targetRef:
    kind: HTTPRoute
    name: route
  defaults:
    strategy: merge
    rules:
      authentication:
        "api-key-authn":
          apiKey:
            selector:
              matchLabels:
                app: toystore
          credentials:
            authorizationHeader:
              prefix: APIKEY
EOF
```
* Run operator
```
make run
```
* Verify nil pointer crash
* Checkout this branch
* Run operator again
```
make run
```
* Verify no nil pointer exception
* Verify `AuthPolicy` is enforced
```
kubectl get authpolicy authz -o yaml | yq '.status'
```

### DIFF
--- a/controllers/auth_policy_status_updater.go
+++ b/controllers/auth_policy_status_updater.go
@@ -2,11 +2,13 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"sync"
 
 	envoygatewayv1alpha1 "github.com/envoyproxy/gateway/api/v1alpha1"
+	"github.com/go-logr/logr"
 	authorinooperatorv1beta1 "github.com/kuadrant/authorino-operator/api/v1beta1"
 	authorinov1beta3 "github.com/kuadrant/authorino/api/v1beta3"
 	"github.com/kuadrant/policy-machinery/controller"
@@ -86,7 +88,7 @@ func (r *AuthPolicyStatusUpdater) UpdateStatus(ctx context.Context, _ []controll
 		if !accepted {
 			meta.RemoveStatusCondition(&newStatus.Conditions, string(kuadrant.PolicyConditionEnforced))
 		} else {
-			enforcedCond := r.enforcedCondition(policy, topology, state)
+			enforcedCond := r.enforcedCondition(policy, topology, state, logger)
 			meta.SetStatusCondition(&newStatus.Conditions, *enforcedCond)
 		}
 
@@ -114,7 +116,7 @@ func (r *AuthPolicyStatusUpdater) UpdateStatus(ctx context.Context, _ []controll
 	return nil
 }
 
-func (r *AuthPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.AuthPolicy, topology *machinery.Topology, state *sync.Map) *metav1.Condition {
+func (r *AuthPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.AuthPolicy, topology *machinery.Topology, state *sync.Map, logger logr.Logger) *metav1.Condition {
 	kObj := GetKuadrantFromTopology(topology)
 	if kObj == nil {
 		return kuadrant.EnforcedCondition(policy, kuadrant.NewErrSystemResource("kuadrant"), false)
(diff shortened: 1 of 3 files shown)


controllers/auth_policy_status_updater.go
```diff
@@ -143,7 +143,10 @@ func (r *AuthPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.AuthPolic
 		if len(kuadrantv1.PoliciesInPath(effectivePolicy.Path, func(p machinery.Policy) bool { return p.GetLocator() == policy.GetLocator() })) == 0 {
 			continue
 		}
-		gatewayClass, gateway, listener, httpRoute, httpRouteRule, _ := kuadrantpolicymachinery.ObjectsInRequestPath(effectivePolicy.Path)
+		gatewayClass, gateway, listener, httpRoute, httpRouteRule, err := kuadrantpolicymachinery.ObjectsInRequestPath(effectivePolicy.Path)
+		if err != nil {
+			continue
```
maleck13, line 148: I worry a little when I see this. Why do we want to swallow the error here?

controllers/authconfigs_reconciler.go
```diff
@@ -71,7 +71,11 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	modifiedAuthConfigs := []string{}
 
 	for pathID, effectivePolicy := range effectivePoliciesMap {
-		_, _, _, httpRoute, httpRouteRule, _ := kuadrantpolicymachinery.ObjectsInRequestPath(effectivePolicy.Path)
+		_, _, _, httpRoute, httpRouteRule, err := kuadrantpolicymachinery.ObjectsInRequestPath(effectivePolicy.Path)
+		if err != nil {
+			logger.Error(err, "failed to reconcile authconfig objects", "path", effectivePolicy.Path)
```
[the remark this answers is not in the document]
maleck13: Could we return an error type that would inform the controller of the type of failure? Failing to reconcile seems like a big deal to me? Is it safe to continue at that point or should the controller just bail and try again?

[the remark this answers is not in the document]
maleck13: ok so looking at https://github.com/KevFan/kuadrant-operator/blob/5d97f9aaf562314bf05e87a6fdbcb81a4328ae15/pkg/policymachinery/utils.go#L58 we return nil but also return an error explaining no HTTPRoute `NewErrInvalidPath("index 3 is not a HTTPRoute")` although I hate the message :) what does index 3 even mean
So does it mean effectively there is nothing to reconcile and so we can continue rather than we failed to reconcile?

controllers/auth_policy_status_updater.go
maleck13: as this only currently ever returns an invalid path error I think we should log them at a debug level to help understand what might be happening in logs same for below

maleck13 (approved):
Spoke with @KevFan and I understand better how this is working and why the error wasn't logged or returned. We agreed on logging the error at info level for debugging purposes as the only error that can be returned here is an InvalidPathError which it would be normal to see with a missing HTTPRoute.

controllers/auth_policy_status_updater.go
KevFan: Updated with debug info log

maleck13 (approved):
/lgtm

PR #1020, opened 2024-11-13
Have only auth accessing actions happen after auth

### LINKED ISSUE #375: Pre-auth rate limit
RFE: add support for pre-auth RL.

The main use case behind this request is to support RL by IP as a measure to mitigate DDoS attacks, which should run before checking authentication, and ideally without having to deploy another gateway only for this.

alexsnaps:
fixes #375


alexsnaps:
This should break all integration tests that assert on that order... Waiting for that to happen and then fix

eguzki (approved):

PR #1027, opened 2024-11-15
Set more lenient timeouts for protection services

adam-cattermole:
The [default connection timeout](https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L172-L176) used by the wasm-shim when the value is omitted is 20ms. This could be too short when set to use external redis for rate limiting, or with complex auth policies. By default I have increased this to:

* 200ms for auth (this matches the value in [envoy ext auth](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_authz/v3/ext_authz.proto))
* 100ms for rate limiting

This is also not exposed by the API in anyway, and any changes to the `WasmPlugin` will be overridden by the operator, so it's not possible to set these values. For now until we decide how to expose this in the API I've added two env vars:
* `AUTH_SERVICE_TIMEOUT`
* `RATELIMIT_SERVICE_TIMEOUT`

These let us override the defaults if a user requires.

### DIFF
--- a/pkg/wasm/types.go
+++ b/pkg/wasm/types.go
@@ -43,7 +43,7 @@ func (c *Config) EqualTo(other *Config) bool {
 	}
 
 	for key, service := range c.Services {
-		if otherService, ok := other.Services[key]; !ok || service != otherService {
+		if otherService, ok := other.Services[key]; !ok || !service.EqualTo(otherService) {
 			return false
 		}
 	}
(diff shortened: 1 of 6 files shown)


eguzki (commented):
LGTM

I have listed 7 env vars so far
* `LOG_LEVEL`
* `LOG_MODE`
* `OPERATOR_NAMESPACE`
* `RELATED_IMAGE_WASMSHIM`
* `RELATED_IMAGE_CONSOLEPLUGIN`
* `AUTH_SERVICE_TIMEOUT`
* `RATELIMIT_SERVICE_TIMEOUT`

We should document those env vars for devs and operator release managers. I was thinking on documenting about how to tune those parameters when releasing using OLM (using the subscription object). As well as when using helm charts. I know currently  it is not possible to tune those env vars when deploying using helm charts, but this is also something to be done. Leaving here for the record, not asking for this PR.

eguzki (approved):

eguzki:
> My stance on environment variables is that they are "backdoors" into the system. That's certainly what I had envisioned for these two new ones. So, _if we really document them_, I think this should be done in such a way that it's clear that this is either "experimental" or whatever, but clearly _not_ supported and/or guaranteed to work moving forward.
> 
> Discussing this issue with @adam-cattermole on Friday, the intent is to go with somewhat sensible defaults and see how make this a properly supported (and documented) feature moving forward... There is more to this than it may seem and this deserves some real thinking. A default per service is definitely _not_ sensible from a Kuadrant perspective.

Fair enough.

Let's not document those env vars yet.

PR #1029, opened 2024-11-18
Fix references to renamed dashboard configmaps

david-martin:
Knock on from renaming of configmaps in https://github.com/Kuadrant/kuadrant-operator/commit/526b0c9b8f6bb623159a7d132b21e87b4230fee6#diff-1f10072a23b74e4257ddbad38fb10afc7c0839c839d0af0b550a40bdbd3c22ed
(configmaps were renamed to have to the same name as the json files)

**NOTE:** This affects local/kubernetes setup of dashboards in grafana if the cmds at https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/README.md are followed. It does *NOT* affect openshift setup as that's a completely different guide.

### DIFF
--- a/config/observability/grafana/grafana_deployment_patch.yaml
+++ b/config/observability/grafana/grafana_deployment_patch.yaml
@@ -22,17 +22,17 @@
 - op: add
   path: /spec/template/spec/volumes/-
   value:
-    name: grafana-controller-resources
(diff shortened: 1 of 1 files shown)


philbrookes (approved):
fixed the issue for me locally.

Issue #866, opened 2024-09-20
Expose Wasm `failureMode` setting


eguzki:
Wasm `failureMode` https://github.com/Kuadrant/wasm-shim/blob/main/src/configuration.rs#L490-L495 defines what to do when the GRPC request with limitador returns an error. To be clear, when limitador returns `OVER_LIMIT` in the RateLimitResponse struct of the GRPC response, this is not considered as an error. 

So when there is an error returned by  the GRPC access:

* `failureMode`: *allow* => The http request resumes the filter chain. So the request is allowed to be routed to destination. In other words, when limitador is down, the traffic is still being routed and not rate limited. 
*  `failureMode`: *deny* => Envoy returns `500 Internal Server Error` to the downstream client.This is the value hardcoded in latest 0.10.0 and earlier versions.  In other words, when limitador is down, the traffic is not being routed and upstream backend does not get traffic from envoy (from that listener). 

The ask is to expose this behavior at the kuadrant level. 

Option 1: In the kuadrant spec
Option 2: An annotation in the kuadrant CR
Option 3: Per policy (so some attribute at the policy level). Related work:  Wasm refactor https://github.com/Kuadrant/wasm-shim/pull/87

Note that if ExtAuth is being handled by wasm, we might want to expose two attributes independently: one for rate limiting, another one for auth. So, for example, one could allow traffic when rate limiting is unavailable and disable traffic routing when external auth is not available (more critical in that sense).

eguzki:
> This one is a bit like [#1027](https://github.com/Kuadrant/kuadrant-operator/pull/1027) ... And probably even directly relates to it. Wondering if we shouldn't address _for now_ just as we did with the timeout, i.e. have it _feasible_ to control it, but through some backdoor mechanism. And consider it to be included in the API properly alongside the timeouts... wdyt?

That's Option 4: operator's env var(s)
LGTM

PR #1030, opened 2024-11-18
Setting FailureMode from ENV

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/866

Reading the values from ENV, with the following defaults:

Auth => Deny
RateLimit => Allow

### DIFF
--- a/pkg/wasm/utils.go
+++ b/pkg/wasm/utils.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/kuadrant/policy-machinery/machinery"
@@ -30,8 +31,31 @@ func AuthServiceTimeout() string {
 	return env.GetString("AUTH_SERVICE_TIMEOUT", "200ms")
 }
 
+func AuthServiceFailureMode() FailureModeType {
+	return parseFailureModeValue("AUTH_SERVICE_FAILURE_MODE", FailureModeAllow)
+}
+
 func RatelimitServiceTimeout() string {
 	return env.GetString("RATELIMIT_SERVICE_TIMEOUT", "100ms")
+}
+
+func RatelimitServiceFailureMode() FailureModeType {
+	return parseFailureModeValue("RATELIMIT_SERVICE_FAILURE_MODE", FailureModeDeny)
+}
+
+func parseFailureModeValue(envVarName string, defaultValue FailureModeType) FailureModeType {
+	value := os.Getenv(envVarName)
+	if value == "" {
+		return defaultValue
+	}
+
+	switch value {
+	case string(FailureModeAllow), string(FailureModeDeny):
+		return FailureModeType(value)
+	default:
+		fmt.Printf("Warning: Invalid value '%s' for %s. Using default value '%s'.\n", value, envVarName, defaultValue)
+		return defaultValue
+	}
 }
 
 func ExtensionName(gatewayName string) string {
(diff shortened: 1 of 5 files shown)


pkg/wasm/types.go
```diff
@@ -83,7 +83,7 @@ const (
 	AuthServiceType      ServiceType = "auth"
 )
 
-// +kubebuilder:validation:Enum:=deny;allow
+// FailureModeType +kubebuilder:validation:Enum:=deny;allow
```
eguzki, line 86: This is breaking kubebuilder validation, isn't it?

eguzki: Actually, that kubebuilder annotation is useless, isn't it?

didierofrivia: It was there before, I just added the name for documentation completion on IDEs

eguzki: I think it is useless...  Anyway, let's not mix doc and annotations

```suggestion
// FailureModeType
+kubebuilder:validation:Enum:=deny;allow
```

pkg/wasm/utils.go
```diff
@@ -30,10 +31,33 @@ func AuthServiceTimeout() string {
...
+	value := os.Getenv(envVarName)
+	if value == "" {
+		return defaultValue
+	}
+
+	switch value {
+	case string(FailureModeAllow), string(FailureModeDeny):
+		return FailureModeType(value)
+	default:
+		fmt.Printf("Warning: Invalid value '%s' for %s. Using default value '%s'.\n", value, envVarName, defaultValue)
```
adam-cattermole, line 56: Seems slightly odd to use standard printing here instead of using a logger but I guess that's because it's not easy to get the logger here?

tests/envoygateway/extension_reconciler_test.go
```diff
@@ -171,13 +171,13 @@ var _ = Describe("wasm controller", func() {
 					wasm.AuthServiceName: {
 						Type:        wasm.AuthServiceType,
 						Endpoint:    kuadrant.KuadrantAuthClusterName,
-						FailureMode: wasm.FailureModeDeny,
+						FailureMode: ptr.To(wasm.AuthServiceFailureMode()),
```
adam-cattermole, line 174: I'm sure you've already spotted but the `Timeout` is a pointer whereas the `FailureMode` is not which is why the tests are failing
```suggestion
						FailureMode: wasm.AuthServiceFailureMode(),
```

adam-cattermole (commented):
Primarily just tests need fixing but otherwise looks good

pkg/wasm/utils.go
didierofrivia: Yup, I wanted to avoid having to create an instance of the logger just for this... but I could if needed.

didierofrivia: I could also remove the message xD

adam-cattermole: I think it's nice to have the message at least to know that if you set it with a typo there's a reason why it's not being set.. but it's also going to look weird with interleaved `Printf` messages with the operator logs every time there's an event that triggers modification to the plugin that goes through this code. I don't want to hold up this PR though not a big issue

adam-cattermole (approved):
lgtm, verified and all appears to be working

pkg/wasm/utils.go
eguzki: I would panic, it is wrong to have an invalid value

eguzki: But happy to merge this as it is as well

didierofrivia: Take a look at https://github.com/Kuadrant/kuadrant-operator/pull/1030/commits/4bffd5ed4778f30f712b031b006b0c2c5bb6705d, an alternative would be creating a different instance just for the logging that setting or leave it to `fmt.Println`

eguzki (approved):

PR #1041, opened 2024-11-21
add small tweak to install

maleck13:

rh-pre-commit.version: 2.2.0


eguzki (approved):
LGTM

That being said, I think we should promote helm charts as the preferred (and recommended) way to install upstream version of kuadrant, either in bare k8s or Openshift.

maleck13:
@eguzki sure but this is a nice option also as you can just use kubectl -k

PR #1042, opened 2024-11-21
guides: auth authzd/spicedb

KevFan:
# Description
Part of https://github.com/Kuadrant/docs.kuadrant.io/issues/25

Introduces an `AuthPolicy` guide describing how an authpolicy can be integrated with spicedb for permission requests.
This can serve as an equivalent / replacement guide to https://github.com/Kuadrant/authorino/blob/main/docs/user-guides/authzed.md guide on https://docs.kuadrant.io/0.11.0/authorino/docs/user-guides/authzed/

# Verification
* Running through the guide should work as expected

### DIFF
--- a/doc/user-guides/auth/authzed.md
+++ b/doc/user-guides/auth/authzed.md
@@ -2,43 +2,24 @@
 
 This guide explains how to configure permission requests for a Google Zanzibar-based [Authzed/SpiceDB](https://authzed.com) instance using gRPC.
 
-## Requisites
+## Prerequisites
 
-- [Docker](https://docker.io)
+You have installed Kuadrant in a [kubernetes](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-kubernetes/) or [OpenShift](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-openshift/) cluster.
 
-## Run the guide ① → ⑥
+## Run the guide ① → ⑦
 
-### ①  Setup
+### ① Deploy Toy Store application
 
-Clone the repo:
+Deploy a simple HTTP application service that echoes back the request data: 
 
 ```sh
-git clone git@github.com:Kuadrant/kuadrant-operator.git && cd kuadrant-operator
+kubectl apply -f https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/refs/heads/main/examples/toystore/toystore.yaml
 ```
 
-Run the following command to create a local Kubernetes cluster with [Kind](https://kind.sigs.k8s.io/), install & deploy Kuadrant:
+### ② Expose the Application
+Create an `HTTPRoute` to expose a `/posts` path for `GET` and `POST` requests to the application:
 
 ```sh
-make local-setup
-```
-
-Request an instance of Kuadrant in the `kuadrant-system` namespace:
-
-```sh
-kubectl -n kuadrant-system apply -f - <<EOF
-apiVersion: kuadrant.io/v1beta1
-kind: Kuadrant
-metadata:
-  name: kuadrant
-spec: {}
-EOF
-```
-
-### ② Deploy the Talker API
-
-```sh
-kubectl apply -f examples/toystore/toystore.yaml
-
 kubectl apply -f - <<EOF
 apiVersion: gateway.networking.k8s.io/v1
 kind: HTTPRoute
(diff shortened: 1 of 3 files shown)


Boomatang (dismissed):
This looks good, guide works as expected. I didn't spot any wrong with the grammar, but I am poor at that.

doc/user-guides/auth/authzed.md
```diff
@@ -0,0 +1,345 @@
+# AuthPolicy Integration with Authzed/SpiceDB
+
+This guide explains how to configure permission requests for a Google Zanzibar-based [Authzed/SpiceDB](https://authzed.com) instance using gRPC.
+
+## Prerequisites
+
+You have installed Kuadrant in a [kubernetes](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-kubernetes/) or [OpenShift](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-openshift/) cluster.
+
+## Run the guide ① → ⑦
```
R-Lawton, line 9: Can we remove the Unicode to follow the same style as the rest of the guides

Boomatang:
@R-Lawton When your here, I am reviewing a different up date and a I remember seeing something about rolling the installation guides into one. Sadly, I can seem to find right now. I think you were the author. There is reference to those guides here. What impact would that have?

doc/user-guides/auth/authzed.md
```diff
@@ -0,0 +1,345 @@
+# AuthPolicy Integration with Authzed/SpiceDB
+
+This guide explains how to configure permission requests for a Google Zanzibar-based [Authzed/SpiceDB](https://authzed.com) instance using gRPC.
+
+## Prerequisites
+
+You have installed Kuadrant in a [kubernetes](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-kubernetes/) or [OpenShift](https://docs.kuadrant.io/latest/kuadrant-operator/doc/install/install-openshift/) cluster.
```
R-Lawton, line 7: these links have now been updated so they wont work when the new version of the docs is released on the website https://docs.kuadrant.io/dev/kuadrant-operator/doc/install/install-openshift/

R-Lawton (changes requested):
Few things i noticed

doc/user-guides/auth/authzed.md
KevFan: Seems most other guides were updated to https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/ratelimiting/authenticated-rl-for-app-developers.md#setup-the-environment,  is this what we want here instead now or do you have an example? 🤔

doc/user-guides/auth/authzed.md
```diff
@@ -0,0 +1,345 @@
...
+      path:
+        type: PathPrefix
+        value: "/posts"
+    backendRefs:
+    - name: toystore
+      port: 80
+EOF
+```
+
+Export the gateway hostname and port for testing:
```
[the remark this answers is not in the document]
KevFan: Cool, I can include setting up the gateway as part of the guide 👍

doc/user-guides/auth/authzed.md
KevFan: The setup looks to still use the `make` targets 🤷

R-Lawton: We have issue to remove the make target altogether for existing docs [Remove/replace using make target as the method of installation for docs. #1093](https://github.com/Kuadrant/kuadrant-operator/issues/1093) that's going to be worked on soon. In this case its the links you have above, have changed in a new version of the website. We now link off to helm. The install-kubernetes doc no longer exists. The openshift doc has been updated but the link is same i just linked to it as a way to get to the dev view of the install guides on the website.

KevFan: Cool 👍 I'll link the setup to either use Helm or if using Openshift 🤔 
A question on this though, is are we expecting this to be cherry picked to the v1 branch? 🤔 
If so, then I guess we should not hard link to the docs website but use relative paths and also not to the dev docs? Or would be cherry pick and have changes to work with the current v1 setup guides ?

R-Lawton: thats a good question that im not sure i know we are working towards doing a v1.1 release later on in q1 but that seems late enough to make these change live. Yeah relative sounds good to me what you could also do is have similar to what we have for prereq in the SCP doc and just say "Kubernetes cluster with Kuadrant operator installed.". Oh yeah we don't want to use the dev view

R-Lawton: Ok found more info, if the doc is directly needed by 1.0 like a fix etc then cherry pick if not wait for the next release it will be seen in the dev view

averevki (approved):

maksymvavilov:

PR #1043, opened 2024-11-21
final couple of tweaks

maleck13:

rh-pre-commit.version: 2.2.0


maleck13:
@eguzki made  couple more small changes and added the catalog image to our olm/catalogsource.yaml when we do a release

eguzki (approved):
LGTM

this config/deploy/olm/kustomization.yaml is only being used [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/make/catalog.mk#L70-L72) and it is already doing that.

so, if you check out the tagged version, and then, from [Deploy kuadrant operator using OLM](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/development.md#deploy-kuadrant-operator-using-olm). you execute `make deploy-catalog` (without overriding `CATALOG_IMG`), the system would pick the tagged one and deploy the right version.

Anyway, this is harmless

Issue #928, opened 2024-10-10
[E2E] kuadrant based DNS policy scale test


philbrookes:
**What**

Allow setup and installation of Kuadrant with multiple DNS operators that are installed namespace scoped
Define a set of gateways and Policies to cause Kuadrant via DNS operator to manage DNS with N different instances across y clusters

**Done**

- [ ] Automate setup and installation of Kuadrant + (n) DNS operators namespace scoped. This should be able to be done across at least 2 clusters
- [ ] Define a set of policies and Gateway resources and add a test to show they get into the correct state 
- [ ] Document how to run and execute this installation and tests

mikenairn:
Moving this to review since i want to get feedback on the proposed approach of using kubeburner for this scale test task https://github.com/Kuadrant/testsuite/pull/615. It doesn't answer all the questions asked in this issue, in particular multi cluster testing, but it is a better approach than rolling our own IMO.

trepel:
@mikenairn I just want to point out to throttling in Azure (not sure about other cloud providers). For instance if creating too many DNS records it might fail with 429 error code, see
https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-subscription-service-limits#public-dns-resource-record-operations

maleck13:
@mikenairn is this done now?

trepel:
I am running scale test (kubeburner) time to time (for each minor version) but only control plane scale test (https://github.com/Kuadrant/testsuite/blob/main/scale_test/control-plane.md). I am not running DNS scale test (https://github.com/Kuadrant/testsuite/blob/main/scale_test/namespaced-dns-operator-deployments.md)

PR #1053, opened 2024-11-27
DNSPolicy scale testing

mikenairn:
DNSPolicy scale testing

part of #928 

_Installation_

Adds a new installation overlay `namespaced-dns-operator` that when applied to a cluster will install multiple instances of the dns operator in different namespaces. Each instance is configured to only watch for dns resources (DNSRecords etc..) in that single namespace.

Templates the kuadrant and dns operator versions in config/install overlays. These should be kept in sync with the branch/tag they are for to allow more easily installing a particular version or latest. These will be automatically updated as part of the release creation.

_Testing_

Adds a `hack/dnspolicy-scale-testing/test.sh` script to run multi cluster testing scenarios manually on kind clusters
Adds basic dns testing config for aws,azure,gcp and inmem providers


Makefile
```diff
@@ -300,10 +300,13 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 dependencies-manifests: export AUTHORINO_OPERATOR_GITREF := $(AUTHORINO_OPERATOR_GITREF)
 dependencies-manifests: export LIMITADOR_OPERATOR_GITREF := $(LIMITADOR_OPERATOR_GITREF)
 dependencies-manifests: export DNS_OPERATOR_GITREF := $(DNS_OPERATOR_GITREF)
+dependencies-manifests: export IMAGE_TAG := $(IMAGE_TAG)
 dependencies-manifests: ## Update kuadrant dependencies manifests.
 	$(call patch-config,config/dependencies/authorino/kustomization.template.yaml,config/dependencies/authorino/kustomization.yaml)
 	$(call patch-config,config/dependencies/limitador/kustomization.template.yaml,config/dependencies/limitador/kustomization.yaml)
 	$(call patch-config,config/dependencies/dns/kustomization.template.yaml,config/dependencies/dns/kustomization.yaml)
+	$(call patch-config,config/install/namespaced-dns-operator/dns-operator/kustomization.template.yaml,config/install/namespaced-dns-operator/dns-operator/kustomization.yaml)
+	$(call patch-config,config/install/standard/kuadrant-version.template.yaml,config/install/standard/kuadrant-version.yaml)
```
mikenairn, line 309: @maleck13 I added your existing kuadrant install to this task. When we create a release, this will update the kuadrant version to the correct value based on the "VERSION" variable currently set. In the case of main, that will be "latest", but for a semantic version it will be the expected tag e.g. "v1.0.0-rc6"

config/install/standard/kuadrant-version.yaml
```diff
@@ -3,4 +3,4 @@ kind: CatalogSource
 metadata:
   name: kuadrant-operator-catalog
 spec:
-  image: quay.io/kuadrant/kuadrant-operator-catalog:v1.0.0-rc6 #change this to the version you want to install
+  image: quay.io/kuadrant/kuadrant-operator-catalog:latest #change this to the version you want to install
```
mikenairn, line 6: This is being changed to "latest" because it's on main, this will now get updated to the correct image tag as part of the release process see https://github.com/Kuadrant/kuadrant-operator/pull/1053/files#r1860933307

Installing different versions should be done by checking out the correct version before applying:
```
git checkout v1.0.0-rc10
kubectl apply -k config/install/standard
```
or, ideally using a remote ref:

```
kubectl apply -k github.com/kuadrant/kuadrant-operator/config/install/standard?ref=v1.0.0-rc10
```

maleck13: I don't think the remote will work as the env files are not present

mikenairn: I'm probably going to close this PR in favour of using kube burner for building the same scale testing scenarios. That said, this image ref on main should probably be pointing to something other than a fixed tag that never gets updated as part of the release. If you are referring to the example apply command i.e. `kubectl apply -k github.com/kuadrant/kuadrant-operator/config/install/standard?ref=v1.0.0-rc10`, it was just an example of how it could be used if you started updating this image correctly as part of releases.

mikenairn:
Closing for now, although some of the changes here might come back in another PR.

Issue #1049, opened 2024-11-26
500s in gateway logs after enabling sidecars in authorino & limitador, and STRICT PeerAuthentication


david-martin:
After doing a `make local-setup` locally, then following https://docs.kuadrant.io/dev/kuadrant-operator/doc/user-guides/secure-protect-connect/, I got to the point of a working AuthPolicy & RatelimitPolicy.

I then enabled mtls (via sidecar injection) in authorino and limitador in an attempt to have encrypted traffic from envoy to/from the kuadrant components.
Here's the cmds I ran.

```bash
kubectl -n kuadrant-system patch deployment authorino \
  -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject":"true"}}}}}'


kubectl -n kuadrant-system patch deployment limitador-limitador \
  -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject":"true"}}}}}'

kubectl apply -f - <<EOF
apiVersion: security.istio.io/v1
kind: PeerAuthentication
metadata:
  name: default
  namespace: kuadrant-system
spec:
  mtls:
    mode: STRICT
EOF
```

However, after enabling this, all responses are now 500s.
I bumped up the log level to debug in the gateway, using this cmd:

```bash
kubectl exec $(kubectl -n kuadrant-system get po -l istio.io/gateway-name=api-gateway -o name
pod/api-gateway-istio-649dc76fc-nklds) -n kuadrant-system -- curl -X POST http://localhost:15000/logging?level=debug
```

Here's a snippet of logs
https://gist.github.com/david-martin/90104685603ecab49aad45cf595d2c92

Here's the Pod spec & config_dump of the gateway
https://gist.github.com/david-martin/43fdbbede84df3b156609081a9b6387f

Discussion on slack at https://kubernetes.slack.com/archives/C05J0D0V525/p1732621388340329

At this time it's unclear if the issue is with the envoy filter configuration that kuadrant manages, or something else.

david-martin:
I was able to get it working locally with some patching of the envoyfilters, though this isn't a viable solution as the filters are maintained by the kuadrant-operator.
I added this to both the`kuadrant-auth-service` &  `kuadrant-ratelimit-service` envoyfilters:

```bash
          transport_socket:
            name: envoy.transport_sockets.tls
            typed_config:
[...]

PR #1054, opened 2024-11-27
Add new guide for configuring mtls between gateway & authorino/limitador

david-martin:
This will be a new document in the 'Installation' section of https://docs.kuadrant.io
This documents the manual steps that were the outcome of an investigation detailed in https://github.com/Kuadrant/kuadrant-operator/issues/1049

A separate [RFC](https://github.com/Kuadrant/architecture/pull/110) is proposed for the possibility of a more automated way to handle enabling mTLS. 
This document is seen as a stopgap, should users desire to enable mTLS in the meantime.

### DIFF
--- a/doc/install/mtls-configuration.md
+++ b/doc/install/mtls-configuration.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This guide includes manual steps to enable mTLS between the gateway and Kuadrant components.
+This guide includes manual steps to enable mTLS between an Istio provided gateway and the Kuadrant components.
 If you use an AuthPolicy or RateLimitPolicy, there will be communication between the gateway and the respective Kuadrant components at request time. This communication happens between the Wasm plugin in Envoy proxy, and Authorino or Limitador.
 At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add this capability as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
 
--- a/doc/install/mtls-configuration.md
+++ b/doc/install/mtls-configuration.md
@@ -4,7 +4,7 @@
 
 This guide includes manual steps to enable mTLS between an Istio provided gateway and the Kuadrant components.
 If you use an AuthPolicy or RateLimitPolicy, there will be communication between the gateway and the respective Kuadrant components at request time. This communication happens between the Wasm plugin in Envoy proxy, and Authorino or Limitador.
-At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add this capability as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
+At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add mTLS capabilities as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
 
 !!! note
 
--- a/doc/install/mtls-configuration.md
+++ b/doc/install/mtls-configuration.md
@@ -1,0 +1,148 @@
+# Configure mTLS between the Gateway and Kuadrant components
+
+## Overview
+
+This guide includes manual steps to enable mTLS between the gateway and Kuadrant components.
+If you use an AuthPolicy or RateLimitPolicy, there will be communication between the gateway and the respective Kuadrant components at request time. This communication happens between the Wasm plugin in Envoy proxy, and Authorino or Limitador.
+At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add this capability as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
+
+!!! note
+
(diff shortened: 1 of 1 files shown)


maleck13 (commented):
This looks good. I am going to try it before I approve. 👍

eguzki:
Very nice! :lock: :100: :closed_lock_with_key:

doc/install/mtls-configuration.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+You should see the number of containers in either pod increase from 1 to 2, as the `istio-proxy` is added to the pods. This change will force all traffic to those pods to go through the proxy. However, mTLS is not enabled yet.
+
+### Envoy Filter
+
+The next step enables mTLS for traffic originating in the gateway (where the Wasm plugin executes), going to the Kuadrant components.
+This requires modifying the EnvoyFilters directly.
+Here is an example `transport_socket` configuration to add to both the `kuadrant-auth-api-gateway` and `kuadrant-ratelimiting-api-gateway` EnvoyFilter resources.
+This should be added below `spec.configPatches[].patch.value`.
```
guicassolato, line 38: I wonder if "below" is unequivocal enough. To be more precise, I think we want to say the value of `spec.configPatches[].patch.value` (which is a YAML object) to be extended with this snippet below. I.e. for `transport_socket` to be added as a sibling of `name`, `type`, ..., `load_assignment`, is that right?

doc/install/mtls-configuration.md
```diff
@@ -0,0 +1,150 @@
...
+metadata:
+  name: default
+  namespace: kuadrant-system
+spec:
+  mtls:
+    mode: STRICT
+EOF
+```
+
+If you prefer to only able mTLS for a specific component, you can modify just the EnvoyFilter and Deployment for that component.
```
guicassolato, line 88:
```suggestion
If you prefer to only enable mTLS for a specific component, you can modify just the EnvoyFilter and Deployment for that component.
```

doc/install/mtls-configuration.md
```diff
@@ -0,0 +1,150 @@
...
+kubectl -n kuadrant-system patch deployment limitador-limitador \
+  -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject":"true"}}}}}'
+```
+
+You should see the number of containers in either pod increase from 1 to 2, as the `istio-proxy` is added to the pods. This change will force all traffic to those pods to go through the proxy. However, mTLS is not enabled yet.
+
+### Envoy Filter
+
+The next step enables mTLS for traffic originating in the gateway (where the Wasm plugin executes), going to the Kuadrant components.
+This requires modifying the EnvoyFilters directly.
```
guicassolato, line 36: I'm actually surprised this works. In theory, any events related to the following CRs should reconcile the EnvoyFilter back to no-TLS form:
- Kuadrant
- GatewayClass
- Gateway
- HTTPRoute
- AuthPolicy
- EnvoyFilter

doc/install/mtls-configuration.md
david-martin: Yes. I'll udpate this to be clearer

doc/install/mtls-configuration.md
david-martin: I'll add a caveat that it may get undone, and they should allow for that by having some job or gitops thing reconcile it back.
I'll also link to the RFC where they can engage on it potential being a feature of kuadrant.

david-martin:
@guicassolato 
I've updated the words to make it clearer what has to be updated in EnvoyFilters.
I've also added a note about the changes potentially being undone, and a link to the rfc for people to engage if they'd like this as a feature.

@maleck13 
I've added some words to the overview note about ensuring service mesh is enabled, and a reference to Istio CNI (including a link to the sail operator docs).
There is also mention of having at least 1 AuthPolicy or RateLimitPolicy in the pre-reqs section (so an EnvoyFilter exists)
This has now been tested and verified on a local kind cluster and Openshift.
I know you had issues with the istio-proxy container coming up locally when setting up things a little differently that I had. Is that something you want to revisit or are you happy to approve the doc as it is?

doc/install/mtls-configuration.md
```diff
@@ -0,0 +1,148 @@
+# Configure mTLS between the Gateway and Kuadrant components
+
+## Overview
+
+This guide includes manual steps to enable mTLS between the gateway and Kuadrant components.
```
maleck13, line 5:
```suggestion
This guide includes manual steps to enable mTLS between an Istio provided gateway and the Kuadrant components.
```

doc/install/mtls-configuration.md
```diff
@@ -0,0 +1,148 @@
...
+If you use an AuthPolicy or RateLimitPolicy, there will be communication between the gateway and the respective Kuadrant components at request time. This communication happens between the Wasm plugin in Envoy proxy, and Authorino or Limitador.
+At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add this capability as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
```
maleck13, line 7:
```suggestion
At the time of writing there is [an RFC](https://github.com/Kuadrant/architecture/pull/110) discussing how to add mTLS capabilities as a feature of the Kuadrant operator. If you are interested in having that feature or influencing how it is delivered, please engage on that pull request.
```

maleck13 (approved):

PR #1070, opened 2024-12-05
Some small fixes for the Install Kuadrant and Sail via OLM guide.

jasonmadigan:
Small tweaks to the kustomize install:

- Add a sample AWS env-vars config
- Fixed an old/outdated path in a comment in one of the kustomize files


config/install/standard/kuadrant-version.yaml
```diff
@@ -3,4 +3,4 @@ kind: CatalogSource
 metadata:
   name: kuadrant-operator-catalog
 spec:
-  image: quay.io/kuadrant/kuadrant-operator-catalog:latest #change this to the version you want to install
+  image: quay.io/kuadrant/kuadrant-operator-catalog:release-v1.0.0 # change this to the version you want to install
```
jasonmadigan, line 6: thinking about it, makes sense as `latest` on main. One less thing to worry about at release time.

jasonmadigan:
cc: @maleck13

maleck13 (approved):

PR #1074, opened 2024-12-09
DNSPolicy: display healthy cond on get

maksymvavilov:
when listing policies display healthy condition 
image


maksymvavilov:
image
this is how it looks

eguzki:
Just minor comment about the PR title. Make explicit it is about the DNS policy, as it is shown as part of the release notes.

philbrookes (approved):

PR #1075, opened 2024-12-10
Improved Kuadrant installation via Olm  and kustomize

maleck13:
**What** This updates the install with more details and hopefully clearer instructions. 

The guide is focused around OLM and kustomize. Having this doesn't mean we can't also have a helm or kustomize without an OLM requirement

**Verfication**

Follow the guide without needing to clone the repo, You should be able to use a basic kind cluster.

**Note**
You will have to set the version in the configure/kustomization.yaml resource paths from `v1.0.1` to `olm-installation` as some of the changes on this branch are not yet on the v1.0.1 branch

Still to do

- [x] Add examples for setting up  requests and limits 
- [x] Add example for configuring topology constraints


config/install/README.md
```diff
@@ -1,101 +1,326 @@
...
+# NAME      REVISIONS   READY   IN USE   ACTIVE REVISION   VERSION   AGE
+# default   1           1       1        Healthy           v1.23.0   3d22h
+```
+
+
+
+At this point Kuadrant is installed and ready to be used as is Istio as the gateway provider. This means AuthPolicy and RateLimitPolicy can now be configured and used to protect any Gateways you create. 
+
+
+## Configure a DNS provider credentials and a Certificate Issuer
```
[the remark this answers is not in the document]
maleck13: not needed. Is optional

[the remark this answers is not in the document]
R-Lawton: who do we want the audience of this doc to be? what are we trying to show or teach people? 

My thinking is if we document different parts of Kuadrant as optional even tho they technically might be, its going to be a pain for us in the long run with keeping things up to date, making docs follow the same installation etc, are we just overcomplicating things?  

Should we just show how to install kuadrant as a whole and if they don't want to use a certain aspect then so be it. We could show in the architecture what components come out of the box and what are optional. 

side note: I think we need to have a convo as a team on how we want to portray the project in our docs as we are currently writing in all different styles and ways. I'm currently writing a doc (doception 😂 ) to suggest a standard and have a kick off point for conversation. 

This install guide is a bit of a double whammy because we also need a convo on what installation methods do we want to stand behind. we have many different types and different docs on how to do it and duplicate docs that go old and obsolete

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
+```
+
+> Note: you may see ` no matching resources found ` if the deployments are not yet present.
 
+Once OLM has finished installing the operators (this can take several minutes). You should see the following in the kuadrant-system namespace:
 
+```bash
+kubectl get deployments -n kuadrant-system
+
+## Output
```
david-martin, line 94:
```suggestion
## Output (kuadrant-console-plugin deployment only installed on OpenShift)
```

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
+kubectl get kuadrant kuadrant -n kuadrant-system -o=wide
 
-Create the `$CLOUD_PROVIDER-credentials.env file` in the cloud provider directory `config/install/configure/$CLOUD_PROVIDER.` e.g. `aws-credentials.env` in the `config/install/configure/aws` directory. Apply the configuration for the desired cloud provider. Example AWS
+# NAME       STATUS   AGE
+# kuadrant   Ready    109s
 
 ```
-kubectl apply -k config/install/configure/aws
+
+You should see the condition with type `Ready` with a message of `kuadrant is ready`.
```
david-martin, line 152:
```suggestion
You should see the condition with type `Ready`. 
```
message not show in wide view

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
+4. [[OpenShift Specific] Setup Observability ](#set-up-observability-openshift-only)
+
 
 ## Prerequisites  
-- Clone the[ Kuadrant-operator](https://github.com/Kuadrant/kuadrant-operator) repo
-- OLM (operator lifecycle manager)
-- cert-manager 
+- OCP or K8s cluster and CLI available.
+- OLM installed [operator lifecycle manager releases](https://github.com/operator-framework/operator-lifecycle-manager/releases)
+- (Optional) Gateway Provider Installed: By default this guide will install the [Sail Operator](https://github.com/istio-ecosystem/sail-operator) that will configure and install an Istio installation. Kuadrant is intended to work with [Istio](https://istio.io) or [Envoy Gateway](https://gateway.envoyproxy.io/) as a gateway provider before you can make use of Kuadrant one of these providers should be installed.
```
david-martin, line 23:
```suggestion
```
To avoid confusion, I'd remove this optional dependency because the guide shows how to install kuadrant with istio/sail. The 2 are entwined in the steps below, such that if the user already has a gateway provider installed that they want to use, there is extra work to decouple the istio/sail pieces.

Alternatively, I'm wondering if we should decouple those steps for the user anyways in this guide?
I think it's reasonable that a user could arrive at the Kuadrant installlation guide with or without a gateway provider already running in their cluster.

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
 
-This will configure Kuadrant and Sail to install their components, set the credentials needed to access DNS zones in the cloud provider, and create a Let's Encrypt cluster issuer configured to use DNS-based validation.
 
-### Validate
 
-Validate Kuadrant is ready via the kuadrant resource status condition
+At this point Kuadrant is installed and ready to be used as is Istio as the gateway provider. This means AuthPolicy and RateLimitPolicy can now be configured and used to protect any Gateways you create. 
+
+
+## Configure DNS and TLS integration
```
david-martin, line 170:
```suggestion
## (Optional) Configure DNS and TLS integration
```

To match the "(Optional) cert-manager" and "(Optional) Access to AWS" prereqs at the top, 
otherwise CRDs don't exist and no credentials.

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
-- cert-manager 
+- OCP or K8s cluster and CLI available.
+- OLM installed [operator lifecycle manager releases](https://github.com/operator-framework/operator-lifecycle-manager/releases)
+- (Optional) Gateway Provider Installed: By default this guide will install the [Sail Operator](https://github.com/istio-ecosystem/sail-operator) that will configure and install an Istio installation. Kuadrant is intended to work with [Istio](https://istio.io) or [Envoy Gateway](https://gateway.envoyproxy.io/) as a gateway provider before you can make use of Kuadrant one of these providers should be installed.  
+- (Optional) cert-manager:
   - [cert-manager Operator for Red Hat OpenShift](https://docs.openshift.com/container-platform/4.16/security/cert_manager_operator/cert-manager-operator-install.html)
   - [installing cert-manager via OperatorHub](https://cert-manager.io/docs/installation/operator-lifecycle-manager/)
-- AWS, Azure or GCP with DNS capabilities. (Optional)
-- Accessible Redis instance, for persistent storage for your rate limit counters. (Optional)
+- (Optional) Access to AWS, Azure or GCP with DNS services.
```
david-martin, line 27: For this guide, it's just AWS that's involved. (see comment below about marking that step as optional).
I'd remove mention of Azure and GCP here, or alternatively add info about how you'd use them instead of AWS in the relevant sections below.

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
+
+generatorOptions:
+  disableNameSuffixHash: true
+  labels:
+    app.kubernetes.io/part-of: kuadrant
+    app.kubernetes.io/managed-by: kustomize
+
+secretGenerator:
+  - name: aws-provider-credentials
+    namespace: cert-manager # assumes cert-manager namespace exists.
```
david-martin, line 214:
```suggestion
    namespace: cert-manager # change this to the namespace where cert-manager is running.
```
In my case, following olm install of cert-manager, it in the `operators` namespace

config/install/README.md
```diff
@@ -1,101 +1,591 @@
...
+            configSecretRef:
+              name: redis-credentials
+
+```
+
+
+Re-Apply the configuration to setup the new secret and configuration:
+
+```bash
+kubectl apply -k $KUADRANT_DIR/configure/
```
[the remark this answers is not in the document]
david-martin: Probably related to changes below https://github.com/Kuadrant/kuadrant-operator/pull/1075/files#diff-36411ee7f394bd28aaa769d3b33d193cde1e03f8f7bb3fd569b2efb1b4a509f0R4
would need to land that on the v1.0.1 branch

david-martin (changes requested):
I got to end of the doc, including the redis, TopologyConstraints, PodDisruptionBudget and Resource Limits pieces.
Comments and suggestions above.

config/install/README.md
maleck13: Yeah but perhaps we can add that as an follow up. I think there is a lot of value in the guide as is for now?

config/install/README.md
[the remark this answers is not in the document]
maleck13: Yes I want to cherry pick these changes to the v1.0.1 branch. In the validation steps I mentioned you would have to use the ref=olm-installation for the resources .

config/install/README.md
maleck13: I will add a link to the examples for other providers

david-martin (approved):

PR #1088, opened 2024-12-18
Combining the two scp into the one document 

R-Lawton:
# WHAT
Removing the old out dated K8s walkthrough in favour of having the one guide that can work for both types of installation


david-martin (approved):
Verified the former openshift SCP doc works on vanilla k8s (local kind cluster) as well.
This makes sense to have 1 guide.

PR #1090, opened 2024-12-20
update location of where kustomize looks for deploying example gateway

laurafitzgerald:
### What
$subject so the user doesn't need to have a local copy of the repo to run the quickstart
also removing the need to have kustomize as a dependency for running the quickstart

### Verification
Temporarily uninstall kustomize from your system.
Run `curl "https://raw.githubusercontent.com/laurafitzgerald/kuadrant-operator/refs/heads/remove-local-repo-depedency/hack/quickstart-setup.sh"| bash` 
Reinstall kustomize

### DIFF
--- a/hack/quickstart-setup.sh
+++ b/hack/quickstart-setup.sh
@@ -73,7 +73,6 @@ export TMP_DIR=$SCRIPT_DIR/tmp/mgc
 export CONTAINER_RUNTIME_BIN=$(containerRuntime)
 export KIND_BIN=kind
 export HELM_BIN=helm
-export KUSTOMIZE_BIN=$(dockerBinCmd "kustomize")
 export SUBNET_OFFSET=1
 export HUB=1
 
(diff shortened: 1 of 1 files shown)


hack/quickstart-setup.sh
```diff
@@ -462,7 +462,7 @@ success "Kuadrant sample configuration deployed."
 
 # Deploy gateway
 info "Deploying example gateway ..."
-kustomize build config/dependencies/istio/gateway | kubectl apply -f -
+kustomize build ${KUADRANT_REPO_RAW}/config/dependencies/istio/gateway | kubectl apply -f -
```
Boomatang, line 465: I haven't tried this but I am always wondering why write this in this fashion, why use a pipe?
``` sh
kubectl apply -k ${KUADRANT_REPO_RAW}/config/dependencie/istio/gateway
```

laurafitzgerald: I believe it's because it's the output of kustomize build that's needed rather than the raw directory.

Boomatang: This I understand, the raw directory will fail if the `-f` flag is used. In my example it is the ` -k / --kustomize ` flag that is used. This requires a kustomization directory and under the hood will run kustomize build. 

If you try using a file with the `-k` it will fail, like how `-f` fails with a directory.

laurafitzgerald: ah okay. I think that's what @david-martin is referring to also here -> https://github.com/Kuadrant/docs.kuadrant.io/pull/161#discussion_r1894023740 so I will make those updates as part of this pr. 👍

Boomatang: ack

laurafitzgerald: These changes are ready for review now.

Boomatang (approved):
This all looks good to me and works.

laurafitzgerald:
@Boomatang I notice the option to merge is Rebase and Merge rather than Merge pull request, I'm blocked from doing the latter. Is it correct to do the former?

Boomatang:
For me I have being using the `Rebase and merge`.

Issue #1101, opened 2025-01-08
Certificates not managed by Kuadrant getting deleted


roivaz:
**Describe the bug**

kuadrant-operator seems to delete any cert-manager Certificate resource that it's not directly managed by Kuadrant. Even without a TLSPolicy in place, each time the kuadrant-operator-controller-manager pod starts, it wipes out any Certificate resources in any namespace.

Logs (debug mode) show the Certificate deletion event, but not the actual deletion operation itself:

```
{"level":"info","ts":"2025-01-07T17:21:34Z","logger":"kuadrant-operator.event logger","msg":"new event","type":"delete","kind":"Certificate","namespace":"3scale-saas","name":"wildcard-stg-saas-3sca-net-cert"}
{"level":"info","ts":"2025-01-07T17:21:34Z","logger":"kuadrant-operator.event logger","msg":"new event","type":"delete","kind":"Certificate","namespace":"3scale-saas","name":"stg-saas-wildcard-apicast-production"}
{"level":"info","ts":"2025-01-07T17:21:34Z","logger":"kuadrant-operator.event logger","msg":"new event","type":"delete","kind":"Certificate","namespace":"3scale-saas","name":"stg-saas-wildcard-apicast-staging"}
{"level":"info","ts":"2025-01-07T17:21:34Z","logger":"kuadrant-operator.event logger","msg":"new event","type":"delete","kind":"Certificate","namespace":"openshift-ingress","name":"default-ingress-wildcard-certificate"}
```

**To Reproduce**
1. Have cert-manager installed in the cluster, with some Certificate resources in place
2. Install kuadrant-operator, a Kuadrant resource and a Gateway
3. Check that as soon as the kuadrant-operator starts, it picks up the Certificates that already exist in the cluster. Then, those Certificates are deleted, I'm assuming that by the kuadrant-operator-controller-manager pod itself. Behavior can be forces by just deleting the pod and letting it start a again.

**Expected behavior**
Any Certificate resource that is not directly managed by Kuadrant should be left intact.

**Additional context**

Pasting the topology here. Note that this is still a work in progress and there are not Kuadrant policies yet, nor there are any HTTPRoutes attached to the Gateway yet.

```
digraph  {

        n5[label="Authorino\nkuadrant-system/authorino",shape="ellipse"];
        n7[label="ClusterIssuer\nletsencrypt-production",shape="ellipse"];
[...]

PR #1102, opened 2025-01-08
fix: only register Certificates managed by Kuadrant to the topology

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/1101

Register only `Certificates` managed by Kuadrant to the topology by filtering by `app: kuadrant-operator` label.
This also removes the need for the custom predicate which now aligns more with other similar resources.

Previous to this, the topology will register all `Certificates` on the cluster and Kuadrant Operator will delete these as unexpected certs.

# Verification
* Checkout main
* Create local cluster
```
make local-env-setup
```
* Create un-managed certificate
```
kind: Certificate
metadata:
  name: tests
  namespace: istio-system
spec:
  dnsNames:
  - '*.test.io'
  issuerRef:
    group: cert-manager.io
    kind: Issuer
    name: selfsigned-issuer
  secretName: tests
  usages:
  - digital signature
  - key encipherment
EOF
```
* Watch for `Certificates`
```
kubectl get certificates -A
```
* Start kuadrant operator
```
make run
```
* Verify bug where unmanaged certificate was deleted
* Stop kuadrant operator
* Checkout this branch
* Re-create the un-managed certificate from the previous step
* Run kuadrant operator
* Verifiy unmanaged certificate **is not** deleted

### DIFF
--- a/controllers/state_of_the_world.go
+++ b/controllers/state_of_the_world.go
@@ -25,7 +25,6 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/utils/env"
 	ctrlruntime "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	ctrlruntimepredicate "sigs.k8s.io/controller-runtime/pkg/predicate"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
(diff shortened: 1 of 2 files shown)


controllers/state_of_the_world.go
```diff
@@ -364,29 +363,12 @@ func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
...
-					return isCertificateOwnedByTLSPolicy(e.ObjectNew)
-				},
-				DeleteFunc: func(e event.TypedDeleteEvent[*certmanagerv1.Certificate]) bool {
-					return isCertificateOwnedByTLSPolicy(e.Object)
-				},
-				GenericFunc: func(e event.TypedGenericEvent[*certmanagerv1.Certificate]) bool {
-					return isCertificateOwnedByTLSPolicy(e.Object)
-				},
-			})),
+			controller.FilterResourcesByLabel[*certmanagerv1.Certificate](fmt.Sprintf("%s=%s", AppLabelKey, AppLabelValue))),
```
mikenairn, line 371: :+1:

mikenairn (commented):
This is what i thought might be the problem and using the label filter brings it in-line with DNSPolicy so lgtm.

Wondering if we should add a test that explicitly creates Certificate resources(not using a TLSPolicy), and ensure processing a TLSPolicy doesn't affect it?

KevFan:
> This is what i thought might be the problem and using the label filter brings it in-line with DNSPolicy so lgtm.
> 
> Wondering if we should add a test that explicitly creates Certificate resources(not using a TLSPolicy), and ensure processing a TLSPolicy doesn't affect it?

Adding that test is a good idea, I'll push a new commit with the test

PR #1083, opened 2024-12-17
Use WASM pull secret if protected registry

maleck13:
closes #1077 
Signed-off-by: craig <cbrookes@redhat.com>

rh-pre-commit.version: 2.2.0
rh-pre-commit.check-secrets: ENABLED

PR #1104, opened 2025-01-09
Cherry picks v1.0.2

maleck13:
cherry-pick fixes from 

https://github.com/Kuadrant/kuadrant-operator/pull/1083
https://github.com/Kuadrant/kuadrant-operator/pull/1102


maleck13:
I am going to merge this as I don't think the failing helm check is relevant here

eguzki:
The helm verification fail is not expected. It is due to a bug on how the makefile computes `AUTHORINO_OPERATOR_BUNDLE_VERSION`, `LIMITADOR_OPERATOR_BUNDLE_VERSION` and `DNS_OPERATOR_BUNDLE_VERSION` which does not respect `make/release.mk`. No harm for this PR, the helm values are correct, it is the verification process that shamefully fails.

PR #1105, opened 2025-01-10
enable build image trigger based on tags with -rc suffix

eguzki:
### Verification steps

Tested with the tag `v1.0.2-dev4` https://github.com/Kuadrant/kuadrant-operator/releases/tag/v1.0.2-dev4


maleck13 (approved):

eguzki:
The `verify bundle` test error is due to the missing `make/release.mk`. Without the release file, the content of the bundle manifests and the (default) versions set on the Makefile do not match, hence the error.

PR #676, opened 2024-05-28
Start subnet offset at 1 to avoid overlap with kind cluster ips

david-martin:
This fixes an issue where the service loadbalancer cluster ip can collide with other ip address used by local kind clusters or the network gateway ip.

I've seen this happen locally with podman (on macos) where the first kind cluster ip when using the quickstart script is 10.89.0.3, with a network gateway (not gateway api) ip of 10.89.0.1.

```
podman inspect $(podman ps -q --filter "name=kuadrant-local-control-plane") | grep -i "10."
...
                         "Gateway": "10.89.0.1",
                         "IPAddress": "10.89.0.3",
```

The first service loadbalancer ip assigned by metallb will be 10.89.0.0, but the next one will be 10.89.0.1 (and so on), which can collide.
This causes connection issues to that service loadbalancer ip.
```
kubectl --context kind-kuadrant-local get svc -A|grep -i loadbalancer
kuadrant-system   api-gateway-istio                                       LoadBalancer   10.96.85.153    10.89.0.2     15021:30253/TCP,443:30654/TCP                     27m
monitoring        prometheus-k8s-lb                                       LoadBalancer   10.96.190.210   10.89.0.1     9090:32689/TCP,8080:30109/TCP                     28m
monitoring        thanos-receive-router-lb                                LoadBalancer   10.96.190.231   10.89.0.0     10901:32399/TCP,10902:32451/TCP,19291:30993/TCP   58m
```

If you bring up a 2nd kind cluster using the quickstart script, the ip address of it will be 10.89.0.4
```
podman inspect $(podman ps -q --filter "name=kuadrant-local-1-control-plane") | grep -i "10."
...
                         "Gateway": "10.89.0.1",
                         "IPAddress": "10.89.0.4",
```
Similarly, this can clash with service loadbalancer ips on the first cluster.

The fix starts the subnet offset for metallb assigned service loadbalancer IPs at .16.

e.g.
1st cluster .16-.31
```shell
kubectl --context kind-kuadrant-local get svc -A|grep -i loadbalancer
kuadrant-system   api-gateway-istio                                       LoadBalancer   10.96.194.138   10.89.0.18    15021:30452/TCP,443:31978/TCP                     18m
monitoring        prometheus-k8s-lb                                       LoadBalancer   10.96.251.8     10.89.0.17    9090:30172/TCP,8080:31329/TCP                     18m
[...]

PR #1110, opened 2025-01-13
Allow setting metallb CIDR value for local env

mikenairn:
The default CIDR value (/28) allows for 16 ips to be assigned which is fine for the majority of dev/test/demo needs. However, when trying to test at scale it is useful to be able to create more Gateways on a single cluster and therefore more IPs can be required.

Changes here allow the CIDR and number of IP values used by the local setup script to be set as needed, default is the same as before (/28 & 16 IPs).

```
make local-setup CIDR=26 NUM_IPS=64
```

### DIFF
--- a/make/development-environments.mk
+++ b/make/development-environments.mk
@@ -9,11 +9,13 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: install-metallb
 install-metallb: SUBNET_OFFSET=1
+install-metallb: CIDR=28
+install-metallb: NUM_IPS=16
 install-metallb: kustomize yq ## Installs the metallb load balancer allowing use of an LoadBalancer type with a gateway
 	$(KUSTOMIZE) build config/metallb | kubectl apply -f -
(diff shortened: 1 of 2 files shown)


Boomatang:
If a users was to hit this issue, what would it look like? How would they identify the issue?

make/development-environments.mk
```diff
@@ -8,12 +8,14 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 	$(KUSTOMIZE) build config/crd | kubectl delete -f -
 
 .PHONY: install-metallb
-install-metallb: SUBNET_OFFSET=1
+install-metallb: SUBNET_OFFSET=0
```
david-martin, line 11: Why is the offset being changed from 0 to 1?
I recall there being some problem with overlapping IPs between clusters and loadbalancers.
Is that still valid? https://github.com/Kuadrant/kuadrant-operator/pull/676

mikenairn: Yeah, you're right, i will revert this.

philbrookes (approved):

PR #1111, opened 2025-01-14
docs: Fix incorrect geo code in DNS user guide

mikenairn:
GEO-US -> US

Only the following codes will be accepted with a "GEO-" prefix:

```
AWS_CONTINENT_CODE_AFRICA        = "AF"
AWS_CONTINENT_CODE_ANTARTICA     = "AN"
AWS_CONTINENT_CODE_ASIA          = "AS"
AWS_CONTINENT_CODE_EUROPE        = "EU"
AWS_CONTINENT_CODE_OCEANIA       = "OC"
AWS_CONTINENT_CODE_NORTH_AMERICA = "NA"
AWS_CONTINENT_CODE_SOUTH_AMERICA = "SA"
```

### DIFF
--- a/doc/user-guides/dns/load-balanced-dns.md
+++ b/doc/user-guides/dns/load-balanced-dns.md
@@ -135,7 +135,7 @@ metadata:
 spec:
   loadBalancing:
     weight: 130
-    geo: GEO-US
+    geo: US
     defaultGeo: false
   targetRef:
     group: gateway.networking.k8s.io


mikenairn:
> As well as updating the docs can we add CEL validation to policy, so the user can't apply the a policy to the cluster with invalid codes?

Probably could add a CEL validation for codes with the "GEO-" prefix since we have a provider agnostic list that for those that we define. That's said, its not something i intend to look into now, or add to this PR. Created an issue for it https://github.com/Kuadrant/kuadrant-operator/issues/1113

thomasmaas (approved):

PR #1119, opened 2025-01-15
add DNS Operator to grafana.com, example.md and kustomize.yaml

philbrookes:
- Updated format of dns-operator dashboard to be compatible with grafana.com
- Added DNS Operator dashboard to grafana.com 
- Updated example.md to show dns-operator dashboard
- Added DNS Operator dashboard to observability kustomize.yaml
- renamed file to accurately show `.json`


examples/dashboards/kustomization.yaml
```diff
@@ -20,6 +20,9 @@ configMapGenerator:
 - name: grafana-controller-resources-metrics
   files:
   - ./controller-resources-metrics.json    
+- name: dns-operator
+  files:
+  - ./dns-operator.json
```
david-martin, line 25: Would you mind updating 2 other places so the configmap will be mounted into grafana on both k8s and openshift?
https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/grafana/grafana_deployment_patch.yaml
and
https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/openshift/grafana/dashboards.yaml

The existing config in those files should hopefully be enough of an example what to add for the dns-operator configmap

mikenairn: Ah, this is why it isn't working for me

david-martin:
> * Updated format of dns-operator dashboard to be compatible with grafana.com

Checking, did you follow the instructions at [Exporting a dashboard for use with Grafana Community Platform or other Grafana Instances ](https://github.com/Kuadrant/kuadrant-operator/tree/main/config/observability#exporting-a-dashboard-for-use-with-grafana-community-platform-or-other-grafana-instances), including the `make dashboard-cleanup` cmd after?
I think the datasource gave us trouble before, and those instructions and cmd were a way to get around that.

mikenairn:
@david-martin Does anything extra need to be done to make sure when it is added that is has the same "kuadrant" tag as the others?
image

david-martin:
> Does anything extra need to be done to make sure when it is added that it has the same "kuadrant" tag as the others?

Yeah, good catch.
You can add the tags in the UI and export again, or it should be safe to add directly to the json.
Here's an example 
https://github.com/Kuadrant/kuadrant-operator/blob/main/examples/dashboards/app_developer.json#L1612-L1614

philbrookes:
Thanks, I've made those changes now.

doc/observability/examples.md
```diff
@@ -9,6 +9,7 @@ There are some example dashboards uploaded to [Grafana.com](https://grafana.com/
 | [App Developer Dashboard](https://grafana.com/grafana/dashboards/21538)      | `21538`       |
 | [Business User Dashboard](https://grafana.com/grafana/dashboards/20981)   | `20981`        |
 | [Platform Engineer Dashboard](https://grafana.com/grafana/dashboards/20982) | `20982` |
+| [DNS Operator Dashboard](https://grafana.com/grafana/dashboards/22695) | `22695` |
```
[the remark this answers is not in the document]
philbrookes: yes, it seems to be working now.

david-martin (approved):
Thanks for fitting the dashboard into the existing structure.

If you have any recommendations on making this process easier in future, I'm happy to help out.

examples/dashboards/kustomization.yaml
```diff
@@ -20,6 +20,9 @@ configMapGenerator:
 - name: grafana-controller-resources-metrics
   files:
   - ./controller-resources-metrics.json    
+- name: dns-operator
```
mikenairn, line 23: I think this needs to be `grafana-dns-operator` since that's what you have called it here https://github.com/Kuadrant/kuadrant-operator/pull/1119/files#diff-ab8359b25bb9e5ee0554a1f8eafd7e20f244e05e52837074eddbb3e0604b411b

philbrookes: oh nice one, thanks!

mikenairn (approved):
Working now 
image

I'd consider renaming that to "DNS Dashboard" or maybe name them all in a similar way to some of the kubernetes ones e.g. "Kuadrant / DNS / Overview" but i don't think we need to decide on that here.

PR #1121, opened 2025-01-16
Update catalogsource.yaml

maleck13:
When installing from the v1.0 branch is should always use the latest patch


maleck13:
@eguzki I updated this manually here but I think it makes sense to always have the 1.0 branch point to the latest 1.0.x catalog

maleck13:
actually we prob want the bundle etc also updated

PR #1124, opened 2025-01-17
Enable specification of the container engine for catalog builds

trepel:
## Overview

Makefile defines CONTAINER_ENGINE env var setting 'docker' as a default value. This env var should be used rather than having 'docker' hard-coded. 

## Verification Steps

`make catalog-build`
`CONTAINER_ENGINE=docker make catalog-build`
`CONTAINER_ENGINE=podman make catalog-build`

The output depends on your docker/podman config but if you have both installed and configured the commands should work. If you e.g. do not use docker then you can see something along the lines of
`ERROR: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?`

### DIFF
--- a/make/catalog.mk
+++ b/make/catalog.mk
@@ -59,7 +59,7 @@ catalog: $(OPM) ## Generate catalog content and validate.
 .PHONY: catalog-build
 catalog-build: ## Build a catalog image.
 	# Build the Catalog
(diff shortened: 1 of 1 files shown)


Boomatang (approved):
Looks good to me, I don't have docker installed, but on main this failed to find docker when the env var was set. On this branch it worked.

PR #1125, opened 2025-01-17
update headings and some format fixes

laurafitzgerald:
Changing from using the word pre-requisites as this might imply to the user when they are reading that they are responsible for those things when the steps needed to install are provided in the guide.
Also some small formatting fixes.

### DIFF
--- a/charts/kuadrant-operator/README.md
+++ b/charts/kuadrant-operator/README.md
@@ -4,57 +4,58 @@ This is the Helm Chart to install the official Kuadrant Kubernetes Operator
 
 ## Installation
 
-### Prerequisites
+### Install Dependencies
+
+Follow the steps below to install the following dependencies for Kuadrant Operator:
 - [Gateway API](https://gateway-api.sigs.k8s.io/)
 - [cert-manager](https://cert-manager.io/)
-- A Gateway Provider (I.E: [Istio](https://istio.io/latest/docs/ambient/install/helm/),
-  [Envoy](https://www.envoyproxy.io/docs/envoy/latest/start/start))
+- A Gateway Provider [Istio](https://istio.io/latest/docs/ambient/install/helm/)
 
-### Install dependencies
+[Envoy Gateway](https://gateway.envoyproxy.io/) is also an option as a gateway provider, **Steps Coming soon!**
 
-1. Install Kubernetes Gateway API:
+#### Install Kubernetes Gateway API:
 
-```sh
-kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.1.0/standard-install.yaml
-```
+ ```sh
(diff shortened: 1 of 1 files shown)


charts/kuadrant-operator/README.md
```diff
@@ -2,59 +2,60 @@
...
-  # Supported values for sail-operator v0.1.0 are [v1.22.4,v1.23.0]
-  version: v1.23.0
-  namespace: istio-system
-  # Disable autoscaling to reduce dev resources
-  values:
-    pilot:
-      autoscaleEnabled: false
-EOF
-```
+    ```sh
```
david-martin, line 24: Why the 4 space indent here.
Are the backticks not enough to render the code block?

charts/kuadrant-operator/README.md
```diff
@@ -2,59 +2,60 @@
...
-### Prerequisites
+Follow the steps below to install the following dependencies for Kuadrant Operator:
 - [Gateway API](https://gateway-api.sigs.k8s.io/)
 - [cert-manager](https://cert-manager.io/)
-- A Gateway Provider (I.E: [Istio](https://istio.io/latest/docs/ambient/install/helm/),
-  [Envoy](https://www.envoyproxy.io/docs/envoy/latest/start/start))
+- A Gateway Provider [Istio](https://istio.io/latest/docs/ambient/install/helm/)
 
-### Install dependencies
+[Envoy](https://www.envoyproxy.io/docs/envoy/latest/start/start) is also an option as a gateway provider, **Steps Coming soon!**
```
david-martin, line 12:
```suggestion
[Envoy Gateway](https://gateway.envoyproxy.io/) is also an option as a gateway provider, **Steps Coming soon!**
```

charts/kuadrant-operator/README.md
laurafitzgerald: it does render as is but the numbered headings are producing an error then. this change isn't essential tho so if you'd prefer to have it removed I can do that.

david-martin: Funny you should mention the numbered headings.
I'm proposing we avoid numbered lists in this [style guide PR](https://github.com/Kuadrant/docs.kuadrant.io/pull/183/files#diff-73ed2b26f52326ccd45aaa81e052bab2f213fb4e521a6f231bbb629d6dfb8b96R95-R97).

Personally, I think that should extend to numbering of headers as well, for the same reason.

re: the spaces on the code blocks,
for consistency would you mind if we removed the indentation, to align with the usage docs in https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#usage

david-martin:
Thanks @laurafitzgerald 
I'll merge this once the ci passes

david-martin (approved):

PR #1133, opened 2025-01-22
UPDATE: Release Document

Boomatang:
Adjust the release document to suit the new release process.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,21 +4,26 @@
 
 To release a version _“v0.W.Z”_ of Kuadrant Operator in GitHub and Quay.io, follow these steps:
 
-1. Kuadrant dependencies need to be released first:
+1. Verify correct versions for the Kuadrant release and its dependencies are listed in the `release.toml` file.
+   The verification should be done as part of the community call prior to cutting the release.[^1]
+   Update versions if required, release of dependency may be required, see step 2.
+   This file is located in the root of the kuadrant operator.
+
+   [^1]: This is the general guide, for hot fixes the discretion for releaser can be used.
+
+2. (Optional) Release Kuadrant dependencies:
    * [Authorino Operator](https://github.com/Kuadrant/authorino-operator/blob/main/RELEASE.md).
    * [Limitador Operator](https://github.com/Kuadrant/limitador-operator/blob/main/RELEASE.md).
    * [DNS Operator](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md).
    * [WASM Shim](https://github.com/Kuadrant/wasm-shim/).
    * [Console Plugin](https://github.com/Kuadrant/kuadrant-console-plugin).
 
-2. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
-   sure to fill all the fields:
+   Once dependencies have being released jump to step 1.
 
+3. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
+   sure to fill all the fields:
     * Branch containing the release workflow file – default: `main`
     * Commit SHA or branch name of the operator to release – usually: `main`
-    * Operator version to release (without prefix) – i.e. `0.W.Z`
-    * Kuadrant dependencies (WASM Shim, Console Plugin, Authorino, Limitador and DNS operators) versions (without prefix) – i.e. `0.X.Y`
-    * If the release is a prerelease
 
 3. Verify that the build [release tag workflow](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and completes for the new tag.
 
@@ -83,3 +88,33 @@ The usual steps are:
     * Copy the bundle files from `github.com/kuadrant/kuadrant-operator/tree/v0.W.Z/bundle`
     * Copy `github.com/kuadrant/kuadrant-operator/tree/v0.W.Z/bundle.Dockerfile` with the proper fix to the COPY commands
       (i.e. remove /bundle from the paths)
+
+--- 
+
+## Release file format.
+
+This example of the `release.toml` file uses tag `v1.0.1` as reference.
+
+```toml
+# FILE: ./release.toml
+[kuadrant]
+channel = "alpha"
+channels = ["alpha",]
+release = "1.0.1"
+
+[dependencies]
+Authorino_bundle = "0.16.0"
+Console_plugin = "0.0.14"
+DNS_bundle = "0.12.0"
+Limitador_bundle = "0.12.1"
+Wasm_shim = "0.8.1"
+
+```
+
+## Automated Verification Processes.
+
+* Check that dependencies release images exist for stated version
+* Check that the git tag does not already exist for the stated release.
+* Automated test, install via OLM and ensure the correct images are on cluster.
+This should handle most of [Verify OLM Deployment](#verify-olm-deployment), except the checking of the logs.
+* Ensure all required fields are stated in the `release.toml`.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,36 +1,68 @@
 # How to release Kuadrant Operator
 
-## Process
+## Kuadrant Operator Release
 
 To release a version _“v0.W.Z”_ of Kuadrant Operator in GitHub and Quay.io, follow these steps:
 
-1. Verify correct versions for the Kuadrant release and its dependencies are listed in the `release.toml` file.
-   The verification should be done as part of the community call prior to cutting the release.[^1]
-   Update versions if required, release of dependency may be required, see step 2.
-   This file is located in the root of the kuadrant operator.
+### Release file format.
+This example of the `release.toml` file uses tag `v1.0.1` as reference.
 
-   [^1]: This is the general guide, for hot fixes the discretion for releaser can be used.
+```toml
+# FILE: ./release.toml
+[kuadrant]
+default_channel = "alpha"
+channels = ["alpha",]
+release = "1.0.1"
 
-2. (Optional) Release Kuadrant dependencies:
+[dependencies]
+Authorino = "0.16.0"
+Console_plugin = "0.0.14"
+DNS = "0.12.0"
+Limitador = "0.12.1"
+Wasm_shim = "0.8.1"
+```
+The `[kuadrant]` section relates to the release version of the kuadrant operator.
+While the `[dependencies]` section relates to the released versions of the subcomponents that will be included in a release.
+There are validation steps during the `make prepare-release` that require the dependencies to be release before generating the release of the Kuadrant operator.
+
+
+### Local steps
+
+1. Create the `kuadrant-vX.Y` branch, if the branch does not already exist. 
+2. Push the `kuadrant-vX.Y` to the remote (kuadrant/kuadrant-operator)
+3. Create the `kuadrant-vX.Y.Z-rc(n)` branch with the updated versions of the `release.toml`.
+4. Run `make prepare-release` on the `kuadrant-vX.Y.Z-rc(n)` 
+
+
+### Remote steps
+
+1. Open a PR against the `kuadrant-vX.Y` branch with the changes from `kuadrant-vX.Y.Z-rc(n)` branch. 
+2. PR verification checks will run.
+3. Get manual review of PR with focus on changes in these areas:
+   * `./bundle.Dockerfile`
+   * `./bundle`
+   * `./config`
+   * `./charts/`
+4. Merge PR
+5. Run the Release Workflow on the `kuadrant-vX.Y`. This does the following:
+   * Creates the GitHub release
+   * Creates tags
+6. Verify that the build [release tag workflow](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/build-images-for-tag-release.yaml) is triggered and completes for the new tag.
+
+## Kuadrant Operator Dependencies
+
+1. Release Kuadrant dependencies as required:
    * [Authorino Operator](https://github.com/Kuadrant/authorino-operator/blob/main/RELEASE.md).
    * [Limitador Operator](https://github.com/Kuadrant/limitador-operator/blob/main/RELEASE.md).
    * [DNS Operator](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md).
    * [WASM Shim](https://github.com/Kuadrant/wasm-shim/).
    * [Console Plugin](https://github.com/Kuadrant/kuadrant-console-plugin).
 
-   Once dependencies have being released jump to step 1.
+2. Update the `release.toml` with the versions of required dependencies.
 
-3. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
-   sure to fill all the fields:
-    * Branch containing the release workflow file – default: `main`
-    * Commit SHA or branch name of the operator to release – usually: `main`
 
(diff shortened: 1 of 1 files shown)


RELEASE.md
```diff
@@ -83,3 +86,56 @@ The usual steps are:
...
+VERSION=1.0.1
+```
+
+## Automated Verification Processes.
+
+* Check that dependencies release images exist for stated version
+* Check that the git tag does not all ready exist for the stated release.
+* Automated test, install via OLM and ensure the correct images are on cluster.
+This should handle most of [Verify OLM Deployment](#verify-olm-deployment), except the checking of the logs.
+* Ensure all required fields are stated in the `release.toml`.
```
Boomatang, line 141: Tracking checks that I believe should be part of the release process. This section may not make it to the final version.

RELEASE.md
```diff
@@ -4,21 +4,24 @@
 
 To release a version _“v0.W.Z”_ of Kuadrant Operator in GitHub and Quay.io, follow these steps:
 
-1. Kuadrant dependencies need to be released first:
+1. Verify correct versions for the Kuadrant release and its dependencies are listed in the `release.toml` file.
+   The verification should be done as part of the community call prior to cutting the release.
```
maleck13, line 8: I don't think it should have to be in the community call. I think it can also be asked in slack. I think in general we should be able to always take the latest release of a given stream. IE latest of a major stream we are on. That said I think it is good to call out a release at the community call. Really just saying it isn't a "requirement" but more of an recommendation

RELEASE.md
```diff
@@ -4,21 +4,24 @@
...
+   Update versions if required, release of dependency maybe required, see step 2.
```
maleck13, line 9:
```suggestion
   Update versions if required, release of dependency may be required, see step 2.
```

RELEASE.md
```diff
@@ -4,21 +4,24 @@
...
    * [WASM Shim](https://github.com/Kuadrant/wasm-shim/).
    * [Console Plugin](https://github.com/Kuadrant/kuadrant-console-plugin).
 
-2. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
-   sure to fill all the fields:
+   Once dependencies have being released jump to step 1.
 
+3. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
+   sure to fill all the fields:
     * Branch containing the release workflow file – default: `main`
```
maleck13, line 23: are there obvious times where these will be different values?

RELEASE.md
```diff
@@ -4,21 +4,24 @@
...
     * Commit SHA or branch name of the operator to release – usually: `main`
-    * Operator version to release (without prefix) – i.e. `0.W.Z`
-    * Kuadrant dependencies (WASM Shim, Console Plugin, Authorino, Limitador and DNS operators) versions (without prefix) – i.e. `0.X.Y`
-    * If the release is a prerelease
```
maleck13, line 21: This might still be useful for RCs

RELEASE.md
```diff
@@ -83,3 +86,56 @@ The usual steps are:
...
+WASM_SHIM_VERSION=0.8.1
+RELATED_IMAGE_WASMSHIM=oci://quay.io/kuadrant/wasm-shim:v0.8.1
+RELATED_IMAGE_CONSOLEPLUGIN=quay.io/kuadrant/console-plugin:v0.0.14
+IMG=quay.io/kuadrant/kuadrant-operator:v1.0.1
+BUNDLE_VERSION=1.0.1
+BUNDLE_IMG=quay.io/kuadrant/kuadrant-operator-bundle:v1.0.1
+CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:v1.0.1
+CHANNELS=alpha
+BUNDLE_CHANNELS=--channels=alpha
+VERSION=1.0.1
```
maleck13, line 132: I guess it's not expected for us to touch these? Does this serve a purpose here ?

RELEASE.md
Boomatang: > we should be able to always take the latest release of a given stream

I am not sure I agree with this, and was there not a problem during the v1 rc testing where limitador v2 got pulled in as it was the latest. 

But that aside. The reason that I mention doing this in the community call is to have a fixed time and location for agreeing on the correct versions. Slack, I don't think is a good place make does decisions. There has being a number of times I have missed stuff in slack because of how noisy slack can be, and I switch the laptop to "do not disturb". The recent patch release is a prime example of this. I didn't see your message till hours later at which stage I believe you were already starting to create the images. I had a change that would have being good to get in as it was a bug fix, and right now I am not sure if that made it in.

If you feel strongly that it should only be a recommendation, I can change the wording, but I don't agree that it is the best idea.

RELEASE.md
Boomatang: At this stage, I am not sure. This is still the draft and that was what the existing was doing so didn't change it. I will look at this again.

RELEASE.md
Boomatang: I don't think it does. It is a choice that whom ever is doing the release would have to make at time of running the release workflow. This choice introduces a chance to make a mistake, and we want to reduce that chance.

I would expect that in the `release.toml` file when stating the version, it would be at that point the version is marked as a pre-release. This can be reviewed by a second set of eyes, reducing the chance for errors.

RELEASE.md
Boomatang: Yeah, this is here as a reference for now. I expect by the time we are done that file will not exist.

RELEASE.md
maleck13: The reason for a recommendation, is sometimes patches need to a be a rapid response. And if it is wednesday and we have to wait a week before being able to first communicate and then cut the release, I fear this is too slow.

RELEASE.md
maleck13: sure ok makes sense that we can infer from the release file

RELEASE.md
eguzki: My take is somewhat in between. I also do not think that a community call is a _must_. But there must be some agreement (or consensus) around the dependency version we want to pick for a given release of kuadrant. Whether this is slack, gmeet or pigeon messenger, I could not care less. We must be agile doing patch releases. I strongly agree on this. But it is a matter of days at least, not hours. So, for me, when we decide to do a patch release, if the patch is out there in one week time frame, this is good enough.

RELEASE.md
```diff
@@ -4,21 +4,24 @@
...
    * [Limitador Operator](https://github.com/Kuadrant/limitador-operator/blob/main/RELEASE.md).
    * [DNS Operator](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md).
    * [WASM Shim](https://github.com/Kuadrant/wasm-shim/).
    * [Console Plugin](https://github.com/Kuadrant/kuadrant-console-plugin).
 
-2. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
-   sure to fill all the fields:
+   Once dependencies have being released jump to step 1.
 
+3. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
```
eguzki, line 21: What is this release operator workflow doing high level?

I think we should have one section for minor releases and other one for patch releases. 

### minor releases

* create `kuadrant-vX.Y.0` branch
* create `kuadrant-vX.Y.0-minor` branch
* Open PR against `kuadrant-vX.Y.0` branch updating kuadrant version and dependency versions (that would trigger tests)
* Once approved, tag it. 
  * The tag would trigger GHA workflow to build images and create release notes in GH.  (no tests run), deploy new release in operatorhub (maybe not for this time.. in the future), deploy helm charts.

### patch releases

* create `kuadrant-vX.Y.Z-patch` branch
* Open PR against `kuadrant-vX.Y.0` branch updating kuadrant version and dependency versions (that would trigger tests)
* Once approved, tag it. 
  * The tag would trigger GHA workflow to build images and create release notes in GH.  (no tests run), deploy new release in operatorhub (maybe not for this time.. in the future), deploy helm charts.

Sounds reasonable? No need to deal with release implementation details. My design goal is that anyone, with no tech knowledge requirements of kuadrant, is able to release kuadrant.

eguzki: Looks like minor and patch release procedures are pretty much the same.. maybe we should merge them in one single... IDK :shrug:

RELEASE.md
```diff
@@ -83,3 +86,56 @@ The usual steps are:
...
     * Copy `github.com/kuadrant/kuadrant-operator/tree/v0.W.Z/bundle.Dockerfile` with the proper fix to the COPY commands
       (i.e. remove /bundle from the paths)
+
+--- 
+
+## Release file format.
+
+This example of the `release.toml` file uses tag `v1.0.1` as reference.
+
+```toml
```
eguzki, line 96: The format LGTM. What tools could we leverage from GHActions (bash) or Makefile to read those?

RELEASE.md
maleck13: Yeah I think that is fine as long as we call out the key difference of using the existing release branch for a patch

RELEASE.md
maleck13: I don't want this to hold things up. So lets leave as is and change as needed in the future

RELEASE.md
maleck13: Yeah these look good and IMO should be added as tasks to the overall top level issue

RELEASE.md
```diff
@@ -83,3 +86,56 @@ The usual steps are:
...
+CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:v1.0.1
+CHANNELS=alpha
+BUNDLE_CHANNELS=--channels=alpha
+VERSION=1.0.1
+```
+
+## Automated Verification Processes.
+
+* Check that dependencies release images exist for stated version
+* Check that the git tag does not all ready exist for the stated release.
```
thomasmaas, line 138:
```suggestion
* Check that the git tag does not already exist for the stated release.
```

RELEASE.md
Boomatang: Sure, that makes more sense than making the note here.

RELEASE.md
```diff
@@ -83,3 +88,33 @@ The usual steps are:
...
+
+```toml
+# FILE: ./release.toml
+[kuadrant]
+channel = "alpha"
+channels = ["alpha",]
+release = "1.0.1"
+
+[dependencies]
+Authorino_bundle = "0.16.0"
```
eguzki, line 106: I would call this `Authorino` instead of `Authorino_bundle`. `bundle` is OLM term, and release process covers OLM and helm charts.

RELEASE.md
```diff
@@ -4,21 +4,26 @@
...
    * [Limitador Operator](https://github.com/Kuadrant/limitador-operator/blob/main/RELEASE.md).
    * [DNS Operator](https://github.com/Kuadrant/dns-operator/blob/main/docs/RELEASE.md).
    * [WASM Shim](https://github.com/Kuadrant/wasm-shim/).
    * [Console Plugin](https://github.com/Kuadrant/kuadrant-console-plugin).
 
-2. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
-   sure to fill all the fields:
+   Once dependencies have being released jump to step 1.
 
+3. Run the GHA [Release operator](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/release.yaml); make
```
eguzki, line 23: This is what I have in mind:

I want to release, so I run "Prepare Release" workflow (workflow number one to be implemented). The input for this "Prepare Release" would be:
* base branch/commit: the base branch/commit from which you want to create the release branch
* version of kuadrant
* [optional} Branch containing the release workflow file – default: `main`

The versions of the dependencies are already in place, no need to specify them. The workflow would create release branch (for minor release), update kuadrant version and build chart and bundle. Then commit, push. I would implement also that a PR is being opened so human intervention is needed to verify and approve. 

Once merged I run "Release" workflow (workflow number two to be implemented). The input for this "Release" would be:
* base branch/commit: the base branch/commit from which you want to create the release

The workflow would create a tag (kuadrant version being read from toml file so avoid inconsistencies between git tag and toml), push it and create Github release from it. The tag would trigger another (existing) GHA workflow to build images (no tests would be run), deploy new release in operatorhub (maybe not for this time.. in the future), deploy helm charts.

sounds reasonable? cc @maleck13

RELEASE.md
maleck13: I think this set of steps makes sense and should be merged into one set with a comment for patch releases needing existing branch.  One other thing I would add is the tag names. IE it should be named v1.2.0-rc1 for example until ready to do final tag and the final tag should be based on a tested RC tag?

RELEASE.md
maleck13: So slight modification WDYT

```
- create kuadrant-vX.Y branch (if doesn't exist)
- create kuadrant-vX.Y.Z-rc(n) branch (with updated versions in release.toml)
- Open PR against kuadrant-vX.Y branch kuadrant-vX.Y.Z-rc(n) branch (that would trigger tests)
- Run the prepare release workflow using the kuadrant-vX.Y.0-rc(n)  branch (that would setup the helm and bundle stuff)
- Get PR approved (so bundle and helm can be seen and version)
- Merge PR
- Run Release Workflow that would create the GH release and final tags etc
```

maleck13: I guess Ideally you wouldn't need an separate prepare release workflow but could just execute that based on your TOML file locally before pushing to the kuadrant-vX.Y.Z-rc(n) branch get that approved and then trigger the release workflow.

eguzki: That sounds good to me as well. 

But that requires that the one doing that has dev tools installed locally (like makefile). I was having in mind somebody without the dev environment (or at least _our_ dev env) installed being able to release. So it is all web based: run prepare release workflow, review approve, run release workflow. Done. But your approach also LGTM

RELEASE.md
Boomatang: As long as we don't the bundle and component version going out of sync this also makes sense to me.

maleck13 (approved):
+1 this looks like a good step forward

maleck13:
@didierofrivia FYI some adjustments to remove the need to input versions into the action each time and to make the release more traceable and verifiable

RELEASE.md
```diff
@@ -1,31 +1,68 @@
 # How to release Kuadrant Operator
 
-## Process
+## Kuadrant Operator Release
 
 To release a version _“v0.W.Z”_ of Kuadrant Operator in GitHub and Quay.io, follow these steps:
 
-1. Kuadrant dependencies need to be released first:
+### Release file format.
```
didierofrivia, line 7: So we are abandoning  the "release button" in favour of creating branch + release file? I reckoned it was very convenient to have a workflow such as https://github.com/Kuadrant/limitador-operator/blob/main/.github/workflows/release.yaml#L4

didierofrivia: I'd say is good to keep both, but not relying on the file to be a requisite for the release... those are 3 extra steps for releasing instead of 1.

didierofrivia: And nothing prevent us for making mistakes in the file neither xD

didierofrivia:
IMHO I wouldn't remove the Release Operator action from the docs, I'll keep it and add the file alternative workflow.

RELEASE.md
Boomatang: I don't think this is an abandonment of the "release button", but more stepping back to reevaluate. Some of the manual local steps can, and possible should be automated again in the future. The main focus is how do we reduce the chance of errors during the release, but also give us an escape route for when the error happen. That escape route would be the PR review.

didierofrivia: in that case, this PR shouldn't remove the automated version, but mention both alternatives

PR #1143, opened 2025-01-29
Checkpoint adding observability feature

### LINKED ISSUE #1065: Observability Configuration Phase 1: Create ServiceMonitors for components in the kuadrant operator
Related RFC https://github.com/Kuadrant/architecture/pull/97

Candidate resources
- https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/prometheus/monitors/operators.yaml
- Istio https://github.com/Kuadrant/kuadrant-operator/blob/main/config/observability/prometheus/monitors/istio/service-monitor-istiod.yaml
- EG https://github.com/Kuadrant/kuadrant-operator/tree/main/config/observability/prometheus/monitors/envoy

Simplifies the installation of kuadrant by removing the manual creation of ServiceMonitors, while also allowing users to disable the feature if they’d prefer to create them themselves.

Adding the capability to the sub-operators (authorino, dns, limitador) is out of scope at this time. The goal is to get the feature in front of users more quickly than the time it would take and risk of prioritising implementation in the individual sub-operators and bubbling that up to the kaudrant operator.

The Kuadrant CR configuration will look like this:

```yaml
observability:
     enable: true
```

As part of this work, update the guide at https://docs.kuadrant.io/dev/kuadrant-operator/doc/user-guides/full-walkthrough/secure-protect-connect/#optional, and any other locations where Service/PodMonitors are mentioned to reference using this new configuration.
A new guide specifically for enabling Observability, explaining what it does and its intended use would be good to write.

david-martin:
Closes #1065

TODO:
- [x] Verify pattern for setting up the reconcile function and monitoring client
- [x] Create Gateway PodMonitor (istio & envoy)
- [x] Refactor duplicate creation blocks
- [x] Add controller test


api/v1beta1/kuadrant_types.go
```diff
@@ -54,6 +54,11 @@ func (p *Kuadrant) GetLocator() string {
 
 // KuadrantSpec defines the desired state of Kuadrant
 type KuadrantSpec struct {
+	Observability Observability `json:"observability,omitempty"`
```
eguzki, line 57: I would make this pointer, so you can discriminate between 
a) not set
b) set to false
c) set to true

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil
+	}
+	logger.Info("observability enable is true")
+
+	// Check if monitoring CRDs exist
+	if ok, err := utils.IsCRDInstalled(r.restMapper, monitoringv1.SchemeGroupVersion.Group, monitoringv1.ServiceMonitorsKind, monitoringv1.SchemeGroupVersion.Version); !ok || err != nil {
+		logger.Info("ServiceMonitor CRD not installed")
+		return nil
+	}
+	logger.Info("ServiceMonitor CRD is installed")
```
eguzki, line 63: nitpick: I would make this debug level, `logger.V(1).Info("ServiceMonitor CRD is installed")`

eguzki (commented):
Code looks like going in a correct direction. 

Few notes to think about:

* Have you considered using a regular dynamic client to do the monitoring resources reconciliation (create, update, delete)?? I do not see any real need to use the monitoring client. I might be missing somehting, though
* When implementing the reconciliation, try to answer the following questions: What happens when the user changes the switch between `enabled: true` to `enabled: false` and vice versa? What if somebody externally updates the monitoring resources manually (or some external controller)?? Should kuadrant controller watch the reconciled resources and revert the changes to make sure it is what kuadrant wants to be? For instance, somebody (let's hope accidentally) removes the montoring resources

I will keep watching this PR and providing review, but I will defer the approval to the current experts @Boomatang and @KevFan as I am no longer as familiar as them with the current codebase.

Boomatang (commented):
Hay Dav, I have left a few comments around the use of the policy machinery and the structure the code should be in. If there is any questions, please feel free to ask. The policy machinery works differently to the controller runtime and I have a lot of assumed knowledge, so some of my comments might be vague.

controllers/state_of_the_world.go
```diff
@@ -404,6 +407,7 @@ func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
 			NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run,
 			NewDataPlanePoliciesWorkflow(b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run,
 			NewKuadrantStatusUpdater(b.client, b.isGatewayAPIInstalled, b.isGatewayProviderInstalled(), b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Subscription().Reconcile,
+			NewObservabilityReconciler(b.client, b.manager.GetRESTMapper(), b.monClient).Subscription().Reconcile,
```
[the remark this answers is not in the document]
david-martin: Checking my understanding is right.
With the suggested approach, the reconciler would only be created if CRDs (ServiceMonitor/PodMonitor) exist in the cluster.
If those CRDs come into existence later, the reconciler will not be created, unless the operator is restarted?

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+		MonClient:  mc,
+	}
+}
+
+func (r *ObservabilityReconciler) Subscription() *controller.Subscription {
+	return &controller.Subscription{ReconcileFunc: r.Reconcile, Events: []controller.ResourceEventMatcher{
+		{Kind: ptr.To(kuadrantv1beta1.KuadrantGroupKind)},
+	},
+	}
+}
```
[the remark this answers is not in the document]
david-martin: > You are missing the event matcher

Assuming no event specified means all events, I think that's the desired behaviour, so as to capture the setting and unsetting of the enable flag ( and if the Kuadrant CR is removed)

> linking function

Thanks for pointing this out. I wasn't aware of this pattern before.
I'll try linking the Kuadrant CR to ServiceMonitors and PodMonitors with specific labels as a first pass. If I think of a more meaningful linking, I'll reach out to discuss.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+
+	kObj := GetKuadrantFromTopology(topology)
+	if kObj == nil {
+		return nil
+	}
+
+	// Check if observability is enabled
+	if !kObj.Spec.Observability.Enable {
+		logger.Info("observability enable is false")
+		// TODO: Remove monitors if they exist
```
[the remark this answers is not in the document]
david-martin: > I assume this TODO would not remove service monitors that the user may have created.

That would be the idea, yes.
Targeting specific labels should be sufficient.

> what do you do if the ServiceMonitor has no parent in the topology, as in kuadrant CR was removed

I'm not sure yet, but a couple of thoughts:

- Would it matter if we can remove the ServiceMonitor based on a label when the Kuadrant CR is removed?
- Is this a problem already when the Kuadrant CR is removed, and it was linked to other resources like authorino and limitador? (I haven't dived into that code yet)

api/v1beta1/kuadrant_types.go
david-martin: Understood.
From the reconcile function point of view, I'm wondering if there would be a difference between a and b. 
I think the behaviour would be the same for both i.e. remove the SerivceMonitors/PodMonitors if they exist.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/utils/ptr"
+)
+
+type ObservabilityReconciler struct {
+	Client     *dynamic.DynamicClient
+	restMapper meta.RESTMapper
+	MonClient  monclient.Interface
+}
+
+func NewObservabilityReconciler(client *dynamic.DynamicClient, rm meta.RESTMapper, mc monclient.Interface) *ObservabilityReconciler {
```
laurafitzgerald, line 25: nit: when it comes to naming, are we using this initialism as a standard in this code base?

controllers/observability_reconciler.go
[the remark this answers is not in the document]
Boomatang: One why the topology can change the clean up code is based around the parents of a resource. If the monitors are children of resources, let it be the kuadrant CR or autorino, limitador, what you can do is scan the root of the topology for any monitors. The clean up question becomes has this resource got any parents? If no, then it can be removed from the cluster.

controllers/observability_reconciler.go
david-martin: do you mean the param names like `rm` for the RESTMapper param?

david-martin:
Superceded by #1164

PR #1146, opened 2025-01-30
test improvement

maksymvavilov:
Improves two Rate Limit Policy tests: 
1. The test does not evaluate one of the conditions (always passes)
2. The test does not check for the second policy. Leaves a possibility of false positive.


maksymvavilov:
@KevFan would you have a moment for a sanity check of this one?

KevFan (approved):
Looks good to me! 👍

PR #1150, opened 2025-01-31
Release process/workflow

Boomatang:
This PR targets the uses of a `release.toml` file for releases.

There are a number of changes that have being done in the hope to simplify future work. 

The first main change is the adding of a `release.toml` file for the release. Below is a sample of what this file will look like. The file will be saved at the root of the project.
``` toml
[kuadrant]
channels = ["alpha"]
default_channel = "alpha"
release = "1.0.1"

[dependencies]
Authorino_bundle = "0.16.0"
Console_plugin = "0.0.14"
DNS_bundle = "0.12.0"
Limitador_bundle = "0.12.1"
Wasm_shim = "0.8.1"
```
The next major change is moving the moving the release code out of the Makefiles and into self contained scripts. Every thing from the tools required to validation. This is an attempt to allow the make the steps required to do a release simpler. 

The location that the tools are being installed to has also being changed to `~/.local/bin`, this does need to be checked to work with the github action. The reason for this change is around there was case where it was very unclear where the tool was being called from and gave no insurances that the correct version of the tool was being used. 

Next change is related to tooling, and that is the removing of the dependency on golang. The required dependency was needed for the installation of tools that also had released binaries. As not all tools could be managed with the golang method I believe it is better to manage the require tools by direct downloads. The removal of golang gives us one less place to maintain a version of a resource. 

The last change is related to the workflow directly, the form is much simpler now with only the branch/tag/sha to use in the release process.

# Validation
To validation that the process is still working as expected. 

- check out this branch
- create a worktree for a past release. This sha was before cutting the release for v1.0.1
```sh
git worktree add /tmp/review 14d99004ab8dfda4bd4f8e50191da28d8bc85aa9
```

- Create the `release.toml` file in the root of the project.
```sh
echo '[kuadrant]
channels = ["alpha"]
default_channel = "alpha"
release = "1.0.1"

[dependencies]
Authorino_bundle = "0.16.0"
Console_plugin = "0.0.14"
DNS_bundle = "0.12.0"
Limitador_bundle = "0.12.1"
Wasm_shim = "0.8.1"' > /tmp/review/release.toml
```

- Run the release script. NOTE: this will download any tools required that are missing from your path
```sh
ROOT=/tmp/review hack/release/release.sh
```

- compare the resulting changes with the commit that was created during the v1.0.1 release.
```sh
git -C /tmp/review diff 1e417b087bd96a5a1f5fc0c3e53e2237b324c1f8
```

The expected changes are as follows
- The `createdAT` timestamp being different
- The `make/release.mk` file being missing.


eguzki:
Before jumpling to the review of the implementation: 
* `The location that the tools are being installed to has also being changed to ~/.local/bin` -> this can break interoperability with other repos that are using same tools (yq, helm, operator-sdk, kustomize) and require specific version. 
* I do not see as an issue the Golang dependency for installing tools. From the perspective of somebody releasing, it would be interacting with GH workflows, so no need to have golang installed. From the perspective of engineering, we all have. And to be honest, golang provides very nice/easy (not simple)  mechanism to install tools locally. 

What are your thoughts?

Boomatang:
After the call today the implementation is changing and some of points to address your comments are not as valid. 

The breaking of interoperability with other repos. So what if it does. Prior to the call today, these scripts were being designed to run in GitHub workflows. There should be no other repos using the tools. The argument over requiring specific version for the different repos, ya  I can see that, but that problem also exists today. I don't add the project bin to my PATH every time I enter the project. So if  I call any of those tools out side the make targets it is highly likely the version don't match. Even if the I was to use one in the project bin there is no guarantee that version would be the correct version. This has happened to me more than once were a version of a tool was updated and the only way I found out was by a failing build. Think we high to much of the complicity around tooling in the make targets. I do like the idea of nix for the management of tools, but I think it better to warn if a version we don't expect is being used.

On the Golang dependency, that is a hard one. Someone doing a release that is all in the GitHub workflows, they should not need it locally. We can agree on that. When I don't think we will agree is on the perspective of engineering. You say we all have, but that is not true. Last week I found my self using a machine that didn't have Golang install. In fairness, I was only wanting to explore the operator running, and do much edits. But I couldn't run `make local-setup` as it requires Golang even tho it builds the operator in container files. That machine still does not have Golang installed, and I still have not explored what I was going to look at. If there was no way to build the operator but only using `go build` locally, or all the dev tools could be installed using Golang, then ya I would have a lesser issue with it. We know the arguments between easy and simple, and I don't want to go there.

.github/workflows/release.yaml
```diff
@@ -1,88 +1,34 @@
 name: Release Operator
 
 on:
-  workflow_dispatch:
```
didierofrivia, line 4: I strongly believe this is should **not** be removed. The new release process that is proposed adds many extra steps, both local and remote, reducing the profile of people that can release the operator and even if it adds some validation step to avoid human error, it doesn't guarantee mistakes won't be made. The release button could be enhance to have an intermediate step that could add some extra validation.

didierofrivia (changes requested):

.github/workflows/release.yaml
```diff
@@ -1,88 +1,34 @@
...
-      consolePluginImageURL:
-        description: ConsolePlugin image URL
-        default: "quay.io/kuadrant/console-plugin:latest"
-        type: string
-      prerelease:
-        description: Is the release a pre-release?
-        required: false
-        type: boolean
+  pull_request:
+    types:
```
eguzki, line 5: no sure we want to generate a release every time a PR against release branches are merged. Thinking about backporting features/fixes from `main` branch

.github/workflows/release-changes.yaml
```diff
@@ -0,0 +1,23 @@
+name: Release Operator
+
+on:
+  pull_request:
+    types:
+    - opened
```
eguzki, line 6: I think that we want to verify that releases are correct for all branches, including release branches. And not only opening a PR. This test should be in place with multiple verification checks, like running `make bundle` and then checking all changes are git committed.

.github/workflows/release.yaml
```diff
@@ -1,88 +1,34 @@
...
-          tagging_message: v${{ inputs.kuadrantOperatorVersion }}
-          commit_options: '--signoff'
+          bash ./hack/release/load_envvar.sh
       - name: Create release
         uses: softprops/action-gh-release@v1
         with:
-          name: v${{ inputs.kuadrantOperatorVersion }}
-          tag_name: v${{ inputs.kuadrantOperatorVersion }}
-          body: "**This release enables installations of Authorino Operator v${{ inputs.authorinoOperatorVersion }}, Limitador Operator v${{ inputs.limitadorOperatorVersion }}, DNS Operator v${{ inputs.dnsOperatorVersion }}, WASM Shim v${{ inputs.wasmShimVersion }} and ConsolePlugin ${{ inputs.consolePluginImageURL }}**"
+          name: v${{ kuadrantOperatorVersion }}
```
eguzki, line 29: where is `kuadrantOperatorVersion` being set? if it is an env var, it should be `env.kuadrantOperatorVersion`

.github/workflows/release.yaml
```diff
@@ -1,88 +1,34 @@
...
-        uses: stefanzweifel/git-auto-commit-action@v5
-        with:
-          commit_message: "Prepared release v${{ inputs.kuadrantOperatorVersion }}"
-          commit_user_name: "github-actions[bot]"
-          commit_user_email: "github-actions[bot]@users.noreply.github.com"
-          branch: release-v${{ inputs.kuadrantOperatorVersion }}
-          create_branch: true
-          tagging_message: v${{ inputs.kuadrantOperatorVersion }}
-          commit_options: '--signoff'
+          bash ./hack/release/load_envvar.sh
```
eguzki, line 25: You're not going to be able to modify the caller's shell because it's in a different process context.

hack/release/operator/make_helm.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
eguzki, line 1: this is about preparing a helm chart, so what about calling the file `make_chart.sh` ?

release.toml
```diff
@@ -0,0 +1,11 @@
+[kuadrant]
+channels = ["alpha"]
+default_channel = "alpha"
+release = "1.0.1"
+
+[dependencies]
+Authorino = "0.16.0"
```
eguzki, line 7: I am much in favor of dependency pinning here, but if I recall it correctly, we were going to address this in a following up work item and needs to be agreed (we might find some resistance). In `main` this file should not exist with well known defaults or exist with `0.0.0`

eguzki (dismissed):
Some comments left

release.toml
```diff
@@ -0,0 +1,11 @@
+[kuadrant]
+channels = ["alpha"]
```
eguzki, line 2: `channel` is OLM specific term. What about `[olm]` section? There can be another section for `helm-chart` to set other optional fields like `appVersion` or maybe others ([Chart.yaml](https://helm.sh/docs/topics/charts/#the-chartyaml-file))

.github/workflows/release.yaml
Boomatang: Far I can add that back in.

.github/workflows/release-changes.yaml
```diff
@@ -0,0 +1,23 @@
+name: Release Operator
```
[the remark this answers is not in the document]
Boomatang: Copy and paste error

.github/workflows/release.yaml
Boomatang: Good point, I will look at it. What is our process for doing backports? Do we have that documented anyway?

.github/workflows/release-changes.yaml
Boomatang: The check that is being done here is ensuring the versions in the `release.toml` are set in the bundles & charts. It might not be good to enable this on all branches just yet as it would fail for any PR that is currently opened with out the `release.toml` file.

If it should be enabled for every branch long term is a different question. And I am not a 100% sure it should, but feel it would natural end up there. Which would be good as it would mean the `verify-bundle` and verify-helm` checks could be removed.

.github/workflows/release.yaml
Boomatang: I will look at this.

.github/workflows/release.yaml
Boomatang: You sure? I am almost sure I got that from GitHubs examples. Will check.

hack/release/operator/make_helm.sh
Boomatang: Sure why not.

release.toml
Boomatang: This file can exist on any branch currently, there are only two make targets that are making use of it. The `prepare-release` and `verify-prepare-release`  will use the file. If the file is not on main, then every time a release is created the releaser would need to create the file, this I could see as being error prone. It would be better to update a existing file.

release.toml
Boomatang: Ya we can do that. I would be inclined to do `[kuadrant.olm]` as it is more directly related to the kuadrant section.

.github/workflows/release-changes.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+on:
+  pull_request:
+    types:
+    - opened
+    branches:
+      - 'release-v[0-9]+.[0-9]+'
```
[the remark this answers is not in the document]
eguzki: I was thinking on the branch `kuadrant-vX.Y.Z-rc(n)`. Does it make sense?

.github/workflows/release-changes.yaml
eguzki: > The check that is being done here is ensuring the versions in the release.toml are set in the bundles & charts. It might not be good to enable this on all branches just yet as it would fail for any PR that is currently opened with out the release.toml file.

Currently we are checking on every PR and `main` that bundles and charts are consistent with versions. If `release.toml` is missing, versions should have a default value and that default value is the one set on bundles and charts. That is what I would do, just food for thought. 

Should the jobs `verify-bundle` and  `verify-helm` be updated to use `release.toml` instead a new workflow? Task overlapping is quite clear.

.github/workflows/release.yaml
Boomatang: Yes, they should be prefixed with `env.`

.github/workflows/release.yaml
eguzki: Currently, we cherry pick commits to the release branch from `main`.

.github/workflows/release-changes.yaml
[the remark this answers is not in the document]
Boomatang: I see what your saying, but I feel it is some different that what we have outlined in the [Local and Remote Steps](https://github.com/Kuadrant/kuadrant-operator/blob/release-process-enhancement/RELEASE.md#local-steps). 

Should we to diagram this process? How I was seeing it there was never going to be a release PR made against the rc branches.

.github/workflows/release-changes.yaml
Boomatang: Some what yes. The current `verify-bundle` and `verify-helm` check that there has being no changes to the CRD data. There is no checking of operand versions. Where as the `verify-prepare-release` will also check that the index can be built for olm.

For updating the current `verify-bundle` and `verify-helm` to use the `release.toml`. I though about this, and decided against it. I would be inclined to deprecate those targets. They both do the same modifications to the initial manifest, which can mean they can fallout of sync. They both are required to run together. I don't see any reason why you may want to run one but not the other. 

There is also the last part where it is hard to know where people has set up automation with them targets. In this case I am thinking of a developers local workflow.  Once the release.toml file is everywhere those targets could be removed.

.github/workflows/release.yaml
eguzki: a patch release. Let's say you release v1.1.0. And continue adding features in `main`. Then we need to create v1.1.1 with some fixes. Those fixes are first added in `main` and then backported to v1.1.1 and v1.1.1 can be released. Makes sense?

.github/workflows/release-changes.yaml
eguzki: > The current verify-bundle and verify-helm check that there has being no changes to the CRD data. There is no checking of operand versions

.github/workflows/release.yaml
eguzki: Oh.. that is then different. It is populating github environment which is a data structure and it is not shell env var. Bear in mind that this approach does not work locally, only on github workflows. That can be limitting.

release.toml
eguzki: I am ok having the `release.toml` in every branch, even `main`. This change is targeting `main` indirectly, so the values of this file should be those in `main`. Maybe `0.0.0` (which can be interpreted as `latest` for docker tags)

release.toml
eguzki: Up to you.

Some suggestion: `kuadrant` seems redundant.. this project is all about kuadrant :)

.github/workflows/release.yaml
Boomatang: I would see this as the next release and not a backport. To me a backport would be when you have v1.1.0 and v1.2.0 released. There is a bug that affects but version and needs to be fixed in both version. So you would end up v1.1.1 and v1.2.1 both having the same change. 

I think the process stays the same, but the doc need a tweak. When doing the [local steps 3](https://github.com/Kuadrant/kuadrant-operator/blob/release-process-enhancement/RELEASE.md#local-steps) the branch would be created from the released version and not main. Then the required commits are cherry picked over. After which I think the process stays the same.

.github/workflows/release.yaml
Boomatang: Yes I hit that issue. The only step in that uses the data structure values is the `softprops/action-gh-release@v1` which is creating the github release. There is no other scripts using that data. 

And of local checking I have add a `--echo` flag that will echo out what would be added to the the structure. So I think we are okay.

hack/release/operator/make_chart.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
didierofrivia, line 1: We have a Make target doing the build too, I'd say this script should read the versions from the toml file and call that target, some changes will be needed, reading from the env var instead of the manager olm file.

.github/workflows/release-operator.yaml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  pull_request:
+    types:
+    - closed
+    branches:
+      - 'release-v[0-9]+.[0-9]+'
+  workflow_dispatch: {}
+
+jobs:
+  if_merged:
```
didierofrivia, line 12: job name should be `release` or `release-operator`, `if_merged` is more a state, condition, etc than a job name.

.github/workflows/release-operator.yaml
```diff
@@ -0,0 +1,35 @@
...
+    if: github.event.pull_request.merged == true
+    name: Release operator
+    runs-on: ubuntu-latest
+    steps:
+      # TODO: not sure if the `Install gettext-base`is still required
```
didierofrivia, line 17: why it was required before?

.github/workflows/release.yaml
Boomatang: There is two branches, the release branch and the RC branch. Changes going to the release branch are what is being released. It is the RC branch where add/modifying stuff should happen.

hack/release/load_github_envvar.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
Boomatang: Sure I can move them over there.

hack/release/operator/make_chart.sh
Boomatang: I am not 100% sure I understand what you are saying. But I can say I didn't want to change the behavior of existing make targets as they are also used in other workflows. You have to either do logic to around the existences of the toml file, or duplicate the code. I went with duplicate till it is proven they are the same.

.github/workflows/release-operator.yaml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request:
+    types:
+    - closed
+    branches:
+      - 'release-v[0-9]+.[0-9]+'
+  workflow_dispatch: {}
+
+jobs:
+  if_merged:
+    if: github.event.pull_request.merged == true
```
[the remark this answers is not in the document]
Boomatang: Good point, I don't know enough about the `enqueued` stuff to answer this question. I will look into it.

.github/workflows/release-operator.yaml
Boomatang: A bit of coping and pasting going on the workflow was copied from the existing release workflow. And there I could not work out why it was needed. I didn't mean to go back to it, and see what happens when it is removed.

hack/release/operator/make_chart.sh
didierofrivia: makes sense, if you want I could do the change in my following PR or we could look at it together

.github/workflows/release.yaml
eguzki: Maybe we got this differently. From the RELEASE.md, when you are ready to release the existing content in `kuadrant-vX.Y.Z`, you create a new ephemeral branch named  `kuadrant-vX.Y.Z-rc(n)` to hold the changes to be reviewed in the PR. After being merged the RC branch can be deleted. I do not see two long lived release branches.

Boomatang: Yes, less step back for a moment, I think we might be saying the same thing but in different ways.

---
We going about making the release for v1.1.0. The branch v1.1 is created which holds the patch version. A v1.1.0-rc1 branch is created from main. A PR is opened against v1.1 from v1.1.0-rc1. Once that PR is merged the GitHub release is created from the v1.1 branch for v1.1.0. After this the v1.1.0-rc1 branch can be deleted.

If a patch v1.1.1 is required, and main has moved on. The v1.1.1-rc1 branch is created. The required commits are cherry picked from main to v1.1.1-rc1. Then a PR is opened against v1.1, once merged the GitHub release is created from v1.1 for v1.1.1. Once this is done the v1.1.1-rc1 branch can be deleted.

At no point should changes land in the the v1.1 branch that didn't come from a PR. We may need logic to ensure that PR is coming from a patch rc branch for that version stream, but that can be an improvement.
---

Does my description align with your  understand?

eguzki: Yes, in general terms we are aligned. But, there is one fact that I see differently. 

>If a patch v1.1.1 is required, and main has moved on. The v1.1.1-rc1 branch is created. The required commits are cherry picked from main to v1.1.1-rc1

For me the `v1.1.1-rc1` branch (and related PR) **only** includes the releasing updates, like versioning, manifests and so on.. The patches cherry-picked from `main` can be added to v1.1 branch in separate PRs, or directly pushed to v1.1 as we have been doing. For me v1.1 can be updated just like we do with `main`. Minor release branches, vX.Y, serves as base branch to create patch releases. In the same way, `main` branch serves as base branch to minor releases.

I also think it is a typo, but just in case, to clarity:
> the branch v1.1 is created which holds the patch version. A v1.1.0-rc1 branch is created from main

v1.1.0-rc1 branch is created from v1.1 at the point we want to release. 

Sorry if this process looks tedious. For me it is totally worthy. We need to be aligned and in the same page. And, like anything else in real life, there are multiple ways of doing it, all valid. We need to go through this helpful discussion in order to reach agreement we all feel comfortable with.

release.yaml
```diff
@@ -0,0 +1,14 @@
+kuadrant:
+  release: "1.0.1"
```
didierofrivia, line 2: Once this goes to main branch, all the versions will be `0.0.0` or something else was decided?

Boomatang: Nothing has being decided yet.

didierofrivia: In that case I'm in favour of `main` --> `0.0.0` as it's reflected currently on our manifests (OLM, Helm)

.github/workflows/release-operator.yaml
```diff
@@ -0,0 +1,30 @@
+name: Release Operator
+
+on:
+  pull_request:
+    types:
+    - closed
+    branches:
+      - 'release-v[0-9]+.[0-9]+'
```
didierofrivia, line 8: Was thinking that since we are targeting this workflow to be run once the candidate release merge into the release branch, we could do the following:

```suggestion
on:
  push:
    branches:
      - 'release-v[0-9]+.[0-9]+'
```

`push` since we'll be only merging once the candidate PR is approved, no?
Thus the conditional on line 13 won't be needed

.github/workflows/release.yaml
didierofrivia: Will add it in my following PR

didierofrivia (commented):
Given that everything regarding the automated process will be included in my following PR, matching the versions on main to `0.0.0` and reviewing the event that triggers the _Release Operator_ are the only changes requested from my side. Maybe @eguzki can mark as resolve or expand on his review?

Good work!

release.yaml
```diff
@@ -0,0 +1,14 @@
...
+  channels: 
+  - "alpha"
+  default_channel: "alpha"
+
+dependencies:
+  Authorino: "0.16.0"
+  Console_plugin: "0.0.14"
+  DNS: "0.12.0"
+  Limitador: "0.12.1"
+  Wasm_shim: "0.8.1"
```
didierofrivia, line 14:
```suggestion
version: "1.0.1"
dependencies:
  authorino-operator: "0.16.0"
  dns-operator: "0.12.0"
  limitador-operator: "0.12.1"
  console-plugin: "0.0.14"
  wasm-shim: "0.8.1"
olm:
  channels: 
  - "alpha"
  default_channel: "alpha"
```

didierofrivia: The suggestion above makes it much leaner, and also matches the real dependency names.

.github/workflows/release-validation.yaml
```diff
@@ -0,0 +1,24 @@
+name: Release Validation
+
+on:
+  pull_request:
+    types:
+    - opened
+    branches:
+      - 'release-v[0-9]+.[0-9]+'
```
didierofrivia, line 8: If this is meant for validating the RC branch, maybe the following will target better the event:

```suggestion
      - 'release-v[0-9]+.[0-9]+-?**'
```

.github/workflows/release-operator.yaml
didierofrivia: I reckon that the advantage would be that the work is not triggered when a PR is closed without merging, so no conditional needed in the job :)

release.yaml
Boomatang: I don't agree with the some of the changes suggested here. The dependency section changes you say are to match the real dependency names. They are abstract dependencies. If you want the real dependency then the full image names should be stated. That would be become interesting in the future as their is the possibility that how the wasm shim is severed will be changed. I don't see an argument for making the change.

On the making it leaner by removing the `kuadrant` section, going to disagree with this also as it hinders expending the configuration in the future. It is also common that the a global level configuration would apply a default to the section under it. Having it sectioned removes any of that ambiguity, it this version or any future version of the configuration.

Change the word `release` to `version`, I can see but also I can see why to leave it as is. When are creating a release we don't talk about the version we are creating but the release we are making. It is only after the release has being completed do we refer to a release as a version. A  number of tools will use the wording around release while other use version wording. I don't see a strong enough reason to make the change.

.github/workflows/release-validation.yaml
Boomatang: I am not a fan of using regex for any of this. I know some is a skill issue on my behalf, but that is why I am not going to change this. What you have suggested is not valid on https://regex101.com/ and as someone that does not have a depth understand of regex that is a tool I would use to validated what your suggesting. If I can't validate it now, a future contributor may also have the same issue.

release.yaml
maleck13: Naming things is hard. On the dependency map, I think I also prefer the component name that is the dependency (Authorino_Operator vs Authorino) as it more clearly conveys what the kuadrant operator interacts with and depends on. With the kuadrant->release vs version how big of a deal is this really? I am not sure anyone is going to be confused by that, whereas with the dependencies it is feasible that someone puts in the version of limitador vs the version of the limitador operator?

.github/workflows/release-validation.yaml
maleck13: The first one would still capture the RC right?
maybe something like? 
```
release-(v[0-9]+.[0-9])+(-[\w]+)? 
```
could be a minor improvement as it would capture the version and the rc for potential use? (by the way my regex foo is also weak)

didierofrivia: Sadly GH actions don't support full Regex, so the suggestion is considering GH limitations. We are currently using it in some workflows, here the PR reckoning the behaviour: https://github.com/Kuadrant/limitador-operator/pull/181 and https://github.com/Kuadrant/kuadrant-operator/pull/1162

eguzki: More food for thought: check out my comment regarding this: https://github.com/Kuadrant/limitador-operator/pull/181#pullrequestreview-2596181105

release.yaml
didierofrivia: I agree, naming is hard! I reckon that in this particular manifest, matching the real dependency name will help self documenting, also keeping consistency in naming conventions, in this case [kebab-case](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case). Regarding the `release` vs `version`, the yaml file is named _release.yaml_, and X.Y.Z is a semantic version, so it seems even if both would suggest the same, the later also aligns with OLM and Helm manifests.

Boomatang: Can someone put the a full example of the naming? This is not some thing I want us to be bike shedding on.

maleck13: what @didierofrivia has in the suggested commit is just that?

release.yaml
```diff
@@ -0,0 +1,14 @@
+kuadrant-operator:
+  version: "0.0.0"
+
+olm:
+  channels: 
+  - "0.0.0"
+  default-channel: "0.0.0"
```
didierofrivia, line 7: `default-channel` and `channels` first (and only so far) entry should be `alpha` I reckon, wdyt @eguzki @maleck13 

```suggestion
  channels: 
  - "alpha"
  default-channel: "alpha"
```

didierofrivia: or `default-channel: "stable"` (?)

utils/release/pre-validation/verify_release_file.sh
```diff
@@ -0,0 +1,31 @@
...
+    v=$(yq --unwrapScalar "$1" "$file" 2>/dev/null)
+    if [[ -z "$v" || "$v" == "null" ]]; then
+        echo "$2 is a required field. Please update."
+        has_error=1
+    fi
+}
+
+check_field ".olm.channels" "olm.channels"
+check_field ".olm.default-channel" "olm.default-channel"
+check_field ".kuadrant.version" "kuadrant.version"
```
didierofrivia, line 24:
```suggestion
check_field ".kuadrant-operator.version" "kuadrant-operator.version"
```

didierofrivia (commented):
I couldn't make it work following the verification steps, apparently there's a discrepancy with the CRD versions included in the bundle but not in the CSV. Someone else should assert the verification steps, then we merge and iterate.  Also the verification workflow might be still redundant.

maleck13:
Verification steps worked well for me

```
 git -C /tmp/review diff 1e417b087bd96a5a1f5fc0c3e53e2237b324c1f8
diff --git a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
index 8f8805f5..1d461e1d 100644
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -109,7 +109,7 @@ metadata:
     capabilities: Basic Install
     containerImage: quay.io/kuadrant/kuadrant-operator:v1.0.1
-    createdAt: "2024-12-05T13:08:35Z"
+    createdAt: "2025-02-26T14:24:23Z"
     description: A Kubernetes Operator to manage the lifecycle of the Kuadrant system
     operators.operatorframework.io/builder: operator-sdk-v1.32.0
diff --git a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml b/bundle/manifests/kuadrant-o
perator.clusterserviceversion.yaml
index 8f8805f5..1d461e1d 100644
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -109,7 +109,7 @@ metadata:
     capabilities: Basic Install
     containerImage: quay.io/kuadrant/kuadrant-operator:v1.0.1
-    createdAt: "2024-12-05T13:08:35Z"
+    createdAt: "2025-02-26T14:24:23Z"
     description: A Kubernetes Operator to manage the lifecycle of the Kuadrant system
     operators.operatorframework.io/builder: operator-sdk-v1.32.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v3
diff --git a/make/release.mk b/make/release.mk
deleted file mode 100644
index afc65131..00000000
--- a/make/release.mk
+++ /dev/null
@@ -1,17 +0,0 @@
-#Release default values
-AUTHORINO_OPERATOR_VERSION=0.16.0
-AUTHORINO_OPERATOR_BUNDLE_IMG=quay.io/kuadrant/authorino-operator-bundle:v0.16.0
-LIMITADOR_OPERATOR_VERSION=0.12.1
-LIMITADOR_OPERATOR_BUNDLE_IMG=quay.io/kuadrant/limitador-operator-bundle:v0.12.1
-DNS_OPERATOR_VERSION=0.12.0
-DNS_OPERATOR_BUNDLE_IMG=quay.io/kuadrant/dns-operator-bundle:v0.12.0
-WASM_SHIM_VERSION=0.8.1
-RELATED_IMAGE_WASMSHIM=oci://quay.io/kuadrant/wasm-shim:v0.8.1
-RELATED_IMAGE_CONSOLEPLUGIN=quay.io/kuadrant/console-plugin:v0.0.14
-IMG=quay.io/kuadrant/kuadrant-operator:v1.0.1
-BUNDLE_VERSION=1.0.1
-BUNDLE_IMG=quay.io/kuadrant/kuadrant-operator-bundle:v1.0.1
-CATALOG_IMG=quay.io/kuadrant/kuadrant-operator-catalog:v1.0.1
-CHANNELS=alpha
-BUNDLE_CHANNELS=--channels=alpha
-VERSION=1.0.1
```

KevFan:
@Boomatang asked me to run through the the verification steps since I'm using a Mac. Verification worked as expected for me also 💯

didierofrivia (approved):
Approving even if it didn't work on my local, since 3 others verified this. Let's merge and iterate

Issue #1123, opened 2025-01-16
Add test Helm upgrade path


maleck13:
**What**

n the Kuadrant operator github actions add a workflow that allows specifying two existing versions (upgrade from and upgrade to) and execute an upgrade from these versions using helm charts.

Issue #1112, opened 2025-01-14
Kuadrant Upgrades


maleck13:
This work is not directly impacted by the work in https://github.com/Kuadrant/kuadrant-operator/issues/1122 but is related to it

> Note while we call out workflows, these actions should be reproducable locally. The workflows should just take the actions a developer would (IE provide inputs and install CLI for example)

**What**

Once we have a release candidate or new release of kuadrant, we should be able to trigger testing the upgrade from the previous version to the new version works for the installation tools we support. 

Ideally we would install the previous version into a cluster such as a kind cluster (although it would be ideal to be able to target a cluster also ) and then trigger the upgrade to the new version. This workflow could take inputs of previous and next version and then test the upgrade from the previous all the way to the next version. 

I can imagine having 2 workflows. 

one for installing kudarant onto a kind cluster at a specific version or if possible a target cluster via the install method (IE OLM) at a specified version. Wondering if we had a drop down of (kind | ocp) and if OCP it expects a kubeconfig from a GH action secret with well known secret names?

and another for performing the actual upgrade that called the first one to setup the cluster this would be useful in general as then installing a new version on to an eng OCP for example would be a simple automated process. 

For doing an OLM based install we can use the follow guide to setup some useful files
https://docs.kuadrant.io/dev/install-olm/

For doing an OLM upgrade, I use the `operator-sdk run bundle-upgrade <target bundle>` command. This seemed to work well and had nice output. So if we have a new bundle, we should be able to install the previous version from the previous catalog image and then trigger an upgrade using the bundle upgrade command. This way we can allow the upstream catalogs to handle the meta data and we can just ensure the upgrade will work

PR #1156, opened 2025-02-03
Add test helm upgrade path

eguzki:
### What

Implements #1123 and is part of the work being done for https://github.com/Kuadrant/kuadrant-operator/issues/1112

It's a Github Action Workflow to test the Kuadrant upgrade process using **helm charts**.

It is being triggered manually (`on.workflow_dispatch`) and accepts as input:
* `kuadrantStartVersion`:  Kuadrant start version (*required*). For example: `1.0.0` (without `v` prefix)
* `kuadrantNamespace`: Namespace where Kuadrant is installed (*optional* defaults to `kuadrant-system`).
* `clusterServer`: Cluster server URL (*optional* defaults to local kind cluster). For example: `https://example.com:443`
* `clusterToken`:  Cluster Server Bearer Token (*optional* defaults to local kind cluster token). For example: `sha256~abcdefghijk12345432432`

The workflow installs published Kuadrant release from `https://kuadrant.io/helm-charts` with the version specified by `kuadrantStartVersion`. As of today, existing release versions:

```
❯ bin/helm search repo kuadrant -l
NAME                       	CHART VERSION	APP VERSION	DESCRIPTION                                       
kuadrant/kuadrant-operator 	1.0.2        	1.0.2      	The Operator to install and manage the lifecycl...
kuadrant/kuadrant-operator 	1.0.1        	1.0.1      	The Operator to install and manage the lifecycl...
kuadrant/kuadrant-operator 	1.0.0        	1.0.0      	The Operator to install and manage the lifecycl...
kuadrant/kuadrant-operator 	0.11.0       	           	The Operator to install and manage the lifecycl...
kuadrant/authorino-operator	0.16.0       	0.16.0     	Kubernetes operator for managing Authorino inst...
kuadrant/authorino-operator	0.15.1       	0.15.1     	Kubernetes operator for managing Authorino inst...
kuadrant/authorino-operator	0.14.0       	           	Kubernetes operator for managing Authorino inst...
kuadrant/authorino-operator	0.13.1       	           	Kubernetes operator for managing Authorino inst...
kuadrant/authorino-operator	0.13.0       	           	Kubernetes operator for managing Authorino inst...
kuadrant/authorino-operator	0.12.0       	           	Kubernetes operator for managing Authorino inst...
kuadrant/dns-operator      	0.12.0       	0.12.0     	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.11.0       	0.11.0     	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.10.1       	0.10.1     	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.10.0       	0.10.0     	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.9.0        	           	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.8.0        	           	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.7.0        	           	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.6.0        	           	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.5.0        	           	Kubernetes operator responsible for reconciling...
kuadrant/dns-operator      	0.4.1        	           	Kubernetes operator responsible for reconciling...
kuadrant/limitador-operator	0.12.1       	0.12.1     	Kubernetes operator for managing Limitador inst...
kuadrant/limitador-operator	0.12.0       	           	Kubernetes operator for managing Limitador inst...
kuadrant/limitador-operator	0.11.0       	           	Kubernetes operator for managing Limitador inst...
```
Then, deploys kuadrant CR and checks status is *Ready*. The kuadrant operator, runs a set of tests to mark the resource status as ready, so effectively, the workflow is relying on those tests to verify installation. 

When the starting point installation is ready, it proceeds with the upgrade process. The upgrade is being carried out running [helm upgrade](https://helm.sh/docs/helm/helm_upgrade/) command. The command waits until all deployments are ready (with a timeout set to 3 mins) and fails if some deployment is not ready after the timeout expires. Upgrade may fail for instance if some referenced docker image is not valid or does not exist. The upgrade verification steps by the workflow relies, once again, on the kuadrant CR status. After the upgrade, the kuadrant CR status should be Ready to pass the test. For example, if for whatever reason, let's say the limitador operator, or limitador instance, is not up and running, kuadrant CR would report on the status and the upgrade test would fail. 

Simple commands are being coded as part of the Github workflow. For high level commands, those are implemented using makefile targets or third party tools so they can be easily run locally to dev/test.

### DIFF
--- a/.github/workflows/test-upgrade.yaml
+++ b/.github/workflows/test-upgrade.yaml
@@ -97,10 +97,19 @@ jobs:
           --version ${{ inputs.kuadrantStartVersion }} \
           --namespace ${{ inputs.kuadrantNamespace }} \
           --create-namespace
+      - name: Deploy Kuadrant
+        run: |
+          kubectl -n ${{ inputs.kuadrantNamespace }} apply -f - <<EOF
+          apiVersion: kuadrant.io/v1beta1
+          kind: Kuadrant
+          metadata:
+            name: kuadrant
+          spec: {}
+          EOF
       - name: Verify Kuadrant installation
         run: |
-          # TODO
-          echo "kuadrant installation ✅" &&  exit 0
+          kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant -n ${{ inputs.kuadrantNamespace }}
+          echo "kuadrant installation ✅"
       - name: Build local chart dependencies for version ${{ steps.upgrade-version.outputs.version }}
         run: |
           make helm-dependency-build
@@ -122,5 +131,5 @@ jobs:
         run: exit 1
       - name: Verify Kuadrant upgrade
         run: |
-          # TODO
-          echo "kuadrant upgrade ✅" &&  exit 0
+          kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant -n ${{ inputs.kuadrantNamespace }}
+          echo "kuadrant upgrade ✅"
--- a/.github/workflows/test-upgrade.yaml
+++ b/.github/workflows/test-upgrade.yaml
@@ -29,6 +29,7 @@ jobs:
       KIND_CLUSTER_NAME: kuadrant-test
       K8S_USER: kuadrant-admin  # can be whatever, it does not matter.
       CLUSTER_NAME: remote-cluster
+      LOCAL_CLUSTER: ${{ inputs.clusterServer == '' && inputs.clusterToken == '' }}
     steps:
       - name: Check out code
         uses: actions/checkout@v4
@@ -47,7 +48,7 @@ jobs:
       - name: Print versions
         run: echo "Installing version ${{ inputs.kuadrantStartVersion }}, upgrading to version ${{ steps.upgrade-version.outputs.version }}"
       - name: Deploy local Kind cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        if: ${{ env.LOCAL_CLUSTER }}
         uses: helm/kind-action@v1.12.0
         with:
           version: v0.23.0
@@ -55,18 +56,18 @@ jobs:
           cluster_name: ${{ env.KIND_CLUSTER_NAME }}
           wait: 120s
       - name: Install kubectl for remote cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         uses: azure/setup-kubectl@v4
         with:
           version: v1.25.3
       - name: Mask cluster token
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterToken != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         run: |
           CLUSTER_TOKEN=$(jq -r '.inputs.clusterToken' $GITHUB_EVENT_PATH)
           echo ::add-mask::$CLUSTER_TOKEN
           echo CLUSTER_TOKEN=$CLUSTER_TOKEN >> $GITHUB_ENV
       - name: Setup kubectl for remote cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         run: |
           kubectl config set-credentials ${{ env.K8S_USER }}/${{ env.CLUSTER_NAME }} --token ${{ env.CLUSTER_TOKEN }}
           kubectl config set-cluster ${{ env.CLUSTER_NAME }} --insecure-skip-tls-verify=true --server=${{ inputs.clusterServer }}
@@ -77,7 +78,7 @@ jobs:
           kubectl cluster-info
           kubectl get nodes
       - name: Deploy pre-requisites on local Kind cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        if: ${{ env.LOCAL_CLUSTER }}
         run: |
           make install-metallb
           make install-cert-manager
@@ -85,7 +86,6 @@ jobs:
           make deploy-eg-gateway
       - name: Install helm and add Kuadrant repo
         run: |
-          make helm
           make helm-add-kuadrant-repo
           bin/helm search repo kuadrant
       - name: Install Kuadrant ${{ inputs.kuadrantStartVersion }}
--- a/.github/workflows/test-upgrade.yaml
+++ b/.github/workflows/test-upgrade.yaml
@@ -25,7 +25,6 @@ jobs:
   helm-charts-upgrade-test:
     runs-on: ubuntu-latest
     name: Helm Charts Upgrade Test
-    if: ${{ github.event_name == 'workflow_dispatch' }}
     env:
       KIND_CLUSTER_NAME: kuadrant-test
       K8S_USER: kuadrant-admin  # can be whatever, it does not matter.
--- a/.github/workflows/test-upgrade.yaml
+++ b/.github/workflows/test-upgrade.yaml
@@ -1,0 +1,126 @@
+---
+name: Kuadrant Upgrade Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      kuadrantStartVersion:
+        description: Kuadrant start version
+        required: true
+        type: string
+      clusterServer:
+        description: Cluster server URL
+        required: false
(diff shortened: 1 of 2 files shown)


.github/workflows/test-upgrade.yaml
```diff
@@ -0,0 +1,135 @@
...
+    inputs:
+      kuadrantStartVersion:
+        description: Kuadrant start version
+        required: true
+        type: string
+      clusterServer:
+        description: Cluster server URL
+        required: false
+        type: string
+      clusterToken:
```
didierofrivia, line 15: I'm not sure how secure is to pass down sensitive info like tokens from inputs, might want to check the docs or at least the full output logs with debug mode enabled.

.github/workflows/test-upgrade.yaml
```diff
@@ -0,0 +1,135 @@
...
+name: Kuadrant Upgrade Test
+
+on:
+  workflow_dispatch:
+    inputs:
+      kuadrantStartVersion:
+        description: Kuadrant start version
+        required: true
+        type: string
+      clusterServer:
```
didierofrivia, line 11: Is having an external k8s cluster desired by these kind of tests? If not, we can just rely on kind

.github/workflows/test-upgrade.yaml
```diff
@@ -0,0 +1,135 @@
...
+      kuadrantNamespace:
+        description: Namespace where Kuadrant is installed
+        required: false
+        default: kuadrant-system
+        type: string
+jobs:
+  helm-charts-upgrade-test:
+    runs-on: ubuntu-latest
+    name: Helm Charts Upgrade Test
+    if: ${{ github.event_name == 'workflow_dispatch' }}
```
didierofrivia, line 28: not needed since the only event listener is `workflow_dispatch` https://github.com/Kuadrant/kuadrant-operator/pull/1156/files#diff-e3c60a7e344a4151f98dbfd60770be22bc6169d9c3c2cb7c4748529a3e9a6333R4-R5

.github/workflows/test-upgrade.yaml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Verify Kuadrant installation
+        run: |
+          kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant -n ${{ inputs.kuadrantNamespace }}
+          echo "kuadrant installation ✅"
+      - name: Build local chart dependencies for version ${{ steps.upgrade-version.outputs.version }}
+        run: |
+          make helm-dependency-build
+      - name: Start upgrade to kuadrant ${{ steps.upgrade-version.outputs.version }}
+        run: |
+          bin/helm upgrade kuadrant charts/kuadrant-operator \
```
didierofrivia, line 118: we are not gonna be upgrading to pre-releases charts, are we?

make/helm.mk
```diff
@@ -89,3 +89,10 @@ helm-sync-package-deleted: ## Sync the deleted helm chart package to the helm-ch
 	  -H "X-GitHub-Api-Version: 2022-11-28" \
 	  https://api.github.com/repos/$(ORG)/$(HELM_REPO_NAME)/dispatches \
 	  -d '{"event_type":"chart-deleted","client_payload":{"chart":"$(CHART_NAME)","version":"$(CHART_VERSION)"}}'
+
+.PHONY: helm-print-chart-version
+helm-print-chart-version: $(HELM) ## Reads chart version
+	@$(HELM) show chart charts/$(REPO) | grep -E "^version:" | awk '{print $$2}'
```
didierofrivia, line 95: given this will show the local info in `charts/Chart.yaml`, this workflow is only meant to be run from a branch (i.e. release branch) that has the new helm chart package for the release, right?

.github/workflows/test-upgrade.yaml
didierofrivia: because if we are, we might need to pass the `--devel` flag

didierofrivia:
It's looking good, I'd like to know if it's meant to run in an external k8s cluster, and if so, we might want to see if the secrets are treated correctly. Also, in that case, a cleaning step might be needed.

.github/workflows/test-upgrade.yaml
eguzki: there is one specific step to have the token redacted `Mask cluster token`. Good enough?

.github/workflows/test-upgrade.yaml
eguzki: Upgrade to pre-release charts can be tested.

The target (version being upgraded to) chart is determined by the git reference where you run the tests. Thus, the chart is available locally in the filesystem, hence the path `charts/kuadrant-operator`.

make/helm.mk
eguzki: right

.github/workflows/test-upgrade.yaml
didierofrivia: Mhmm I don't know honestly, but we could try with a PR to this branch, see if running the workflow in debugging mode exposes the token

.github/workflows/test-upgrade.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Deploy local Kind cluster
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        uses: helm/kind-action@v1.12.0
+        with:
+          version: v0.23.0
+          config: utils/kind-cluster.yaml
+          cluster_name: ${{ env.KIND_CLUSTER_NAME }}
+          wait: 120s
+      - name: Install kubectl for remote cluster
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer != '' }}
```
didierofrivia, line 58: What we could do is having a job scoped env: 

```yaml
env:
  LOCAL_CLUSTER: ${{ inputs.clusterServer == '' && inputs.clusterToken != ''}}
```

thus in checks like this, we could do:

```suggestion
        if: ${{ !env.LOCAL_CLUSTER }}
```

also, the `github.event_name == 'workflow_dispatch'` is not needed in the conditional

wdyt?

.github/workflows/test-upgrade.yaml
didierofrivia: From the https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-secrets


> **Audit how secrets are handled**
> 
>     Audit how secrets are used, to help ensure they’re being handled as expected. You can do this by reviewing the source code of the repository executing the workflow, and checking any actions used in the workflow. For example, check that they’re not sent to unintended hosts, or explicitly being printed to log output.
>     View the run logs for your workflow after testing valid/invalid inputs, and check that secrets are properly redacted, or not shown. It's not always obvious how a command or tool you’re invoking will send errors to STDOUT and STDERR, and secrets might subsequently end up in error logs. As a result, it is good practice to manually review the workflow logs after testing valid and invalid inputs. For information on how to clean up workflow logs that may unintentionally contain sensitive data, see [Using workflow run logs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/using-workflow-run-logs#deleting-logs).

eguzki: This is not a secret, this is an input from a workflow_dispatch, so it needs to be masked explicitly

.github/workflows/test-upgrade.yaml
eguzki: doing that, yes

.github/workflows/test-upgrade.yaml
didierofrivia: That's true, I was only concerned since the inputs are not masked, and passed around would show in the logs when the debug mode is enabled, we could check with a workflow dispatch and see how it shows

make/helm.mk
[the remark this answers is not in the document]
Boomatang: @didierofrivia General question on why we should stop using the $REPO var?

didierofrivia (approved):
Let’s try it out with an alpha release and check the logs for any possible leaked token

make/helm.mk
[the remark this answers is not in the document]
didierofrivia: because even if $REPO is the same name of the packaged chart (in this case), it's not clear that is, in fact, the name of the chart... also it doesn't tell a good story for anyone new reading the code ;)

PR #1162, opened 2025-02-07
Triggering builds on full semver tags

didierofrivia:
This PR modifies the event listener for pushing tags following a GH action regex matching semver with pre release and build metadata parts, like vx.y.z-alpha1, vx.y.z-rc1, vx.y.z-dev, etc. Previously it only matched vx.y.z

Following https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Notes:

This will also match vx.y.zwhatever, GH actions regex is limited and we can't use a proper regex for semver like in https://semver.org/

### DIFF
--- a/.github/workflows/build-images-for-tag-release.yaml
+++ b/.github/workflows/build-images-for-tag-release.yaml
@@ -3,7 +3,7 @@ name: Build and Publish Images For Tag Release
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


eguzki (approved):
LGTM 

For the record, same comment as https://github.com/Kuadrant/limitador-operator/pull/181#pullrequestreview-2596181105

Issue #1158, opened 2025-02-04
Observability Configuration Phase 1: Documentation


david-martin:
Follow on from https://github.com/Kuadrant/kuadrant-operator/issues/1065 , update the guide at https://docs.kuadrant.io/dev/kuadrant-operator/doc/user-guides/full-walkthrough/secure-protect-connect/#optional, and any other locations where Service/PodMonitors are mentioned to reference using this new configuration.
A new guide specifically for enabling Observability, explaining what it does and its intended use would be good to write.

Issue #1141, opened 2025-01-28
Observability Configuration Phase 2: Set OTLP Collector for all components


david-martin:
Currently, if using tracing, the OTLP collector needs to be configured in multiple places (istio, limitador, authorino) as per https://docs.kuadrant.io/dev/kuadrant-operator/doc/observability/tracing/ 

Expose the collector url in the spec of the Kuadrant CR, like this:

```yaml
observability:
  otlp_collector: rpc://temp:6379
```

The kuadrant operator should set all configuration points of components to use the configured collector.
As part of this work, update the docs guide to reflect how to enable tracing using this new configuration.

PR #1164, opened 2025-02-10
Add observability feature flag to Kuadrant CRD

### LINKED ISSUE #1167: Add new observability reconciler to the Kuadrant operator

david-martin:
Closes #1167
Supercedes #1143 (not using a fork to allow all tests to have access to secrets)
Also adds a new user guide as part of #1158 (separate PR to add it to docs site will follow)

A new field is introduced to the Kuadrant CRD `spec.observability.enable`.
If `true`, various ServiceMonitors and PodMonitors will be created:

- Kuadrant operator servicemonitor
- Authorino operator servicemonitor
- Limitador operator servicemonitor
- DNS operator servicemonitor

Also, for each Gateway in the toplogy, the following monitors are created (depending on the Gateway provider):

- Istio Gateways
    - Istiod ServiceMonitor
    - Istio gw pod (envoy) stats PodMonitor
- Envoy Gateway Gateways
    - Envoy Gateway ServiceMonitor
    - Envoy Proxy Stats PodMonitor 

To accomplish this, a new reconciler is added `observability_reconciler.go`
All monitors are added to the topology.
The 4 core component monitors are linked to the root Kuarant CR in the topology.
The other monitors are not linked, but do show up in the topology.

fyi @jasonmadigan , as this will affect the topology view in the openshift console plugin.

A set of integration tests are included.
1 for the core component monitors, 1 for istio gateways and 1 for envoy gateways.
In either of these tests, a Kuadrant CR is created without the new flag set, and monitors verified as *not* created.
Then the flag is set, and monitors verified as created.
Finally, the flag is unset, and monitors verified as deleted.


david-martin:
@Boomatang @eguzki I'd appreciate another look at these changes when you get a chance.
I've made a number of changes since your initial review on (now closed) #1143

make/development-environments.mk
```diff
@@ -15,6 +15,10 @@ install-metallb: kustomize yq ## Installs the metallb load balancer allowing use
 	kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=60s
 	./utils/docker-network-ipaddresspool.sh kind $(YQ) ${SUBNET_OFFSET} | kubectl apply -n metallb-system -f -
 
+.PHONY: install-observability-crds
+install-observability-crds: $(KUSTOMIZE)
+	$(KUSTOMIZE) build ./config/observability/| docker run --rm -i docker.io/ryane/kfilt -i kind=CustomResourceDefinition | kubectl apply --server-side -f -
```
Boomatang, line 20: The `docker` command needs to be changed to use the CONTAINER_ENGINE. As it is this will not work with podman

david-martin:
@jasonmadigan here's an example of the topology with the various monitors created, and using an istio gateway

image

Boomatang (commented):
I have looked over this, and there is a comments that I wanted to make.

I have not ran the code. I seem to have some issue with checking out the branch. I assume that is a me problem.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,435 @@
...
+func NewObservabilityReconciler(client *dynamic.DynamicClient, rm meta.RESTMapper) *ObservabilityReconciler {
+	return &ObservabilityReconciler{
+		Client:     client,
+		restMapper: rm,
+	}
+}
+
+func (r *ObservabilityReconciler) Subscription() *controller.Subscription {
+	return &controller.Subscription{ReconcileFunc: r.Reconcile, Events: []controller.ResourceEventMatcher{
+		{Kind: ptr.To(kuadrantv1beta1.KuadrantGroupKind)},
```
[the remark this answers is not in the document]
david-martin: > I would have assumed delete events would want to be watched

That makes sense.
If one is deleted, it should be recreated.
I'll update this.

> In the case of update, are you allowing the user to override all the values

I think it could get overly complex if we start to worry about updates to the monitors.
Ignoring them sounds like a reasonable approach, provided we document this is the behaviour (I can add this to the guide in this PR, if we're in agreement)

> should create and delete events not be watched on gatewayclasses
> Would you not want to watch for all events on the gateways?

Agreed. 
I'll make an update to do this.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,435 @@
...
+	kuadrantObservabilityLabel = "kuadrant-observability"
+	kOpMonitorName             = "kuadrant-operator-monitor"
+	dnsOpMonitorName           = "dns-operator-monitor"
+	authOpMonitorName          = "authorino-operator-monitor"
+	limitOpMonitorName         = "limitador-operator-monitor"
+	istiodMonitorName          = "istiod-monitor"
+	istiodMonitorNS            = "istio-system"
+	istioPodMonitorName        = "istio-pod-monitor"
+	envoyGatewayMonitorName    = "envoy-gateway-monitor"
+	envoyGatewayMonitorNS      = "envoy-gateway-system"
```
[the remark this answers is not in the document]
david-martin: My understanding is these are the default namespaces where the gateway system pieces are installed.
For istio at least, that can be overwritten. I haven't seen if that's possible in EG.
Either way, I think we can mention this in the documentation to avoid unexpected behaviour.

An alternative would be to add some dynamic lookup logic to detect the namespace, but I think that could be added at a later point if users find the observability feature useful, but need to run things in a different namespace.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,435 @@
...
+					logger.Error(err, "failed to create envoy gateway monitor", "status", "error")
+					return err
+				}
+				if err := r.createMonitor(ctx, monitorObjs, envoyStatsMonitorSpec, gateway.GetNamespace(), logger); err != nil {
+					logger.Error(err, "failed to create envoy stats monitor", "status", "error")
+					return err
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
david-martin: Thanks for the suggestion.
That makes sense. I'll move it outside the loop

Boomatang:
Hay @david-martin, I finally managed to clone this and when I tried to run it with `make local-setup` I got the following error.
```
{"level":"error","ts":"2025-02-11T17:25:20Z","logger":"kuadrant-operator","msg":"Unhandled Error","logger":"UnhandledError","error":"pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watch *v1.PodMonitor: failed to list *v1.PodMonitor: podmonitors.monitoring.coreos.com is forbidden: User \"system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager\" cannot list resource \"podmonitors\" in API group \"monitoring.coreos.com\" at the cluster scope","stacktrace":"k8s.io/client-go/tools/cache.DefaultWatchErrorHandler\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:158\nk8s.io/client-go/tools/cache.(*Reflector).Run.func1\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:308\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/backoff.go:226\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/backoff.go:227\nk8s.io/client-go/tools/cache.(*Reflector).Run\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:306\nk8s.io/client-go/tools/cache.(*controller).Run.(*Group).StartWithChannel.func2\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/wait.go:55\nk8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/wait.go:72"}
```

controllers/observability_reconciler.go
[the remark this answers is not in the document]
Boomatang: The mention of this in documentation should be reasonable. Is there any point in making an issues now for the dynamic lookup logic so we know have a record that it should be done.

controllers/observability_reconciler.go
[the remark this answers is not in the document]
Boomatang: I am always in favor of allowing the user to modify resource. So if we document the behavior as create and delete only, I would be happy.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,435 @@
...
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
```
eguzki, line 20: You need to grant permission to the operator to read/write/update monitoring resources, the operator is dying because it does not have permissions to watch resources:

```
{"level":"info","ts":"2025-02-12T10:22:15Z","logger":"kuadrant-operator","msg":"pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: failed to list *v1.ServiceMonitor: servicemonitors.monitoring.coreos.com is forbidden: User \"system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager\" cannot list resource \"servicemonitors\" in API group \"monitoring.coreos.com\" at the cluster scope"}
{"level":"error","ts":"2025-02-12T10:22:15Z","logger":"kuadrant-operator","msg":"Unhandled Error","logger":"UnhandledError","error":"pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:243: Failed to watch *v1.ServiceMonitor: failed to list *v1.ServiceMonitor: servicemonitors.monitoring.coreos.com is forbidden: User \"system:serviceaccount:kuadrant-system:kuadrant-operator-controller-manager\" cannot list resource \"servicemonitors\" in API group \"monitoring.coreos.com\" at the cluster scope","stacktrace":"k8s.io/client-go/tools/cache.DefaultWatchErrorHandler\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:158\nk8s.io/client-go/tools/cache.(*Reflector).Run.func1\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:308\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/backoff.go:226\nk8s.io/apimachinery/pkg/util/wait.BackoffUntil\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/backoff.go:227\nk8s.io/client-go/tools/cache.(*Reflector).Run\n\t/go/pkg/mod/k8s.io/client-go@v0.31.1/tools/cache/reflector.go:306\nk8s.io/client-go/tools/cache.(*controller).Run.(*Group).StartWithChannel.func2\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/wait.go:55\nk8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/wait/wait.go:72"}
```

It can be done adding a line here and then running "make generate && make manifests" (then `make bundle && make helm-build`)

```
//+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;podmonitors,verbs=get;list;watch;create;update;patch;delete
``` 
(I did not try that line, could contain typos)

eguzki (changes requested):

api/v1beta1/topology.go
```diff
@@ -61,3 +62,33 @@ func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
...
+func LinkKuadrantToServiceMonitor(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantGroupKind), controller.ObjectAs[machinery.Object])
+
+	return machinery.LinkFunc{
+		From: KuadrantGroupKind,
+		To:   schema.GroupKind{Group: monitoringv1.SchemeGroupVersion.Group, Kind: monitoringv1.ServiceMonitorsKind},
+		Func: func(child machinery.Object) []machinery.Object {
+			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
+				// ServiceMonitors outside the Kuadrant namespace will be outside the topology tree
+				return kuadrant.GetNamespace() == child.GetNamespace()
```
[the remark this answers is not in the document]
david-martin: Unless my understanding is flawed, will monitors only with the `kuadrant-observability=true` label be considered/filtered as well based on the changes in state_of_the_world.go?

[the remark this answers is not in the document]
david-martin: I did a quick test locally with an extra servicemonitor without the label, in the kaudrant namespace.
Here's all monitors first:

```shell
kubectl get servicemonitor,podmonitor -A
NAMESPACE         NAME                                                              AGE
kuadrant-system   servicemonitor.monitoring.coreos.com/authorino-operator-monitor   103s
kuadrant-system   servicemonitor.monitoring.coreos.com/dns-operator-monitor         103s
kuadrant-system   servicemonitor.monitoring.coreos.com/kuadrant-operator-monitor    103s
kuadrant-system   servicemonitor.monitoring.coreos.com/limitador-operator-monitor   103s
kuadrant-system   servicemonitor.monitoring.coreos.com/test-aaaa                    43s
```

And the topology view, which only shows the 4 labelled monitors
image

controllers/observability_reconciler.go
david-martin: Thanks @eguzki

api/v1beta1/topology.go
```diff
@@ -61,3 +62,33 @@ func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
...
+
+func LinkKuadrantToServiceMonitor(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(KuadrantGroupKind), controller.ObjectAs[machinery.Object])
+
+	return machinery.LinkFunc{
+		From: KuadrantGroupKind,
+		To:   schema.GroupKind{Group: monitoringv1.SchemeGroupVersion.Group, Kind: monitoringv1.ServiceMonitorsKind},
+		Func: func(child machinery.Object) []machinery.Object {
+			return lo.Filter(kuadrants, func(kuadrant machinery.Object, _ int) bool {
+				// ServiceMonitors outside the Kuadrant namespace will be outside the topology tree
```
eguzki, line 74: Did you consider filtering by label instead of namespace? The operator might not be installed where the kuadrant CR is being created. For instance in Openshift, operators (by default) go to `openshift-operator` whereas the kuadrant CR is usually created in `kuadrant-system`

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -109,7 +109,7 @@ metadata:
     capabilities: Basic Install
     categories: Integration & Delivery
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2024-12-09T09:56:27Z"
+    createdAt: "2025-02-10T11:02:45Z"
```
eguzki, line 112: I expect kuadrant operator new required permission to lie here. Try running `make bundle` and see if this file is changed.

make/development-environments.mk
```diff
@@ -15,6 +15,10 @@ install-metallb: kustomize yq ## Installs the metallb load balancer allowing use
 	kubectl -n metallb-system wait --for=condition=ready pod --selector=app=metallb --timeout=60s
 	./utils/docker-network-ipaddresspool.sh kind $(YQ) ${SUBNET_OFFSET} | kubectl apply -n metallb-system -f -
 
+.PHONY: install-observability-crds
+install-observability-crds: $(KUSTOMIZE)
+	$(KUSTOMIZE) build ./config/observability/| $(CONTAINER_ENGINE) run --rm -i docker.io/ryane/kfilt -i kind=CustomResourceDefinition | kubectl apply --server-side -f -
```
eguzki, line 20: RH does not meet `docker.io` license agreements. Can we still use it within kuadrant? 

If not, we can find an alternative way to do this, either using `yq, `jq` or something else (maybe kustomize??)

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,420 @@
...
+		logger.V(1).Info("deleting any existing monitors", "kuadrant", kObj != nil)
+		r.deleteAllMonitors(ctx, monitorObjs, logger)
+		return nil
+	}
+
+	// Create all monitors
+	logger.V(1).Info("observability enabled, creating monitors")
+
+	// Kuadrant Operator monitor
+	r.createMonitor(ctx, monitorObjs, kOpMonitorSpec, kObj.Namespace, logger)
```
eguzki, line 344: check my comment about namespaces. `OPERATOR_NAMESPACE` env var could be a better option. It can be used via global variable `operatorNamespace` in this `controllers` package.

eguzki: Same for the other operators. I think we can make a "safe" assumption (this will shoot back in our face sooner or later) that all operators are being installed in the same namespace as the kuadrant operator.

eguzki:
My next comment is by no means a request for change. Just a way to record my thinking. 

Exposing the monitoring resources _of the operators_ from the Kuadrant CR does make much sense to me. Kuadrant operators are cluster wide, hence _singletons_ in the cluster. Kuadrant only supports one instance of the Kuadrant CRD, as of today. That's why it works as it is implemented today. Tomorrow? who knows? I can see that one customer might want a _dedicated_ kuadrant instance to isolate from from kuadrant issues generated by other customers. Kuadrant instance being: authorino instance, limitador instance and instrumentation of a specific set of gateways (so kuadrant instance injects specific wasm module to that set of gateways). 

I can totally see some `spec.observability` on Limitador CR or Authorino CR for data plane components. Kuadrant CR is more about the data plane configuration that the control plane configuration. In my mind. 

A different approach would be having the kuadrant operator deploying monitoring resources for the operators always by default. Unless monitoring CRDs (aka APIs)  have not been installed in the cluster. A cluster admin could opt-out of this default behavior using some operator environment variable that can be added from the operator OLM subscription resource or some Helm [Value](https://helm.sh/docs/chart_template_guide/values_files/) specific for that.

eguzki:
Just in case from my previous comments there is any little doubt, I want to explicitly acknowledge the awesome work being done here :medal_military: :medal_sports:

david-martin:
@Boomatang @eguzki I believe I've addressed the latest round of comments and resolved each.
Would you mind giving the changes another pass to see if we're closer to landing them?
Thanks

david-martin:
Bump @Boomatang @eguzki if you get a chance. Thank you

controllers/state_of_the_world.go
```diff
@@ -404,6 +447,7 @@ func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
 			NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run,
 			NewDataPlanePoliciesWorkflow(b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run,
 			NewKuadrantStatusUpdater(b.client, b.isGatewayAPIInstalled, b.isGatewayProviderInstalled(), b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Subscription().Reconcile,
+			NewObservabilityReconciler(b.client, b.manager.GetRESTMapper(), operatorNamespace).Subscription().Reconcile,
```
Boomatang, line 450: I have being question my self why only the observability reconciler requires the `GetRESTMapper`, while no other reconcile does. It seems like the only place that it is required is setting up for the `r.Client.Resource`, and that is done in other reconcilers also. So like in a previous comment I am wondering would the better thing to do be set up vars like there is in `pkg/istio/utils.go`. This I believe would remove the dependency on the `GetRESTMapper`.

Boomatang (commented):
Some comments on the changes from the last time. My biggest concern is the use of the `GetRESTMapper`

api/v1beta1/topology.go
```diff
@@ -5,9 +5,13 @@ import (
 	limitadorv1alpha1 "github.com/kuadrant/limitador-operator/api/v1alpha1"
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/samber/lo"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+
+	kuad "github.com/kuadrant/kuadrant-operator/pkg/kuadrant"
```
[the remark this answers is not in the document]
david-martin: I'll swamp some things around with some params so `kuadrant` can be used

controllers/state_of_the_world.go
david-martin: I'm not completely following what you're suggesting.
The code in observability_reconciler.go that relies on the restmapper is used to create a monitor (Pod or Service), and is generic in that it needs the restmapping capability to set up the resource client based on the kind.
Are you suggesting something in utils that sets up those clients ahead of time?

david-martin: > My biggest concern is the use of the GetRESTMapper
What is the concern with using it?

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,422 @@
...
+		Name: dnsOpMonitorName,
+		Labels: map[string]string{
+			kuadrant.ObservabilityLabel:    "true",
+			"control-plane":                "controller-manager",
+			"app.kubernetes.io/name":       "servicemonitor",
+			"app.kubernetes.io/instance":   "controller-manager-metrics-monitor",
+			"app.kubernetes.io/component":  "metrics",
+			"app.kubernetes.io/created-by": "dns-operator",
+			"app.kubernetes.io/part-of":    "dns-operator",
+			"app.kubernetes.io/managed-by": "kustomize",
```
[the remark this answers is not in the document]
david-martin: They look generated perhaps.
I took the monitors from the operators.yaml file

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,422 @@
...
+		restMapper: rm,
+		namespace:  namespace,
+	}
+}
+
+func (r *ObservabilityReconciler) Subscription() *controller.Subscription {
+	return &controller.Subscription{ReconcileFunc: r.Reconcile, Events: []controller.ResourceEventMatcher{
+		{Kind: &kuadrantv1beta1.KuadrantGroupKind},
+		{Kind: ptr.To(schema.GroupKind{Group: monitoringv1.SchemeGroupVersion.Group, Kind: monitoringv1.ServiceMonitorsKind}), EventType: ptr.To(controller.DeleteEvent)},
+		{Kind: ptr.To(schema.GroupKind{Group: monitoringv1.SchemeGroupVersion.Group, Kind: monitoringv1.PodMonitorsKind}), EventType: ptr.To(controller.DeleteEvent)},
```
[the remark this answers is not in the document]
david-martin: Not really. It's used in 2 places.
Maybe if it was used in more.
Also, I can't think of a good place to put it that isn't a some utils file. (pkg/observability/utils ?)
I'd prefer if there was a better pattern for this kind of thing that I could follow, like a schemas or kinds file.

controllers/observability_reconciler.go
```diff
@@ -0,0 +1,422 @@
...
+				r.createMonitor(ctx, monitorObjs, envoyGatewayMonitorSpec, envoyGatewayMonitorNS, logger)
+				r.createMonitor(ctx, monitorObjs, envoyStatsMonitorSpec, gateway.GetNamespace(), logger)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (r *ObservabilityReconciler) createMonitor(ctx context.Context, monitorObjs []machinery.Object, monitor client.Object, ns string, logger logr.Logger) {
```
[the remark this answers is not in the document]
david-martin: You are correct.
There is an error in the log line where it is outputting `monitor.GetNameSpace()` instead of `ns`
I'll fix this.

Though the `ns` param isn't strictly needed, I prefer passing it in rather than having to update the namespace in the spec object, opting for the spec to never have a namespace, given that it's dynamic based on various things.

controllers/state_of_the_world.go
david-martin: Discussed offline, the result of which is a refactor of the schemas out to pkg/observability/schemas.go file

controllers/observability_reconciler.go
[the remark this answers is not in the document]
david-martin: Discussed offline, the result of which is a refactor of the schemas out to pkg/observability/schemas.go file

controllers/state_of_the_world.go
```diff
@@ -392,6 +396,46 @@ func (b *BootOptionsBuilder) getAuthorinoOperatorOptions() []controller.Controll
...
 
+func (b *BootOptionsBuilder) getObservabilityOptions() []controller.ControllerOption {
+	var opts []controller.ControllerOption
+	var err error
+
+	b.isPrometheusOperatorInstalled, err = utils.IsCRDInstalled(b.manager.GetRESTMapper(), monitoringv1.SchemeGroupVersion.Group, monitoringv1.ServiceMonitorsKind, monitoringv1.SchemeGroupVersion.Version)
+	if err != nil || !b.isPrometheusOperatorInstalled {
+		b.logger.Info("prometheus operator is not installed (ServiceMonitor CRD not found), skipping related watches and reconcilers", "err", err)
+		return opts
+	}
```
KevFan, line 407: We also report missing dependency on the affected CRs. Since the field for enabling/disabling obsersability is at the `Kuadrant` CR, do we want to report if the prometheus dependency is missing on the `Kuadrant` CR status also? 🤔 

https://github.com/Kuadrant/kuadrant-operator/blob/5f5dce36389a3f54b3480b599fe2a7aac85927f2/controllers/kuadrant_status_updater.go#L155-L181

controllers/state_of_the_world.go
Boomatang: I was expecting to see the `GetRESTMapper` removed, but with how the delete all is implemented it makes sense why it is still there. I have no issue with it being there, only a little surprised that's all.

controllers/state_of_the_world.go
Boomatang: I am not sure if that is what we want to do all the time. As this is a feature that is enabled in the kuadrant CR I think updating the status to say dependencies are missing would only be required if the feature is enabled. 

I think the status should be addressed but it could be a follow up from this. This is introducing the configuration in the kuadrant CR which I would like to see main so more people could have a look.

KevFan: Yes, I agree that it should be reported only when the feature is enabled and can be done as a follow up

Boomatang (approved):
Looks good, service monitor are created as expected. A follow up on the status will be required

david-martin:
Thanks @Boomatang @KevFan 

I've added words to https://github.com/Kuadrant/kuadrant-operator/issues/1141 about reporting on the status in future.

Issue #1117, opened 2025-01-14
Add a test OLM upgrade path workflow


maleck13:
**What**

In the Kuadrant operator github actions add a workflow that allows specifying two existing versions (upgrade from and upgrade to) and execute an upgrade from these versions using OLM.

Ideally this workflow would simply call a script or make command that could also be ran locally or in another CI flow against any target cluster that has OLM installed on it. For this workflow it should be fine to test on a kind cluster that has OLM installed on it.

Boomatang:
For RHOAM I have done a some work to be able to do this locally. You could give it a version "A" as a start point and version "X" as a end point. It would work out the chain based on the release versions. There maybe some work there that we can reuse. 
https://github.com/Boomatang/orc-dev-cli/blob/main/docs/commnads/rhoam.md#index

maleck13:
Interesting. I was also looking with @eguzki at the upgrade command from operator-sdk. We could possibly avoid having to mess with Catalogs https://sdk.operatorframework.io/docs/cli/operator-sdk_run_bundle-upgrade/

Boomatang:
One of the reason, I went at the time with creating standalone catalogs was for local testing. I felt that you would want to be able to quickly switch back to past versions. I am not sure if the upgrade command in the operator-sdk would allow that. It would also be the case if there is a pipeline that does upgrade testing and is failing, we may need to be able to install the previous state before the upgrade was tried. Operator-sdk can simplify this all the better. 

One thing that should be considered is be able do the upgrade from n-2, or more. There was some issues that were notice when going from n-2 to n, but the issue didn't happen with n-1 to n. What ever system is used would need to be able to go from the oldest version we are willing to support to the latest.

maleck13:
Might be possible to achieve this with a set of calls to the bundle upgrade command ? like install catalog v1.0 run upgrade to v1.0.1 then run upgrade to v1.02? Anyway lets discuss the options in further detail and decide on a direction forward

PR #1165, opened 2025-02-11
Add test OLM  upgrade path

eguzki:
### What

Implements #1117 and is part of the work being done for https://github.com/Kuadrant/kuadrant-operator/issues/1112

It's a Github Action Workflow to test the Kuadrant upgrade process using **OLM** APIs.

It is being triggered manually (`on.workflow_dispatch`) and accepts as input:
* `kuadrantStartVersion`:  Kuadrant start version (*required*). For example: `1.0.0` (without `v` prefix)
* `kuadrantNamespace`: Namespace where Kuadrant is installed (*optional* defaults to `kuadrant-system`).
* `clusterServer`: Cluster server URL (*optional* defaults to local kind cluster). For example: `https://example.com:443`
* `clusterToken`:  Cluster Server Bearer Token (*optional* defaults to local kind cluster token). For example: `sha256~abcdefghijk12345432432`

The workflow goes as follows:

* It Installs  Kuadrant release with the version specified by `kuadrantStartVersion`. It's being installed traditionally using OLM: OperatorGroup + CatalogSource (with the catalog build for the `kuadrantStartVersion`) + Subscription. 
* Waits for subscription, CSV and deployments to be ready
* Deploys kuadrant CR and checks status is *Ready*. The kuadrant operator, runs a set of tests to mark the resource status as ready, so effectively, the workflow is relying on those tests to verify installation. 
* When the starting point installation is ready, it proceeds with the upgrade process. The upgrade is being carried out running [operator-sdk run bundle-upgrade](https://sdk.operatorframework.io/docs/cli/operator-sdk_run_bundle-upgrade//) command. The command, basically, builds an ephemeral index adding an upgrade path from the existing installed bundle to the given bundle as input param. Then, waits until all deployments are ready (with a timeout set to 3 mins) and fails if some deployment is not ready after the timeout expires. Upgrade may fail for instance if some referenced docker image is not valid or does not exist. The upgrade verification steps by the workflow relies, once again, on 1) upgraded CSV readiness, 2) upgraded deployment image and 3) the kuadrant CR status. For example, if for whatever reason, let's say the limitador operator, or limitador instance, is not up and running, kuadrant CR would report on the status and the upgrade test would fail. The upgrade command may fail if kuadrant dependencies cannot be upgraded as well. This issue is not obvious from the upgrade command output, so digging is required. 

Simple commands are being coded as part of the Github workflow. For high level commands, those are implemented using makefile targets or third party tools so they can be easily run locally to dev/test. 

### Verification Steps

Prepare a new release, let's say v1.0.2

```
git checkout -b testing-upgrade
make prepare-release VERSION=1.0.2 AUTHORINO_OPERATOR_VERSION=0.16.0 LIMITADOR_OPERATOR_VERSION=0.12.1 DNS_OPERATOR_VERSION=0.12.0 WASM_SHIM_VERSION=0.8.1 RELATED_IMAGE_CONSOLEPLUGIN=quay.io/kuadrant/console-plugin:v0.0.18
```
Commit and push the branch (`testing-upgrade` in the example).

Then, run workflow manually (cannot be done from github UI as it has not been merged to `main` branch). This run would test upgrade from `v1.0.1` to `v1.0.2`:

```
gh workflow run test-upgrade.yaml --ref testing-upgrade -f kuadrantStartVersion=1.0.1
```
Check the [upgrade test workflow run](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/test-upgrade.yaml) to complete successfully.

### DIFF
--- a/.github/workflows/test-upgrade.yaml
+++ b/.github/workflows/test-upgrade.yaml
@@ -135,23 +135,20 @@ jobs:
   olm-upgrade-test:
     runs-on: ubuntu-latest
     name: OLM Upgrade Test
-    if: ${{ github.event_name == 'workflow_dispatch' }}
     env:
       KIND_CLUSTER_NAME: kuadrant-test
       K8S_USER: kuadrant-admin  # can be whatever, it does not matter.
       CLUSTER_NAME: remote-cluster
+      LOCAL_CLUSTER: ${{ inputs.clusterServer == '' && inputs.clusterToken == '' }}
     steps:
       - name: Check out code
         uses: actions/checkout@v4
-      - name: Install helm
-        run: |
-          make helm
       - name: Install yq tool
         run: |
           # following sub-shells running make target should have yq already installed
           make yq
       - name: Deploy local Kind cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        if: ${{ env.LOCAL_CLUSTER }}
         uses: helm/kind-action@v1.12.0
         with:
           version: v0.23.0
@@ -159,18 +156,18 @@ jobs:
           cluster_name: ${{ env.KIND_CLUSTER_NAME }}
           wait: 120s
       - name: Install kubectl for remote cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         uses: azure/setup-kubectl@v4
         with:
           version: v1.31.0
       - name: Mask cluster token
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterToken != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         run: |
           CLUSTER_TOKEN=$(jq -r '.inputs.clusterToken' $GITHUB_EVENT_PATH)
           echo ::add-mask::$CLUSTER_TOKEN
           echo CLUSTER_TOKEN=$CLUSTER_TOKEN >> $GITHUB_ENV
       - name: Setup kubectl for remote cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer != '' }}
+        if: ${{ !env.LOCAL_CLUSTER }}
         run: |
           kubectl config set-credentials ${{ env.K8S_USER }}/${{ env.CLUSTER_NAME }} --token ${{ env.CLUSTER_TOKEN }}
           kubectl config set-cluster ${{ env.CLUSTER_NAME }} --insecure-skip-tls-verify=true --server=${{ inputs.clusterServer }}
@@ -181,11 +178,11 @@ jobs:
           kubectl cluster-info
           kubectl get nodes
       - name: Deploy OLM on local Kind cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        if: ${{ env.LOCAL_CLUSTER }}
         run: |
           make install-olm
       - name: Deploy pre-requisites on local Kind cluster
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
+        if: ${{ env.LOCAL_CLUSTER }}
         run: |
           make install-metallb
           make install-cert-manager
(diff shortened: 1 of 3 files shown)


.github/workflows/test-upgrade.yaml
```diff
@@ -5,19 +5,19 @@ on:
   workflow_dispatch:
     inputs:
       kuadrantStartVersion:
-        description: Kuadrant start version
+        description: Kuadrant start version (format X.Y.Z)
```
didierofrivia, line 8: do we mean to upgrade from a `pre-version` tag?

.github/workflows/test-upgrade.yaml
```diff
@@ -133,3 +132,180 @@ jobs:
         run: |
           kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant -n ${{ inputs.kuadrantNamespace }}
           echo "kuadrant upgrade ✅"
+  olm-upgrade-test:
+    runs-on: ubuntu-latest
+    name: OLM Upgrade Test
+    if: ${{ github.event_name == 'workflow_dispatch' }}
```
didierofrivia, line 138: Not needed if we mean to run only this event type

.github/workflows/test-upgrade.yaml
```diff
@@ -133,3 +132,180 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Install helm
+        run: |
+          make helm
+      - name: Install yq tool
+        run: |
+          # following sub-shells running make target should have yq already installed
+          make yq
+      - name: Deploy local Kind cluster
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.clusterServer == '' }}
```
didierofrivia, line 154: as in the helm version, we could do the job scoped env:

```yaml
env:
  LOCAL_CLUSTER: ${{ inputs.clusterServer == '' && inputs.clusterToken != ''}}
```

thus in checks like this, we could do:

```suggestion
        if: ${{ !env.LOCAL_CLUSTER }}
```

The `github.event_name == 'workflow_dispatch'` wouldn't be needed neither, wdyt?

.github/workflows/test-upgrade.yaml
eguzki: I do not see a use case where you would want to test upgrade _from_ a release candidate to something else. The common use case would be upgrade _from_ a existing release to some candidate release or existing release.

.github/workflows/test-upgrade.yaml
eguzki: ok, doing like that, lgtm

.github/workflows/test-upgrade.yaml
```diff
@@ -133,3 +132,180 @@ jobs:
...
+    name: OLM Upgrade Test
+    if: ${{ github.event_name == 'workflow_dispatch' }}
+    env:
+      KIND_CLUSTER_NAME: kuadrant-test
+      K8S_USER: kuadrant-admin  # can be whatever, it does not matter.
+      CLUSTER_NAME: remote-cluster
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Install helm
```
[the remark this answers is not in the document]
eguzki: yes, it's needed. When running in local cluster, some commands use helm. What I am going to do is to add dependency of `helm` on those task and I can remove the explicit installation of helm in this job, which is indeed, confusing.

eguzki:
> I tried running the verification steps and it failed. https://github.com/Kuadrant/kuadrant-operator/actions/runs/13286208550/job/37095372107

You tried to upgrade kuadrant from 1.0.1 to 0.0.0. Not surprised it failed. Did you prepare a new release?

Boomatang:
I copied and pasted the instructions in the verification steps.

maleck13:
> > I copied and pasted the instructions in the verification steps.
> 
> Your workspace in the pushed commit does not have the `make/release.mk` that is generated with `make prepare-release` target.
> 
> ```
> ❯ git checkout c0a85bbed5f05268ff474a273a062a0dfb06a349
> HEAD is now at c0a85bbe Workflow testing
> ❯ cat make/release.mk
> cat: make/release.mk: No such file or directory
> ```
> 
> Might be you did not add all the changes to the commit. Could it be?

I This worked great for me using the verification steps. I did have to add the `make/release.mk` file manually as it wasn't added as part of the `prepare release` command.
https://github.com/Kuadrant/kuadrant-operator/actions/runs/13365901249/job/37323516174

maleck13 (approved):
LGTM

PR #1173, opened 2025-02-24
Bump Authorino package to v0.20.0

guicassolato:


guicassolato:
Specifically, I wanted to see if bumping the version of the AuthConfig to the one from `github.com/kuadrant/authorino@v0.20.0` would hit the threshold for maximum number of bytes stored in the CRD metadata.

I ran into this issue today while client-side applying an updated example from the [Policy Machinery](https://github.com/kuadrant/policy-machinery) repo:

```
The CustomResourceDefinition "authpolicies.kuadrant.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes
```

`local-setup` and other make targets, as well as some CI workflows, do server-side apply the CRDs, which is the initial remediation to this problem. In case there are other places to fix (e.g. tests, docs, etc), we may want to spot them beforehand and thus save users and contributors some unnecessary headache.

didierofrivia (approved):
🏛️

PR #1178, opened 2025-02-27
Refactor observability controller

eguzki:
### What

Following up the work being done in https://github.com/Kuadrant/kuadrant-operator/pull/1164. Refactor to, hopefully, simplify the code to maintain. 

* Monitoring resources loaded in memory are no longer static. They are instantiated every time. Reason being that go client might update objects on creation. 
* Implemented mutator pattern with `createOnly` mutator, having the same functionality: resources can externally be updated and the operator will not revert the changes back. 
* ~Added error handling. When resource creation reports error, the error is being returned and propagated up in the stack.~
* istiodMonitor and envoyGatewayMonitor are not created repeatedly for each gateway. They are single instance per Istio control plane / EnvoyGateway control plane. 

### Verification steps

* Setup environment based on EnvoyGateway

```shell
make local-setup GATEWAYAPI_PROVIDER=envoygateway
```

* Apply the Kuadrant custom resource with observability enabled

```yaml
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec:
  observability:
    enable: true
EOF
```
* Wait for kuadrant instance to be ready
```shell
kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant-sample -n kuadrant-system
```

* Verify the monitoring resources are there
```shell
k get servicemonitor,podmonitor -A
```

The expected result:

```shell
NAMESPACE              NAME                                                              AGE
envoy-gateway-system   servicemonitor.monitoring.coreos.com/envoy-gateway-monitor        148m
kuadrant-system        servicemonitor.monitoring.coreos.com/authorino-operator-monitor   148m
kuadrant-system        servicemonitor.monitoring.coreos.com/dns-operator-monitor         148m
kuadrant-system        servicemonitor.monitoring.coreos.com/kuadrant-operator-monitor    148m
kuadrant-system        servicemonitor.monitoring.coreos.com/limitador-operator-monitor   148m

NAMESPACE        NAME                                                   AGE
gateway-system   podmonitor.monitoring.coreos.com/envoy-stats-monitor   148m
```

Filtering by labels `kuadrant-observability=true` should give the same output. 
```shell
k get servicemonitor,podmonitor -l kuadrant.io/observability=true -A
```
* Turn monitoring off

```shell
kubectl patch kuadrant kuadrant-sample --type=merge --patch '{"spec": {"observability": null}}' -n kuadrant-system
```
* Verify the monitoring resources are gone
```shell
k get servicemonitor,podmonitor -l kuadrant.io/observability=true -A
```

The expected result:
```shell
No resources found
```

### DIFF
--- a/controllers/observability_reconciler.go
+++ b/controllers/observability_reconciler.go
@@ -12,14 +12,15 @@ import (
 	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/utils/ptr"
+	ctrlruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
 	"github.com/kuadrant/kuadrant-operator/pkg/kuadrant"
 	"github.com/kuadrant/kuadrant-operator/pkg/observability"
+	"github.com/kuadrant/kuadrant-operator/pkg/reconcilers"
 )
 
 //+kubebuilder:rbac:groups=monitoring.coreos.com,resources=servicemonitors;podmonitors,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


controllers/observability_reconciler.go
```diff
@@ -344,20 +379,32 @@ func (r *ObservabilityReconciler) Reconcile(ctx context.Context, _ []controller.
 	logger.V(1).Info("observability enabled, creating monitors")
 
 	// Kuadrant Operator monitor
-	kOpMonitorSpec.SetNamespace(r.namespace)
-	r.createMonitor(ctx, monitorObjs, kOpMonitorSpec, observability.ServiceMonitorsResource, logger)
+	kOpMonitor := kOpMonitorBuild(r.namespace)
+	err := r.createServiceMonitor(ctx, kOpMonitor, logger)
+	if err != nil {
```
david-martin, line 384: The errors are deliberately only logged, not propagated up the stack, to allow other monitors to possibly be still created.

eguzki: Ok, I will leave as it is then. However, if there is a net issue or API server downtime issue, the operator would not retry until there is another event :shrug:

eguzki:
There is an issue with the hardcoded namespaces of istio and envoygateway. The namespace of those control planes cannot be assumed to be those hardcoded. I will address this issue in a follow up PR

eguzki:
There is an issue with the lifecycle of the monitors when a gateway is deleted. The monitoring resource does not get deleted.  I will address this issue in a follow up PR

Boomatang:
> There is an issue with the lifecycle of the monitors when a gateway is deleted. The monitoring resource does not get deleted. I will address this issue in a follow up PR

@eguzki is that an exist issue?

Patryk-Stefanski:
@Boomatang, I created the follow-up issues below. They seem like good first issues that I could get started with, @eguzki could we sync on these when you're back?

- https://github.com/Kuadrant/kuadrant-operator/issues/1189
- https://github.com/Kuadrant/kuadrant-operator/issues/1190

eguzki:
> uld we sync on these when you're

Yes, those are good first issues. Send invite and we sync.

Patryk-Stefanski:
The verification steps were successful. Turning the observability option on and off created and deleted the monitoring resources as expected.  /lgtm

Patryk-Stefanski (approved):

Issue #1175, opened 2025-02-25
Use of latest tags in the release.yml


Boomatang:
# What 
Current implementation of the make prepare-release that is using the release.yml file does not support the latest tag for the components.  

# Why
By enabling support for the latest tags a number of features can be unlocked, such as version builds for make local-setup

# How
Logic needs to be added to the release scripts that allow the use of the latest tag. For the version in the release.yml file the versions will be set as `0.0.0` which will be the latest tag.

# Done
The `make prepare-release` can be called with a release.yml file that contains `0.0.0`

maleck13:
marking as done. Re-open if not

Boomatang:
Reopening, this is not done, only started to work on it today. It would be for the next sprint.

maleck13:
I think we can mark this as done?

PR #1181, opened 2025-03-03
Latest tags support for release.yaml

Boomatang:
closes: #1175 

Add support of the latest tag in the prepare-release script.
This allows the removal of individual checks.


maleck13:
@didierofrivia may leave this one for you as I think you can prob give a more meaningful review. Almost there with this work now great to see it coming together.

didierofrivia (commented):
The `verify-manifests`, `verify-helm-charts` and  `verify-bundle` should not be removed nor refactor to a `verify-release` since those are triggered for every branch PR, and push not only to a release branch but also to main, meaning they assert those manifest are kept in a healthy state, not only for a release, but for general CI/CD.   Regarding the version replacement to `latest` tags LGTM.

Boomatang:
@dd can I ask why you think the the manifest will not be kept in a healthy state if the pipeline is refactored to use `verify-release` over the other 3. The same checks are preformed. The only thing that would be changing is every branch will be required to have the release.yaml file. 

If there is some other reason that you can think of I would please let me know.

didierofrivia:
@Boomatang for anyone new to the project, a workflow that's named `verify-release` will assume that it's only for releases, and it's executed for every PR... I'd rather have it named `verify-manifests`, which is what it does, and include kustomize/olm and helm ones.

Boomatang:
@didierofrivia  so you are saying your not happy with the naming check, and not the removal of the 3 in favor of 1. I don't think the naming of the check should hold this PR up, as it is something that could be refactored later. I would be in favor of calling it  `verify-releasability`, but the main reason I used `verify-release` was because to fix the issue would be by running `make release`. What do you want to do? Change the name, refactor later, or something else.

didierofrivia:
@Boomatang for the time being, just rename it to `verify-manifests`

didierofrivia (approved):
Approving, let's schedule a refactor afterwards

PR #1183, opened 2025-03-04
bump Go version from 1.22.x to 1.24.x in workflow files

### LINKED ISSUE #1182: Workflow Go version is EOL
# What
Go 1.22 is end of life, with the latest being 1.24. 
The workflows use a public version of go.

# Why
The use of a EOL version may cause issue as we are public version of go that does not have extended support.

# How
Update the workflows to the news version of go that we can use.

# Done
All workflows are updated to use a support version of go.

Brijeshthummar02:
fixes #1182


Brijeshthummar02:
@Boomatang just quick follow up as it's almost a week and haven't heard back any thing from you.

eguzki (changes requested):
We need to hold on this, until we know official dates of the release of the image `registry.redhat.io/ubi9/go-toolset` for Golang 1.24. This image is being used for RedHat Connectivity Link. If kuadrant was using Golang 1.24 and there was no go-toolset available for 1.24, we would not be able to build RedHat Connectivity Link.

PR #1185, opened 2025-03-05
Release process enhancement

didierofrivia:


maleck13 (dismissed):
These changes all look good. 

/lgtm

eguzki:
Let's :eye:  this.

Please, wait for my review before merging it

release.yaml
```diff
@@ -0,0 +1,14 @@
+kuadrant-operator:
```
guicassolato, line 1: I wonder if this file should be named something like `release.yaml.tmpl` or `release.yaml.in` in the `main` branch, and renamed/duplicated as `release.yaml` only in the release branches.

.github/workflows/test.yaml
```diff
@@ -284,28 +283,30 @@ jobs:
...
-      - name: Run make verify-manifests
-        run: |
-          make verify-manifests
+      - name: Check bundles and manifests
+        run: make verify-prepare-release
 
-  verify-bundle:
-    name: Verify bundle
-    if: needs.pre-job.outputs.should_skip != 'true'
+  verify-manifests:
```
eguzki, line 291: I think it is better to have N verify jobs. Two reasons:
* the UX is better in Github. You know very quickly what failed
* Each tests is run in isolation: one test cannot interfere the other test. For instance, `make bundle` could be changing files from helm charts and the helm charts test could be failing due to bundle test.  Test isolation is always a desired property.

RELEASE.md
```diff
@@ -1,30 +1,85 @@
...
+To release a version _“vX.Y.Z”_ of Kuadrant Operator in GitHub and Quay.io, there are two options, a manual and an automated process.
+For both processes, first make sure every [Kuadrant Operator dependency](https://github.com/Kuadrant/kuadrant-operator/blob/main/RELEASE.md#kuadrant-operator-dependencies) has been already released.
+
+## Automated Workflow
+
+1. Run the GHA  [Automated Release](https://github.com/Kuadrant/kuadrant-operator/actions/workflows/automated-release.yaml)
+   filling the following fields:
+   - gitRef: Select the branch/tag/commit where you want to cut a release from.
+   - kuadrantOperatorVersion: the [Semantic Version](https://semver.org/) of the desired release.
+   - authorinoOperatorVersion: Authorino Operator bundle version (X.Y.Z)
```
eguzki, line 12:
```suggestion
   - authorinoOperatorVersion: Authorino Operator version (X.Y.Z)
```

RELEASE.md
```diff
@@ -1,30 +1,85 @@
...
+   - limitadorOperatorVersion: Limitador Operator bundle version (X.Y.Z)
```
eguzki, line 13:
```suggestion
   - limitadorOperatorVersion: Limitador Operator version (X.Y.Z)
```

RELEASE.md
```diff
@@ -1,30 +1,85 @@
...
+   - dnsOperatorVersion: DNS Operator bundle version (X.Y.Z)
```
eguzki, line 14:
```suggestion
   - dnsOperatorVersion: DNS Operator version (X.Y.Z)
```

eguzki (commented):
Looking good

.github/workflows/test.yaml
Boomatang: I am not sure I fully agree with the each test running in isolation. There is a strong interaction between the bundles and helm charts when it comes to releasing. Would see this test more of an integration test over a unit test. When the `make prepare-release` would resolve both issue I don't see the need to have them split.

You might know very quickly what failed, but does that tell you as quickly how to fix it. I also think the make target in the test should not be used, but instead use `make verify-prepare-release`.

utils/release/release.sh
```diff
@@ -0,0 +1,67 @@
...
+}
+
+# Task runner function
+run_tasks() {
+    local dir_name="$1"
+    local tasks=()
+
+    log "RUNNING: $dir_name"
+
+    tasks=($(find "$script_dir/$dir_name" -mindepth 1 -maxdepth 1 -perm -001 | sort))
```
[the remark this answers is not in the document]
guicassolato: Why `sort` then? Deterministic output?

release.yaml
guicassolato: Mainly communication, to call the file for what it is. This file should only exist in release branches. The only meaning it has in `main` is as a template.

Plus, we don't want people modifying and committing this file in `main`. Some distraction and it may end up with some of values set something other than `0.0.0`, which is ugly, to say the least.

utils/release/release.sh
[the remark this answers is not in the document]
Boomatang: Good question. Initial the order did matter, and there was numbering prefixes. This is something could be removed, but I currently don't see the harm in having it there. The deterministic output is a nice side affect. I don't think there is a need for a change here. More of an understanding of what is happening.

release.yaml
Boomatang: I would not say it should only exist on release branches, there are possible advantages to how the nightly pipelines could make use of the file. None of that work has being done as of yet. There is also a hope to get `make local-setup` to be able to consume that file, so it is not truly  about releases.

utils/release/dependencies/dependencies-manifests.sh
```diff
@@ -0,0 +1,20 @@
...
+  else
+    echo "v$version"
+  fi
+}
+
+authorino_version=$(mod_version $(yq '.dependencies.authorino-operator' $env/release.yaml))
+dns_version=$(mod_version $(yq '.dependencies.dns-operator' $env/release.yaml))
+limitador_version=$(mod_version $(yq '.dependencies.limitador-operator' $env/release.yaml))
+
+AUTHORINO_OPERATOR_GITREF=$authorino_version envsubst < $env/config/dependencies/authorino/kustomization.template.yaml > $env/config/dependencies/authorino/kustomization.yaml
```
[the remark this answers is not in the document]
Boomatang: So then  I am hear that for now the best step forward is to document the use of the `release.sh`.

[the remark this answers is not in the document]
Boomatang: https://github.com/Kuadrant/kuadrant-operator/issues/1219

.github/workflows/test.yaml
eguzki: The source of truth is in golang code and `/config` plus few other places like Makefile. Those are verification steps to make sure the source of truth is consistent with the shipped manifests (bundle or helm charts). So, IMO, independent and isolated tests are good to have. I can live without it as well.

eguzki: In the future, the Makefile will no longer be a source of truth.. it will be replaced by release.yaml. Because now there are two ways to generate the manifests and eventually we need to merge into one.

eguzki (changes requested):
Image building is broken.

The image building system is implemented in `.github/workflows/build-images-for-tag-release.yaml`. This workflow still assumes the existence of `release.mk` file, which no longer exists. So Image build fails. One example run failing: https://github.com/Kuadrant/kuadrant-operator/actions/runs/13954148232/job/39060935703

I (tried to) fix it partially in https://github.com/Kuadrant/kuadrant-operator/pull/1223. But the catalog and bundle build is very likely to fail as well. 

I can open another PR with the fix for the bundle and catalog builds or leave someone else to do this.

.github/workflows/test.yaml
Boomatang: Agreed eventually it needs to merge into one. The next step on the road I see is how the nightly test pipeline generate their builds.

didierofrivia: @eguzki I could add a commit splitting the verification step in 3, as it was before

utils/release/release.sh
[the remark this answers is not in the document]
didierofrivia: I can think of some processes that might gain from executing in order, the ones that have GH API interactions for example. Better to do the least possible

RELEASE.md
```diff
@@ -1,30 +1,85 @@
...
+
+## Manual Workflow
+
+### Local steps
+
+1. Create the `release-vX.Y` branch, if the branch does not already exist. 
+2. Push the `release-vX.Y` to the remote (kuadrant/kuadrant-operator)
+3. Create the `release-vX.Y.Z-rc(n)` branch with `release-vX.Y` as the base.
+4. Cherry-pick commits to the `kudrant-vX.Y.Z-rc(n)` from the relevant sources, i.e. `main`.
+5. Update the applicable version in the [release.yaml](https://github.com/Kuadrant/kuadrant-operator/blob/main/RELEASE.md#release-file-format).
```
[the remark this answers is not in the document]
Boomatang: If I understand what you are point at. There will be no long lived `release-vX.Y.Z` branch. They act the same as the rc branches. Where their propose is to allow reviewing of changes going into the `release-vX.Y` branches.

eguzki (dismissed):

Boomatang (dismissed):
Looking good to me

eguzki (approved):

Boomatang (approved):

PR #1197, opened 2025-03-10
Fix verifications

didierofrivia:
This PR re introduces the make targets that were asserting the validity of manifests such as _bundle_, _controller-gen_ and _helm-charts_ in order to be used for non release manifest verification. Now the test.yaml will conditionally trigger the verification manifests for a release consuming the release shell scripts _or_ the verification of manifests how it was done before, without running the whole release scripts that consumes GH API, checks agains the release.yaml file, etc.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,15 +1,15 @@
 name: Test
 
 on:
+  workflow_dispatch:
   push:
-    branches: [ 'main', 'release-*' ]
+    branches: ['main', 'release-v*']
   pull_request:
     branches: ['*']
   merge_group:
     types: [ checks_requested ]
   schedule:
     - cron: "15 1 * * *"
-  workflow_dispatch:
 
 jobs:
   pre-job:
@@ -272,10 +272,9 @@ jobs:
         if: always()
         run: |
           kubectl get pods -n kuadrant-system
-
   verify-release:
-    name: Validate release data 
-    if: needs.pre-job.outputs.should_skip != 'true'
+    name: Validate release data
+    if: startsWith(github.ref_name, 'release-v') && needs.pre-job.outputs.should_skip != 'true'
     needs: pre-job
     runs-on: ubuntu-latest
     steps:
@@ -288,6 +287,26 @@ jobs:
         uses: actions/checkout@v4
       - name: Check bundles and manifests
         run: make verify-prepare-release
+
+  verify-manifests:
+    name: Validate set of manifests
+    if: !startsWith(github.ref_name, 'release-v') && needs.pre-job.outputs.should_skip != 'true'
+    needs: pre-job
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install gettext-base
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y gettext-base
+      - name: Set up Go 1.22.x
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.22.x
+        id: go
+      - name: Checkout code at git ref
+        uses: actions/checkout@v4
+      - name: Check bundles and manifests
+        run: make verify-manifests
 
   verify-fmt:
     name: Verify fmt
@@ -369,7 +388,7 @@ jobs:
     # This check adds a list of checks to one job to simplify adding settings to the repo.
     # If a new check is added in this file, and it should be retested on entry to the merge queue,
     # it needs to be added to the list below aka needs: [ existing check 1, existing check 2, new check ].
-    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-release, verify-fmt, test-scripts, verify-generate, verify-go-mod ]
+    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-fmt, test-scripts, verify-generate, verify-go-mod ]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/make/verify.mk
+++ b/make/verify.mk
@@ -16,7 +16,28 @@ verify-go-mod: ## Verify go.mod matches source code
 	go mod tidy
 	git diff --exit-code ./go.mod
 
-.PHONY: verify-prepare-release
+.PHONY: verify-controller-manifests
+verify-controller-manifests: manifests ## Verify controller-gen manifests update.
+	git diff --exit-code ./config
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./config)" ]
+
+.PHONY: verify-bundle
+verify-bundle: bundle ## Verify bundle update.
+	git diff --exit-code ./bundle
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./bundle)" ]
+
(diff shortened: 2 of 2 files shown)


eguzki (approved):

eguzki (approved):

make/verify.mk
```diff
@@ -16,7 +16,28 @@ verify-go-mod: ## Verify go.mod matches source code
...
+
+.PHONY: verify-bundle
+verify-bundle: bundle ## Verify bundle update.
+	git diff --exit-code ./bundle
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./bundle)" ]
+
+.PHONY: verify-helm-charts
+verify-helm-charts: helm-build ## Verify helm charts update.
+	git diff --exit-code ./charts
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory ./charts)" ]
```
Boomatang, line 32: There is a number of things here that we need to discussed.
The two main things are:

The use of the bundle make target was only left in the codebase to not break the nightly pipelines. And the next action should be getting the nightly pipelines to support the release.yaml file. Which could mean the removal of these base make targets completely.

The second thing that is interesting is around the developer experiences within the project. On a number of occasions now I have being asked why is there so many make targets within the project. They say it is hard to know what should be done, what order to do things in, where to even start. These comments came from people within the project, but also external to the project. They are the more interesting ones.  

So I am going to ask. Is there a different way that this can be done that reduces the number of make targets, and stream lines the experiences of others.

.github/workflows/test.yaml
```diff
@@ -273,21 +271,6 @@ jobs:
...
-    steps:
-      - name: Set up Go 1.22.x
-        uses: actions/setup-go@v4
-        with:
-          go-version: 1.22.x
-        id: go
-      - name: Checkout code at git ref
-        uses: actions/checkout@v4
-      - name: Check bundles and manifests
-        run: make verify-prepare-release
```
Boomatang, line 290: Why remove this from all PRs?

.github/workflows/verify-main.yaml
```diff
@@ -0,0 +1,59 @@
...
+          sudo apt-get install -y gettext-base
+      - name: Set up Go 1.22.x
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.22.x
+        id: go
+      - name: Checkout code at git ref
+        uses: actions/checkout@v4
+      - name: Check bundles and manifests
+        run: make verify-manifests
```
Boomatang, line 46: Why use this make target over the verify-prepare-release? If the verification calls to github checking the existing of versions seems to expensive I would argue that is not a good enough reason. There are will be external call to quay by opm render when the bundle make target is called. 

There is also the feature of the scripts used in the prepare-release make target to grep and only run a given script. If the cost of a check was that high problematic (thinking getting rate limited), it would make more sense to add the feature to filter out scripts over trying to maintain two versions. Then how I would see that being used would be something like `make verify-prepare-release --exclude script-a script-b`. Stating the list of scripts that would be exclude also makes it easier for someone in 6 months, a year to understand what is and is not happening.

Boomatang (commented):
A part from the question I am asking in the other comments there is one question that I have that does not fit neatly to any line change. Why is there two more workflows being added? They don't seem to give any value that the `test.yaml` can not give. They are increasing the stuff that would need to be maintained. So are they really needed?

make/verify.mk
didierofrivia: So would you say a the number of shell scripts introduced is better for a developer experience? Also, pretty much every make target one can easily check what is its purpose with `make help`

.github/workflows/test.yaml
```diff
@@ -1,15 +1,13 @@
...
 
 on:
+  workflow_dispatch:
   push:
-    branches: [ 'main', 'release-*' ]
+    branches: ['main', 'release-v*']
   pull_request:
     branches: ['*']
-  merge_group:
-    types: [ checks_requested ]
```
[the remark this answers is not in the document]
didierofrivia: Back now in the latest refactor

.github/workflows/verify-main.yaml
didierofrivia: It's on the description... one could argue why replacing all the makefiles with the shell scripts, but well, we are beyond that point now.

make/verify.mk
```diff
@@ -16,7 +16,28 @@ verify-go-mod: ## Verify go.mod matches source code
...
+
+.PHONY: verify-manifests ## Verify controller-gen, bundle and helm charts manifests.
+verify-manifests: ## Verify manifests update.
+	make verify-controller-manifests
+	make verify-bundle
+	make verify-helm-charts
```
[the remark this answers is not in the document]
didierofrivia: Because I was thinking that we might gain from quick fail, instead of building all the manifest and check, building one at a time and do the check

make/verify.mk
eguzki: `make bundle` should stay. It should be implemented with whatever you like and support the release.yaml file.

eguzki: Same for helm charts. `make bundle` is well known by everyone developing k8s controllers and very useful for daily development.

.github/workflows/test.yaml
```diff
@@ -1,15 +1,15 @@
...
   push:
-    branches: [ 'main', 'release-*' ]
+    branches: ['main', 'release-v*']
   pull_request:
     branches: ['*']
   merge_group:
     types: [ checks_requested ]
   schedule:
     - cron: "15 1 * * *"
-  workflow_dispatch:
```
eguzki, line 12: Why remove? I think it is handy to run the tests anytime.

make/verify.mk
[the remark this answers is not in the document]
eguzki: Each verify make targets knows which folder to test. `git diff --exit-code .` can catch changes that may not belong to manifests and return false positives

.github/workflows/test.yaml
didierofrivia: it's hoisted, at line [4](https://github.com/Kuadrant/kuadrant-operator/pull/1197/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R4 ) ... sorry, stupid change

make/verify.mk
eguzki: > The second thing that is interesting is around the developer experiences within the project. On a number of occasions now I have being asked why is there so many make targets within the project. They say it is hard to know what should be done, what order to do things in, where to even start. These comments came from people within the project, but also external to the project. They are the more interesting ones.

Boomatang: While  they maybe well known targets, that does not mean that it is correct that we allow users to make mistakes by only generating one at a time. I have seen a number of times prs failing because the bundle was not generated, then the author fixes that issue. With the pr to fail again because the author didn't run the targets for the helm charts. I believe there should be one for creation and one for verification.

But I am not going to hold this up on want I believe to be correct.

make/verify.mk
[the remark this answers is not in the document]
Boomatang: While do agree, I also can agree. If any of the make targets are changing source outside the folder that they are expected to be working, causing the false positive, that would mean there is other issue with the code generation would it not. If there is any source code changes doing the verification steps seems like a bad thing. If it was possible that there may be changes generate not related to manifest what checks exist today to catch does changes, or are we merging code that we don't verify the generation of.

PR #1198, opened 2025-03-11
Prepare release 1.1.0-alpha1

Boomatang:
The following release candidate of Kuadrant Operator version 1.1.0-alpha1 includes:
   * Authorino Operator version 0.16.0
   * Console Plugin version 0.0.19
   * DNS Operator version 0.12.0
   * Limitador Operator version 0.12.1
   * WASM Shim version 0.8.1


Boomatang:
This release is based off the release-process-enhancement, and rebased off main.

There is an error with the PR which was created to see if it can be spotted in a review.

Boomatang:
The conflicts were not expected. Not sure why it was in conflict. The error that I know I introduced was the limitador version being wrong. 

The reason for not doing the `alpha1` is simple. The gh cli would not allow me to push to the release-v1.1.0-alpha1 branch without doing a force push. So I was very much, no  we want this to be as clean as possible. Why introduce that difference.

didierofrivia:
@Boomatang are we actually releasing this? if that's so reviewers might expect a conflict free PR. 
Regarding the alpha1 branch, that's weird, since I can't find it :/  I remember deleting it.

The limitador version for anyone not know the exact versions to be expected in the release would be an issue to spot, since the only way would be to review the manifests match the release.yaml file.

Boomatang:
The reason I added the version error was to run through the steps required to update the version. The conflict on release.yaml has shown one issue. The repo branch protections were stopping the update.

didierofrivia (approved):

PR #1202, opened 2025-03-13
Fix Workflow Configuration

Boomatang:
The workflow steps has no field `needs`. This was causing the action to fail.


didierofrivia (approved):
re run the failing tests, then port the commit to the `release-process-enchancement` branch too

PR #1205, opened 2025-03-13
Fix workflow

Boomatang:
Wrong target was being used


didierofrivia (changes requested):
`target_commitish` can't be the tag, checking their schema...


> Commitish value that determines where the Git tag is created from. Can be any branch or commit SHA.

PR #1206, opened 2025-03-13
Kubebuilder v4 migration

alexsnaps:
Manual upgrade to [kubebuilder v4](https://go.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4)
I would "revert" [these changes](f726b2f29dbe5034b7b51a8833c0b96ab09451ef), once I introduce the first client lib code for the plugins in `pkg/plugin/`...
But maybe I was a little eager in moving all that was in `pkg/` to `internal/` and there might be something from that we already want in `pkg/`... let me know if you think that's the case and I'll do the change!


alexsnaps:
>All modified and coverable lines are covered by tests ✅

Whatever Codecov complains about can only be wrong... so I'll ignore, so should you probably. I won't start this here again, but can we please 🔥 that thing?

alexsnaps:
I can't get to regenerate bundles (`make bundle`)... Deleted things in `bin/`, but didn't help... any clues?

maleck13 (dismissed):
scanned this all looks good

adam-cattermole (approved):
Re-approving after rebase

PR #1213, opened 2025-03-17
Extension mechanism, with support for Out-of-Process with gRPC over unix sockets

alexsnaps:


alexsnaps:
@adam-cattermole I think this lays the foundation of the wiring. We can sync if needed. Please review and/or ask for changes as you see fit, happy to make this better.
I'll be wiring the loading of extensions through config next and then wire a somewhat useful `kuadrantService`... (thinking accepting a CEL expression that then resolves and adds subscriptions)

internal/extension/manager.go
```diff
@@ -0,0 +1,112 @@
...
+
+func NewManager(names []string, location string, logger logr.Logger) (Manager, error) {
+	var extensions []Extension
+	var err error
+
+	service := newExtensionService()
+	logger = logger.WithName("extension")
+
+	for _, name := range names {
+		if oopExtension, e := NewOOPExtension(name, location, service, logger); e != nil {
```
adam-cattermole, line 50:
```suggestion
		if oopExtension, e := NewOOPExtension(name, location, service, logger); e == nil {
```

PR #1223, opened 2025-03-19
workflows/build-images-for-tag-release.yaml: fix operator image build

eguzki:
### What
`workflows/build-images-for-tag-release.yaml` not working because tag versions are still expected to be read from release.mk file, which no longer exists when a commit is tagged.


Boomatang (approved):
This looks good

didierofrivia (approved):
🙏🏼

eguzki:
Thanks, working on the fix for the catalog. The bundle looks good.

Issue #1211, opened 2025-03-14
Auto generation of release notes failure


Boomatang:
**Describe the bug**
During the release process the release notes are auto generated.
These note do not reflect that actual change, and is comparing the wrong version.

What I have seen is the release notes is trying to compare the difference between `v0.1.0` and version being released.

**To Reproduce**
Hard to reproduce without trying to create a release.

**Expected behavior**
It would be expected the at the change would be made between the version in previous in the chain. 
This would mean that if v1.1.2 was released it would compare with v1.1.1. I mean the chain, and not the previous release as the previous release could be v1.2.1 or v1.1.4, if there is back porting of fixes happening

Boomatang:
@didierofrivia you mention to me during the alpha-1 test run. Can you outline the manual steps that were previously carried out to resolve this issue on past release?

Boomatang:
I also have found this PR to softprops/action-gh-release which I think we would be need to be merged in order to fix this issue. The pull request has being opened for nearly two years now, so I am guess it will not be resolved any time soon.

https://github.com/softprops/action-gh-release/pull/372

Without that fix, how do we go about resolving the issue for our use case.

Boomatang:
I was thinking about what we could do, and now I am overly happy with. The first thing was update release doc to warn of this issue. That change has being merged, so that should help.

I am thinking we should stop trying to use the auto generation of the release note feature of `action-gh-release`, at least until the issue is fixed on their end. That action does does allow the adding of the release notes body so we could replace it with notes on how to generate the note correctly. But again that would be a stop gap. 

There is also the possibility to generate the change log note via the GitHub api, and at that to the body of the release. However that would be a bit of work, and would raise the question why use the `action-gh-release` at all. If we are going to go to the effort of manually generation the release body via api calls we could as easily create the release using the api directly. The api allows the setting of the previous tag which is what we are needing to do. There would still be some work required to find what the previous tag should be, but in every case I think that would be required.

- https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release

eguzki:
Looks like promising workaround. Give it a try and let us know

PR #1227, opened 2025-03-21
FIX: Release Note Generation

Boomatang:
closes: #1211 

The softprops/action-gh-release does not correctly pick up the tag for comparing when creating the release notes. Here the GitHub api is directly used to generate the release notes. softporps/action-gh-release is still used to create the actual release. Lines below dotted lines will be ignored, and an empty title aborts the creation process.

### DIFF
--- a/.github/workflows/release-operator.yaml
+++ b/.github/workflows/release-operator.yaml
@@ -39,6 +39,7 @@ jobs:
           name: ${{ env.kuadrantOperatorTag }}
           tag_name: ${{ env.kuadrantOperatorTag }}
           body: "${{ env.releaseBody }}"
+          # TODO: https://github.com/Kuadrant/kuadrant-operator/issues/1229
           # The generate of release notes does not pick up the correct tag to compare.
           # There is an open PR within softprops/action-gh-release that would solve the problem.
           # Till then we cannot use the generate release notes feature.
--- a/.github/workflows/release-operator.yaml
+++ b/.github/workflows/release-operator.yaml
@@ -28,6 +28,10 @@ jobs:
           tag_name="${{ env.kuadrantOperatorTag }}"
           git tag $tag_name
           git push origin $tag_name
+      - name: Generate release body
+        id: generate_release_body
+        run: |
+          GITHUB_TOKEN=${{ secrets.KUADRANT_WORKFLOWS_PAT }} bash ./utils/release/github-release-changelog.sh
       - name: Create release
         id: create_release
         uses: softprops/action-gh-release@v2
@@ -35,6 +39,10 @@ jobs:
           name: ${{ env.kuadrantOperatorTag }}
           tag_name: ${{ env.kuadrantOperatorTag }}
           body: "${{ env.releaseBody }}"
-          generate_release_notes: true
+          # The generate of release notes does not pick up the correct tag to compare.
+          # There is an open PR within softprops/action-gh-release that would solve the problem.
+          # Till then we cannot use the generate release notes feature.
+          # https://github.com/softprops/action-gh-release/pull/372
+          generate_release_notes: false
           target_commitish: ${{ env.releaseBranch }}
           prerelease: ${{ env.prerelease }}
--- a/utils/release/github-release-changelog.sh
+++ b/utils/release/github-release-changelog.sh
@@ -2,7 +2,7 @@
 
 # Access token will be required. Check to ensure that is it provied
 if [[ -z "$GITHUB_TOKEN" ]]; then
-	echo "GITHUB_TOKEN most be set"
+	echo "GITHUB_TOKEN must be set"
 fi
 auth_header="-H Authorization: Bearer $GITHUB_TOKEN"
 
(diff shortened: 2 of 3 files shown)


.github/workflows/release-operator.yaml
```diff
@@ -28,13 +28,21 @@ jobs:
...
         uses: softprops/action-gh-release@v2
         with:
           name: ${{ env.kuadrantOperatorTag }}
           tag_name: ${{ env.kuadrantOperatorTag }}
           body: "${{ env.releaseBody }}"
-          generate_release_notes: true
+          # The generate of release notes does not pick up the correct tag to compare.
+          # There is an open PR within softprops/action-gh-release that would solve the problem.
+          # Till then we cannot use the generate release notes feature.
+          # https://github.com/softprops/action-gh-release/pull/372
```
didierofrivia, line 45: Could you add a `TODO` annotation so we can easily track this?

utils/release/github-release-changelog.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+# Access token will be required. Check to ensure that is it provied
+if [[ -z "$GITHUB_TOKEN" ]]; then
+	echo "GITHUB_TOKEN most be set"
```
didierofrivia, line 5:
```suggestion
	echo "GITHUB_TOKEN must be set"
```

utils/release/github-release-changelog.sh
```diff
@@ -0,0 +1,39 @@
...
+fi
+auth_header="-H Authorization: Bearer $GITHUB_TOKEN"
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+source $script_dir/shared.sh
+
+# Get latest release version
+log "Getti previous release version tag"
+previous_tag_name=$(curl -L "https://api.github.com/repos/kuadrant/kuadrant-operator/releases/latest" -H "Accept: apllication/vnd.github+json" | yq '.tag_name')
```
didierofrivia, line 14: I'm not entirely sure that latest endpoint work with pre-releases, but also probably not required since the release notes might be wanted to reflect the full changes from the stable releases.

utils/release/github-release-changelog.sh
```diff
@@ -0,0 +1,39 @@
...
+if [[ -z "$GITHUB_TOKEN" ]]; then
+	echo "GITHUB_TOKEN most be set"
+fi
+auth_header="-H Authorization: Bearer $GITHUB_TOKEN"
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+source $script_dir/shared.sh
+
+# Get latest release version
+log "Getti previous release version tag"
```
didierofrivia, line 13:
```suggestion
log "Get previous release version tag"
```

didierofrivia: or

```suggestion
log "Getting previous release version tag"
```

eguzki, line 13: What about `latest` instead of previous which is more accurate?
```suggestion
log "Getting latest release version tag"
```

utils/release/github-release-changelog.sh
```diff
@@ -0,0 +1,39 @@
...
+data=$(curl -L "https://api.github.com/repos/kuadrant/kuadrant-operator/releases/generate-notes" -X POST -H "Accept: apllication/vnd.github+json" -H "Authorization: Bearer $GITHUB_TOKEN" -H "X-GitHub-Api-Version: 2022-11-28" -d "$payload")
+
+release_body=$(echo $data | yq '.body')
+
+if [[ $_log == "1" ]]; then
+	log "releaseBody=$release_body"
+fi
+
+if [[ $dry_run == "0" ]]; then
+	echo "releaseBody=$release_body" >> "$GITHUB_ENV"
```
eguzki, line 38: This `releaseBody` is shadowing same GITHUB_ENV added from `load_github_envvar.sh`. Should we remove releaseBody from load_github_envvar.sh??

utils/release/github-release-changelog.sh
Boomatang: From my testing the version that will be returned it the on marked as `latest` in the GitHub release page, not the newest entry in the release list. Currently there is a `v1.1.0-alpha1` release, but the change log would be generated comparing the changes from `v1.0.2`. For now I don't see an issue with this. Where it will run into issue is if we need to support back ports on previous minor streams.

utils/release/github-release-changelog.sh
Boomatang: I was going to use the word latest here, but I also felt it might be a bit misleading as the script is running to generate the new latest release.

utils/release/github-release-changelog.sh
Boomatang: Checking, I have a feeling we can't remove it from the `load_github_envvar.sh`, but might need to rename it.

utils/release/github-release-changelog.sh
eguzki: your call. `latest` expresses more accurately what the script is doing.

didierofrivia (approved):
👍🏼

PR #1231, opened 2025-03-24
updating paths for integration and codecov tests afer kubebuilder v4 migration

laurafitzgerald:
cc @alexsnaps 

This change removes the ./pkg directory from integration tests. It's not causing any problems, just an error log.
The codecov changes I'm waiting to see how that goes with the change pushed. I think it's needed.


alexsnaps (dismissed):

make/integration-tests.mk
```diff
@@ -1,4 +1,4 @@
-INTEGRATION_COVER_PKGS = ./pkg/...,./internal/controller/...,./api/...
+INTEGRATION_COVER_PKGS = ./internal/controller/...,./api/...
```
eguzki, line 1: what about other packages at `internal` ?

.github/codecov.yaml
```diff
@@ -38,26 +38,26 @@ component_management:
       paths:
         - api/v1beta2
     - component_id: common
-      name: pkg/common (u)
+      name: internal/common (u)
```
eguzki, line 41: does not exist

```
❯ ls internal/
authorino  controller  envoygateway  gatewayapi  istio  kuadrant  log  observability  openshift  policymachinery  ratelimit  reconcilers  utils  wasm
```

laurafitzgerald: @KevFan could i get your input on these changes. Looking back at the original copy of this file there was an entry for 
```
flag_management:
  individual_flags:
    - name: unit
      paths:
        - pkg/**
        - api/**
      carryforward: true
    - name: integration
      paths:
        - controllers/**
      carryforward: true
``` 
which doesn't seem to be present anymore. 

Lots of files have moved around now after the kubebuilder 4 migration so I want to be sure we're updating this file in the right way rather than simply updating the paths. 
Do we still want to maintain this configuration of unit test vs integration test coverage. I don't know the repo well enough to make an informed decision. 
cc @eguzki

eguzki: I would just have a coverage component per folder in  `internal`. The component helps when analyzing coverage results as  you can check coverage trends per component.

alexsnaps: I'd just 🔥 codecov... 
Who uses those metrics? This is _an honest_ question! It seems like whatever this thing says, we merge either way, so what do we use this for?

eguzki: While we discuss and take a decision on this matter, can you review https://github.com/Kuadrant/kuadrant-operator/pull/1277? (I am using codecov metrics when analyzing coverage to see if more tests are needed)

eguzki:
superseded by https://github.com/Kuadrant/kuadrant-operator/pull/1277

Boomatang:
Close as superseded by #1277, @eguzki reopen this if I have wrongly closed it.

PR #1232, opened 2025-03-24
Fix custom catalog images build workflow build

eguzki:


.github/workflows/build-images-for-tag-release.yaml
```diff
@@ -138,18 +138,41 @@ jobs:
         run: |
           # following sub-shells running make target should have yq already installed
           make yq
-      - name: Read operator bundle image reference
-        id: parsed-operator-bundle
+      - name: Set environment variables
+        id: set_env_variables
+        run: |
+          bash ./utils/release/load_github_envvar.sh
+      - name: Build catalog variables
```
didierofrivia, line 143: maybe something like

```suggestion
- name: Set catalog image variables
```

.github/workflows/build-images-for-tag-release.yaml
```diff
@@ -138,18 +138,41 @@ jobs:
...
-        run: make catalog
+        run: |
+          # Pinning the channel used in the catalog to "stable".
+          # It should not matter for catalogs built for a single release
+          make catalog \
+            BUNDLE_IMG=${{ steps.catalog-variables.outputs.bundleImg }} \
+            LIMITADOR_OPERATOR_BUNDLE_IMG=${{ steps.catalog-variables.outputs.limitadorBundleImg }} \
+            AUTHORINO_OPERATOR_BUNDLE_IMG=${{ steps.catalog-variables.outputs.authorinoBundleImg }} \
+            DNS_OPERATOR_BUNDLE_IMG=${{ steps.catalog-variables.outputs.dnsOperatorBundleImg }} \
+            CHANNEL=stable
```
didierofrivia, line 173: This will set the channel to stable to pre-releases as well, no?

eguzki: yes

didierofrivia: we might want to change that behaviour, since historically we've been assigning the `alpha` or `preview` to those... maybe we can get the first value from the `channels` env variable exported from the release.yaml file? for the time being? We've never published a catalog with many channels so far. What do you reckon?

eguzki:
ready for review @Boomatang @didierofrivia

.github/workflows/build-images-for-tag-release.yaml
```diff
@@ -21,7 +21,6 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v4
       - name: Set environment variables
-        id: set_env_variables
```
Boomatang, line 24: General question: Why have some steps with an id and others like this one without an id?

utils/release/post-validation/verify_default_olm_channel.sh
```diff
@@ -8,14 +8,14 @@ if [[ "$OLM_DEFAULT_CHANNEL" == "null" ]]; then
   echo "OLM default channel is not set in release.yaml"
   bundle_default_channel=$(yq '.annotations | has("operators.operatorframework.io.bundle.channel.default.v1")' $env/bundle/metadata/annotations.yaml)
   if [[ "$bundle_default_channel" == "true" ]]; then
-    echo "🚨 OLM default channel is set in bundle and should not be"
+    >&2 echo "🚨 OLM default channel is set in bundle and should not be"
```
Boomatang, line 11: General question: Why pipe the output to standard error?

Boomatang (approved):
This looks good to me. All the images are created in quay as expected.

I do have two general questions for my own understanding, but they are non blocking questions.

.github/workflows/build-images-for-tag-release.yaml
eguzki: `id` is used when you want to reference it from another step, for instance if generates some outputs. It is harmless anyway.

didierofrivia (approved):

Issue #1200, opened 2025-03-12
Custom GatewayClass names


shaneutt:
**Is your feature request related to a problem? Please describe.**

Feature Request

**Describe the solution you'd like**

Today Kuadrant will look for specific known and supported GatewayClass names, such as "istio" in order to attach itself to inject policies. It is however possible for Istio and other implementations to be deployed with non-default name, so the purpose of this issue is to request support for configuring custom GatewayClass names for supported implementations.

maleck13:
@guicassolato @eguzki may be best placed to comment. It would seem reasonable to allow a user to specify a list of GatewayClasses that they want us to interact with. Not sure if we still check specifically for the controller names?

maleck13:
Yes I think you are correct @guicassolato it was my mistake

eguzki:
why not? 👍

maleck13:
Conversation
https://kubernetes.slack.com/archives/C05J0D0V525/p1741936555648029

We ended up on using EnvVars 

proposal is to add support for 2 new env vars that can take multiple comma separated values

```
ISTIO_CONTROLLER_NAMES=c1,c2
ENVOYGATEWAY_CONTROLLER_NAMES=c3,c4
```

maleck13:
Interesting idea @eguzki . @shaneutt does an annotation like this fit your requirement?

shaneutt:
This seems functional, my only hesitation is a general desire to not have "forever annotations". I lean a bit towards preferring this be an interim solution for that reason. That said, if this is the best thing to do for now then 👍

cc @Miciah

maleck13:
@eguzki Given a pretty specific use case , while adding env vars does require patching the deployment of the operator or at least us providing a configmap to use to load values? I also tend to shy away from annotations for most things. In time I could imagine we allow this config option to be surfaced up through the Kuadrant CR  if the usage of it becomes more core to the use of Kuadrant.

eguzki:
Sure thing @maleck13 and @shaneutt . I was just trying to shape the system to something more easier to manage. But provided your (always welcome) feedback, happy to have this custome behavior ruled by an env var. The env var ends up in the deployment and can be customized with Helm Values or OLM subscription.

PR #1240, opened 2025-03-26
Custom gateway controller names

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/1200

This PR introduces a way to configure multiple controller names, as in [GatewayClass.ControllerName](https://gateway-api.sigs.k8s.io/api-types/gatewayclass/#gatewayclass-parameters) exposing an environment variable per Gateway implementation, `ISTIO_GATEWAY_CONTROLLER_NAMES` and `ENVOY_GATEWAY_GATEWAY_CONTROLLER_NAMES` respectively. Which configures the operator to also reconcile with multiple custom names, i.e:

`ISTIO_GATEWAY_CONTROLLER_NAMES= istio.io/gateway-controller-alpha1, istio.io/gateway-controller-rc2`

Kuadrant operator will reconcile not only the controllers defined in the env var, but also the default one `istio.io/gateway-controller`

**Notes**
* Should we document this? and if so, how/where?
* To configure this on Helm Charts/OLM would be best to have a different issue, since it's a much bigger topic, at least with helm values.yml

### DIFF
--- a/internal/controller/data_plane_policies_workflow.go
+++ b/internal/controller/data_plane_policies_workflow.go
@@ -2,6 +2,9 @@ package controllers
 
 import (
 	"fmt"
+	"strings"
+
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
@@ -48,6 +51,9 @@ var (
 		{Kind: &kuadrantenvoygateway.EnvoyPatchPolicyGroupKind},
 		{Kind: &kuadrantenvoygateway.EnvoyExtensionPolicyGroupKind},
 	}
+
+	istioGatewayControllerNames        = getGatewayControllerNames("ISTIO_GATEWAY_CONTROLLER_NAMES", defaultIstioGatewayControllerName)
+	envoyGatewayGatewayControllerNames = getGatewayControllerNames("ENVOY_GATEWAY_GATEWAY_CONTROLLER_NAMES", defaultEnvoyGatewayGatewayControllerName)
 )
 
 //+kubebuilder:rbac:groups=kuadrant.io,resources=authpolicies,verbs=get;list;watch;update;patch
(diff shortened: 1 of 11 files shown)


internal/controller/data_plane_policies_workflow.go
```diff
@@ -119,3 +127,24 @@ func gatewayComponentsToSync(gateway *machinery.Gateway, componentGroupKind sche
...
+	})
+
+	if envValue == defaultGatewayControllerName {
+		return gatewayControllers
+	}
+	return append(gatewayControllers, gatewayapiv1.GatewayController(defaultGatewayControllerName))
+}
+
+func defaultGatewayControllerName(controllerName gatewayapiv1.GatewayController) gatewayapiv1.GatewayController {
+	if slices.Contains(istioGatewayControllerNames, controllerName) {
```
guicassolato, line 144: Can we please use `github.com/samber/lo` here and avoid importing another package (even if stdlib)?

internal/controller/data_plane_policies_workflow.go
```diff
@@ -119,3 +127,24 @@ func gatewayComponentsToSync(gateway *machinery.Gateway, componentGroupKind sche
 	}
 	return nil
 }
+
+func getGatewayControllerNames(envName string, defaultGatewayControllerName string) []gatewayapiv1.GatewayController {
+	envValue := env.GetString(envName, defaultGatewayControllerName)
+	gatewayControllers := utils.Map(strings.Split(envValue, ","), func(c string) gatewayapiv1.GatewayController {
```
guicassolato, line 133: Can we please use `github.com/samber/lo` here and avoid importing another package?

didierofrivia: is there any equivalent to `strings.Split` in `lo` ?

eguzki: There is nothing wrong with `utils.Map`. As long as it is there, we are free to use it. 

@guicassolato if you think `github.com/samber/lo` is a better library than 4 lines of code on our own repo, be my guest and open a PR deleting the `utils.Map` function from the repo.

didierofrivia: I think he means to avoid including `strings` and `slices` and use the already imported `github.com/samber/lo`

eguzki: `strings` and `slices` packages come from the standard library, which is always more stable and performant than any other third party lib. 

@guicassolato can you clarify your request?

didierofrivia: Maybe we could decide if it makes sense to remove the slice utils in favor of `lo` (not as part of this PR tho)

internal/controller/data_plane_policies_workflow_test.go
```diff
@@ -0,0 +1,35 @@
...
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+func TestGetGatewayControllerNames(t *testing.T) {
+	t.Setenv("ISTIO_GATEWAY_CONTROLLER_NAMES", "istio-alpha1 , istio-alpha2  ")
+
+	istioGwCtrlNames := getGatewayControllerNames("ISTIO_GATEWAY_CONTROLLER_NAMES", "default-istio")
+	envoyGwGwCtrlNames := getGatewayControllerNames("ENVOY_GATEWAY_GATEWAY_CONTROLLER_NAMES", "default-envoy")
+
+	fmt.Println(istioGwCtrlNames)
```
eguzki, line 17: :fire: ???

eguzki (approved):
LGTM

:1st_place_medal:

eguzki (approved):

maleck13 (approved):

guicassolato (approved):

PR #1242, opened 2025-03-26
fix(ratelimit): transform counter "`variable.expression`s" properly

alexsnaps:
~~Tho, that fix isn't really correct. We should parse the expression and apply an AST transformation that has all "root `ident`s" be accessed through `descriptors[0]` instead.~~

~~This would "fix" variables/counter qualifiers such as `user.id`, prefixing with `descriptors[0].`...
But e.g. "real" expressions, e.g. `"foo/" + user.id` would just fail completely. So I guess that's _not_ good. I don't know that we really documented these tho, but we need to properly transform the AST to make this "correct".~~

update: As of 9c451e8, this was actually doing the proper thing and using the AST to do the rewrite

### DIFF
--- a/internal/cel/transformer.go
+++ b/internal/cel/transformer.go
@@ -45,12 +45,8 @@ func TransformCounterVariable(expression string) (*string, error) {
 			}
 		} else if expr.Kind() == ast.CallKind {
 			call := expr.AsCall()
-			if call.FunctionName() == "_[_]" {
-				stack.push(call.Args()[0])
-			} else {
-				for _, arg := range call.Args() {
-					stack.push(arg)
-				}
+			for _, arg := range call.Args() {
+				stack.push(arg)
 			}
 		} else if expr.Kind() == ast.SelectKind {
 			stack.push(expr.AsSelect().Operand())
--- a/internal/cel/transformer.go
+++ b/internal/cel/transformer.go
@@ -3,6 +3,7 @@ package transformer
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 
 	"github.com/google/cel-go/common"
@@ -25,6 +26,7 @@ func parseExpression(expression string) (*ast.AST, error) {
 }
 
 func TransformCounterVariable(expression string) (*string, error) {
+	knownAttributes := []string{"request", "source", "destination", "connection", "metadata", "filter_state", "auth", "ratelimit"}
 	var err error
 	var p *ast.AST
 	if p, err = parseExpression(expression); err != nil {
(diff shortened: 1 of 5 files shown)


internal/cel/transformer.go
```diff
@@ -0,0 +1,45 @@
...
+
+	p, iss := prsr.Parse(common.NewTextSource(expression))
+	if len(iss.GetErrors()) > 0 {
+		return nil, errors.New("got errors parsing")
+	}
+
+	return p, nil
+}
+
+func SafeToSimplyPrefix(expression string) bool {
```
alexsnaps, line 28: This probably makes it a whole lot safer already... while I try to get an AST transformation going.

internal/cel/transformer.go
```diff
@@ -0,0 +1,126 @@
...
+
+	decls := make(map[string]ast.CallExpr)
+	toReplace := make([]ast.OffsetRange, 0)
+	stack := newStack()
+	stack.push(p.Expr())
+
+	for next := stack.pop(); next != nil; next = stack.pop() {
+		expr := *next
+		if expr.Kind() == ast.IdentKind && expr.AsIdent() != "descriptors" {
+			if decls[expr.AsIdent()] == nil {
```
alexsnaps, line 42: Should this be "simplified", and become a "[known attribute](https://github.com/Kuadrant/architecture/blob/main/rfcs/0002-well-known-attributes.md)" list that get replaced instead?
i.e. `request`, `auth`, `source`, `destination`, ... 

and for all the others, we're leaving it to the user to do "the proper" thing?

alexsnaps: I've done that in [this commit](https://github.com/Kuadrant/kuadrant-operator/pull/1242/commits/fd035feb05aae8183806e0239d10dd501701363b)

internal/cel/transformer.go
```diff
@@ -0,0 +1,128 @@
...
+
+func parseExpression(expression string) (*ast.AST, error) {
+	prsr, err := parser.NewParser()
+	if err != nil {
+		return nil, err
+	}
+
+	p, iss := prsr.Parse(common.NewTextSource(expression))
+	if len(iss.GetErrors()) > 0 {
+		return nil, errors.New("got errors parsing")
```
maleck13, line 22: Would it be useful to use `iss.ToDisplayString()` in the error returned here?

internal/cel/transformer.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return p, nil
+}
+
+func TransformCounterVariable(expression string) (*string, error) {
```
maleck13, line 28: as this is very limitador v1 to v2 specific, a code comment could be useful here

maleck13 (dismissed):
This looks good to me. Will be honest had to spend a bit of time reading the transform logic "to get it" but it is readable and understandable. Good set of tests also 👍

maleck13 (approved):

PR #1145, opened 2025-01-29
`RateLimit.Conditions` need explicit scope

alexsnaps:
With Limitador v2.0, conditions need to specify where the variables are to be looked up. Variables coming from the envoy RLP are populated in the `descriptors` root binding. While the protocol allows for multiple descriptors (each with multiple entries), the wasm-shim will only ever use the first one, i.e. index 0.

Fixes #1144

PR #1243, opened 2025-03-27
revert RateLimit.Conditions need explicit scope from #1145

eguzki:
Reverting the following commit from #1145 
```
commit 42d065f2881b341c0d4123de292eb19e0d150f37
Author: Alex Snaps <alex@wcgw.dev>
Date:   Wed Jan 29 13:28:50 2025 -0500

    RateLimit.Conditions need explicit scope

    With Limitador v2.0, conditions need to specify where the variables
    are to be looked up. Variables coming from the envoy RLP are populated
    in the  root binding. While the protocol allows for
    multiple descriptors (each with multiple entries), the wasm-shim will
    only ever use the first one, i.e. index 0.
``


Boomatang (approved):
This is looking good to me.

didierofrivia (approved):
⛑️

PR #1247, opened 2025-03-31
Prepare release 1.1.0

Boomatang:
The following release candidate of Kuadrant Operator version 1.1.0 includes:
  * Authorino Operator version 0.18.0
  * Console Plugin version 0.1.0
  * DNS Operator version 0.13.0
  * Limitador Operator version 0.13.1
  * WASM Shim version 0.9.0

### DIFF
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -108,14 +108,14 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Integration & Delivery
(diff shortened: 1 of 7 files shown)


trepel (commented):
I reviewed the code, it looks good to me. I also think that the branches are correct (this PR is from `release-v1.1.0` to `release-v1.1`). I grepped the code and there is no "-rc2" occurrence left in `release-v1.1`.

The `quay.io/kuadrant/kuadrant-operator:v1.1.0` image is not in place yet (it will only appear after this one gets merged) so not possible to install and test this on cluster.

Not sure what else to check before approving this. I think that this is good to go.

eguzki (approved):

PR #1255, opened 2025-04-04
remove unuseful janitors as final steps

eguzki:
@KevFan @Boomatang what is the utility of such reconcilers? If none, better to be removed.


eguzki:
All tests passed, so either those reconcilers were not useful or are not covered from tests

Boomatang (approved):
Those Janitors never did anything, very happy to see them removed. I do wonder what we expected we were going to need them for.

PR #1257, opened 2025-04-04
feat(extensions): Emerging EmbeddedPlugins

alexsnaps:
See #1236


internal/extension/oop.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+func (p *OOPExtension) Start() error {
+	p.logger.Info("starting...")
+
+	if err := p.startServer(); err != nil {
+		return err
+	}
+
+	cmd := exec.Command(p.executable, p.socket) // #nosec G204
```
didierofrivia, line 76: Are we passing the socket to the command args for any reason in particular?

internal/extension/oop.go
```diff
@@ -0,0 +1,160 @@
...
+
+	"github.com/go-logr/logr"
+	"google.golang.org/grpc"
+)
+
+const defaultUnixSocket = ".grpc.sock"
+
+type OOPExtension struct {
+	name       string
+	executable string
```
didierofrivia, line 37: If we would like to send arguments to the `exec.Cmd` we could change the signature to maybe a `[]string`, being `executable[0]` the command and the rest the args.

alexsnaps: This proposes another approach, as this actually does pass arguments, well _one_ argument. That's the `socket`. So instead of "just" passing "any arg", the API here specifies the semantic of the arg. And actually how the value is passed to the child is abstracted from the API. As it turns out today the socket is passed (see above as to why) via an argument, but it could use any other mean to achieve the same goal, the caller of this API should _not_ know about how that's achieved

internal/extension/oop.go
alexsnaps: Cause the child need to know where to connect to. So rather than doing this by convention (and if it changes, you better _not_ use an "old" client with a new kuadrant operator or vice-versa), this does it by "dependency injection" and injects the endpoint to connect to as an argument passed to the child process... are you proposing doing this differently? e.g. `stdin`? My thinking is that this is "static" and with passing it as an arg, this can very easily be debugged at runtime (e.g. even using `ps` on the pod)

didierofrivia: I reckon it's good only if we are not meaning to pass any other args, if that's so, we can deconstruct some []string of args, and pass the socket as the last one

didierofrivia: (or first one)

alexsnaps: I'm really confused by what you are proposing or asking here. This line is creating a command, using the first (or second... `args[1]` as the `socket`, i.e. the location where the unix socket is to be found, e.g. `/tmp/socket` - where as `args[0]` will always be the executable's name). The "socket" is a positional argument as of now, but nothing outside of this implementation concerns itself with that. So it could also be `[]string{socket}` and be deconstructed as `[]string{socket}...`, is that what you are asking for? Or do you want to append (or prepend) other random args? That could be injected by the caller then? What would these args be good for? Where would they come from? 

Cause that's what I'm trying to explain that I didn't want to do, "how" these values are passed to the executable, the caller should be unaware of. I'd rather add a specific field to the struct, say `config string`, that then would be passed as before or after the socket one, should we decide that there is always a config be passed to all extensions. 

But maybe you mean something completely different. Am I using `exec.Command(name string, arg ...string) *Cmd` somehow in an unidiomatic way? What am I missing?

didierofrivia: I guess if the idea is that the caller should not be concerned, and at the time being we are not meant to pass anything else to the exec.Command, then I'm not proposing to change anything.

internal/extension/manager.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return Manager{
+		extensions: extensions,
+		service:    service,
+		logger:     logger,
+	}, err
+}
+
+func (m *Manager) Start() error {
+	var err error
```
didierofrivia, line 69: to avoid defining the empty `var err error` one could returned a _named_ variable as in

```suggestion
func (m *Manager) Start() (err error) {

```

alexsnaps:
So, your suggestion failed the build, then I tried to fix it and it failed again... but then I also saw that the pattern is used across the entire file (i.e. _not_ using named returns) _and_ it can't be used across all the file as you still need the `var`s (including on the first line), because of the "don't use `else` if that returns and is the last statement" lint. Which leads me to think that this currently applies the most consistent style across this file.

alexsnaps:
@didierofrivia just to be clear, this requires a new review (which would be true whatever, other than I would have rejected your suggestion), given the merge queue

alexsnaps:
cc @adam-cattermole in the queue!

PR #1272, opened 2025-04-09
Prepare release 1.2.0-rc2

Boomatang:
The following release candidate of Kuadrant Operator version 1.2.0-rc2 includes:
  * Authorino Operator version 0.18.0
  * Console Plugin version 0.1.1
  * DNS Operator version 0.14.0
  * Limitador Operator version 0.14.0
  * WASM Shim version 0.9.0

### DIFF
--- a/release.yaml
+++ b/release.yaml
@@ -1,5 +1,5 @@
 kuadrant-operator:
-  version: "1.2.0-rc1"
+  version: "1.2.0-rc2"
 olm:
   channels:
     - "alpha"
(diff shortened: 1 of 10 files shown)


release.yaml
```diff
@@ -1,5 +1,5 @@
 kuadrant-operator:
-  version: "1.2.0-rc1"
+  version: "1.2.0-rc2"
 olm:
   channels:
     - "alpha"
```
eguzki, line 5: we can leave this as alpha. The workflow generating the catalog has `stable` hardcoded https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/build-images-for-tag-release.yaml#L173

eguzki (approved):

PR #1273, opened 2025-04-09
Add initial parts to build and deploy an embedded extension

adam-cattermole:
Deploy/build with extensions:
```sh
WITH_EXTENSIONS=true make local-setup
```

Restart with extensions:
```sh
WITH_EXTENSIONS=true make local-deploy && kubectl rollout restart -n kuadrant-system deployment/kuadrant-operator-controller-manager
```

Set the `WITH_EXTENSIONS` flag in the deployment:
```sh
kubectl set env -n kuadrant-system deployments/kuadrant-operator-controller-manager WITH_EXTENSIONS="true"
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -409,9 +409,6 @@ bundle: $(OPM) $(YQ) manifests dependencies-manifests kustomize operator-sdk ## 
 	# Set desired ConsolePlugin image
 	V="$(RELATED_IMAGE_CONSOLEPLUGIN)" \
 	$(YQ) eval '(select(.kind == "Deployment").spec.template.spec.containers[].env[] | select(.name == "RELATED_IMAGE_CONSOLEPLUGIN").value) = strenv(V)' -i config/manager/manager.yaml
-	#  Set WITH_EXTENSIONS env var
-	V="$(WITH_EXTENSIONS)" \
-	$(YQ) eval '(select(.kind == "Deployment").spec.template.spec.containers[].env[] | select(.name == "WITH_EXTENSIONS").value) = strenv(V)' -i config/manager/manager.yaml
 	# Set desired operator image
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	# Update CSV
--- a/Makefile
+++ b/Makefile
@@ -173,6 +173,8 @@ endif
 ## gatewayapi-provider
 GATEWAYAPI_PROVIDER ?= istio
 
+WITH_EXTENSIONS ?= false
+
 all: build
 
 ##@ General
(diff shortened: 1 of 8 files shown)


extension.Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -ldflags "-X main.version=${VERSION} -X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" -o manager cmd/main.go
+
+
+RUN mkdir -p extensions/myextension
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -o extensions/myextension/myextension cmd/extension/main.go
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
adam-cattermole, line 40: Had issues with Podman using a local image as a base. I initially wanted to use the normal process to build the image, but depending on the value of `IMG` add our extension executable into it using a builder layer. Would not work in the case of `IMG` set to quay.io/kuadrant/kuadrant-operator:dev as this doesn't exist in quay and wasn't able to pull the image

internal/controller/example_extension_reconciler.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func NewExampleReconciler(client *dynamic.DynamicClient, logger logr.Logger) *ExampleExtensionReconciler {
+	return &ExampleExtensionReconciler{
+		client: client,
+		logger: logger,
+	}
+}
+
+func (e *ExampleExtensionReconciler) Reconcile(_ context.Context, _ reconcile.Request) (reconcile.Result, error) {
```
adam-cattermole, line 27: Right now this is a standard controller, but we need to create a wrapper that provides the KuadrantCtx that this would implement

Makefile
```diff
@@ -173,6 +173,8 @@ endif
 ## gatewayapi-provider
 GATEWAYAPI_PROVIDER ?= istio
 
+WITH_EXTENSIONS ?= false
```
adam-cattermole, line 176: Instead of a boolean, maybe we could use a list of extensions, which gets fed into the extension manager as the set to start? if list is empty disabled?

cmd/main.go
```diff
@@ -151,6 +153,19 @@ func main() {
...
+			setupLog.Error(err, "unable to create extension manager")
+			os.Exit(1)
+		}
+		if err = extManager.Start(); err != nil {
+			setupLog.Error(err, "unable to start extension manager")
+			os.Exit(1)
+		}
+	}
+
 	stateOfTheWorld := controllers.NewPolicyMachineryController(mgr, client, log.Log)
```
alexsnaps, line 169: I think we'll eventually need to move this up, and pass that to the `extManager`, I'll work on that as soon as this is merged!

alexsnaps (dismissed):
afaict this remains effect free unless `WITH_EXTENSIONS`, so lgtm for what we're trying to achieve right now

cmd/main.go
adam-cattermole: Yep 100% was thinking that

adam-cattermole:
I guess the one thing with my changes actually is the `WITH_EXTENSIONS` being included in the manager template would mean this would end up being included in the bundle and added to deployments in releases, I'm now wondering whether I should remove that and we should just patch the deployment with a temporary make target. wdyt @alexsnaps?

adam-cattermole:
> I guess the one thing with my changes actually is the WITH_EXTENSIONS being included in the manager template would mean this would end up being included in the bundle (through the CSV) and added to deployments in releases, I'm now wondering whether I should remove that and we should just patch the deployment with a temporary make target

Decided to go ahead with this in 407f7da, we can update the deployment with (not sure there's a need for a make target in the end):

```sh
kubectl set env -n kuadrant-system deployments/kuadrant-operator-controller-manager WITH_EXTENSIONS="true"
```

alexsnaps (approved):

Issue #1154, opened 2025-01-31
Update documentation to reflect the new method of configuration.


laurafitzgerald:
Update https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/install/mtls-configuration.md to reflect the new method of configuration. Remove the manual steps due to maintenance, going out of date, personal label removal

Issue #1152, opened 2025-01-31
mTLS Configuration


laurafitzgerald:
Based on discussion in https://github.com/Kuadrant/kuadrant-operator/issues/1130 ,
add configuration to the Kuadrant CR to enable Zero Trust configuration.

maleck13:
@laurafitzgerald I would be in favour of calling this mTLS rather than zero trust. Kinda like calling a spade a spade :) cc @guicassolato

laurafitzgerald:
> [@laurafitzgerald](https://github.com/laurafitzgerald) I would be in favour of calling this mTLS rather than zero trust. Kinda like calling a spade a spade :) cc [@guicassolato](https://github.com/guicassolato)

ack. I don't have a strong opinion on the naming. I believe @alexsnaps had mentioned it being called zero-trust. Happy to go with whatever naming folks prefer. 👍

PR #1276, opened 2025-04-10
Update documentation to reflect the new method of mTLS configuration

eguzki:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/1154

### DIFF
--- a/doc/install/mtls-configuration.md
+++ b/doc/install/mtls-configuration.md
@@ -17,132 +17,36 @@ At the time of writing there is [an RFC](https://github.com/Kuadrant/architectur
 
 ## Enabling mTLS
 
-### Kuadrant components
+In order to ensure that communications between the gateway and the kuadrant components are secured,
+set kuadrant's custom resource `spec.mtls.enable` field to `true`.
 
-As the Kuadrant components (Authorino & Limitador) are already part of the service mesh in Istio, mTLS can be enabled after an Envoy proxy sidecar is deployed alongside them.
-To do this, apply the Istio sidecar label to both Deployment templates.
+Example:
 
-```bash
-kubectl -n kuadrant-system patch deployment authorino \
-  -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject":"true"}}}}}'
-
-kubectl -n kuadrant-system patch deployment limitador-limitador \
-  -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject":"true"}}}}}'
+```yaml
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: kuadrant-sample
+spec:
+  mtls:
+    enable: true
 ```
-
-You should see the number of containers in either pod increase from 1 to 2, as the `istio-proxy` is added to the pods. This change will force all traffic to those pods to go through the proxy. However, mTLS is not enabled yet.
-
-### Envoy Filter
-
-The next step enables mTLS for traffic originating in the gateway (where the Wasm plugin executes), going to the Kuadrant components.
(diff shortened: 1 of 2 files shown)


doc/install/mtls-configuration.md
```diff
@@ -17,132 +17,36 @@ At the time of writing there is [an RFC](https://github.com/Kuadrant/architectur
 
 ## Enabling mTLS
 
-### Kuadrant components
+In order to ensure that communications between the gateway and the kuadrant components are secured,
+set kuadrant's custom resource `spec.mtls.enable` field to `true`.
 
-As the Kuadrant components (Authorino & Limitador) are already part of the service mesh in Istio, mTLS can be enabled after an Envoy proxy sidecar is deployed alongside them.
```
eguzki, line 22: It's a shame to drop all this documentation. But from a user perspective it is implementation details and should not be included in a user guide. 

Thinking on some architecture doc to record implementation details like this. However, also hesitating to do so because implementation changes imply maintenance of such architecture docs

doc/install/mtls-configuration.md
```diff
@@ -12,137 +12,42 @@ At the time of writing there is [an RFC](https://github.com/Kuadrant/architectur
 
 ## Prerequisites
 
+- Have [Istio](https://istio.io/) as the Gateway API provider installed.
 - You have installed Kuadrant in a Kubernetes cluster.
 - Additionally, you have at least 1 AuthPolicy or RateLimitPolicy attached to your Gateway or HTTPRoute.
```
maleck13, line 17: hmm this is unfortunate perhaps we should look at an enhancement here to allow MTLS without needing an additional policy in place?

doc/reference/kuadrant.md
```diff
@@ -7,11 +7,26 @@
...
+| **Field** | **Type**                          | **Required** | **Description**                      |
+|-----------|-----------------------------------|:------------:|--------------------------------------|
+| `observability`    | [Observability](#observability)     | No | Kuadrant observability configuration. |
+| `mtls`  | [mTLS](#mtls) |      No      | Two way authentication between kuadrant components. |
+
+#### mTLS
+
+| **Field** | **Type**                          | **Required** | **Description**                      |
+|-----------|-----------------------------------|:------------:|--------------------------------------|
+| `enable`    | Boolean     |  No | Enable mutual authentication on every communication between any kuadrant components. Default: `false`|
```
maleck13, line 21:
```suggestion
| `enable`    | Boolean     |  No | Enable mutual authentication on every communication between any the gateway and the kuadrant data plane components. Default: `false`|
```

maleck13 (approved):
minor reword just to make it clear it is only the data plane components and not for example the dns-operator

doc/install/mtls-configuration.md
david-martin: If there are no policies, then there's no kuadrant data plane components to ensure mtls between?

eguzki: let me add a note to make it very clear

david-martin: Ack.
Sounds better to *not* have this requirement called out, provided the implementation will kick in when needed even if mtls was configured before a policy existed (which I believe it should)

eguzki: pushed a note. Let me know what you think.

> provided the implementation will kick in when needed even if mtls was configured before a policy existed (which I believe it should)

Currently not implemented. I captured this feature in https://github.com/Kuadrant/kuadrant-operator/issues/1284 and included as child of the parent issue: mTLS Configuration #1152

david-martin (approved):

david-martin:
fyi @smccarthy-ie

PR #1277, opened 2025-04-10
fix codecov tests

eguzki:
### What 

Update codecov configuration to match the new folder layout after  Kubebuilder v4 migration   https://github.com/Kuadrant/kuadrant-operator/pull/1231

Supersedes https://github.com/Kuadrant/kuadrant-operator/pull/1231


didierofrivia (approved):
We might want to tune a bit more so we can agree on the expectations of the coverage

PR #1287, opened 2025-04-14
Extensions OOP Logging round 2

didierofrivia:
Screenshot 2025-04-16 at 15 55 37

Notes:
* Is not possible to feed the `logr.LogSink` directly with Bytes or a JSON string, since it doesn't have anything to write to it. Its interface defines Info and Error similar to the logr wrapper, but with an extra log level int
* Since at the time we are not discerning the log level from the kuadrant-operator, the OOP extensions will use the same log level, thus I've decided to use the _logr_ instance methods to log, instead of directly its _sink_
* I've found that Unmarshalling the bytes coming from the OOP extension Stderr Pipe was necessary to correctly interact with the `logr` instance, and thus later to add extra info in our future SDK bits.

### DIFF
--- a/internal/extension/oop.go
+++ b/internal/extension/oop.go
@@ -18,6 +18,7 @@ package extension
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
@@ -33,13 +34,6 @@ import (
 )
 
 const defaultUnixSocket = ".grpc.sock"
-
-type logLevel int
-
-const (
-	LogLevelInfo logLevel = iota
-	LogLevelError
-)
 
 type OOPExtension struct {
 	name       string
@@ -103,7 +97,7 @@ func (p *OOPExtension) Start() error {
 
 	go func() {
 		if e := cmd.Wait(); e != nil {
-			p.logStderr(1, fmt.Sprintf("Extension %q finished with an error", p.name), e)
+			p.logger.Error(e, fmt.Sprintf("Extension %q finished with an error", p.name))
 			close(stopChan)
 		}
 	}()
@@ -201,8 +195,12 @@ func (p *OOPExtension) monitorStderr(stderr io.ReadCloser, stopChan <-chan struc
 			}
 
 			if scanner.Scan() {
-				lvl, text, err := parseStderr(scanner.Bytes())
-				p.logStderr(lvl, text, err)
+				logLine, err := unmarshalLogEntry(scanner.Bytes())
+				if err != nil {
+					p.logger.Error(err, "failed to parse extension log entry")
+					return
+				}
+				p.logStderr(logLine)
 				lastReadTime = time.Now()
 			} else if err := scanner.Err(); err != nil {
 				p.logger.Error(err, "failed to read stderr")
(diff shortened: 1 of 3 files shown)


internal/extension/oop.go
```diff
@@ -201,8 +195,12 @@ func (p *OOPExtension) monitorStderr(stderr io.ReadCloser, stopChan <-chan struc
 			}
 
 			if scanner.Scan() {
-				lvl, text, err := parseStderr(scanner.Bytes())
-				p.logStderr(lvl, text, err)
+				logLine, err := unmarshalLogEntry(scanner.Bytes())
+				if err != nil {
+					p.logger.Error(err, "failed to parse extension log entry")
+					return
```
alexsnaps, line 201: So in "non dev mode", (i.e. unstructured or regular logging) this would fail?
Do we then say that an extension needs to _always_ log in structured and the parent would "convert" when needed?

internal/extension/oop.go
```diff
@@ -214,31 +212,53 @@ func (p *OOPExtension) monitorStderr(stderr io.ReadCloser, stopChan <-chan struc
...
+	LogLevelError logLevel = "error"
+)
+
+type oopLogEntry struct {
+	Level         logLevel               `json:"level"`
+	Msg           string                 `json:"msg"`
+	Error         string                 `json:"error,omitempty"`
+	Timestamp     string                 `json:"ts"`
+	KeysAndValues map[string]interface{} `json:"-,omitempty"`
+}
```
alexsnaps, line 243: Is this defined anywhere in the `zap` or whatever defines "structured logging"? I'd rather have compile time checks should this change.

internal/extension/oop.go
```diff
@@ -214,31 +212,53 @@ func (p *OOPExtension) monitorStderr(stderr io.ReadCloser, stopChan <-chan struc
...
-	// Convert first byte to integer and validate log range
-	logLvl = logLevel(logLine[0])
-	if logLvl < LogLevelInfo || logLvl > LogLevelError {
-		// At the moment only differencing from Info and Error
-		return LogLevelError, "", fmt.Errorf("first byte is not a valid log level range 0-1. log: %q", string(logLine))
-	}
+const (
+	LogLevelInfo  logLevel = "info"
+	LogLevelError logLevel = "error"
+)
```
alexsnaps, line 235: Should we reuse [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/log/log.go#L41-L58) ?

internal/extension/oop.go
```diff
@@ -214,31 +212,53 @@ func (p *OOPExtension) monitorStderr(stderr io.ReadCloser, stopChan <-chan struc
...
+	}
+	switch logLine.Level {
 	case LogLevelInfo:
-		p.logger.Info(logString)
+		p.logger.Info(logLine.Msg, keysAndValues...)
+	case LogLevelError:
+		p.logger.Error(fmt.Errorf("%v", logLine.Error), logLine.Msg, keysAndValues...)
 	default:
-		p.logger.Error(err, logString)
+		p.logger.Error(fmt.Errorf("unknown LogLevel %v", logLine.Level), logLine.Msg, keysAndValues...)
```
alexsnaps, line 226: Further log levels do exists, see [the slog](https://pkg.go.dev/github.com/go-logr/logr#FromContextAsSlogLogger) API

alexsnaps (commented):
The route you are going down here I fear would result in a whole lot of coupling and more work (to do and maintain). But I may well be missing pros and focusing on cons.

To me it looks both easier and simpler to to "link" child processes and parent's logging to the env vars and "simply forward". 

```go
	sync := zapcore.Lock(zapcore.AddSync(os.Stdout))
	logger := log.NewLogger(log.WriteTo(sync)) // that's kuadrant current log package

	ctx := logr.NewContext(context.TODO(), logger)
	logr.FromContextAsSlogLogger(ctx).Warn("Danger here!")

	go func() {
		logger.Error(nil, "EFGH")
	}()

	n, e := sync.Write([]byte{65, 66, 67, 68}) // lets you write the bytes straight to whatever that logger was wired to log to
	if n != 4 || e != nil {
		t.Errorf("Failed to write to zap logger")
	}

	time.Sleep(time.Second) // wait for the go routines to end and observe no tearing.
```

didierofrivia:
Thanks @alexsnaps for the review and the explanation, I'm closing this PR in favor of https://github.com/Kuadrant/kuadrant-operator/pull/1292, will develop further.

PR #1288, opened 2025-04-15
delete unwanted service and pod monitors during observability reconciliation

Patryk-Stefanski:
### Verification steps

1. Setup environment based on EnvoyGateway

`make local-setup GATEWAYAPI_PROVIDER=envoygateway`

2. Apply the Kuadrant custom resource with observability enabled

```
---
kind: Kuadrant
metadata:
  name: kuadrant-sample
spec:
  observability:
    enable: true
EOF
```

3. Wait for kuadrant instance to be ready
`kubectl wait --timeout=300s --for=condition=Ready kuadrant kuadrant-sample -n kuadrant-system`

4. Verify the monitoring resources are there
`oc get servicemonitor,podmonitor -A`

5. Check existing gateway
`oc get gateways -A `

6. Delete existing gatway
`oc delete gateway kuadrant-ingressgateway -n gateway-system`

7. Check podmonitor
`oc get podmonitor -A`

8. The envoy-stats pod monitor should be deleted

9. Check the logs of the kuadrant operator and you should see relevant log 
`{"level":"info","ts":"2025-04-15T15:36:28Z","logger":"kuadrant-operator.ObservabilityReconciler","msg":"deleting unwanted monitors","monitors":"envoy-stats-monitor","namespace":"gateway-system"}
`

10. Similarly enough deleting the envoy gatewayclass should delete the `envoy-gateway-monitor` service monitor

### DIFF
--- a/internal/controller/observability_reconciler.go
+++ b/internal/controller/observability_reconciler.go
@@ -399,26 +399,72 @@ func (r *ObservabilityReconciler) Reconcile(baseCtx context.Context, _ []control
 	gatewayClasses := topology.Targetables().Items(func(o machinery.Object) bool {
 		return o.GroupVersionKind().GroupKind() == machinery.GatewayClassGroupKind
 	})
+
+	wantedSMs := []monitoringv1.ServiceMonitor{
+		*kOpMonitor,
+		*dnsOpMonitor,
+		*authOpMonitor,
+		*limitOpMonitor,
+	}
+
+	var wantedPMs []monitoringv1.PodMonitor
+	var monitorsToDelete []machinery.Object
+
 	for _, gatewayClass := range gatewayClasses {
 		gateways := topology.All().Children(gatewayClass)
 		gwClass := gatewayClass.(*machinery.GatewayClass)
 		if lo.Contains(istioGatewayControllerNames, gwClass.GatewayClass.Spec.ControllerName) {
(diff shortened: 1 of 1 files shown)


eguzki (approved):
Nice verification steps. Working well.

Consider adding some integration test where deleting a gateway, the associated pod monitor is also deleted. 

The code looks good. But does not take advantage of the topology. The code would be easier to implement  if it were just iterating over existing pod monitors and service monitors and delete those that do not have a parent (gateway or istio/envoygateway control plane instance). Not asking to implement that on this PR. Just a thought. 

LGTM

Patryk-Stefanski:
I looked into using owner references for this, but that doesn't work across different namespaces. And as for looking through existing monitors that don't have an associated gateways, it probably would be more complicated than the current approach, as we don't know what monitors are associated with a gateway.

Thinking of it now, we could annotate or label the monitors with the associated gateway, which would make the cleanup a bit easier 🤔, but then again I don't think relying on labels for cleanup is a great thing.

eguzki:
Not referring Ownerrefs or labels. For the reasons you point out correctly.

I am referring to the topology. Right now the service monitors and pod monitors are linked to the kuadrant CR, so there is a path between the kuadrant CR and the monitors. It would require gateway and monitors to be linked, so linking function should be smarter.

PR #1292, opened 2025-04-17
OOP Logging - Round 2

didierofrivia:
This PR changes the way the child out of process forwards its logs to the parent (kuadrant operator). The logging of the child OOP is written to the `stderr` which the parent will read and write directly using a _syncer writer_.

The messages are logged as shown below:

Screenshot 2025-04-18 at 16 10 50

### DIFF
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"time"
 
+	"go.uber.org/zap/zapcore"
+
 	extpb "github.com/kuadrant/kuadrant-operator/pkg/extension/grpc/v0"
 
 	"github.com/go-logr/logr"
@@ -31,6 +33,7 @@ type Manager struct {
 	extensions []Extension
 	service    extpb.ExtensionServiceServer
 	logger     logr.Logger
+	sync       zapcore.WriteSyncer
 }
 
 type Extension interface {
@@ -39,7 +42,7 @@ type Extension interface {
 	Name() string
 }
 
-func NewManager(names []string, location string, logger logr.Logger) (Manager, error) {
+func NewManager(names []string, location string, logger logr.Logger, sync zapcore.WriteSyncer) (Manager, error) {
 	var extensions []Extension
 	var err error
 
@@ -47,7 +50,7 @@ func NewManager(names []string, location string, logger logr.Logger) (Manager, e
 	logger = logger.WithName("extension")
 
 	for _, name := range names {
-		if oopExtension, e := NewOOPExtension(name, location, service, logger); e == nil {
+		if oopExtension, e := NewOOPExtension(name, location, service, logger, sync); e == nil {
 			extensions = append(extensions, &oopExtension)
 		} else {
 			if err == nil {
@@ -59,9 +62,10 @@ func NewManager(names []string, location string, logger logr.Logger) (Manager, e
 	}
 
 	return Manager{
-		extensions: extensions,
-		service:    service,
-		logger:     logger,
+		extensions,
+		service,
+		logger,
+		sync,
 	}, err
 }
 
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -19,9 +19,8 @@ package extension
 import (
 	"context"
 	"fmt"
+	"io"
 	"time"
-
-	"go.uber.org/zap/zapcore"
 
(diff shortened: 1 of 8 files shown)


internal/log/log.go
```diff
@@ -32,7 +32,8 @@ import (
 
 var (
 	// Log is the base logger
-	Log logr.Logger = logr.New(ctrllog.NullLogSink{})
+	Log  logr.Logger = logr.New(ctrllog.NullLogSink{})
+	Sync             = zapcore.WriteSyncer(nil)
```
didierofrivia, line 36: I could also refactor and add a `LogWrapper` type, holding them together but I thought this was the least intrusive change

internal/extension/manager.go
```diff
@@ -39,15 +42,15 @@ type Extension interface {
 	Name() string
 }
 
-func NewManager(names []string, location string, logger logr.Logger) (Manager, error) {
+func NewManager(names []string, location string, logger logr.Logger, sync zapcore.WriteSyncer) (Manager, error) {
```
alexsnaps, line 45: I'm thinking I'd make that a `io.Writer` here. That makes it easier to work with, including in tests.

internal/extension/manager.go
```diff
@@ -31,6 +33,7 @@ type Manager struct {
 	extensions []Extension
 	service    extpb.ExtensionServiceServer
 	logger     logr.Logger
+	sync       zapcore.WriteSyncer
```
alexsnaps, line 36: `io.Writer` here too

internal/log/log.go
alexsnaps: If you make this one `io.Writer` too here, then you could wire a `bytes.Buffer` here. Or just have it `nil` directly.

internal/log/log.go
```diff
@@ -116,6 +117,11 @@ func SetLogger(logger logr.Logger) {
 	klog.SetLogger(logger)
 }
 
+// SetSync sets the zapcore.WriterSyncer used to write the logs.
+func SetSync(sync zapcore.WriteSyncer) {
```
alexsnaps, line 121: Same again, `io.Writer` is all we actually need here.

alexsnaps (commented):
No need to couple to `zapcore` here, `io.Writer` is all that's needed. 
Let's then add tests (which will be much simpler to write with that type btw)

alexsnaps:
I think you forgot the matching [changes to `cmd/extension/main.go` reading the env vars](https://github.com/Kuadrant/kuadrant-operator/blob/main/cmd/extension/main.go#L21-L27) as done [in `cmd/main.go` here](https://github.com/Kuadrant/kuadrant-operator/blob/main/cmd/main.go#L61-L62)

adam-cattermole (approved):
This looks good, I'd like to test it with my newer changes. Only one question but not blocking

internal/log/log.go
```diff
@@ -32,7 +32,8 @@ import (
 
 var (
 	// Log is the base logger
-	Log logr.Logger = logr.New(ctrllog.NullLogSink{})
+	Log  logr.Logger = logr.New(ctrllog.NullLogSink{})
+	Sync io.Writer
```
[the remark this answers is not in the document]
adam-cattermole: Ah I see, actually it's set as part of the init and then read from this global var to pass to the extensions in main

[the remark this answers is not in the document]
adam-cattermole: I wonder, should it have a default initialisation similar to the Log? (not sure what that might look like)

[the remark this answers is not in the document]
didierofrivia: Yup, I basically followed how it was done with `Log`, I also considered having a `LogWrapper` or `CustomLogger` type to enclose both but wasn't sure if it was a change for this PR

internal/log/log_test.go
```diff
@@ -61,3 +62,16 @@ func TestToMode(t *testing.T) {
...
 	}()
 }
+
+type writerMock struct{}
+
+func (w *writerMock) Write(p []byte) (n int, err error) {
+	return len(p), nil
+}
+
+func TestSetSync(t *testing.T) {
```
alexsnaps, line 72: Unsure about the value of this test tbh...

didierofrivia: not much value, kind of making codecov happy... which TBH, is a bit annoying

adam-cattermole (approved):

Issue #1321, opened 2025-04-22
# Create workflow Structure


Boomatang:
In the state of the world a new workflow needs to be added for the three distinct features.

Until the kuadrant CR spec has being updated, this workflow should be enabled by a feature flag.

For now an annotation in the kuadrant CR is what will enable this.

PR #1327, opened 2025-04-22
ADD: Resilient Deployment Workflow Structure

Boomatang:
closes: #1321 

Base Branch: https://github.com/Boomatang/kuadrant-operator/tree/Resilient_Deployment

Adds the skeleton structure of the workflow required for the resilientdeployment feature. 
Currently, there is only logging on the debug level.

The workflow is disabled if the kuadrant CR does not have the following
annotation.
`kuadrant.io/experimental-dont-use-resilient-data-plane: "true"`

When the annotation is set there should be a number of debug logs that contain "resilience" in the logs.

### DIFF
--- a/internal/controller/resilient_deployment_workflow.go
+++ b/internal/controller/resilient_deployment_workflow.go
@@ -1,0 +1,201 @@
+package controllers
+
+// TODO: When feature complete, remove all experimental code references.
+
+import (
+	"context"
+	"sync"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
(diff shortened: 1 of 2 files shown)


eguzki:
LGTM

I was wondering if it makes sense to do this kind of preparation PR's (that do nothing), instead of implementing iteratively complete feature of resilience for rate limiting, auth, ...

Boomatang:
To me this is part of the iterative process. It lays the ground work for others to build on.

eguzki:
Not blocking this. In my book, the iterative process is something else. The contributions to `main` branch should be fully implemented features/fixes or whatever _thing_ even if they are very tiny additions. But completed. 

:+1:

maleck13 (commented):
also agree changes look fine, but we might want to group the PRs so that when merged that feature is present? 
I get why you are splitting them but perhaps have gone too far down that road :)

Boomatang:
Closed, do as part of https://github.com/Kuadrant/kuadrant-operator/pull/1368

PR #1328, opened 2025-04-22
feat(extensions): Wiring extensions into the state-of-the-world

alexsnaps:


alexsnaps:
@adam-cattermole I know this misses a whole lot of tests... but I'd already appreciate some feedback.

One issue I have is [with `kuadrantLib.DAG` here](https://github.com/Kuadrant/kuadrant-operator/blob/04edb07e2529a58af33d60d09aadba2d49f0f216/pkg/cel/ext/kuadrant.go#L27). This would ideally now be a "snapshot" of the DAG and... should it be kept consistent for... a single expression evaluation? a "transaction" of such CEL expressions? And, once that's answered, ... how? Given golang's execution model, what's the pattern there? having that snapshot part of the almighty `Context`? Cause a "thread-local" would have no real semantic here...

adam-cattermole (approved):
Nice, looks good to me. We can add tests later as it's dead code without extensions being enabled

PR #1346, opened 2025-04-24
UPDATE: Kuadant CR API

Boomatang:
Add the `resilience.counterStorage` to the kuadrant spec.
This uses the `storage` object from the limitador API


eguzki (commented):
Looks good.

Wondering if it makes sense to merge this NO-OP PR...

Boomatang:
@eguzki I was hoping to make small, easy to review PRs that were only focused on one thing. The fact this is a NO-OP PR never crossed my mind as being an issue. 

Also sorry I pushed a commit to this PR after you had give the LGTM, I did not realize you had gave it an lgtm.

eguzki (approved):

maleck13 (commented):
changes here also look good. I added a comment to the RFC about possible adding a enabled: true default option , but we can discuss and add after if we agree.

Issue #1318, opened 2025-04-22
# Set limitador storage


Boomatang:
## What
When the `spec.resilience.counterStorage` is configured in the kuadrant CR, configure the `spec.storage`in the limitador CR.
The configuration should be mirrored from the kuadrant CR to the limitador CR.
While the feature is enabled any modification the user does to the `spec.storage` in the limitador CR should be reverted. 

## Why
For a resilient deployment counters need to be able to with stand restarts of limitador. 
This can be done by configuring a storage option.

## How
When the `spec.resilience.counterStorage` is configured in the kuadrant CR, get the limitador CR from the topology.
Modify the limitador CR `spec.storage` to have only the values of the `spec.resilience.counterStorage`.

If the user modifies the `spec.storage` in the limitador CR when the `spec.resilience.counterStorage` is configured, the user modifications get reverted.

## Done
- [ ] When the `spec.resilience.counterStorage` is configured in the kuadrant CR, the configuration is mirrored to the limitador CR `spec.storage`.
- [ ] When the `spec.resilience.counterStorage` is configured in the kuadrant CR, and the user modifies the configuration in the limitador CR the changes get reverted.

PR #1347, opened 2025-04-24
Resilient Deployment set counter storage

Boomatang:
Closes: #1318

Requires: 
- #1327 
- #1346

Base Branch: https://github.com/Boomatang/kuadrant-operator/tree/Resilient_Deployment

- **ADD: Resilient Deployment Workflow Structure**
- **ADD: isConfigured (counter storage)**
- **ADD: mirror counterStorage to limitador storage**

Allow the configuration of limitadors `spec.storage` from the kuadrant CR.
The change will maintain the configuration defined in the kuadrant CR.

As this is a work in progress area the experimental annotation is required.

Example kaudrant CR
```yaml
kind: Kuadrant
metadata:
  name: kuadrant
  annotations:
    kuadrant.io/experimental-dont-use-resilient-data-plane: "true"
spec:
  resilience:
    counterStorage:
      disk: {}
```

### DIFF
--- a/internal/controller/resilient_deployment_workflow.go
+++ b/internal/controller/resilient_deployment_workflow.go
@@ -1,0 +1,201 @@
+package controllers
+
+// TODO: When feature complete, remove all experimental code references.
+
+import (
+	"context"
+	"sync"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
+// WARNING: level varible is only here for the basic dev work and should not end up in the finished feature
+// FIXME: don't merge to main with value of zero, set to one.
+var level = 1
+
+const (
+	ExperimentalResilienceFeature = "ExperimentalResilienceFeature"
+	ResilienceFeatureAnnotation   = "kuadrant.io/experimental-dont-use-resilient-data-plane"
+)
+
+func NewResilienceDeploymentWorkflow() *controller.Workflow {
+	return &controller.Workflow{
+		Precondition:  NewResilienceDeploymentPrecondition().Subscription().Reconcile,
+		Tasks:         NewResilienceDeploymentTasks(),
+		Postcondition: NewResilienceDeploymentPostcondition().Subscription().Reconcile,
+	}
+}
+
+// INFO: Precontion Section
+
+func NewResilienceDeploymentPrecondition() *ResilienceDeploymentPrecondition {
+	return &ResilienceDeploymentPrecondition{}
+}
+
+type ResilienceDeploymentPrecondition struct{}
+
+func (r *ResilienceDeploymentPrecondition) Subscription() controller.Subscription {
+	return controller.Subscription{
+		ReconcileFunc: r.run,
+		Events: []controller.ResourceEventMatcher{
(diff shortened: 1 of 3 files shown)


internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,253 @@
...
+		return nil
+	}
+
+	if !r.isConfigured(kObj) {
+		logger.V(level).Info("CounterStorage not configured", "status", "exiting")
+		return nil
+	}
+	logger.V(level).Info("CounterStorage configured", "status", "contiune")
+
+	lObj.Spec.Storage = kObj.Spec.Resilience.CounterStorage
```
eguzki, line 176: What is the benefit of implementing the limitador object reconciliation in two different reconcilers that race and increasing the probability of "stale update" error rising ?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,253 @@
...
+	err := r.updateLimitador(ctx, lObj)
```
eguzki, line 177: update unconditionally?

internal/controller/resilient_deployment_workflow.go
Boomatang: I think the befit is the separation  of concerns is the big benefit. While the reconcile may be tangential to the reconcile of the limitador CR limits based on the policies, there is no dependency between the two. 

But you are right it raises the risk of a stale update. This I was thinking about when do a different part. I am not sure if a reconcile should be writing to the cluster. The machinery policy workflows allow for a post condition. I think all the tasks should write the resources they want to write to the cluster to the state *sync.map that all tasks have access to. Then a the post condition writes all the resources to the cluster.

While there may be a raised risk, I feel the benefit is worth the risk. I don't think lumping all the actions based on a resource is the long term answer.

internal/controller/resilient_deployment_workflow.go
Boomatang: When it said out loud again, this might not be the best thing. After the last failure of the built in go comparison function to compare the limits for the limitador CR I felt this was the easier option. 

I will go back, and take a second look.

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+}
+
+func (r *ResilienceRateLimitingReconciler) reconcile(ctx context.Context, _ []controller.ResourceEvent, _ *machinery.Topology, _ error, state *sync.Map) error {
+	logger := controller.LoggerFromContext(ctx).WithName("ResilienceRateLimitingReconciler")
+
+	logger.V(level).Info("ResilienceRateLimitingReconciler Task", "status", "started")
+	defer logger.V(level).Info("ResilienceRateLimitingReconciler Task", "status", "completed")
+	if !experimentalFeatureEnabledSate(state) {
+		logger.V(level).Info("Experimental resilience feature is not enabled, early exit", "status", "exiting")
```
maleck13, line 124: nit: using the word exit here might be a bit confusing. Like `os.Exit(1)`  // Exit causes the current program to exit with the given status code.

Perhaps `Experimental resilience feature is not enabled. Nothing to do", "status", "completed"`

internal/controller/resilient_deployment_workflow.go
maleck13: I am not convinced. If I were to think about where to look for counter storage handling, it would prob be in the limitador reconciler  ?

Issue #1319, opened 2025-04-22
# UnSet limitador storage


Boomatang:
## What
When the `spec.resilience.counterStorage` configuration is removed the kuadrant CR, the kuadrant-operator should also remove the `spec.storage` from the limitador CR.
While the `spec.resilience.counterStorage` is not configurated in the kuadrant CR, the user should be able to 

## Why
The kuadrant-operator should clean up any configurations it added to cluster when the feature was enabled.
At the same time the user should be allowed to configure the limitador CR `spec.storage` directly without the kuadrant-operator reverting the change.

This also allows upgrades to be non-breaking

## How
When the `spec.resilience.counterStorage` configuration is removed the kuadrant CR, get the limitador CR from the topology and remove the `spec.storage` section.

To ensure the operator does not revert the user defined configuration, this needs to be explored.

## Done
When the `spec.resilience.counterStorage` configuration is removed the kuadrant CR, the `spec.storage` is remove

PR #1348, opened 2025-04-24
Resilient Deployment unset counter storage

Boomatang:
Closes: #1319

Requires: 
- #1327
- #1346
- #1347

- **ADD: Resilient Deployment Workflow Structure**
- **ADD: History**
- **ADD: isConfigured (counter storage)**
- **ADD: mirror counterStorage to limitador storage**
- **ADD: counterStorage UnSet**

This PR introduces a method of clean up related to the `spec.resilience.counterStorage` in the kuadrant CR.
When the field is fully removed from the kuadrant CR the field is also removed from the limitador CR.

There was the introduction of a history object.
This history allows the operator to know if the feature was enable the last time the workflow task was triggered.
This insight supports upgrades by not reverting user defined changes in the limitador CR if the counter storage had never being configured in the kuadrant CR.

**Edge Case**
There is one case where this does fail.
If the operator is scaled down, and the user removes the counter storage configuration from the kuadrant CR, when the operator starts it will not know that it needs to tidy up the limitador CR configuration.
It assumes the existing configuration in the limitador CR was user created.

### DIFF
--- a/internal/controller/history.go
+++ b/internal/controller/history.go
@@ -1,0 +1,54 @@
+package controllers
+
+import (
+	"context"
+	"sync"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
+type History struct {
+	kuadrant *kuadrantv1beta1.Kuadrant
+}
+
+var (
+	history *History
+	mu      sync.RWMutex
+)
+
+func initializeHistory() {
+	mu.Lock()
+	defer mu.Unlock()
+	if history == nil {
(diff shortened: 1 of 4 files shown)


internal/controller/history.go
```diff
@@ -0,0 +1,54 @@
...
+	"sync"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
+type History struct {
+	kuadrant *kuadrantv1beta1.Kuadrant
```
maleck13, line 14: This has a bit of a code smell and seems to indicate a possible issue in the design of the topology? Like if the pod crashes does this not mean history is rewritten from the point the pod comes back to be the current kuadrant state?

PR #1349, opened 2025-04-24
ADD: integration test for counter storage

Boomatang:
Requires:
- #1347
- #1348

This adds integration tests for counter storage configuration
in the kuadrant resource. As there is interaction with the kuadrant
resource the test can **NOT** be run in parallel.


Boomatang:
Closed, do as part of https://github.com/Kuadrant/kuadrant-operator/pull/1368

PR #1351, opened 2025-04-24
feat(extension): Guard the DAG in a blocking atomic ref

alexsnaps:
Replaced the DAG "static" with a "Blocking version"

### DIFF
--- a/internal/extension/reconciler.go
+++ b/internal/extension/reconciler.go
@@ -44,7 +44,7 @@ func (c *nilGuardedPointer[T]) get() *T {
 func (c *nilGuardedPointer[T]) getWait() T {
 	// First try a quick non-blocking check
 	if val := c.ptr.Load(); val != nil {
-		return *c.ptr.Load()
+		return *val
 	}
 
 	c.mu.Lock()
@@ -68,9 +68,7 @@ func (c *nilGuardedPointer[T]) getWaitWithTimeout(timeout time.Duration) (*T, bo
 	timer := time.NewTimer(timeout)
 	defer timer.Stop()
 
-	done := make(chan struct{})
-	var result *T
-	var success bool
+	result := make(chan *T, 1)
 
 	go func() {
 		c.mu.Lock()
@@ -80,14 +78,13 @@ func (c *nilGuardedPointer[T]) getWaitWithTimeout(timeout time.Duration) (*T, bo
 			c.cond.Wait()
 		}
 
-		result = c.ptr.Load()
-		success = true
-		close(done)
+		val := c.ptr.Load()
+		result <- val
 	}()
 
 	select {
-	case <-done:
-		return result, success
+	case val := <-result:
+		return val, true
 	case <-timer.C:
 		return c.ptr.Load(), false
 	}
--- a/internal/extension/reconciler.go
+++ b/internal/extension/reconciler.go
@@ -4,12 +4,97 @@ import (
 	"context"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/kuadrant/policy-machinery/controller"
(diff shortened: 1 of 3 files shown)


internal/extension/reconciler.go
```diff
@@ -4,12 +4,97 @@ import (
...
+func (c *nilGuardedPointer[T]) get() *T {
+	return c.ptr.Load()
+}
+
+// getWait blocks until the pointer is set to a non-nil value and then returns that value.
+func (c *nilGuardedPointer[T]) getWait() T {
+	// First try a quick non-blocking check
+	if val := c.ptr.Load(); val != nil {
+		return *c.ptr.Load()
+	}
```
adam-cattermole, line 48: Could we use the val we got from the first load? just wondering if there's a chance for it to have been changed and it to be nil on second one:
```suggestion
        val := c.ptr.Load()
	if val != nil {
		return *val
	}
```

internal/extension/reconciler.go
```diff
@@ -4,12 +4,97 @@ import (
...
+		success = true
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		return result, success
+	case <-timer.C:
+		return c.ptr.Load(), false
+	}
```
adam-cattermole, line 93: Is there a similar problem here or am I misunderstanding? maybe something like this fixes it?

```suggestion
	result := make(chan *T, 1)

	go func() {
		c.mu.Lock()
		defer c.mu.Unlock()

		for c.ptr.Load() == nil {
			c.cond.Wait()
		}

		val := c.ptr.Load()
		result <- val
	}()

	select {
	case val := <-result:
		return val, true
	case <-timer.C:
		return c.ptr.Load(), false
	}
```

internal/extension/reconciler.go
alexsnaps: Yeah, we could, but if you look at the `.Set` it takes a value. So it can only ever go from `nil` to actual value, so while the value _could_ change in between the 2 `Load()`s, it merely would be "a newer value"

internal/extension/reconciler.go
alexsnaps: as I read it, same code indeed...

internal/extension/reconciler.go
```diff
@@ -4,12 +4,94 @@ import (
...
+// newNilGuardedPointer creates a new nilGuardedPointer.
+func newNilGuardedPointer[T any]() *nilGuardedPointer[T] {
+	cgp := &nilGuardedPointer[T]{}
+	cgp.cond = sync.NewCond(&cgp.mu)
+	return cgp
+}
+
+// set sets the pointer to x and signals any goroutines waiting for a non-nil value.
+func (c *nilGuardedPointer[T]) set(x T) {
+	c.ptr.Store(&x)
```
guicassolato, line 30: `get()` returns `*T`. Should `set(x)` expect `*T` as well?

```suggestion
func (c *nilGuardedPointer[T]) set(x *T) {
	c.ptr.Store(x)
```

internal/extension/reconciler.go
adam-cattermole: Yep sure I see that now. Woops sorry, I broke the indentation...

internal/extension/reconciler.go
alexsnaps: Well, no, because I don't want it settable to `nil` again, the idea is _only_ support `nil` -> `T` transitions. `get` (and the timeout'ing version) have to return a pointer, as they both could still return `nil`, `getWait` can only ever return a `T`. Now it's probably worth noting this used to store (and guard) a `StateAwareDAG`, so this only "copies" two pointers around

adam-cattermole (approved):

PR #1360, opened 2025-04-30
Extension subscriptions

alexsnaps:
WIP still, needs to actually consume these updates and push them down the stream

### DIFF
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"os"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -126,7 +127,8 @@ func (m *Manager) HasSynced() bool {
 }
 
 type extensionService struct {
-	dag *nilGuardedPointer[StateAwareDAG]
+	dag           *nilGuardedPointer[StateAwareDAG]
+	subscribtions sync.Map
 	extpb.UnimplementedExtensionServiceServer
 }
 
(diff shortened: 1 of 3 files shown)


internal/extension/manager.go
```diff
@@ -141,19 +143,34 @@ func newExtensionService(dag *nilGuardedPointer[StateAwareDAG]) extpb.ExtensionS
...
+		dag := <-channel
+
+		opts := []cel.EnvOption{
+			kuadrant.CelExt(&dag),
+		}
+		if env, err := cel.NewEnv(opts...); err == nil {
+			s.subscribtions.Range(func(sub, _ interface{}) bool {
+				if prg, err := env.Program(sub.(subscription).cAst); err == nil {
+					if _, _, err := prg.Eval(sub.(subscription).input); err == nil {
+						_ = stream.Send(&extpb.Event{})
```
alexsnaps, line 157: I could have some input on this @adam-cattermole 
How do we want to go about this? Thinking pushing any `error` down to the extension might actually be desirable here, wdyt?

internal/extension/manager.go
```diff
@@ -141,19 +145,44 @@ func newExtensionService(dag *nilGuardedPointer[StateAwareDAG]) extpb.ExtensionS
...
 	for {
-		time.Sleep(time.Second * 5)
-		if err := stream.Send(&extpb.Event{}); err != nil {
-			return err
+		dag := <-channel
+		opts := []cel.EnvOption{
+			kuadrant.CelExt(&dag),
 		}
+
+		s.mutex.Lock()
```
maleck13, line 155: using defer would also be an option for the unlock

internal/extension/manager.go
```diff
@@ -141,19 +145,44 @@ func newExtensionService(dag *nilGuardedPointer[StateAwareDAG]) extpb.ExtensionS
...
+		if env, err := cel.NewEnv(opts...); err == nil {
```
maleck13, line 156: if there is an error in these cases why not return it? Is the err not of value?

adam-cattermole: I don't want to answer for Alex, but we were discussing this on a call last week - for all of these error cases it would be somewhat useful to pass the error message back to the client instead of just logging it and returning locally.

adam-cattermole: fwiw we agreed I'd handle the error sending in a follow up

internal/extension/manager.go
adam-cattermole: I think this doesn't work here because defer won't run as it's in an infinite for loop? would only work if it was wrapped in an anonymous func iiuc

alexsnaps: My IDE seemed to have disagreed with you, cause that's what I wrote initially. I know I have to educate myself. But I think it is related to it being in a `for` loop? Is `defer` postponing to lexical scope ending, or function ending? I _think_ it's the former, i.e. pushing it onto the stack, which here wouldn't be what we want, right?

PR #1362, opened 2025-05-06
extend .proto definitions

maksymvavilov:
Define gateway API types as proto messages and expand our own `Policy` type. 
Changes things around and added fields that seemed useful. 
I did not went to the `v0.1`:  we should have a discussion about how we want to version things first imo. 
I also added a script to use `protoc` to generate descriptions (hopefully, it will work for others) 
```
./pkg/extension/grpc/generate_proto.sh v0
```


pkg/cel/ext/kuadrant.go
```diff
@@ -110,3 +111,36 @@ func (l kuadrantLib) ProgramOptions() []cel.ProgramOption {
 func (kuadrantLib) LibraryName() string {
 	return "kuadrant.cel.ext.kuadrant"
 }
+
+func getRegistryWithTypes() *types.Registry {
+	registry, _ := types.NewRegistry(
+		// common.proto
+		&v0.Metadata{},
```
alexsnaps, line 118: Should these all be `v1` _of Gateway API types_?

maksymvavilov: Can you elaborate, please? 
Those meant to be the results of `.proto` message definitions (the ones that implement the `proto.Message` interface) as the registry is used to populate types for our CEL env. 
Also, the `cel.SomethingBinding()` should be enforcing the `ref.Val` type and we use this registry to convert into that :thinking:

alexsnaps: I'm saying, should we move these from:

```golang
	v0 "github.com/kuadrant/kuadrant-operator/pkg/extension/grpc/v0"
```

to `v0` (both the package _and_ its location) to whatever version this represent? I guess we can't use the dot (`.`), so something like if this is version 1.2 of the Gateway API, to `v1_2` or something?

pkg/cel/ext/kuadrant.go
```diff
@@ -110,3 +111,36 @@ func (l kuadrantLib) ProgramOptions() []cel.ProgramOption {
...
+		// policy.proto
+		&v0.Policy{},
+		&v0.PolicyStatus{},
+
+		// kuadrant.proto
+		&v0.PingRequest{},
+		&v0.PongResponse{},
+		&v0.ResolveRequest{},
+		&v0.ResolveResponse{},
+		&v0.Event{},
```
alexsnaps, line 143: There tho, would remain `v0` for now tho!

alexsnaps: ... but I don't think they should be added to this list tho! Why would they be known types to the CEL interpreter?

maksymvavilov: yep, that makes little sense to have them. I'll get rid of them

pkg/cel/ext/kuadrant.go
maksymvavilov: moved to v1 all the new things. I also copied what we had in `v0/kuadrant.proto` to `v1/kuadrant.proto`. Was thinking to leave it, but decided not to: want to keep all imports in the repository (the `google/...` folders). 
Don't know if it matters, but manually copying the `import` folders should allow anyone to regenerate `.pb` files

PR #1364, opened 2025-05-08
Extensions: more wiring

adam-cattermole:
Builds on #1360 

* Add support to pass the cel.expr.Value back instead of strings from Resolve
* Fix a couple issues in the manager - we may need to extend the timeout even more on the `getWaitWithTimeout`
* Passing errors back in the subscribe stream, for now these are logged "internally" within the client controller
* Updated the example reconciler with a Policy that matches the interface (we map an AuthPolicy to it for now), and resolve the first gateway name

If you follow the guide here [authenticated rl guide](https://docs.kuadrant.io/1.2.x/kuadrant-operator/doc/user-guides/ratelimiting/authenticated-rl-for-app-developers/) you'll see the following in the logs (the gateway is called `external` in that example):

```
WITH_EXTENSIONS=true make local setup
kubectl set env -n kuadrant-system deployments/kuadrant-operator-controller-manager WITH_EXTENSIONS="true"
```

```
{"level":"info","ts":"2025-05-07T16:43:21Z","logger":"example-extension-controller.ExampleExtensionReconciler","msg":"Reconciling ExampleExtension"}
{"level":"info","ts":"2025-05-07T16:43:21Z","logger":"example-extension-controller.ExampleExtensionReconciler","msg":"Resolved","out":"external"}
```

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -24,7 +24,8 @@ comment:
 
 # Find more at https://docs.codecov.com/docs/ignoring-paths
 ignore:
-  - "**/*.deepcopy.go"  # ignore controller-gen generated code
+  - "**/*.deepcopy.go" # ignore controller-gen generated code
+  - "**/*.pb.go"
 
 component_management:
   individual_components:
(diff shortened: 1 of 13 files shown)


didierofrivia (dismissed):
🧑‍🍳

didierofrivia (dismissed):

didierofrivia (approved):
👹

PR #1367, opened 2025-05-21
docs: CoreDNS on OpenShift (AWS)

### LINKED ISSUE #1366: Update Core DNS guide for using it with AWS based cluster
https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/dns/core-dns.md

mikenairn:
closes #1366 

Update coredns user guide docs to include additional steps needed to expose the coredns service when using an OpenShift cluster on AWS.


mikenairn:
We could alternatively provide [different kustomizations](https://github.com/Kuadrant/dns-operator/pull/446) to use during installation, but for now i just updated this guide with the manual steps required to change the default installation.

maleck13 (approved):

Issue #1314, opened 2025-04-22
# Set topologySpreadConstraints (limitador deployment)


Boomatang:
## What
Add the [Pod Topology Spread Constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) to the limitador deployment configuration when the `spec.resilience.authorization` feature is enabled in the kuadrant CR.

## Why
To be more resilient the deployment need to react to network failures. 
The Pod Topology Spread Constraints allows us to do that.

## How
Update the limitador deployment with the following configuration.
```yaml
spec:
  template:
    spec:
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: kubernetes.io/hostname
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              limitador-resource: limitador
        - maxSkew: 1
          topologyKey: kubernetes.io/zone
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              limitador-resource: limitador
```

## Done
- [ ] When the `spec.resilience.authorization` feature is enabled the following configuration should be in the limitador deployment.
```yaml
kind: Deployment
metadata:
  name: limitador
spec:
  template:
    spec:
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: kubernetes.io/hostname
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              limitador-resource: limitador
        - maxSkew: 1
          topologyKey: kubernetes.io/zone
          whenUnsatisfiable: ScheduleAnyway
          labelSelector:
            matchLabels:
              limitador-resource: limitador
```

Issue #1313, opened 2025-04-22
# UnSet limitador resourceRequirements.requests


Boomatang:
## What
When the `spec.resilience.rateLimiting` is disabled in the kuadrant CR, the `spec.resourceRequirements` below needs to be removed from the limitador CR.
```yaml
spec:
  resourceRequirements:
    requests:
      cpu: 10m
      memory: 10Mi
```
Once the feature is disabled the user should be allowed to configure the `spec.resourcesRequirements` how they see fit without the kuadrant-operator reverting the changes.

## Why
It is important that the kuadrant-operator undoes any configuration changes it has done to the cluster when the feature is disabled.
When the user modifies the limitador configuration with the feature disabled the kuadrant-operator should not revert the changes. 
This allows for upgrades to not affect the existing users configurations.

## How
When the `spec.resilience.rateLimiting` is disabled in the kuadrant CR, remove `spec.resourceRequirements.requests.cpu` and `spec.resourceRequirements.requests.cpu` from the limitador CR.
The limitador CR can be found in the topology.

How to ensure the user can modify the `spec.resourceRequirements` when the feature is disabled needs to be explored.

## Done
When the `spec.resilience.rateLimiting` is disabled in the kuadrant CR, the `spec.resourceRequirements.requests` are removed from the limitador CR.
When the `spec.resilience.rateLimiting` is disabled in the kuadrant CR, and the user modifies `spec.resourceRequirements.requests` in the limitador CR, the kuadrant-operator does not revert the changes.

Issue #1312, opened 2025-04-22
# Set limitador resourceRequirements.requests


Boomatang:
## What
When the `spec.resilience.rateLimiting` is enabled in the kuadrant CR, the `spec.resourceRequirements` is configured as below.
```yaml
spec:
  resourceRequirements:
    requests:
      cpu: 10m
      memory: 10Mi
```

## Why
The `resources.requests` on a pod help the kubernetes seclude pods to different nodes.
The `spec.resourceRequirements` the limitador CR allows the limitador-operator to configure the deployment.

## How
When the `spec.resilience.rateLimiting` is enabled in the kuadrant CR, get the limitador CR from the topology.
If the `spec.resourceRequirements.requests` is not configured with both `cpu` and `memory` configure them, and write the configuration back to the cluster.

If the values are already configured, don't change the user defined configuration.

## Done
- [ ] When the `spec.resilience.rateLimiting` is enabled, and no user configuration exist in the limitador CR, the `spec.resource.Requirements.request.cpu` is set to `10m`
- [ ] When the `spec.resilience.rateLimiting` is enabled, and no user configuration exist in the limitador CR, the `spec.resource.Requirements.request.memory` is set to `10Mi`
- [ ] When the `spec.resilience.rateLimiting` is enabled, and user configuration exists in the limitador CR, the operator does not override any of the existing configuration.

Issue #1311, opened 2025-04-22
# Limitador Pod Disruption Budget Status


Boomatang:
## What
If the user modifies the `spec.pdb.maxUnavailable` in the limitador CR while the `spec.resilience.rateLimiting` is enabled in the kuadrant CR.
The status of the kuadrant CR should inform the user that the spec is different from the default kuadrant-operator spec.

## Why
When the user is debugging a setup it is important to give the user information on what is different from the default configuration.

## How
Can to the limitador CR will trigger the reconciles.
When the `spec.resilience.rateLimiting` feature is enabled, compare the values of the `spec.pdb.maxUnavailable` to the kuadrant-operator defaults.
Update the status to reflect these changes.

The shape of the status is TBD.

## Done
- [ ] When the `spec.resilience.rateLimiting` is enabled, and the user modifies `spec.pdb.maxUnavailable` the status in the kuadrant CR is updated to reflect the changes.
- [ ] When the `spec.resilience.rateLimiting` is disabled, and the user modifies `spec.pdb.maxUnavailable` there should be no status added to the kuadrant CR.

PR #1368, opened 2025-05-21
Resilient deployment (RateLimiting)

Boomatang:
This cover resilient deployment for limitador.

closes:
- #1306 
- #1307 
- #1308 
- #1309 
- #1310 
- #1311 
- #1312 
- #1313 
- #1314 
- #1315 
- #1326 
- #1318 
- #1319 
- #1320 

For validation this pr implements the [Resilient Deployment of data plane components](https://docs.kuadrant.io/1.1.x/install-olm/#resilient-deployment-of-data-plane-components).

The main goal is to allow the user to have existing configuration, and not have kuadrant remove that configuration on upgrade. The user should be allowed to modify the configuration to suit their needs.


api/v1beta1/kuadrant_types.go
```diff
@@ -111,6 +127,24 @@ func (m *MTLS) IsAuthorinoEnabled() bool {
 	return m.Enable && ptr.Deref(m.Authorino, m.Enable)
 }
 
+// +kubebuilder:validation:XValidation:rule="has(self.rateLimiting) ? (self.rateLimiting == true && has(self.counterStorage)) || (self.rateLimiting == false && has(self.counterStorage)) || (self.rateLimiting == false && !has(self.counterStorage)) : true",message="resilience.counterStorage needs to be explictly configured when using resilience.rateLimiting."
```
maleck13, line 130:
```suggestion
// +kubebuilder:validation:XValidation:rule="has(self.rateLimiting) ? (self.rateLimiting == true && has(self.counterStorage)) || (self.rateLimiting == false && has(self.counterStorage)) || (self.rateLimiting == false && !has(self.counterStorage)) : true",message="resilience.counterStorage needs to be explicitly configured when using resilience.rateLimiting."
```

doc/user-guides/resilience/counterStorage.md
```diff
@@ -0,0 +1,247 @@
...
+spec:
+  resilience:
+    counterStorage:
+        disk:
+          optimize: disk
+```
+
+## Disabling Counter Storage
+
+When the counter storage configuration is removed, the Kuadrant Operator will revert the `Limirtador` CR back to the default of using In-Memory counters.
```
maleck13, line 243:
```suggestion
When the counter storage configuration is removed, the Kuadrant Operator will revert the `Limitador` CR back to the default of using In-Memory counters.
```

internal/controller/history.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"sync"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
+type History struct {
```
maleck13, line 13: Do we still need this?

internal/controller/kuadrant_status_updater.go
```diff
@@ -145,6 +162,117 @@ func mtlsLimitador(kObj *kuadrantv1beta1.Kuadrant, state *sync.Map) *bool {
...
+		create = append(create, cond)
+	} else {
+		remove = append(remove, ResilienceInfoPDBConditionType)
+	}
+
+	// Get the status of the topology spread constraints.
+	deployment := GetDeploymentForParent(topology, kuadrantv1beta1.LimitadorGroupKind)
+	if deployment != nil {
+		message := make([]string, 0)
+		for _, item := range deployment.Spec.Template.Spec.TopologySpreadConstraints {
```
maleck13, line 230: Should we check if the length of topology constraints matches 2?

internal/controller/ratelimit_workflow_helpers.go
```diff
@@ -54,7 +55,23 @@ func GetLimitadorFromTopology(topology *machinery.Topology) *limitadorv1alpha1.L
 	}
 
 	limitador := limitadorObj.(*controller.RuntimeObject).Object.(*limitadorv1alpha1.Limitador)
-	return limitador
+	return limitador.DeepCopy()
+}
+
+func GetMachineryObjectFromTopology(topology *machinery.Topology, groupkind schema.GroupKind) machinery.Object {
```
maleck13, line 61: This seems possibly in the wrong place? Seems a quite useful utility that might be used in many locations?

tests/common/kuadrant/counter_storage_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+		}, testTimeOut)
+	})
+
+	Context("counterStorage is configured", Serial, func() {
+		It("user modifies the storage configuration in the limitador resource", func(ctx SpecContext) {
+			kuadrantKey := client.ObjectKey{Name: "kuadrant-sample", Namespace: testNamespace}
+			limitadorKey := client.ObjectKey{Name: kuadrant.LimitadorName, Namespace: testNamespace}
+
+			By("Apply kurdrant resouce with counterStorage configured")
```
maleck13, line 153:
```suggestion
			By("Apply kuadrant resource with counterStorage configured")
```

tests/common/kuadrant/rate_limiting_test.go
```diff
@@ -0,0 +1,809 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			kObj.Spec.Resilience = &kuadrantv1beta1.Resilience{
+				RateLimiting:   true,
+				CounterStorage: &limitadorv1alpha1.Storage{},
+			}
+			err = k8sClient.Update(ctx, kObj)
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(tests.KuadrantIsReady(testClient(), kuadrantKey)).WithContext(ctx).Should(Succeed())
+
+			By("limitador resource keeps initial configuration")
```
maleck13, line 240: should we also check the status is correct on the kuadrant object?

api/v1beta1/kuadrant_types.go
```diff
@@ -71,6 +72,17 @@ func (k *Kuadrant) IsMTLSAuthorinoEnabled() bool {
 	return k.Spec.MTLS.IsAuthorinoEnabled()
 }
 
+func (k *Kuadrant) BasicResilienceStatus() ResilienceStatus {
+	result := ResilienceStatus{}
+	result.RateLimiting = ptr.To((k.Spec.Resilience != nil && k.Spec.Resilience.RateLimiting))
+	if k.Spec.Resilience != nil && k.Spec.Resilience.CounterStorage != nil {
```
eguzki, line 78: nit:
```go
result.CounterStorage = ptr.To(k.Spec.Resilience != nil && k.Spec.Resilience.CounterStorage != nil)
```

tests/common/kuadrant/rate_limiting_test.go
```diff
@@ -0,0 +1,809 @@
...
+			Eventually(func(g Gomega) {
+				limitadorKey := client.ObjectKey{Name: kuadrant.LimitadorName, Namespace: testNamespace}
+				lObj := &limitadorv1alpha1.Limitador{}
+				err := k8sClient.Get(ctx, limitadorKey, lObj)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(lObj.Spec.ResourceRequirements.Requests.Cpu().Value()).To(Equal(userCpu.Value()))
+				g.Expect(lObj.Spec.ResourceRequirements.Requests.Memory().Value()).To(Equal(userMemory.Value()))
+			},
+			).WithContext(ctx).Should(Succeed())
+			By("")
```
maleck13, line 556: empty By() here. Were you planning on testing the kuadrant status here also?

maleck13 (commented):
Open questions on the history still. Also few minor comments. It is looking pretty good at this stage and once comments resolved and failing checks I will verify and we can look to merge

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,564 @@
...
+
+// INFO: Local Functions
+
+func isExperimentalFeatureEnabled(topology *machinery.Topology) bool {
+	k := GetKuadrantFromTopology(topology)
+	if k == nil {
+		return false
+	}
+
+	if val, exists := k.GetAnnotations()[ResilienceFeatureAnnotation]; exists {
```
eguzki, line 463: So, to enable this I have to define an annotation like this:

```yaml
annotations:
  kuadrant.io/experimental-dont-use-resilient-data-plane: true
```

Should the annotation name be changed?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,564 @@
...
+	state.Store(ExperimentalResilienceFeature, isExperimentalFeatureEnabled(topology))
+
+	return nil
+}
+
+// INFO: Task Section
+
+func NewResilienceDeploymentTasks(client *dynamic.DynamicClient) []controller.ReconcileFunc {
+	return []controller.ReconcileFunc{
+		NewResilienceAuthorizationReconciler().Subscription().Reconcile,
```
eguzki, line 80: is the authorization resilient configuration in the scope of this PR?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,564 @@
...
+	return controller.Subscription{
+		ReconcileFunc: r.reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: &kuadrantv1beta1.KuadrantGroupKind},
+			{Kind: &kuadrantv1beta1.LimitadorGroupKind},
+		},
+	}
+}
+
+func (r *ResilienceRateLimitingReconciler) reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
```
eguzki, line 133: it would be nice to keep this method smaller. Too much done within it. This is not a request, just a "good to have" advice.

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,564 @@
...
+	}
+	lObj := GetLimitadorFromTopology(topology)
+	if lObj == nil {
+		logger.V(level).Info("limitador resource has not being created yet.")
+		return nil
+	}
+
+	wasConfigured := false
+	if history.kuadrant != nil {
+		wasConfigured = history.kuadrant.Spec.Resilience.IsRateLimitingConfigured()
```
eguzki, line 159: do we really need to know if in the past it was configured?

Usually controllers take actions based on the current state of the cluster. Having to code different paths from the previous steps to the current state might be unbearable and hard to maintain.

eguzki (commented):
left some comments

internal/controller/kuadrant_status_updater.go
Boomatang: Are the only allowed values for the topology spread constraints "kubernetes.io/hostname" and "kubernetes.io/zone"? If custom topology keys are allowed, we should not assume the list will only have a length of 2.

doc/user-guides/resilience/counterStorage.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+When configured, Kuadrant copies the `counterStorage` configuration to the Limitador CR `spec.storage`, and maintains the configuration.
+Errors raised by the configuration will be reflected in the Kuadrant status block.
+
+## Counter Storage Configuration
+<!-- Reference: https://github.com/Kuadrant/limitador-operator/blob/main/doc/storage.md -->
+
+Rate limit counters are stored in a backend storage. 
+This is in contrast to the storage of the limits themselves, which are always stored in ephemeral memory.
```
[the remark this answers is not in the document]
Boomatang: Far point, I didn't think there was need for any rewording as it is part of the limitador-operator docs. Does this mean that the state is also wrong in the original documentation.

internal/controller/resilient_deployment_workflow.go
Boomatang: I am not sure if it needs to be changed. As soon as the all the work in the area is done the need for the annotation would be removed. This is most there so we can do iterative work in to main without affecting existing user. This is an undocumented feature.

internal/controller/resilient_deployment_workflow.go
Boomatang: I hear you. I guess this would be in the "Make it work" stage of "Make it work, Make it right, Make it right".

tests/common/kuadrant/rate_limiting_test.go
```diff
@@ -0,0 +1,813 @@
...
+				limitadorKey := client.ObjectKey{Name: kuadrant.LimitadorName, Namespace: testNamespace}
+				lObj := &limitadorv1alpha1.Limitador{}
+				err := k8sClient.Get(ctx, limitadorKey, lObj)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(*lObj.Spec.Replicas).To(Equal(1))
+			},
+			).WithContext(ctx).Should(Succeed())
+
+		}, testTimeOut)
+	})
```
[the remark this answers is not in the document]
Boomatang: Root cause of this issue was found, there was a logic bug in the recenciler checks.

api/v1beta1/kuadrant_types.go
Boomatang: reworked

internal/controller/state_of_the_world.go
```diff
@@ -649,6 +650,7 @@ func initWorkflow(client *dynamic.DynamicClient) *controller.Workflow {
 		Precondition: NewEventLogger().Log,
 		Tasks: []controller.ReconcileFunc{
 			NewTopologyReconciler(client, operatorNamespace).Reconcile,
+			InitializeHistoryFunc,
```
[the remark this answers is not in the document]
Boomatang: Reworked

internal/controller/history.go
Boomatang: Reworked

internal/controller/kuadrant_status_updater.go
```diff
@@ -145,6 +244,117 @@ func mtlsLimitador(kObj *kuadrantv1beta1.Kuadrant, state *sync.Map) *bool {
...
+		return nil, remove
+	}
+
+	lObj := GetLimitadorFromTopology(topology)
+	if lObj == nil {
+		remove = append(remove, ResilienceInfoRRConditionType, ResilienceWarningRRConditionType, ResilienceInfoPDBConditionType, ResilienceInfoTSCConditionType)
+		return nil, remove
+	}
+
+	if lObj.Spec.Replicas != nil && *lObj.Spec.Replicas < LimitadorReplicas {
```
maleck13, line 264: have you considered adding some functions here such as something like `userModifiedReplicas(dep *Deployment, expected int)(bool, string)` where the bool represents if it user changed and the string the message to use?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,625 @@
...
+		},
+	}
+}
+
+func (r *ResilienceDeploymentPrecondition) run(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
+	logger := controller.LoggerFromContext(ctx).WithName("ResilienceDeploymentPrecondition")
+	logger.V(level).Info("ResilienceDeployment Precondition", "status", "started")
+	defer logger.V(level).Info("ResilienceDeployment Precondition", "status", "completed")
+
+	state.Store(ExperimentalResilienceFeature, isExperimentalFeatureEnabled(topology))
```
maleck13, line 72: would it be better to register an errorHandler here and just return a known error if the precondition fails?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,625 @@
...
+func (r *ResilienceAuthorizationReconciler) Subscription() controller.Subscription {
+	return controller.Subscription{
+		ReconcileFunc: r.reconcile,
+		Events: []controller.ResourceEventMatcher{
+			{Kind: &kuadrantv1beta1.KuadrantGroupKind},
+		},
+	}
+}
+
+func (r *ResilienceAuthorizationReconciler) reconcile(ctx context.Context, _ []controller.ResourceEvent, _ *machinery.Topology, _ error, state *sync.Map) error {
```
maleck13, line 103: one thought I had here was the resourceEvents contain the old and new object so perhaps these could be used to compare what has changed. Perhaps not needed 🤷

internal/controller/kuadrant_status_updater.go
```diff
@@ -111,19 +117,112 @@ func (r *KuadrantStatusUpdater) updateKuadrantStatus(ctx context.Context, kObj *
...
+	return status
+}
+
+func userDefinedRateLimitingResiliences(lObj *limitadorv1alpha1.Limitador, topology *machinery.Topology) bool {
+	if lObj == nil {
+		return false
+	}
+
+	requests := false
+	if lObj.Spec.ResourceRequirements != nil && lObj.Spec.ResourceRequirements.Requests != nil {
```
maleck13, line 170: I am a bit confused by this check. Do we not also need to check the kuadrant object? Seems these would be set if the kuadrant resilience had been defined? Seems like we are saying it user defined if set at all on the limitador obj?

internal/controller/kuadrant_status_updater.go
```diff
@@ -111,19 +117,112 @@ func (r *KuadrantStatusUpdater) updateKuadrantStatus(ctx context.Context, kObj *
...
+	resilienceError := false
+	stateResilienceError, ok := state.Load("resilienceError")
+	if ok {
+		resilienceError = stateResilienceError.(bool)
+	}
+
+	if resilienceError {
+		return kObj.Status.Resilience
+	}
+
```
maleck13, line 135: This block reads oddly. If there is a resilience error we just return whatever is in the existing kuadrant status and don't report it? I feel like I must be making a mistake here? Also this piece of code could be simplified right?

```
if ok && stateResilienceError.(bool){
// do something with status
  return kObj.Status.Resilience
}
```

maleck13: Yes so testing this if an error occurs the status is not changed at all. Which means it is not clear to the end user

````
pec:
  resilience:
    counterStorage:
      redis:
        configSecretRef:
          name: redis-credentials
    rateLimiting: true
status:
  conditions:
  - lastTransitionTime: "2025-06-13T10:21:25Z"
    message: Kuadrant is ready
    reason: Ready
    status: "True"
    type: Ready
  mtlsAuthorino: false
  mtlsLimitador: false
  observedGeneration: 1
  resilience:
    counterStorage: kuadrant defined
    rateLimiting: kuadrant defined
```

I think we should be changing the ready condition here. Kuadrant is not ready in the way the user has asked for it to be configured 

```
  - lastTransitionTime: "2025-06-13T10:21:25Z"
    message: Failed to configure resilience for ratelimitng 
    reason: NotReady
    status: "False"
    type: Ready

File: internal/controller/kuadrant_status_updater.go (hunk 20 above)
Comment (maleck13) [reply internal/controller/kuadrant_status_updater.go:170]: So checking this locally. Changing the values manually. I never get to see `rateLimiting: user defined`

```yaml
conditions:
  - lastTransitionTime: "2025-06-16T09:45:41Z"
    message: Deployment has unavailable replicas
    reason: LimitadorNotReady
    status: "False"
    type: Ready
  - lastTransitionTime: "2025-06-16T09:40:56Z"
    message: Number of Limitador replicas (3) greater than minimum default
    reason: UserModifiedLimitadorReplicas
    status: Unknown
    type: Resilience.Info.RateLimiting.Replicas
  - lastTransitionTime: "2025-06-16T09:44:04Z"
    message: Limitador recource Pod Disruption Budget differs from default configuration
    reason: UserModifiedLimitadorPodDisruptionBudget
    status: Unknown
    type: Resilience.Info.RateLimiting.PBD
  - lastTransitionTime: "2025-06-16T09:45:41Z"
    message: Limitador depoloyment TopologySpreadConstraints for key "kubernetes.io/zone"
      is user modified.
    reason: UserModifiedLimitadorTopologySpreadConstraints
    status: Unknown
    type: Resilience.Info.RateLimiting.TopologySpreadConstraints
  mtlsAuthorino: false
  mtlsLimitador: false
  observedGeneration: 10
  resilience:
    counterStorage: kuadrant defined
    rateLimiting: kuadrant defined  # remains as kuadrant defined??
```

internal/controller/ratelimit_workflow_helpers.go
Boomatang: moved

api/v1beta1/kuadrant_types.go
```diff
@@ -153,6 +188,12 @@ func (r *KuadrantStatus) Equals(other *KuadrantStatus, logger logr.Logger) bool
 		return false
 	}
 
+	if !reflect.DeepEqual(r.Resilience, other.Resilience) {
+		diff := cmp.Diff(r.Resilience, other.Resilience)
+		logger.V(0).Info("Resilience not equal", "difference", diff)
```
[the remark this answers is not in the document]
Boomatang: It should have being the same, missed it in my review.

internal/controller/kuadrant_status_updater.go
```diff
@@ -145,6 +244,117 @@ func mtlsLimitador(kObj *kuadrantv1beta1.Kuadrant, state *sync.Map) *bool {
...
+			Message: fmt.Sprintf("Number of Limitador replicas (%v) greater than minimum default", *lObj.Spec.Replicas),
+			Reason:  "UserModifiedLimitadorReplicas",
+			Status:  metav1.ConditionUnknown,
+		}
+		create = append(create, cond)
+	} else {
+		remove = append(remove, ResilienceInfoRRConditionType)
+	}
+
+	if lObj.Spec.PodDisruptionBudget != nil && lObj.Spec.PodDisruptionBudget.MaxUnavailable != nil && lObj.Spec.PodDisruptionBudget.MaxUnavailable.IntVal != LimitadorPDB {
```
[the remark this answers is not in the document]
Boomatang: This one does make more sense than the replicas above because of the `else if`. I will look at a bit of a tidy on it.  The issue with the object in question is the limitador CR and not a deployment CR still applies.

internal/controller/resilient_deployment_workflow.go
Boomatang: Possibly. My mental issue with doing that is saying if the experimental feature is disabled, that is an error. The question that runs around my head is how long do you expect this feature to be experimental? Is configuring a errorHandler the best use of our time?

internal/controller/resilient_deployment_workflow.go
Boomatang: I think it will give you the old and new version of the object. This however can not be used as the old version that you will be give is the old version on the cluster, and the last version you successfully reconciled. If you could compare the new version with the past reconciled version, then yes we could have used that.

internal/controller/kuadrant_status_updater.go
Boomatang: The reading of the block being odd is fair, will address. It is correct the condition type `Ready` should be updated while the reconcile is looping. Will look to address.

internal/controller/kuadrant_status_updater.go
```diff
@@ -23,7 +25,11 @@ import (
 )
 
 const (
-	ReadyConditionType string = "Ready"
+	ReadyConditionType               string = "Ready"
+	ResilienceInfoRRConditionType    string = "Resilience.Info.RateLimiting.Replicas"
```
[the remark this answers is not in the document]
maleck13: The value

internal/controller/resilient_deployment_workflow.go
maleck13: Ok so what exactly does it give you then?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,627 @@
...
+		if err != nil {
+			logger.Error(err, "failed to parse resurce memory string", "status", "error")
+		}
+
+		if lObj.Spec.ResourceRequirements == nil {
+			lObj.Spec.ResourceRequirements = &corev1.ResourceRequirements{}
+		}
+
+		if lObj.Spec.ResourceRequirements.Requests.Cpu().Value() == 0 {
+			if lObj.Spec.ResourceRequirements.Requests == nil {
```
maleck13, line 232: This doesn't look right. We do a nil check after using the object in the previous check?

maleck13:
**Verification**

I did the following things post code review.

✅  Create Kuadrant CR with no resilience ensure deploys and becomes ready. I did wonder if it may be better to just leave resilience status empty in this situation.

```yaml
status:
  conditions:
  - lastTransitionTime: "2025-06-20T10:57:15Z"
    message: Kuadrant is ready
    reason: Ready
    status: "True"
    type: Ready
  mtlsAuthorino: false
  mtlsLimitador: false
  observedGeneration: 1
  resilience:
    counterStorage: undefined
    rateLimiting: undefined

✅  Update the replicas and resources on limitador directly and see that the kuadrant resource reports correctly

```yaml
status:
  conditions:
  - lastTransitionTime: "2025-06-20T11:09:33Z"
    message: Deployment does not have minimum availability.
    reason: LimitadorNotReady
    status: "False"
    type: Ready
  mtlsAuthorino: false
  mtlsLimitador: false
  observedGeneration: 1
  resilience:
    counterStorage: undefined
    rateLimiting: user defined

✅ Define resilience true and counter storage via kuadrant see that limitador is updated to our defaults

```yaml
 spec:
    resilience:
      counterStorage:
        redis:
          configSecretRef:
            name: redis-credentials
      rateLimiting: true
  status:
    conditions:
    - lastTransitionTime: "2025-06-20T11:14:49Z"
      message: Kuadrant is ready
      reason: Ready
      status: "True"
      type: Ready
    mtlsAuthorino: false
    mtlsLimitador: false
    observedGeneration: 2
    resilience:
      counterStorage: kuadrant defined
      rateLimiting: kuadrant defined
---
spec:
    pdb:
      maxUnavailable: 1
    replicas: 2
    resourceRequirements:
      requests:
        cpu: 10m
        memory: 10Mi
    storage:
      redis:
        configSecretRef:
          name: redis-credentials
```
❌  change limitador directly and see we don't overwrite it again and we report status correctly

I expected a status warning here that the resources were not as expected

```yaml
spec:
    pdb:
      maxUnavailable: 1
    replicas: 2
    resourceRequirements:
      requests:
        cpu: 5m
        memory: 10Mi

---
status:
    conditions:
    - lastTransitionTime: "2025-06-20T11:30:52Z"
      message: Kuadrant is ready
      reason: Ready
      status: "True"
      type: Ready
    mtlsAuthorino: false
    mtlsLimitador: false
    observedGeneration: 2
    resilience:
      counterStorage: kuadrant defined
      rateLimiting: kuadrant defined
```

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,623 @@
...
+		}
+
+		if lObj.Spec.ResourceRequirements.Requests == nil {
+			lObj.Spec.ResourceRequirements.Requests = make(corev1.ResourceList)
+		}
+
+		if lObj.Spec.ResourceRequirements.Requests.Cpu().Value() == 0 {
+			cpu, err := resource.ParseQuantity(Resource10m)
+			if err != nil {
+				logger.Error(err, "failed to parse resurce cpu string", "status", "error")
```
maleck13, line 230: should we return the error here rather than attempt to update ? I see it is highly unlikely given we are parsing a constant but it seems odd to just log and attempt to update anyway?

internal/controller/resilient_deployment_workflow.go
```diff
@@ -0,0 +1,623 @@
...
+			}
+			lObj.Spec.ResourceRequirements.Requests[corev1.ResourceCPU] = cpu
+			update = true
+		}
+
+		if lObj.Spec.ResourceRequirements.Requests.Memory().Value() == 0 {
+			memory, err := resource.ParseQuantity(Resource10Mi)
+			if err != nil {
```
maleck13, line 238: same as above

Boomatang:
Wont be done

PR #1381, opened 2025-06-09
Removing references to gateway service monitors which we no longer create

Patryk-Stefanski:


david-martin (approved):
Thanks for fixing this up @Patryk-Stefanski

PR #1382, opened 2025-06-09
FIX: Return error bug

Boomatang:
The return was assuming the list should never be empty, but this is not
the case the list can be empty when the cluster is being configured.


internal/controller/authorino_istio_integration_reconciler.go
```diff
@@ -72,7 +72,8 @@ func (a *AuthorinoIstioIntegrationReconciler) Run(baseCtx context.Context, _ []c
 
 	effectiveAuthPolicies, ok := state.Load(StateEffectiveAuthPolicies)
 	if !ok {
-		return ErrMissingStateEffectiveAuthPolicies
+		logger.V(0).Info("reconciling authorino integration in istio", "status", "exiting", "error", ErrMissingStateEffectiveAuthPolicies)
```
eguzki, line 75: nit: it's kind of contradictory to log (in debug mode) some error and the code does not return an error. 

If it's not an error, we could consider setting `effectiveAuthPolicies` to an empty list and let the reconcile do it's job with the input of empty effective policy list. 

BTW, this `AuthorinoIstioIntegrationReconciler` reconciler is scheduled to run after the computation of the effective policies, so it should be totally unexpected to fail loading them from the _state_. Am I right? Maybe the issue is in the effective policy computation reconciler that should _always_ set the state, at least with an empty list.

eguzki: > I have no problem switching the log to a debug.
My bad, when I see logger.V(something) I assume its a log level other than info. As `logger.V(0).Info()` is the same as `logger.Info()`

 > I am assuming that if there is no effective policies we don't need to run the reconcile.

It has to. The reconcile does it's job based on the current status. If the job is aborted, some work would be undone. For example: if the cluster has some policies and then you remove them, then some authorino deployment configuration needs to be changed. Empty list has a meaning and implication. 

> Looking at the subscriptions for this AuthorinoIstioIntegrationReconciler and EffectiveAuthPolicyReconciler this is not guarantee that the effective policies would have being computed. This error I was trigger with modify the authorino CR when there are no authPolicies on the cluster.

This is indeed some bug we have. The contract with the effective policy reconcilers should be then that if no state is set, then empty list.

PR #1387, opened 2025-06-16
Moving action merge functionality from wasm-shim to kuadrant operator

### LINKED ISSUE #1380: Move action merge functionality from wasm-shim to kuadrant-operator
We should move the functionality of merging actions with the same scope and service name from the [wasm shim](https://github.com/Kuadrant/wasm-shim/blob/main/src/ratelimit_action.rs#L414) to the kuadrant operator. We should also check for duplicate data entries ( same key different value) within these actions.

This will allow us to catch potential issues with the action sets config for the wasm shim early on.

Patryk-Stefanski:
Note: This will be merged after https://github.com/Kuadrant/kuadrant-operator/pull/1379 
Update: We will also wait for https://github.com/Kuadrant/kuadrant-operator/pull/1409 to merge before merging this

Closes #1380 

This PR adds the merge functionality to actions, rate limiting actions now use conditional data which allows for  multiple predicates and data to be bundled together under a single action.

1. Setup a kind cluster:
```
make local-env-setup
```

2. Install the new CRD:
```
make install
```

3. Run the operator locally:
```
make run
```

4. Request a kuadrant instance:
```bash
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

5. Setup an llm-d-sim and expose via Gateway API:
``` bash
kubectl apply -f - <<EOF
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: llm-sim
  labels:
    app: llm-sim
spec:
  replicas: 1
  selector:
    matchLabels:
      app: llm-sim
  template:
    metadata:
      labels:
        app: llm-sim
    spec:
      containers:
      - name: simulator
        image: ghcr.io/llm-d/llm-d-inference-sim:v0.1.1
        args:
          - --model
          - meta-llama/Llama-3.1-8B-Instruct
          - --port
          - "8000"
          - --max-loras
          - "2"
          - --lora
          - food-review-1
        ports:
          - containerPort: 8000
---
apiVersion: v1
kind: Service
metadata:
  name: llm-sim
spec:
  selector:
    app: llm-sim
  ports:
    - port: 80
      targetPort: 8000
      protocol: TCP
---
kind: HTTPRoute
metadata:
  name: llm-sim
spec:
  hostnames:
    - llm.example.com
  parentRefs:
      namespace: gateway-system
      sectionName: http
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/"
      backendRefs:
        - name: llm-sim
          port: 80
EOF
```

6. Create RatelimitPolicies
```bash
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1
kind: RateLimitPolicy
metadata:
  name: token-limits
  namespace: gateway-system
spec:
  targetRef:
    kind: Gateway
    name: kuadrant-ingressgateway
  limits:
    free:
      rates:
        - limit: 20000
          window: 24h
      when:
        - predicate: >
            request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "free")
      counters:
    gold:
      rates:
        - limit: 200000
          window: 24h
      when:
        - predicate: >
            request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "gold")
      counters:
```

 7. Verify the WasmPlugin Configuration

```bash
kubectl get wasmplugin -A -o yaml
```

Before 

```
actionSets:
      - actions:
        - data:
          - expression:
              key: limit.free__b4bc821c
              value: "1"
          - expression:
              key: auth.identity.userid
              value: auth.identity.userid
          predicates:
          - |
            request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "free")
          scope: default/llm-sim
          service: ratelimit-service
        - data:
          - expression:
              key: limit.gold__cd359ced
              value: "1"
          - expression:
              key: auth.identity.userid
              value: auth.identity.userid
          predicates:
          - |
            request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "gold")
          scope: default/llm-sim
          service: ratelimit-service
```

After
``` 
 actionSets:
      - actions:
        - conditionalData:
          - data:
            - expression:
                key: limit.free__b4bc821c
                value: "1"
            - expression:
                key: auth.identity.userid
                value: auth.identity.userid
            predicates:
            - |
              request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "free")
          - data:
            - expression:
                key: limit.gold__cd359ced
                value: "1"
            - expression:
                key: auth.identity.userid
                value: auth.identity.userid
            predicates:
            - |
              request.auth.identity.metadata.annotations["kuadrant.io/groups"].split(",").exists(g, g == "gold")
          scope: default/llm-sim
          service: ratelimit-service
```


internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -206,21 +206,29 @@ func (r *EnvoyGatewayExtensionReconciler) buildWasmConfigs(ctx context.Context,
 		// rate limit
 		if effectivePolicy, ok := effectiveRateLimitPoliciesMap[pathID]; ok {
 			rlAction := buildWasmActionsForRateLimit(effectivePolicy, isRateLimitPolicyAcceptedAndNotDeletedFunc(state))
-			if hasAuthAccess(rlAction) {
-				actions = append(actions, rlAction...)
+			mergedRlActions, err := mergeAndVerify(rlAction)
```
eguzki, line 209: have you considered merging after both RLP and TRLP actions have been added? Would it work? The advantage I see is that you could merge actions from both type of policies altogether as well. Why not?

eguzki (commented):
Left some comments

internal/wasm/types.go
```diff
@@ -147,6 +148,14 @@ type Action struct {
 	ServiceName string `json:"service"`
 	Scope       string `json:"scope"`
 
+	Predicates []string `json:"predicates,omitempty"`
+
+	// ConditionalData data contains the predicates and data that will be sent to the service
+	// +optional
+	ConditionalData []ConditionalData `json:"conditionalData,omitempty"`
```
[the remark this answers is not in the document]
Patryk-Stefanski: Yeah I don't see how it would have issues on upgrade, as it's rebuilding the config on reconcile.

internal/wasm/types.go
```diff
@@ -157,41 +166,58 @@ type Action struct {
...
+	}
+
+	if !reflect.DeepEqual(c.Predicates, other.Predicates) {
+		return false
+	}
+
+	for i := range c.Data {
+		if !c.Data[i].EqualTo(other.Data[i]) {
+			return false
+		}
```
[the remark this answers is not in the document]
Boomatang: True, I see your argument. Following up on this then, does the order matter to us? The reason I ask this  during some of the performance test there seemed to be cases were everything was really slow. It turned out the limitador CR was being rewritten to cluster  because the list limits in the spec was coming back in a different order, than what the generation code was comparing with. From what I see if the `.Data` list do not match we rewrite the wasmPlugin back to the cluster. What knock on affect would that have on the system with reloading the wasm configuration, because the data is in a different order. In the case of the limitador CR issue, you would see limitador reload the its configuration every time.

So to recap, I am happy with how the `.EqualTo` function is working. The question now is should we be using that function if we don't care about the order of the data? If the data is in a different order, but the same sets of values, we don't want to be writing to the cluster.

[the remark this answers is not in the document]
Patryk-Stefanski: AFAIK, the order doesn't matter. And as for the rewriting issue, if we see this happen, I think that points to a problem in our code somewhere. We're better off fixing the issue at its source rather than using sorting to mask it.

Patryk-Stefanski:
@eguzki @Boomatang I've addressed the comments, could you have another look

eguzki:
Change the PR name, as this goes to autogenerated release notes

internal/controller/istio_extension_reconciler_test.go
```diff
@@ -0,0 +1,547 @@
+package controllers
```
eguzki, line 1: missing `//go:build unit`

internal/controller/istio_extension_reconciler_test.go
```diff
@@ -0,0 +1,547 @@
...
+											Value: "user_id",
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedLen: 3,
```
eguzki, line 298: the expected length is 4. Only sub-sequent actions can be merged. Due to data dependency rules, you never know if one rate limit action needs a pre-auth action.

eguzki (changes requested):
only subsequent actions can be merged. 

Consider trying to implement like the wasm module does: https://github.com/Kuadrant/wasm-shim/blob/main/src/runtime_action_set.rs#L43-L60

internal/controller/istio_extension_reconciler_test.go
Patryk-Stefanski: gotcha, changed the logic to only merge subsequent actions with same service names

eguzki (approved):
LGTM

The verification steps are outdated and need to be updated. Actually, with the examples of the verification steps, no action merging is actually happening, as actions have different service: `ratelimit-check-service` vs `ratelimit-service`. Use regular  rate limit policy with two limits and you should see merging "in action".

eguzki:
If we merge this PR, it will break integration with the wasm module. It would require changes in the wasm module to adapt. It's ok to break integration at this time, it's for the best. However,  we should wait and hold this PR on until the ongoing work on the wasm shim is being merged. There are 3 stacked PR's ready for review on the wasm module that would need refactor if we merge this today. 

Sounds reasonable?

eguzki (approved):

PR #1411, opened 2025-07-07
Enabling extensions

didierofrivia:
This PR set extensions enabled by default. It also exposes the `WITH_EXTENSIONS` env in the manager deployment yaml for self documenting purposes. At the time being, extensions will be shipped within the Kuadrant Operator image.

### DIFF
--- a/make/integration-tests.mk
+++ b/make/integration-tests.mk
@@ -11,7 +11,7 @@ test-bare-k8s-integration: clean-cov generate fmt vet ginkgo ## Requires only ba
 	mkdir -p $(PROJECT_PATH)/coverage/bare-k8s-integration
 #	Check `ginkgo help run` for command line options. For example to filtering tests.
 # Run in series as limit of 1 Kuadrant CR (procs=1)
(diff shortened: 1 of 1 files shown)


make/integration-tests.mk
```diff
@@ -11,7 +11,7 @@ test-bare-k8s-integration: clean-cov generate fmt vet ginkgo ## Requires only ba
 	mkdir -p $(PROJECT_PATH)/coverage/bare-k8s-integration
 #	Check `ginkgo help run` for command line options. For example to filtering tests.
 # Run in series as limit of 1 Kuadrant CR (procs=1)
-	OPERATOR_NAMESPACE=kuadrant-system $(GINKGO) \
+	OPERATOR_NAMESPACE=kuadrant-system WITH_EXTENSIONS=false $(GINKGO) \
```
didierofrivia, line 14: An alternative would be to build the extensions too, even if we don't have specific tests.. what do you reckon?

didierofrivia: or disabled by default, and in the deployments set the `WITH_EXTENSIONS` to `true`

eguzki (dismissed):

Issue #896, opened 2024-10-01
Remove `topology` configmap on kuadrant operator uninstall


eguzki:
Link the lifecycle of the kuadrant operator with the `topology` configmap. Currently, when the operator starts, it will create and maintain the `topology` configmap in the namespace where the operator is deployed. It is missing the implementation of the removal of the `topology` configmap when the operator is deleted. 

> Note: The ConsolePlugin resource, together with the associated deployment and service, follows the life events of the `topology` configmap. When the topology is created, so it is the ConsolePlugin (on openshift). Whereas when the topology is deleted, so it is the ConsolePlugin and associated resources.  The ConsolePlugin was introduced in https://github.com/Kuadrant/kuadrant-operator/pull/884

maleck13:
Is this essential for v1?

eguzki:
I would say it is not.

eguzki:
`Release` attribute now is unset

david-martin:
I'm wondering when to actually delete the topology configmap.
Specifically, what does it mean 'when the operator is deleted'.
I'm not sure how we'd hook into that part of the lifecycle.
Perhaps we tie it to the created/deletion of the Kuadrant CR?

Thoughts @eguzki

Boomatang:
I don't think you can (or should) tie this to the creation and deletion of the Kuadrant CR. What happens in the console plugin if that topology is not there? If you create a policy when the kuadrant CR is not present, the operator will accepted it, but raise a `MissingResource` error stating that kuadrant is not installed. The resource is in the topology at this stage. 

In a prefect world the console plugin would show any policies that have errors in the graph. I don't think that happens today. 

I think the removal has to be linked into the lifecycle of the operator. Could it be a simple as adding an ownership to the configmap that links back to the kuadrant deployment CR. 

There is a future world that is some way away, but might be worth the consideration. In the future there will be an [Open Architecture](https://github.com/Kuadrant/architecture/pull/115) that could used to remove the need for the configmap all together.

guicassolato:
> I don't think you can (or should) tie this to the creation and deletion of the Kuadrant CR.
+1

One can have multiple resources, including Kuadrant policies, without necessarily having a Kuadrant CR. Even if Kuadrant is not fully functional without it⸺and _especially_ because it's not functional⸺this state should be reflected in the topology.

It's also possible to have multiple Kuadrant CRs in a cluster. Users will also want to see that in the topology graph.

[...]

PR #1412, opened 2025-07-08
Topology Configmap

orlafoley:
See original issue [here](https://github.com/Kuadrant/kuadrant-operator/issues/896).

Added owner reference to link configmap and kuadrant-operator lifecycles.


eguzki:
> > will be linking the lifecycle of the topology to the Kuadrant CR
> 
> @eguzki No, not quite. Linking the topology configmap to the kuadrant-operator Deployment resource.

The comment mentions kuadrant custom resource, not deployment resource.
```
Could it be a simple as adding an ownership to the configmap that links back to the kuadrant deployment CR.
```

Anyway, instead of hardcoding the deployment name, that name might be passed as env var and inferred from the deployment metadata, as the namespace. Just giving ideas worth exploring https://github.com/Kuadrant/kuadrant-operator/blob/main/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml#L565C1-L568C52
```yaml
                - name: OPERATOR_NAMESPACE
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.namespace
```

internal/controller/topology_reconciler.go
```diff
@@ -80,3 +92,26 @@ func (r *TopologyReconciler) Reconcile(ctx context.Context, _ []controller.Resou
...
+		return nil, err
+	}
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+	deploy, err := clientset.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
```
KevFan, line 109: This is fine but I wonder should we add the Kuadrant `Deployment` resource into the topology instead, and try to get this resource via the topology to update to keep the implementation pattern of using the topology consistent? Otherwise this would constantly get deployment via the api server on each reconcile 🤔 

Makes this change at lot of complex though. Do @Boomatang or anyone else have any thoughts on this?

If not, I think we should use `r.Client` instead of creating another k8s client for this 🤔

eguzki: > Otherwise this would constantly get deployment via the api server on each reconcile

Fortunately, thanks to client caching (at least for the controller runtime client), `Get` operations are very cheap. No need to reach API server on each reconcile loop.. 

In any case, I agree that adding the Deployment to the topology is one way to approach this. You could then use the topology to fetch the Deployment and add the OwnerReference. However, adding the Deployment to the topology isn't entirely straightforward—it requires some coding.

An alternative is to pass the already-available dynamic client to the getOwnerRef function.

Another option is to provide the controller-runtime client (from the manager), as shown here: [Kuadrant example](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/controller/observability_reconciler.go#L278), to the topology reconciler. You can then use it to fetch the Deployment and add the OwnerReference if it doesn't already exist.

That said, personally, I find it much easier to add the ConfigMap to the operator bundle than to reconcile OwnerReferences—no coding required. 🙂

PR #1416, opened 2025-07-10
Fixing make catalog target

Patryk-Stefanski:
Verification
1. Try running `make catalog` without changes, notice errors
2. Run with changes and notice it suceeds


Boomatang (dismissed):
Catalog file is being generated

PR #1419, opened 2025-07-14
set default channel value in makefile for make catalog

Patryk-Stefanski:


azgabur (commented):
So the default channel will be now `alpha`
Will the nightly runs https://github.com/Kuadrant/kuadrant-operator/actions/workflows/build-images-nightly.yaml have channel `preview`?
And than I expect released (tagged with `release-` or `vx.y.z`) will have channel `stable` ?

azgabur (approved):
In that case, good :+1:

PR #1434, opened 2025-07-18
Add a UseHTTPRouteHosts to DNSPolicy 

maleck13:
This allows the hosts specified in the HTTPRoutes attached to a listener to be used as the DNSNames rather than the listener hosts. It means that if you have a wildcard listener you can delegate host definition to the attached HTTPRoutes.

Unfortunately gateway API does not indicate what routes are attached to which hosts so we have to calculate this. 

**Verficiation**

- Create a Gateway with a wildcard listener host
- Create a HTTPRoute with a set of different routes some that match the wild card some that don't
- Create a DNSPolicy with `useHTTPRouteHosts` set to false 
- Validate that the DNSRecord created is a wildcard DNS record
- Change the value of useHTTPRouteHosts
- Validate that there are DNSRecords created for each host that matches the wildcard and that the wildcard DNSRecord is removed


maleck13:
related to https://github.com/Kuadrant/architecture/pull/126

internal/controller/effective_dnspolicies_reconciler.go
```diff
@@ -50,6 +54,128 @@ func (r *EffectiveDNSPoliciesReconciler) Subscription() controller.Subscription
...
+				if !isWildCardListener(listener) {
+					//exact match were done
+					if *listener.Hostname == routeHost {
+						logger.V(1).Info("exact listener host match found", "listener:", listener.GetLocator(), "route host", routeHost)
+						listernerID = listener.GetLocator()
+						listenerName = string(listener.Name)
+						break
+					}
+				}
+				if isWildCardListener(listener) {
```
maleck13, line 115: open question here should we remove this and allow *.a.b.com to be specified in a HTTPRoute and created. Seems against the concept

Issue #1449, opened 2025-07-23
Extension Controller should reconcile Kuadrant objects


didierofrivia:
The extension controller should be in charge of reconciling the Kuadrant policies and Gateway API objects that belong to the DAG.

PR #1451, opened 2025-07-23
Extensions Controller reconciling resources

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/1449

This PR hides the `BaseReconciler` implementation behind the `ExtensionsController`. Thus, the extensions will interact with the KuadrantContext instead.
In a future refactor, ideally we will use the Policy Machinery instead


cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -26,14 +26,13 @@ import (
...
 	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
-	"github.com/kuadrant/kuadrant-operator/internal/reconcilers"
 	extcontroller "github.com/kuadrant/kuadrant-operator/pkg/extension/controller"
 	"github.com/kuadrant/kuadrant-operator/pkg/extension/types"
 	"github.com/kuadrant/kuadrant-operator/pkg/extension/utils"
 )
 
 type OIDCPolicyReconciler struct {
-	*reconcilers.BaseReconciler
+	kCtx   types.KuadrantCtx
```
didierofrivia, line 35: I found it convenient to attach it to the reconciler instead of passing it around.

adam-cattermole: I like this, I wonder if there would be a way to enhance this as a default behaviour as a follow up

pkg/extension/controller/controller.go
```diff
@@ -196,6 +198,24 @@ func (ec *ExtensionController) AddDataTo(ctx context.Context, requester exttypes
 	return err
 }
 
+func (ec *ExtensionController) GetClient() client.Client {
+	return ec.manager.GetClient()
+}
+
+func (ec *ExtensionController) GetScheme() *runtime.Scheme {
+	return ec.manager.GetScheme()
+}
```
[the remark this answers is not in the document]
didierofrivia: makes sense, yeah

PR #1454, opened 2025-07-24
fix panic in unit test

maksymvavilov:
Fixes panic in the unit test (is happening on main at the moment)


maksymvavilov:
@alexsnaps [here](https://github.com/Kuadrant/kuadrant-operator/actions/runs/16491771255/job/46627955736) is the run that made me notice it. I've pulled latest master afterwards and had it failing there as well. Haven't dig deep into why that was happening (it should not have been an issue since we call `Start()` which will init `cmd` before we access the `cmd` var 🤔 )

PR #1464, opened 2025-07-28
dns policy controller test - bump timeout for deletion of dns records

Patryk-Stefanski:
The `should remove DNS records on policy deletion` test often flakes, waiting for the deletion of the DNS record. Debugging this a bit, most of the flakes happen due to the deletion taking slightly longer than the 10-second timeout, bumping it to 15 seconds should decrease the flakes.


tests/common/dnspolicy/dnspolicy_controller_test.go
```diff
@@ -671,7 +671,7 @@ var _ = Describe("DNSPolicy controller", func() {
 
 				err := k8sClient.Get(ctx, client.ObjectKeyFromObject(gateway), gateway)
 				g.Expect(err).NotTo(HaveOccurred())
-			}, tests.TimeoutMedium, time.Second).Should(Succeed())
+			}, time.Second*15, time.Second).Should(Succeed())
```
maksymvavilov, line 674: Could we use one of [those](https://github.com/Kuadrant/kuadrant-operator/blob/127b9ef1e1ac8c4359f5b2b3ecac9e28cfdebd42/tests/commons.go#L42-L43)? 
```
const (
	TimeoutMedium        = time.Second * 10
	TimeoutLong          = time.Second * 30
...
)
```

maksymvavilov: Also streange - only medium and long and now we need the... long-medium? Should we just rename medium into "short" and have a new "medium" for 15 sec? 🤔

Patryk-Stefanski: Done

maksymvavilov (approved):
This will fix at least one flaky test

PR #1469, opened 2025-07-29
Refactor the extensions to use base for logger/client/scheme

### LINKED ISSUE #1457: Remove scheme and client from kuadrantCtx
Follow up to this comment here: https://github.com/Kuadrant/kuadrant-operator/pull/1451#discussion_r2230593146

> These are both already on the regular ctx (like the logger) but are now accessible through the kuadrantCtx when in reality they're nothing to do with Kuadrant. Perhaps we should scrap these and just get them from ctx like the logger in a follow up

* Remove these functions from kuadrantCtx
* Store the scheme and client next to the logger in the relevant reconcilers
* Retrieve at the start of the reconcile, similar to the logger

adam-cattermole:
* Removed the Client and Scheme from the KuadrantCtx which aren't related to KuadrantCtx.
* Added new ExtensionsBase that contains the logger, client, and scheme
* `r.Configure(ctx)` sets all these from the context and we can access them directly, updated both OIDC and Plan policy to use them

Resolves #1457

### DIFF
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -8,7 +8,6 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/go-logr/logr"
 	authorinov1beta3 "github.com/kuadrant/authorino/api/v1beta3"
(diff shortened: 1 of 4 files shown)


didierofrivia (approved):
🍪

PR #1482, opened 2025-08-01
OIDCPolicy status refinement

didierofrivia:
TODO

### DIFF
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -194,29 +194,11 @@ func (r *OIDCPolicyReconciler) calculateStatus(pol *kuadrantv1alpha1.OIDCPolicy,
 		Conditions: slices.Clone(pol.Status.Conditions),
 	}
 
-	availableCond := r.readyCondition(specErr)
+	acceptedCond := extcontroller.AcceptedCondition(pol, specErr)
 
-	meta.SetStatusCondition(&newStatus.Conditions, *availableCond)
+	meta.SetStatusCondition(&newStatus.Conditions, *acceptedCond)
 
 	return newStatus
-}
-
-func (r *OIDCPolicyReconciler) readyCondition(specErr error) *metav1.Condition {
-	cond := &metav1.Condition{
-		Type:    kuadrantv1alpha1.StatusConditionReady,
-		Status:  metav1.ConditionTrue,
-		Reason:  "Ready",
-		Message: "OIDCPolicy is ready",
-	}
-
-	if specErr != nil {
-		cond.Status = metav1.ConditionFalse
-		cond.Reason = "ReconciliationError"
-		cond.Message = specErr.Error()
-		return cond
-	}
-
-	return cond
 }
 
 func (r *OIDCPolicyReconciler) reconcileMainAuthPolicy(ctx context.Context, pol *kuadrantv1alpha1.OIDCPolicy, igw *ingressGatewayInfo) error {
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -9,6 +9,8 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/samber/lo"
+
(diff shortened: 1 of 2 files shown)


cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -138,29 +140,31 @@ func (r *OIDCPolicyReconciler) Reconcile(ctx context.Context, request reconcile.
...
-	if err := r.reconcileCallbackAuthPolicy(ctx, pol, igw); err != nil {
+
+	if err = r.reconcileCallbackAuthPolicy(ctx, pol, igw); err != nil {
 		r.Logger.Error(err, "Failed to reconcile callback AuthPolicy")
-		return err
+		return nil, err
 	}
 
-	return nil
+	return mainAuthPol, nil
```
adam-cattermole, line 163: Doesn't the status depend on all of these things being in a good state? why is only the main auth policy checked?

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -117,14 +119,14 @@ func (r *OIDCPolicyReconciler) Reconcile(ctx context.Context, request reconcile.
...
 	r.Logger.V(1).Info("Resolving ingress gateway info", "ingressGatewayData", ingressGatewayData)
 
-	specErr := r.reconcileSpec(ctx, oidcPolicy, &ingressGatewayData)
-
-	statusResult, statusErr := r.reconcileStatus(ctx, oidcPolicy, specErr)
+	mainAuthPol, specErr := r.reconcileSpec(ctx, oidcPolicy, &ingressGatewayData)
 
 	if specErr != nil {
 		return reconcile.Result{}, specErr
 	}
```
adam-cattermole, line 126: if there's an error reconciling spec shouldn't status reflect it?

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
didierofrivia: True, was just lazy condition hehe. Will update

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -138,37 +135,39 @@ func (r *OIDCPolicyReconciler) Reconcile(ctx context.Context, request reconcile.
...
-	if err := r.reconcileCallbackAuthPolicy(ctx, pol, igw); err != nil {
+	callbackPol, err := r.reconcileCallbackAuthPolicy(ctx, pol, igw)
+	if err != nil {
 		r.Logger.Error(err, "Failed to reconcile callback AuthPolicy")
-		return err
+		return nil, err
 	}
 
-	return nil
+	return []*kuadrantv1.AuthPolicy{mainAuthPol, callbackPol}, nil
```
adam-cattermole, line 158: I don't mind this as such but the API of this function does seem weird, returning a list of policies.. would an alternative be to be calculating the status step-by-step as you create the resources and return that instead?

didierofrivia: Yup, the alternative would be calculating and updating one by one, which would mean a couple of added set of queries to the cluster. Would you reckon that to be more suitable ?

didierofrivia: oh, gotcha, you mean just the validation/calculation not updating the resource

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
adam-cattermole: We don't have to do this here, but I wonder if in future instead we should wrap the user defined reconcile function in the extensions framework `controllers.go` to set a default (overrideable) status in this case

adam-cattermole: Ah can be addressed likely in https://github.com/Kuadrant/kuadrant-operator/issues/1484

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
adam-cattermole: Nice, the direction you took looks really good

PR #1512, opened 2025-08-19
Cel Validating auth for Policies

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/1499

This PR introduces the validation of cel expressions in Policies, and the reporting in the Status (RLP at the moment).

**Verification Steps**:

1. Create kind cluster with Kuadrant

2. Apply Kuadrant CR
```sh
kubectl create -n kuadrant-system -f - <<EOF
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```
3. Follow the steps of the ["Authenticated RL for app devs"](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/ratelimiting/authenticated-rl-for-app-developers.md) but don't create the AuthPolicy nor the secrets

4. Check the status of the RLP:
```sh
kubectl get ratelimitpolicy/toystore -n toystore -o yaml | yq .status
```

It should show:

```yaml
conditions:
  - lastTransitionTime: "2025-08-21T14:37:02Z"
    message: RateLimitPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2025-08-21T14:37:02Z"
    message: |-
      validation issues: [ERROR: <input>:1:1: undeclared reference to 'auth' (in container '')
       | auth.identity.userid == 'alice'
       | ^]
    reason: InvalidCelExpression
    status: "False"
    type: Enforced
```
5. Apply the guide example `AuthPolicy`

6. Check the status of the RLP again:
```sh
kubectl get ratelimitpolicy/toystore -n toystore -o yaml | yq .status
```

and this time would look similar to:

```yaml
conditions:
  - lastTransitionTime: "2025-08-21T14:37:02Z"
    message: RateLimitPolicy has been accepted
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2025-08-21T16:03:13Z"
    message: RateLimitPolicy has been successfully enforced
    reason: Enforced
    status: "True"
    type: Enforced
observedGeneration: 1
```

Notes:
- It's prepared to validate other policies, even if the RLP Status is been updated
- The root validator should have the correct cel types, future work.

### DIFF
--- a/internal/cel/kuadrant_validator.go
+++ b/internal/cel/kuadrant_validator.go
@@ -1,0 +1,58 @@
+package cel
+
+import (
+	"github.com/google/cel-go/cel"
+
+	"github.com/kuadrant/kuadrant-operator/internal/wasm"
+)
+
+const (
+	AuthPolicyKind           = "AuthPolicy"
+	RateLimitPolicyKind      = "RateLimitPolicy"
+	TokenRateLimitPolicyKind = "TokenRateLimitPolicy"
+
+	AuthPolicyName = "auth"
+	RateLimitName  = "ratelimit"
+)
+
+func NewRootValidatorBuilder() *ValidatorBuilder {
+	builder := NewValidatorBuilder()
+	// TODO: correct cel types
+	builder.AddBinding("request", cel.AnyType)
+	builder.AddBinding("source", cel.AnyType)
+	builder.AddBinding("destination", cel.AnyType)
+	builder.AddBinding("connection", cel.AnyType)
+	return builder
+}
+
+func ValidateWasmAction(action wasm.Action, validator *Validator) error {
+	pol := policyKindFromWasmServiceName(action.ServiceName)
+	for _, predicate := range action.Predicates {
+		if _, err := validator.Validate(pol, predicate); err != nil {
+			return err
+		}
+	}
+	for _, conditionalData := range action.ConditionalData {
+		for _, predicate := range conditionalData.Predicates {
+			if _, err := validator.Validate(pol, predicate); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func policyKindFromWasmServiceName(serviceName string) string {
+	switch serviceName {
+	case wasm.AuthServiceName:
+		return AuthPolicyKind
+	case wasm.RateLimitServiceName:
+		return RateLimitPolicyKind
+	case wasm.RateLimitCheckServiceName:
+		return TokenRateLimitPolicyKind
+	case wasm.RateLimitReportServiceName:
+		return TokenRateLimitPolicyKind
+	default:
+		return RateLimitPolicyKind
+	}
+}
--- a/internal/cel/kuadrant_validator.go
+++ b/internal/cel/kuadrant_validator.go
@@ -15,6 +15,57 @@ const (
 	RateLimitName  = "ratelimit"
 )
 
+var StateCELValidationErrors = "CELValidationErrors"
+
+type Issue struct {
+	policyKind string
+	pathID     string
+	err        error
+}
+
+func NewIssue(action wasm.Action, pathID string, err error) *Issue {
+	return &Issue{
+		policyKind: policyKindFromWasmServiceName(action.ServiceName),
+		pathID:     pathID,
+		err:        err,
+	}
+}
+
+func (i *Issue) GetError() error {
+	return i.err
+}
+
+type IssueCollection struct {
+	issues map[string]map[string][]*Issue
+}
+
+func NewIssueCollection() *IssueCollection {
+	return &IssueCollection{
+		issues: make(map[string]map[string][]*Issue),
+	}
+}
+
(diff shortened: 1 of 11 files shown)


KevFan (approved):
Left a minor nit and question, that I believe can be addressed later if we wanted.

Otherwise, I've verified steps working as expected and changes looks good. Approving if we want to merge sooner rather than later 👍

tests/istio/extension_reconciler_test.go
```diff
@@ -316,6 +317,56 @@ var _ = Describe("Rate Limiting WasmPlugin controller", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Eventually(tests.RouteIsAccepted(ctx, testClient(), client.ObjectKeyFromObject(httpRoute))).WithContext(ctx).Should(BeTrue())
 
+			//create authpolicy
```
KevFan, line 320: I wonder if a more rounded test to also test the unhappy path also would be to move the creation of this authpolicy to **after** the rlp created at https://github.com/Kuadrant/kuadrant-operator/blob/d43c0910d9a6a6974ecba0d5f6177aa7c6a5b3b3/tests/istio/extension_reconciler_test.go#L414

Before creating the authpolicy, we can test that the RLP is not enforced with a CEL validation error, and that no `WasmPlugin` CR/config was created 🤔 

Once the authpolicy is created, everything will be in place and RLP will be enforced and WasmPlugin config is created.

This would essentially test the verification steps described in the PR 🤔

internal/cel/kuadrant_validator.go
```diff
@@ -0,0 +1,109 @@
...
+		err:        err,
+	}
+}
+
+func (i *Issue) GetError() error {
+	return i.err
+}
+
+type IssueCollection struct {
+	issues map[string]map[string][]*Issue
```
adam-cattermole, line 39: I'm not strongly opinionated here but I wonder if the clarity in just having this as a `[]*Issue` where your `GetByPolicyKind` does a `lo.Filter` on `issue.policyKind` instead of this several layer nested structure could be more readable, wdyt?

alexsnaps:
Probably worth addressing these. With @didierofrivia away, can either of you @KevFan or @adam-cattermole do those, then the other can review and merge? Otherwise let's wait on @didierofrivia 's return next week.

adam-cattermole:
I'm happy to take these changes on, will put in a PR to address

internal/cel/kuadrant_validator.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"github.com/google/cel-go/cel"
+
+	"github.com/kuadrant/kuadrant-operator/internal/wasm"
+)
+
+const (
+	AuthPolicyKind           = "AuthPolicy"
+	RateLimitPolicyKind      = "RateLimitPolicy"
+	TokenRateLimitPolicyKind = "TokenRateLimitPolicy"
```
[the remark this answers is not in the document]
KevFan: Ah I see, in that case, I'm happy to leave it as it 👍

internal/controller/istio_extension_reconciler.go
```diff
@@ -235,6 +237,10 @@ func (r *IstioExtensionReconciler) buildWasmConfigs(ctx context.Context, state *
 	paths := lo.UniqBy(allPaths, func(e lo.Entry[string, []machinery.Targetable]) string { return e.Key })
 
 	wasmActionSets := kuadrantgatewayapi.GrouppedHTTPRouteMatchConfigs{}
+	validatorBuilder := celvalidator.NewRootValidatorBuilder()
+
+	// clear the stored wasm action errors state
+	state.Delete(celvalidator.StateCELValidationErrors)
```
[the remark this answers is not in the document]
adam-cattermole: I _think_ you're right, the state is [initialised empty](https://github.com/Kuadrant/policy-machinery/blob/main/controller/controller.go#L268) each time the reconcile is propagated

internal/controller/istio_extension_reconciler.go
```diff
@@ -285,7 +294,31 @@ func (r *IstioExtensionReconciler) buildWasmConfigs(ctx context.Context, state *
...
+			if err := celvalidator.ValidateWasmAction(action, validator); err != nil {
+				logger.V(1).Info("WASM action is invalid", "action", action, "path", pathID, "error", err)
+				celValidationIssues.Add(celvalidator.NewIssue(action, pathID, err))
+			} else {
+				validatedActions = append(validatedActions, action)
+			}
+		}
+
+		if !celValidationIssues.IsEmpty() {
+			state.Store(celvalidator.StateCELValidationErrors, celValidationIssues)
```
adam-cattermole, line 314: I just noticed this line here (and the initialisation of the `IssueCollection` happen within the loop over the paths, effectively overwriting the issues with those from the last path only, I'll shift these bits outside the loop

PR #1526, opened 2025-08-26
Fix for the flakey OOPExtensionForwardsLog test

adam-cattermole:
Seen this test failing a good few times in CI, trying to fix it here:

* The message array was written to by multiple goroutines at the same time
* There was an infinite for loop waiting for the process to exit which has been replaced by a channel that indicates when it's done
* I have set the socket to an actual file on disk to allow concurrent test runs; the `/bin/ps` command still errors the same as before, but the server can be started listening to a valid socket

```sh
cd internal/extension && go test -run TestOOPExtensionForwardsLog -v -race -count=50
```

Couple captured flakes here:
- [run 1](https://github.com/Kuadrant/kuadrant-operator/actions/runs/17237355462/job/48905265417)
- [run 2](https://github.com/Kuadrant/kuadrant-operator/actions/runs/17237355462/job/48911044856)


internal/extension/oop_test.go
```diff
@@ -56,25 +57,62 @@ func TestOOPExtensionManagesExternalProcess(t *testing.T) {
...
-	writer := &writerMock{}
+	writer := newWriterMock()
 
 	logger := funcr.New(func(_, args string) {
 		writer.Write([]byte(args))
 	}, funcr.Options{})
 
+	socketPath := fmt.Sprintf("/tmp/kuadrant-test-oop-%d.sock", os.Getpid())
+	os.Remove(socketPath)
+	defer os.Remove(socketPath)
```
KevFan, line 110: Is this intentional to remove and also defer remove? 🤔

adam-cattermole: Oops yeah I think I was doing this because it hadn't been cleaned up from a previous experiment but the defer should do

KevFan (dismissed):
Looks good to me 👍

alexsnaps (approved):

didierofrivia (approved):
🙏🏼

PR #1572, opened 2025-09-17
feat: propagate authorino jwks uri changes

KevFan:


cmd/extensions/oidc-policy/api/v1alpha1/oidcpolicy_types.go
```diff
@@ -72,10 +72,21 @@ type Auth struct {
 }
 
 // Provider defines the settings related to the Identity Provider (IDP)
+//
+//	+kubebuilder:validation:XValidation:rule="!(has(self.jwksURL) && self.jwksURL != '' && has(self.issuerURL) && self.issuerURL != '')",message="Use one of: jwksURL, issuerURL"
+//	+kubebuilder:validation:XValidation:rule="!(has(self.jwksURL) && self.jwksURL != '') || (has(self.authorizationEndpoint) && self.authorizationEndpoint != '' && has(self.tokenEndpoint) && self.tokenEndpoint != '')",message="When jwksURL is set, authorizationEndpoint and tokenEndpoint must also be provided"
```
KevFan, line 77: Assuming when jwksURL is set, authorizationEndpoint and tokenEndpoint must also be provided. What about redirectURI, is the default enough?

didierofrivia: I reckon that the default redirectURI would be enough if it's the same one defined in the IDP config (i.e: [https/http]://[hostname]/callback/auth)

didierofrivia (approved):
Checked with my own Gitlab app since the Ingress IP of our local gateways was different causing the issue with the wrong redirect URI. Also, patching the deployment with the env `{"name": "AUTH_SERVICE_TIMEOUT", "value": "5s"}` in order to give the wasm shim more time to authorino to exchange the token. It works like a charm! 🥇

PR #1603, opened 2025-10-02
Extensions enabled by default (again)

adam-cattermole:
There's several changes here to be able to:

- Build the operator image conditionally including the extensions by the flag
- Build the manifests of the extensions as well as the main operator when the flag is set
- Delete and rebuild the bundle/manifests, conditionally including the manifests for the extensions based on the flag
- And added a kill switch env var, where at runtime we can disable the extensions manager by setting `WITH_EXTENSIONS=false`


config/rbac/kustomization.yaml
```diff
@@ -1,21 +1,21 @@
...
+  - role_binding.yaml
+  - leader_election_role.yaml
+  - leader_election_role_binding.yaml
+  # Comment the following 4 lines if you want to disable
+  # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
+  # which protects your /metrics endpoint.
+  # - auth_proxy_service.yaml
+  # - auth_proxy_role.yaml
+  # - auth_proxy_role_binding.yaml
+  - auth_proxy_client_clusterrole.yaml
```
adam-cattermole, line 18: I was unsure about this - on main we currently have this role included in the bundle and have done for a long time but perhaps this is a remnant from the past? https://github.com/Kuadrant/kuadrant-operator/blob/main/bundle/manifests/kuadrant-operator-metrics-reader_rbac.authorization.k8s.io_v1_clusterrole.yaml

Dockerfile
```diff
@@ -31,11 +32,26 @@ ENV VERSION=${VERSION:-unknown}
...
+    else \
+    echo "Skipping extensions build"; \
+    fi
+
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/extensions /extensions
```
didierofrivia, line 54: I reckon there's no harm in copying an empty dir when `WITH_EXTENSIONS` is set to `false`

Makefile
```diff
@@ -378,7 +379,7 @@ docker-build: ## Build docker image with the manager.
 		--build-arg GIT_SHA=$(GIT_SHA) \
 		--build-arg DIRTY=$(DIRTY) \
 		--build-arg VERSION=v$(VERSION) \
-		--build-arg QUAY_IMAGE_EXPIRY=$(QUAY_IMAGE_EXPIRY) \
+		--build-arg WITH_EXTENSIONS=$(WITH_EXTENSIONS) \
```
adam-cattermole, line 385: I removed the `QUAY_IMAGE_EXPIRY` here as it's passed twice

config/manager/controller_manager_config.yaml
```diff
@@ -0,0 +1,17 @@
...
+    kind: ControllerManagerConfig
+    health:
+      healthProbeBindAddress: :8081
+    metrics:
+      bindAddress: :8080
+    webhook:
+      port: 9443
+    leaderElection:
+      leaderElect: true
+      resourceName: f139389e.kuadrant.io
```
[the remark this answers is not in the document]
didierofrivia: I reckon that now that leader election config is in the main.go file: https://github.com/Kuadrant/kuadrant-operator/blob/main/cmd/main.go#L130-L136

config/manager/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ resources:
 - manager.yaml
 - metrics_service.yaml
 - console-plugin-images.yaml
+- controller_manager_config.yaml
```
didierofrivia, line 5: maybe we can remove this file, it was part of the update of controller runtime: https://github.com/Kuadrant/kuadrant-operator/pull/246

config/rbac/kustomization.yaml
didierofrivia: mhmm, this was probably a leftover when the kube-rbac-proxy sidecar was removed. Maybe we could just comment it and remove the file..we might need to assert auth capabilities are working tho

adam-cattermole: I suspected the same to be honest - done, I've removed this

PR #1606, opened 2025-10-03
UPDATE: dns dashboard

Boomatang:
Closes: https://github.com/Kuadrant/dns-operator/issues/542

Add new panel related to the multi cluster delegated dns records.

To verify these changes the dns-operator needs to be set up in a mutli cluster setup. 
I do not currently have scripts for doing this with kuadrant, but do for the dns-operator. I would be happy to walk someone through the steps in a call. Or show them the dashboard configure.

Screenshot From 2025-10-03 14-15-31

### DIFF
--- a/examples/dashboards/dns-operator.json
+++ b/examples/dashboards/dns-operator.json
@@ -580,6 +580,340 @@
       ],
       "title": "Provider request rate by response code (req/s)",
       "type": "timeseries"
+    },
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 26
+      },
+      "id": 10,
+      "panels": [],
+      "title": "Delegated Records",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PA454A88371DD8FEF"
+      },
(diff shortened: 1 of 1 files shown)


Boomatang:
> Here "kind-kuadrant-local-3" is secondary and will always be 0 and the panel seems to be specifically for primaries.

@mikenairn This is confusing me a bit. I know in my set up, the one that I used for the screen shoot there was two primary and a single secondary. Which I think is the same as your setup. I used the make target in the dns-operator to set the cluster up. And I have never seen the secondaries being listed in the manner that your showing. A secondary cluster should not even be exposing the metric that is used.

mikenairn:
> > Here "kind-kuadrant-local-3" is secondary and will always be 0 and the panel seems to be specifically for primaries.
> 
> @mikenairn This is confusing me a bit. I know in my set up, the one that I used for the screen shoot there was two primary and a single secondary. Which I think is the same as your setup. I used the make target in the dns-operator to set the cluster up. And I have never seen the secondaries being listed in the manner that your showing. A secondary cluster should not even be exposing the metric that is used.

OK. It might have been down to the order i created things in. I created all the clusters, added observability, and then updated all the config (made the third cluster a secondary). Not a big deal i guess, but would it go away eventually?

Boomatang:
It will go away sort of. I think the default time frame for the dashboard is an hour. Once that time has elapsed, the metric would be drop. But if a user was to change that default time to a longer period,  then of course it would come back again.

PR #1615, opened 2025-10-06
Fix extensions bundle generation

adam-cattermole:
Trying a different approach here;

The CRDs are always present in the bundle rather than being generated conditionally, but the permissions are now added to the operators CSV when run `WITH_EXTENSIONS`. This should remove the need to apply any extensions manifests separately - the original roles were generated but missing the appropriate RoleBinding

Follows #1603


crstrn13 (approved):
LGTM :shipit: :tada:  Also installed manually and the crds were there

didierofrivia (approved):

Issue #1628, opened 2025-10-15
Allow settings default TTL for DNSRecords created by DnsPolicy


azgabur:
**Is your feature request related to a problem? Please describe.**
This is only problem for testing as the default TTL is now 60 seconds which slows down our dns testing significantly. We would like to set a lower TTL to speed up testing. We are also unable to query authoritative dns servers to circumvent the TTL.

**Describe the solution you'd like**
Add an environment variable similar to `AUTH_SERVICE_TIMEOUT` which would set the default TTL to specified value.

Or add a field to DnsPolicy which could set this on per-policy basis, but as discussed this is maybe not good idea currently.

PR #1668, opened 2025-11-07
Add env variable to set default dns TTL

azgabur:
closes https://github.com/Kuadrant/kuadrant-operator/issues/1628
requires https://github.com/Kuadrant/dns-operator/pull/630

Additionally refactors tests to not use hard coded value but imports constant


Boomatang (changes requested):
We need to address the error handling if the env vars can not be parsed into ints

Boomatang:
To get the validate release data it seems like you need to run `make bundle` and `make helm-build`. They should generate the changes that are missing causing the check to fail.

Boomatang (dismissed):
Changes look good to me.

PR #1675, opened 2025-11-12
[backport] align trlp validation steps with those of rlp  

eguzki:
Backports #1634 to release branch `release-v1.3`


eguzki:
We can safely ignore the Validate Release data. It believes that we want to release an existing tag (v1.3.0). It's not something that should be tested in release branches. When we bump version, the release verification should pass.

jasonmadigan (approved):

PR #1718, opened 2025-12-23
makefile: refactor dependency tooling install

KevFan:


KevFan:
> I get the following error when installing `egctl`
> 
> ```
> ❯ make egctl 
> Downloading github.com/envoyproxy/gateway/cmd/egctl@v1.2.6
> go: github.com/envoyproxy/gateway/cmd/egctl@v1.2.6 (in github.com/envoyproxy/gateway@v1.2.6):
> 	The go.mod file for the module providing named packages contains one or
> 	more replace directives. It must not contain directives that would cause
> 	it to be interpreted differently than if it were the main module.
> make: *** [make/envoy-gateway.mk:19: /home/eguzki/git/kuadrant/kuadrant-operator/bin/egctl-v1.2.6] Error 1
> ```
> 
> You might want to upgrade to `v1.6.1`. However, that raises a new issue which is compilation of the `btrfs` dependency that requires headers to be installed in the local system.
> 
> Your call, keep trying with the `go-install-tool` method, or fallback to downloading the precompiled binary (which requires platform awareness).

Good catch 👍  Hmm, yeah, I've reverted to the previous method as this PR is more about being able to detect changes in tooling locally primarily rather than install consistency

Issue #1716, opened 2025-12-11
Set tracing endpoint of authorino/limitador cr via kuadrant cr


KevFan:
With the tracing endpoint field being available via the kuadrant cr, should this value be passed to the managed limitador and authorino cr?

This will remove the steps for patching the authorino and limitador cr seperatly also.
- https://docs.kuadrant.io/1.3.x/kuadrant-operator/doc/observability/tracing/#kuadrant-tracing-configuration

alexsnaps:
I'd be tempted to say we should... now should we treat this like the Server-side applies again, or in some similar way? i.e. should users be able to configure those components (e.g. limitador just as much) independently themselves regarding these config bits?

KevFan:
Yeah, I think we should to still give users the fine tune control if they want it

PR #1738, opened 2026-01-20
feat: set tracing config via kuadrant cr

KevFan:
# Description
Closes: https://github.com/Kuadrant/kuadrant-operator/issues/1716

This PR implements tracing configuration propagation from the `Kuadrant` CR to child `Authorino` and `Limitador` CRs, enabling centralized observability configuration while maintaining user flexibility.

  ### Propagation Logic

  The tracing field management follows a conditional ownership model:

  #### When Kuadrant CR has tracing configured:

  - Kuadrant propagates tracing configuration from `Kuadrant.spec.observability.tracing `to child resources (Authorino and Limitador CRs)
  - However, if users have already configured tracing directly on Authorino or Limitador CRs, Kuadrant always cedes ownership
  - Uses Server-Side Apply (SSA) without forcing ownership; conflicts result in ceding the field to the user
  - User-defined tracing configurations in child CRs are preserved across reconciliation loops, even if Kuadrant CR tracing is subsequently updated
  - If Kuadrant CR tracing fields are cleared/set to nil, only non-user-managed child CR tracing fields are reset to empty values

  #### When Kuadrant CR has no tracing configured:

  - Ownership is ceded to users - they can set tracing fields directly on Limitador and Authorino CRs
  - Kuadrant operator omits spec.tracing from SSA patches, leaving the field unmanaged
  - User modifications to child CR tracing fields are preserved across reconciliation loops

# Verification
- Passing integration tests should be sufficient
- Otherwise, play around with setting the tracing fields on `Kuadrant`, `Authorino` and `Limitaodr` CRs, and the above propagation logic should apply

### DIFF
--- a/api/v1beta1/kuadrant_types.go
+++ b/api/v1beta1/kuadrant_types.go
@@ -128,7 +128,10 @@ type DataPlane struct {
 // It enables tracing spans to be exported to external tracing systems (e.g., Jaeger, Zipkin).
 type Tracing struct {
 	// DefaultEndpoint is the default URL of the tracing collector backend where spans should be sent.
-	// Can be overridden per-gateway in future versions.
+	// This endpoint is used by Auth (Authorino), RateLimiting (Limitador) and WASM services for exporting trace data.
+	// If tracing endpoints have been configured directly in Authorino or Limitador CRs, those take precedence
+	// over this default value.
+	// Note: Per-gateway overrides are not currently supported.
 	DefaultEndpoint string `json:"defaultEndpoint,omitempty"`
 
 	// Insecure controls whether to skip TLS certificate verification.
--- a/bundle/manifests/kuadrant.io_kuadrants.yaml
+++ b/bundle/manifests/kuadrant.io_kuadrants.yaml
@@ -133,7 +133,10 @@ spec:
                       defaultEndpoint:
                         description: |-
                           DefaultEndpoint is the default URL of the tracing collector backend where spans should be sent.
-                          Can be overridden per-gateway in future versions.
+                          This endpoint is used by Auth (Authorino), RateLimiting (Limitador) and WASM services for exporting trace data.
+                          If tracing endpoints have been configured directly in Authorino or Limitador CRs, those take precedence
+                          over this default value.
+                          Note: Per-gateway overrides are not currently supported.
                         type: string
                       insecure:
                         description: Insecure controls whether to skip TLS certificate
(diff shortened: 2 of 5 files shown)


internal/controller/authorino_reconciler.go
```diff
@@ -58,23 +59,87 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
+			logger.Error(err, "failed to destruct authorino", "status", "error")
+			return err
+		}
+
+		// This is required since Destruct resulted in the spec.tracing.insecure field being omitted
+		if kobj.Spec.Observability.Tracing != nil {
+			if err := unstructured.SetNestedField(unstructuredAuthorino.Object, kobj.Spec.Observability.Tracing.Insecure, "spec", "tracing", "insecure"); err != nil {
+				return err
+			}
+		}
```
adam-cattermole: I'm not sure I'm following exactly what's going on here - would have to investigate in my IDE, the field ends up dropped when we destruct?

KevFan: Yes, I noticed when we `Destruct`, the value is dropped when the value is `false` (might be due to how the unmarshalling is done in the [policy-machinery](https://github.com/Kuadrant/policy-machinery/blob/1ca17a075dd29f85690c2dead64d78c13f9d16a4/controller/runnable.go#L286-L293)). 

This can become an issue say if authorino has tracing set initially:

```
kubectl apply -f - <<EOF   
apiVersion: operator.authorino.kuadrant.io/v1beta1
kind: Authorino
metadata:
  name: authorino
  namespace: default
spec:
  tracing:
    endpoint: "grpc://authorino:4719"
    insecure: true
EOF
```

Then Kuadrant sets tracing with `insecure: false`:
```
kubectl apply -f - <<EOF   
kind: Kuadrant
metadata:
  name: kuadrant-sample
  namespace: default
spec:
  observability: 
    tracing:
      defaultEndpoint: grpc://kuadrant:4719
EOF
```
Then, we set authorino tracing endpoint to empty:
```
kubectl apply -f - <<EOF   
apiVersion: operator.authorino.kuadrant.io/v1beta1
kind: Authorino
metadata:
  name: authorino
  namespace: default
spec:
  tracing:
    endpoint: ""
    insecure: true
EOF
```

In this istance, kuadrant should now own and manage authorino's tracing fields but  `insecure` field in authorino will not get the default `false` value from kuadrant cr without this and will stay `true` 🤔

KevFan: Just to loop back on this, this field being dropped looks to be intended behaviour from the [marshaller](https://github.com/golang/go/issues/13284#issuecomment-157288425) and is actually due to the `omitempty` on the struct definition 🤔 
- https://github.com/Kuadrant/authorino-operator/blob/42a22574c592af7dc11d56368f499b92579b09b8/api/v1beta1/authorino_types.go#L108

KevFan: Yes, I think we don't need to include this if the overarching tracing field is reset to`nil` also 🤔

PR #1822, opened 2026-03-16
Extensions gRPC proto cache and client

### LINKED ISSUE #1841: Operator Reflection and Caching
Implement gRPC reflection client and proto caching infrastructure in the operator.

## Tasks

- [ ] Implement `ProtoCache` with thread-safe get/set/delete
  - [ ] Unit test: concurrent access
  - [ ] Unit test: delete by cluster
- [ ] Implement `ReflectionClient` with recursive dependency fetching
  - [ ] Unit test: reflection failure (unreachable, no service)
  - [ ] Unit test: 30-second timeout prevents hanging on slow upstreams
- [ ] Update `RegisterUpstreamMethod` handler to trigger reflection
  - [ ] Unit test: cache populated on success
  - [ ] Unit test: error returned on reflection failure
- [ ] Update `ClearPolicyData` to clean up proto cache
  - [ ] Unit test: cache entries deleted when policy removed

## Reference

Design doc: [grpc-reflection-dynamic-messages-design.md](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/proposals/grpc-reflection-dynamic-messages-design.md#phase-21-operator-reflection-and-caching)

### LINKED ISSUE #1842: Operator Descriptor Service
Implement gRPC descriptor service to serve cached proto descriptors to wasm clients.

## Tasks

- [ ] Define `descriptor_service.proto` with `GetServiceDescriptors` RPC
  - [ ] Run `make generate` to generate Go code
- [ ] Implement `GetServiceDescriptors` handler
  - [ ] Unit test: batch fetch returns multiple descriptors
  - [ ] Unit test: missing descriptors return error
- [ ] Implement `StartDescriptorServer` TCP server
  - [ ] Unit test: server starts and handles requests
- [ ] Add Kubernetes Service manifest (`config/manager/grpc_service.yaml`)
- [ ] Update operator deployment to expose port 50051
  - [ ] Integration test: verify service is reachable from pod
- [ ] Implement EnvoyFilter generation for Istio (`istio_extension_reconciler.go`)
  - [ ] Unit test: EnvoyFilter created with correct cluster config
  - [ ] Integration test: verify cluster accessible from gateway pod
- [ ] Implement EnvoyPatchPolicy generation for Envoy Gateway (`envoy_gateway_extension_reconciler.go`)
  - [ ] Unit test: EnvoyPatchPolicy created with correct cluster config
  - [ ] Integration test: verify cluster accessible from gateway pod

## Reference

Design doc: [grpc-reflection-dynamic-messages-design.md](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/proposals/grpc-reflection-dynamic-messages-design.md#phase-22-operator-descriptor-service)

adam-cattermole:
Cleaned up the PoC code for the reflection client and cache.

- We now validate that we can retrieve protos from the registered upstream on registration
- Store the protos in a new `ProtoCache`
- These are then provided to the wasm-shim via the descriptor service, hosted on port 50051
- When an upstream is registered, we also register this descriptor service (the kuadrant operators grpc port) via the envoy filters.

See related https://github.com/Kuadrant/wasm-shim/pull/316

## Verification

Deploy kuadrant:
```sh
make local-setup
kubectl set env -n kuadrant-system deployment/kuadrant-operator-controller-manager \
  RELATED_IMAGE_WASMSHIM=quay.io/kuadrant/wasm-shim:dynamic-grpc
make apply-extensions-manifests
make local-apply-extensions

kubectl apply -f - <<EOF
kind: Kuadrant
metadata:
  name: kuadrant
  namespace: kuadrant-system
EOF
```

Deploy threat assessment service:
```sh
kubectl create namespace security
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/threat-assessment-service/refs/heads/main/examples/threatpolicy/threat-service/configmap.yaml
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/threat-assessment-service/refs/heads/main/examples/threatpolicy/threat-service/deployment.yaml
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/threat-assessment-service/refs/heads/main/examples/threatpolicy/threat-service/service.yaml
```

Gateway:
```sh
export KUADRANT_GATEWAY_NS=api-gateway # Namespace for the example Gateway
export KUADRANT_GATEWAY_NAME=external # Name for the example Gateway
export KUADRANT_DEVELOPER_NS=toystore # Namespace for an example toystore app
kubectl create ns ${KUADRANT_GATEWAY_NS}
kubectl apply -f - <<EOF
kind: Gateway
metadata:
  name: ${KUADRANT_GATEWAY_NAME}
  namespace: ${KUADRANT_GATEWAY_NS}
  labels:
    kuadrant.io/gateway: "true"
spec:
  gatewayClassName: istio
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      allowedRoutes:
        namespaces:
          from: All
EOF
```

HTTPRoute and Toystore:
```sh
kubectl create ns ${KUADRANT_DEVELOPER_NS}
kubectl apply -f https://raw.githubusercontent.com/Kuadrant/Kuadrant-operator/main/examples/toystore/toystore.yaml -n ${KUADRANT_DEVELOPER_NS}
kubectl apply -f - <<EOF
kind: HTTPRoute
metadata:
  name: toystore
  namespace: ${KUADRANT_DEVELOPER_NS}
spec:
  parentRefs:
  - name: ${KUADRANT_GATEWAY_NAME}
    namespace: ${KUADRANT_GATEWAY_NS}
  hostnames:
  - api.toystore.com
  rules:
  - matches:
    - path:
        type: Exact
        value: "/toy"
      method: GET
    backendRefs:
    - name: toystore
      port: 80
EOF
```

Create AuthPolicy (for now so we have an actionset):
```sh
kubectl apply -f - <<EOF
apiVersion: kuadrant.io/v1
kind: AuthPolicy
metadata:
  name: toystore
  namespace: ${KUADRANT_DEVELOPER_NS}
spec:
  targetRef:
    kind: HTTPRoute
    name: toystore
  rules:
    authentication:
      "api-key-users":
        apiKey:
          selector:
            matchLabels:
              app: toystore
          allNamespaces: true
        credentials:
          authorizationHeader:
            prefix: APIKEY
    response:
      success:
        filters:
          "identity":
            json:
              properties:
                "userid":
EOF
```

ThreatPolicy:
```sh
kubectl apply -f - <<EOF
kind: ThreatPolicy
metadata:
  name: external-threatpolicy
  namespace: api-gateway
spec:
  targetRef:
    kind: Gateway
    name: external
  threshold: 5
EOF
```

Check gateway logs to see protobuf pulled by the shim, and envoy filter created.

## Summary by CodeRabbit

* **New Features**
  * Added a gRPC descriptor service and runtime support; WASM configs can reference dynamic gRPC-backed services via descriptor lookup.

* **Chores**
  * Exposed descriptor service on port 50051 across manifests, Helm chart and kustomize; controller now honours the descriptor port via environment.

* **Documentation**
  * Updated design proposal and implementation checklist to reference the descriptor service and related flows.

* **Tests**
  * Added and expanded tests for reflection, proto-cache, descriptor fetch, registration and WASM injection.


Dockerfile
```diff
@@ -41,9 +41,7 @@ RUN if [ "$WITH_EXTENSIONS" = "true" ]; then \
     mkdir -p extensions/plan-policy && \
     CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -o extensions/plan-policy/plan-policy cmd/extensions/plan-policy/main.go && \
     mkdir -p extensions/telemetry-policy && \
-    CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -o extensions/telemetry-policy/telemetry-policy cmd/extensions/telemetry-policy/main.go && \
-    mkdir -p extensions/threat-policy && \
-    CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -o extensions/threat-policy/threat-policy cmd/extensions/threat-policy/main.go; \
+    CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} go build -a -o extensions/telemetry-policy/telemetry-policy cmd/extensions/telemetry-policy/main.go; \
```
adam-cattermole, line 44: /cc @philbrookes; removed the threat policy from the regular Dockerfile here

cmd/extensions/threat-policy/internal/controller/threatpolicy_reconciler.go
```diff
@@ -73,7 +73,9 @@ func (r *ThreatPolicyReconciler) reconcileSpec(ctx context.Context, pol *v1alpha
 	r.Logger.Info("registering upstream", "url", threatServiceURL)
 
 	if err := kuadrantCtx.RegisterUpstreamMethod(ctx, pol, types.UpstreamConfig{
-		URL: threatServiceURL,
+		URL:     threatServiceURL,
+		Service: "threat.v1.ThreatAssessmentService",
+		Method:  "AssessRequest",
 	}); err != nil {
```
adam-cattermole, line 79: Updated the request in the threat policy extension to use the service and method _for now_, while we decide on how to proceed with #1848

internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -285,6 +287,19 @@ func buildUpstreamEnvoyPatchPolicy(gateway *machinery.Gateway, upstreams []exten
...
+	descriptorServiceHost := fmt.Sprintf("kuadrant-operator-grpc.%s.svc.cluster.local", operatorNamespace)
+	descriptorServicePort, _ := env.GetInt("EXTENSIONS_DESCRIPTOR_SERVICE_PORT", 50051)
+
+	descriptorPatches, err := kuadrantenvoygateway.BuildEnvoyPatchPolicyClusterPatch("kuadrant-operator-grpc", descriptorServiceHost, descriptorServicePort, false, func(h string, p int, _ bool) map[string]any {
+		return buildClusterPatch("kuadrant-operator-grpc", h, p, false)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to build cluster patch for descriptor service: %w", err)
+	}
+	allPatches = append(allPatches, descriptorPatches...)
```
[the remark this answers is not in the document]
adam-cattermole: I've added a comment to the code here, the `kuadrant-operator-grpc` is not able to collide with extensions upstreams as they must be prefixed with `ext-`

internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -65,6 +66,17 @@ func (r *EnvoyGatewayExtensionReconciler) Reconcile(ctx context.Context, _ []con
...
+	// reconcile for each gateway based on the desired wasm plugin policies calculated before
+	gateways := lo.Map(topology.Targetables().Items(func(o machinery.Object) bool {
+		return o.GroupVersionKind().GroupKind() == machinery.GatewayGroupKind
+	}), func(g machinery.Targetable, _ int) *machinery.Gateway {
+		return g.(*machinery.Gateway)
+	})
+
+	// Reconcile EnvoyPatchPolicy cluster patches for registered upstreams FIRST
+	// This ensures clusters exist in Envoy before EnvoyExtensionPolicy tries to use them
+	r.reconcileUpstreamClusters(ctx, topology, gateways)
```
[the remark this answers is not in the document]
adam-cattermole: Updated the comment here - it was inaccurate. Regardless of whether the envoy filter has been created or not we cannot know if this has been picked up by the gateway provider and propagated to envoys cluster configuration - we just have to assume it will _eventually_

pkg/extension/grpc/generate_proto.sh
```diff
@@ -17,5 +17,6 @@ fi
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 protoc -I=$SCRIPT_DIR -I=$SCRIPT_DIR/$1 --go_out=$SCRIPT_DIR $SCRIPT_DIR/$1/common.proto
 protoc -I=$SCRIPT_DIR -I=$SCRIPT_DIR/$1 --go_out=$SCRIPT_DIR --go-grpc_out=$SCRIPT_DIR $SCRIPT_DIR/$1/kuadrant.proto
+protoc -I=$SCRIPT_DIR -I=$SCRIPT_DIR/$1 --go_out=$SCRIPT_DIR --go-grpc_out=$SCRIPT_DIR $SCRIPT_DIR/$1/descriptor_service.proto
```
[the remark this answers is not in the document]
adam-cattermole: This is following the existing structure

pkg/extension/grpc/v1/descriptor_service.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+package kuadrant.v1;
+
+option go_package = "/v1";
```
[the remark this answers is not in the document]
adam-cattermole: This is not new or related to the changes in this PR

internal/extension/registry.go
```diff
@@ -370,10 +376,15 @@ func (r *RegisteredDataStore) DeleteSubscription(policy ResourceID, expression s
...
-func (r *RegisteredDataStore) SetUpstream(key RegisteredUpstreamKey, entry RegisteredUpstreamEntry) {
+func (r *RegisteredDataStore) SetUpstream(key RegisteredUpstreamKey, entry RegisteredUpstreamEntry, fds *descriptorpb.FileDescriptorSet) {
 	r.upstreamsMutex.Lock()
 	defer r.upstreamsMutex.Unlock()
 	r.registeredUpstreams[key] = entry
+	cacheKey := ProtoCacheKey{
+		ClusterName: entry.ClusterName,
+		Service:     entry.Service,
+	}
+	r.protoCache.Set(cacheKey, fds)
```
adam-cattermole, line 389: This has been addressed by updating the RegisteredUpstreamKeys, there should no longer be any updates

internal/extension/manager.go
```diff
@@ -259,23 +329,55 @@ func (s *extensionService) Ping(_ context.Context, _ *extpb.PingRequest) (*extpb
...
+
+		fds, found := s.registeredData.GetProtoDescriptor(cacheKey)
+		if !found {
+			return nil, fmt.Errorf("descriptors not found for cluster=%q service=%q", serviceRef.ClusterName, serviceRef.Service)
+		}
+
+		fdsBytes, err := proto.Marshal(fds)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal descriptors for cluster=%q service=%q: %w", serviceRef.ClusterName, serviceRef.Service, err)
+		}
```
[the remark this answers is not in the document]
adam-cattermole: We'll leave this as a follow up, this is a new API and subject to change

tests/istio/upstream_cluster_reconciler_test.go
```diff
@@ -93,18 +105,51 @@ var _ = Describe("Upstream cluster EnvoyFilter controller", Serial, func() {
...
+					endpointAddr := lbEndpoint["endpoint"].(map[string]any)
+					address := endpointAddr["address"].(map[string]any)
+					socketAddress := address["socket_address"].(map[string]any)
+
+					operatorNamespace := env.GetString("OPERATOR_NAMESPACE", "kuadrant-system")
+					descriptorServicePort, _ := env.GetInt("EXTENSIONS_DESCRIPTOR_SERVICE_PORT", 50051)
+					expectedHost := fmt.Sprintf("kuadrant-operator-grpc.%s.svc.cluster.local", operatorNamespace)
+
+					Expect(socketAddress["address"]).To(Equal(expectedHost))
+					Expect(socketAddress["port_value"]).To(BeNumerically("==", descriptorServicePort))
```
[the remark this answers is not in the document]
adam-cattermole: The rhs of the expression is already the fallback on error.

internal/extension/reflection.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func (rc *ReflectionClient) FetchServiceDescriptors(ctx context.Context, url, serviceName, methodName string) (*descriptorpb.FileDescriptorSet, error) {
+	ctx, cancel := context.WithTimeout(ctx, rc.timeout)
+	defer cancel()
+
+	conn, err := grpc.NewClient(url, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to %s: %w", url, err)
+	}
```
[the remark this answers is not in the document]
adam-cattermole: We pass in the parsed host to the `FetchServiceDescriptors` so the scheme is already stripped at this point

internal/extension/registry.go
```diff
@@ -370,10 +378,15 @@ func (r *RegisteredDataStore) DeleteSubscription(policy ResourceID, expression s
...
```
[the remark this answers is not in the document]
adam-cattermole: `DeleteUpstream` is currently only used in tests, the only way that entries are deleted currently is via `ClearPolicy`

internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -165,6 +167,18 @@ func (r *EnvoyGatewayExtensionReconciler) reconcileUpstreamClusters(ctx context.
...
+		gatewayClassObj, found := lo.Find(gatewayClassObjs, func(obj machinery.Targetable) bool {
+			return obj.GroupVersionKind().GroupKind() == machinery.GatewayClassGroupKind
+		})
+		if found {
+			gatewayClass := gatewayClassObj.(*machinery.GatewayClass)
+			if !lo.Contains(envoyGatewayGatewayControllerNames, gatewayClass.Spec.ControllerName) {
+				continue
+			}
+		}
+
```
crstrn13, line 181: I think we have this logic scattered across the repository in several places already. Maybe not for this PR, but extracting a helper function would be ideal.

internal/controller/istio_extension_reconciler.go
```diff
@@ -169,6 +171,18 @@ func (r *IstioExtensionReconciler) reconcileUpstreamClusters(ctx context.Context
...
+		gatewayClassObj, found := lo.Find(gatewayClassObjs, func(obj machinery.Targetable) bool {
+			return obj.GroupVersionKind().GroupKind() == machinery.GatewayClassGroupKind
+		})
+		if found {
+			gatewayClass := gatewayClassObj.(*machinery.GatewayClass)
+			if !lo.Contains(istioGatewayControllerNames, gatewayClass.Spec.ControllerName) {
+				continue
+			}
+		}
+
```
crstrn13, line 185: same as for envoy gateways.

internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -286,6 +300,25 @@ func buildUpstreamEnvoyPatchPolicy(gateway *machinery.Gateway, upstreams []exten
...
+		descriptorServicePort = defaultDescriptorPort
+	}
+
+	descriptorPatches, err := kuadrantenvoygateway.BuildEnvoyPatchPolicyClusterPatch("kuadrant-operator-grpc", descriptorServiceHost, descriptorServicePort, false, func(h string, p int, _ bool) map[string]any {
+		return buildClusterPatch("kuadrant-operator-grpc", h, p, false)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to build cluster patch for descriptor service: %w", err)
+	}
+	allPatches = append(allPatches, descriptorPatches...)
```
crstrn13, line 323: The descriptor service config (env var reads for `OPERATOR_NAMESPACE`, `EXTENSIONS_DESCRIPTOR_SERVICE_PORT`, host formatting) is duplicated identically in the Istio reconciler's `buildUpstreamEnvoyFilter`. This builder function also feels like the wrong place to resolve env vars — they're static config that should be resolved once, not on every reconcile call.                  

Worth extracting the descriptor service config into a shared struct/function, and having each provider's builder just consume it. That would eliminate the duplication and keep env var resolution out of the build logic.

internal/extension/manager.go
```diff
@@ -133,6 +144,65 @@ func (m *Manager) Stop() error {
 	return err
 }
 
+func (m *Manager) startDescriptorServer() error {
+	const defaultPort = 50051
+	descriptorPort, portErr := env.GetInt("EXTENSIONS_DESCRIPTOR_SERVICE_PORT", defaultPort)
+	if portErr != nil {
+		m.logger.Error(portErr, "invalid EXTENSIONS_DESCRIPTOR_SERVICE_PORT, using default", "default", defaultPort)
+		descriptorPort = defaultPort
+	}
```
crstrn13, line 153: Reading the same static config.

internal/extension/manager_test.go
crstrn13, line 1: Consider adding the following tests for RegisterUpstreamMethod:                                                                                                                                       
                                                                                                                                                                                                        
  Multi-registration scenarios:                                                                                                                                                                         
  - Multiple methods on the same service from the same policy — verifies both are stored independently and share the same cluster name and proto cache entry
  - Multiple methods from different policies on the same service — verifies independent storage and that clearing one policy doesn't remove the shared proto cache while the other still references it  
  - Re-registration (same policy, URL, service, method twice) — verifies idempotent overwrite, not duplication                                                                                        
                                                                                                                                                                                                        
  Failure scenarios:                                                                                                                                                                                    
  - Upstream unreachable during reflection — the old _DialFailure test was removed when the dialer was replaced with the reflection fetcher, but there's no replacement covering network-level failures 
  (as opposed to _InvalidMethod which only tests method-not-found)                                                                                                                                      
  - Partial failure across multiple registrations — first method succeeds, second fails. Verifies the first registration remains intact, no partial entry is stored, and the change notifier fires only
  once                                                                                                                                                                                                  
                                                                                                                                                                                                        
  These don't add significant line coverage (~0.7%), but they lock in behavioral invariants around cache sharing, idempotency, and failure isolation that single-method happy-path tests don't exercise.

philbrookes (dismissed):

internal/controller/envoy_gateway_extension_reconciler.go
adam-cattermole: Yeah I considered this, but also think this use of the topology is kind of expected - the second bit is tied to flow control so whilst we could put it in a function we'd still need logic for flow control in the reconciler.

There might be a way to retrieve the gatewayclass earlier to configure the workflow running though.. or to grab once and re-use it in the wasm config bit, but once again this is only the first 3 lines.

internal/controller/envoy_gateway_extension_reconciler.go
adam-cattermole: I see your point on this, but I'm not sure the alternative is any better. The env lookups are actually trivial; (map lookup). To remove it we'd need some kind of struct that inits once and then somehow thread that into the reconcilers? unless you're suggesting some kind of global init once struct? Either way feels more indirection than the duplication costs unless I'm missing something

crstrn13: fair point.

internal/controller/envoy_gateway_extension_reconciler.go
crstrn13: fair point.

internal/extension/manager_test.go
adam-cattermole: > Multi-registration scenarios

I'll add some tests for these, I think though "Multiple methods from different policies on the same service" is covered by `TestClearPolicy_ProtoCacheCleanup`

> Upstream unreachable during reflection — the old _DialFailure test was removed when the dialer was replaced with the reflection fetcher, but there's no replacement covering network-level failures
(as opposed to _InvalidMethod which only tests method-not-found)

This is covered by the new package and tests `reflection_test.go`:
* `TestReflectionClient_FetchServiceDescriptors_InvalidURL`
* `TestReflectionClient_FetchServiceDescriptors_Timeout`

Could add the partial failure test just to catch regressions

adam-cattermole: Just added those now

crstrn13 (dismissed):
LGTM! 🎉 :shipit:

philbrookes (approved):

PR #1856, opened 2026-04-01
docs: DNS-based routing guide for egress gateway

### LINKED ISSUE #1847: [Egress] Investigate DNS-Based Routing to Egress Gateway
## DNS-Based Routing Investigation

## Description

Investigate and document DNS configuration options for routing pod traffic through the egress gateway to external services. There are two approaches to consider:

**Approach 1: Internal hostname** — Application uses a local hostname (e.g., `my.api.local`) that resolves to the egress gateway. The gateway rewrites and routes to the real external service (`my.api.com`). TLS is simpler — app sends HTTP, gateway handles TLS origination. However, the application must be aware of the internal hostname.

**Approach 2: External hostname** — Application uses the real hostname (`my.api.com`) transparently. DNS is configured so the hostname resolves to the egress gateway IP instead of the real service IP. Transparent to the application but requires DNS interception. TLS requires careful handling — if the app initiates HTTPS, certificate validation will fail unless the gateway terminates and re-originates TLS.

See also: [existing external-api guide](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/misc/external-api.md)

## Acceptance Criteria

1. Investigate feasibility of both routing approaches (internal hostname and external hostname), including TLS considerations for each
2. Validate at least one working approach with documented results
3. Verify pods successfully resolve hostnames to the egress gateway IP
4. Validate end-to-end flow: application uses hostname → resolves to gateway IP → HTTPRoute matches → policies apply → traffic reaches external service
5. Test existing DNSPolicy (no API changes) attached to the egress gateway as one of the DNS routing approaches — determine if it can create the necessary DNS records pointing to the gateway
6. For the external hostname approach: document cluster DNS forwarding configuration steps (e.g., CoreDNS ConfigMap), including platform-specific variations where DNS is managed via a dedicated operator/CR
7. Document trade-offs between the approaches (internal hostname, external hostname, DNSPolicy-based)
8. Document limitations and considerations (e.g., ClusterIP vs LoadBalancer for egress gateway)
9. Determine whether existing infrastructure suffices or code changes are necessary

## Context

The RFC identifies DNS-based routing as a core investigation item. DNSPolicy API changes and egress-specific automation are tracked as future work — this task validates the manual DNS configuration patterns and existing DNSPolicy capabilities that users can leverage today.

## Deliverable

Documentation/guide for DNS-based routing to the egress gateway, to be included in the end-to-end documentation (Kuadrant/kuadrant-operator#1804).

maksymvavilov:
## Summary

- Adds `doc/user-guides/egress/dns-routing.md` documenting four approaches for routing pod traffic through an Istio egress gateway to external services
- Investigates and validates DNS configuration patterns as part of the egress gateway epic

### Approaches documented

| # | Approach | Result |
|---|----------|--------|
| 1 | Internal hostname (URLRewrite) | Validated — simplest TLS, no DNS changes, app uses gateway service name |
| 2 | External hostname (CoreDNS rewrite) | Validated — transparent to app, requires CoreDNS config + static ServiceEntry IPs (split DNS) |
| 3 | DNSPolicy on egress gateway | Investigated — not useful for egress today |
| 4 | Sidecar-based routing (VirtualService) | Validated — fully transparent, no split DNS, requires mesh enrollment |

### Key findings

- All three working approaches use existing Kuadrant infrastructure — no code changes needed
- RateLimitPolicy was tested and confirmed working on egress (429 after limit)
- DNSPolicy is designed for ingress DNS advertisement and cannot solve the egress split DNS problem
- CoreDNS `rewrite` directive is preferred over `hosts` for production (avoids hardcoded gateway IPs)

Closes #1847

## Test plan

- [x] Approach 1: `curl http://egress-gateway-istio.gateway-system.svc.cluster.local/get` → 200, Host rewritten to httpbin.org, TLS originated
- [x] Approach 2: CoreDNS rewrite + static ServiceEntry → `curl http://httpbin.org/get` → 200 via egress gateway
- [x] Approach 3: DNSPolicy investigation — documented why it's not applicable
- [x] Approach 4: Sidecar injection + VirtualService → `curl http://httpbin.org/get` → 200 via egress gateway, non-sidecar pods unaffected
- [x] RateLimitPolicy on egress: 5 requests → 200, 6th → 429

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive user guide on DNS routing for egress gateways, covering the recommended internal-hostname approach with host-header rewriting, an alternative CoreDNS-based pod DNS integration with split-DNS behaviour, and other options (per-pod hostAliases and sidecar virtual-service routing). Includes prerequisites, example configurations, validation steps and TLS considerations for each approach.

### DIFF
--- a/doc/user-guides/egress/dns-routing.md
+++ b/doc/user-guides/egress/dns-routing.md
@@ -1,0 +1,334 @@
+# DNS Routing for Egress Gateway
+
+This guide covers routing approaches for directing pod traffic through an Istio egress gateway to external services. The approaches are independent of how the egress gateway was deployed.
+
+## Prerequisites
+
+- A running Istio egress gateway (Gateway, ServiceEntry, DestinationRule, HTTPRoute)
+
+> For a ready-made setup, see the [Egress Gateway Setup](egress-gateway.md) guide.
+
+The examples in this guide use:
+
+| Resource | Value |
+|----------|-------|
+| Gateway namespace | `gateway-system` |
+| Gateway name | `kuadrant-egressgateway` |
+| Istio-created Service | `kuadrant-egressgateway-istio.gateway-system.svc.cluster.local` |
+| External service | `httpbin.org` |
+
+---
+
+## Recommended: Internal Hostname
+
+The application uses the egress gateway's cluster-internal service name as the hostname. The gateway rewrites the host header and routes the request to the real external service with TLS origination.
+
+This is the recommended approach because it requires zero DNS configuration, works on any platform, and has the simplest TLS model.
+
+### How It Works
+
+```text
+Application: curl http://kuadrant-egressgateway-istio.gateway-system.svc.cluster.local/get
+    -> Kubernetes DNS resolves to egress gateway ClusterIP
+    -> Egress gateway receives plain HTTP request
+    -> HTTPRoute matches, URLRewrite filter sets Host header to httpbin.org
+    -> DestinationRule initiates TLS to httpbin.org
+    -> External service receives the request over HTTPS and responds
+```
+
+### DNS Configuration
+
+No additional DNS configuration is needed. When Istio provisions the egress gateway, it creates a Kubernetes Service named `kuadrant-egressgateway-istio` in the gateway's namespace. Kubernetes automatically adds this service to the cluster DNS, so any pod can resolve it without additional configuration:
+
+```bash
+kubectl exec test-client -n egress-test -- nslookup kuadrant-egressgateway-istio.gateway-system.svc.cluster.local
+# Name:      kuadrant-egressgateway-istio.gateway-system.svc.cluster.local
+# Address:   <gateway-clusterip>
+```
+
+If a friendlier hostname is preferred (e.g., `httpbin-egress`), create an ExternalName Service that aliases the gateway service:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: httpbin-egress
+  namespace: app-namespace
+spec:
+  type: ExternalName
+  externalName: kuadrant-egressgateway-istio.gateway-system.svc.cluster.local
+```
+
+Applications in `app-namespace` can then use `http://httpbin-egress/get`.
+
+### HTTPRoute with Host Rewrite
+
+This approach replaces the HTTPRoute deployed by the [egress gateway setup](egress-gateway.md). Instead of matching the external hostname, it matches the gateway's internal service name and rewrites the `Host` header to the real external hostname:
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: httpbin-internal-hostname
+  namespace: gateway-system
+spec:
+  parentRefs:
+    - name: kuadrant-egressgateway
+      namespace: gateway-system
+  hostnames:
+    - kuadrant-egressgateway-istio.gateway-system.svc.cluster.local
+  rules:
+    - filters:
+        - type: URLRewrite
+          urlRewrite:
+            hostname: httpbin.org
+      backendRefs:
+        - group: networking.istio.io
+          kind: Hostname
+          name: httpbin.org
+          port: 443
+```
+
+The `URLRewrite` filter changes the `Host` header from the internal hostname to `httpbin.org`, and the `Hostname` backend (provided by Istio's ServiceEntry) routes the request to the real external service.
+
+> **Note:** The Gateway listener does not restrict by hostname, so it accepts traffic on any host, including the internal service name.
+
+### TLS Considerations
+
+This approach has the simplest TLS model:
+
+1. **App to gateway:** Plain HTTP. The application sends unencrypted HTTP to the egress gateway's ClusterIP. Traffic stays within the cluster network.
+2. **Gateway to external service:** TLS origination. The DestinationRule establishes a TLS connection to the external service with the appropriate SNI.
+
+The application does not need to handle TLS certificates for the external service at all.
+
+### Validated Result
+
+Tested by sending a request from a pod in the `egress-test` namespace:
+
+```bash
+kubectl exec test-client -n egress-test -- \
+  curl -s http://kuadrant-egressgateway-istio.gateway-system.svc.cluster.local/get
+```
+
+Response confirms the flow works end-to-end:
+
+```json
+{
+  "headers": {
+    "Host": "httpbin.org",
+    "X-Envoy-Original-Host": "kuadrant-egressgateway-istio.gateway-system.svc.cluster.local",
+    "X-Envoy-Decorator-Operation": "httpbin.org:443/*"
+  },
+  "url": "https://httpbin.org/get"
+}
+```
+
+- `Host: httpbin.org` - URLRewrite correctly rewrote the host header
+- `X-Envoy-Original-Host` - preserves the original internal hostname
+- `url: https://httpbin.org/get` - TLS origination happened (app sent HTTP, gateway upgraded to HTTPS)
+
+---
+
+## Alternative: Pod DNS with Kuadrant CoreDNS
+
+The application uses the real external hostname (`httpbin.org`) transparently. Workload pods are configured with a custom `dnsConfig` that uses a kuadrant CoreDNS instance as their nameserver. A DNSPolicy on the egress gateway creates DNS records that resolve the external hostname to the gateway IP.
+
+Use this approach when the application cannot be changed to use an internal hostname.
+
+### How It Works
+
+```text
+Application: curl http://httpbin.org/get
+    -> Pod dnsConfig directs query to kuadrant CoreDNS
+    -> Kuadrant CoreDNS resolves httpbin.org to egress gateway IP
+    -> Egress gateway receives HTTP request with Host: httpbin.org
+    -> HTTPRoute matches on httpbin.org
+    -> DestinationRule initiates TLS to httpbin.org
+    -> External service responds
+```
+
+Split DNS is handled by design: workload pods use kuadrant CoreDNS (which returns the gateway IP), while the gateway pod uses standard cluster DNS (which returns real external IPs). These separate resolution paths prevent routing loops.
+
+### Prerequisites
+
+In addition to the [common prerequisites](#prerequisites), this approach requires:
+
+- **dns-operator** installed (included with Kuadrant)
+- **Kuadrant CoreDNS** deployed - see the [CoreDNS integration guide](https://github.com/kuadrant/dns-operator/blob/main/docs/coredns/coredns-integration.md) for deployment instructions
+
+### CoreDNS Configuration
+
+The kuadrant CoreDNS instance needs a zone block for the external hostname and a forward block for all other queries:
+
+```text
+httpbin.org {
+    kuadrant
+}
+. {
+    forward . <cluster-dns-ip>
+}
+```
+
+Replace `<cluster-dns-ip>` with your cluster's DNS service IP:
+
+- **Kubernetes:** `kubectl get svc kube-dns -n kube-system -o jsonpath='{.spec.clusterIP}'`
+- **OpenShift:** `kubectl get svc dns-default -n openshift-dns -o jsonpath='{.spec.clusterIP}'`
+
+The IP depends on your cluster's service subnet configuration - do not assume a default value.
+
+The `httpbin.org` zone block serves egress records via the kuadrant plugin. The `.` block forwards all other queries (including `*.svc.cluster.local`) to the cluster DNS, so pods can still resolve cluster-internal services.
+
+Add one zone block per external hostname that should be routed through the egress gateway.
+
+### Gateway Listener Hostname
+
+For this approach, the egress gateway listener must specify the external hostname. DNSPolicy reads listener hostnames to create DNS records:
+
+```yaml
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: kuadrant-egressgateway
+  namespace: gateway-system
+spec:
+  gatewayClassName: istio
+  listeners:
+    - name: httpbin
+      port: 80
+      protocol: HTTP
+      hostname: httpbin.org
+      allowedRoutes:
+        namespaces:
+          from: All
+```
+
+Add one listener per external hostname that should be routed through the egress gateway.
+
+### Provider Secret and DNSPolicy
+
+Create a CoreDNS provider secret with the external hostname as the zone:
+
+```bash
+kubectl create secret generic dns-provider-credentials-coredns \
+  --namespace=gateway-system \
+  --type=kuadrant.io/coredns \
+  --from-literal=ZONES="httpbin.org"
+```
+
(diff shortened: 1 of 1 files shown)


doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,516 @@
+# DNS-Based Routing to an Egress Gateway
+
+This guide explains how to route pod traffic through an Istio egress gateway to external services using DNS configuration. It presents a recommended approach and an alternative for different trade-offs.
+
+> **Note:** This guide is part of the [Egress Gateway epic](https://github.com/Kuadrant/architecture/issues/145) and will evolve as related work progresses — including the [egress gateway test environment](https://github.com/Kuadrant/kuadrant-operator/issues/1799), [credential injection](https://github.com/Kuadrant/kuadrant-operator/issues/1800), and [end-to-end documentation](https://github.com/Kuadrant/kuadrant-operator/issues/1804).
```
mikenairn: I think this can be removed

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,516 @@
...
+
+**Related:**
+- [RFC 0016 — Egress Gateway Support](https://github.com/Kuadrant/architecture/blob/main/rfcs/0016-egress-gateway.md)
+- [Existing guide: Using Gateway API with APIs outside the cluster](../misc/external-api.md)
```
mikenairn: IMO these re not needed here. The existing guide one is just confusing, it's similar in the sense its using the same istio resources, but nothing to do with egress traffic.

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,516 @@
...
+
+## Prerequisites
```
mikenairn, line 5: IMO this should have a similar prerequisite to the first one [here](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/misc/external-api.md#pre-requisites) 

`Kuadrant and Gateway API installed (with Istio as the gateway provider)`

Anything beyond that should be explained in the doc itself.

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,516 @@
...
+When Istio processes the Gateway, it creates a Deployment and a Service. The examples in this guide use:
+
+| Resource | Value |
+|----------|-------|
+| Gateway namespace | `gateway-system` |
+| Gateway name | `egress-gateway` |
+| Istio-created Service | `egress-gateway-istio.gateway-system.svc.cluster.local` |
+| External service | `httpbin.org` |
+
+> For egress gateway environment setup, see [Egress Gateway Test Environment](https://github.com/Kuadrant/kuadrant-operator/issues/1799).
```
mikenairn: Why are we linking to an issue here? Getting a cluster to the state that it has a Gateway configured as an istio egress gateway should be part of this guide i would have thought, no? Similar to how [this one](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/user-guides/misc/external-api.md) walks you through the creation of the `Gateway`, `ServiceEntry` and `DestinationRule`, which it looks like you are doing below.

doc/user-guides/egress/egress-gateway.md
```diff
@@ -0,0 +1,516 @@
...
+metadata:
+  name: httpbin-external
+  namespace: gateway-system
+spec:
+  host: httpbin.org
+  trafficPolicy:
+    tls:
+      mode: SIMPLE
+      sni: httpbin.org
+```
```
mikenairn: If we are going to have more "egress" guides, i know we will have at least one other related to Auth, then we could consider making these steps part of a "Configuring an Istio Egress Gateway" and making it a prerequisite  of all "egress" guides. This could explain a bit better (not in great detail) the actual config needed to make it "egress":

  1. Gateway - ClusterIP service type, external hostname listeners                                                                                                                                                                                                                         
  2. ServiceEntry - location: MESH_EXTERNAL, external host
  3. DestinationRule - TLS origination config                                                                                                                                                                                                                                              
  4. HTTPRoute - backendRef.kind: Hostname pointing to ServiceEntry

doc/user-guides/egress/egress-gateway.md
```diff
@@ -0,0 +1,516 @@
...
+
+### HTTPRoute
+
+An HTTPRoute attached to the egress gateway routes traffic to the external service. The exact HTTPRoute configuration depends on the routing approach — see each approach section below for the specific HTTPRoute definition.
+
+Kuadrant policies (AuthPolicy, RateLimitPolicy) attach to the Gateway or HTTPRoute in the same way as for ingress.
+
```
mikenairn: If this is a guide, should we not being applying this is part of the guide so the user can see it working?

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,516 @@
...
+kind: Service
+metadata:
+  name: httpbin-egress
+  namespace: app-namespace
+spec:
+  type: ExternalName
+  externalName: egress-gateway-istio.gateway-system.svc.cluster.local
+```
+
+Applications in `app-namespace` can then use `http://httpbin-egress/get`.
```
mikenairn, line 60: Probably worth noting that this is automatically added to the internal pod dns configuration and so is automatically resolvable from within the pod without the need for any other dns changes.

mikenairn: And, since this is a guide, can we not show that working?

mikenairn (commented):
I kind of stopped half way. I want to check what we are doing here, this doc seems to be part guide part informational (Goes into other various ways you could configure things but doesn't demonstrate it).

If it is a guide I'd expect it to detail exactly what the guide is going to show and all resources being applied to show it working, and some commands to demonstrate it working the way described.

Should this be split in two, guide vs dns configuration?

doc/user-guides/egress/dns-routing.md
maksymvavilov: it is gone now

doc/user-guides/egress/dns-routing.md
maksymvavilov: They are now much smaller and reference a script that will set you up. Is it close enough?

doc/user-guides/egress/dns-routing.md
maksymvavilov: This is addressed - there is a script for that now.

doc/user-guides/egress/egress-gateway.md
maksymvavilov: There is a script now that will deploy all of those. 
This doc is the only one that takes on the burden of explaining what the script deployed and what makes this an "egress". 
I'd say it belongs here: explain what was deployed in the context of what will follow (the dns configuration). 
None of the upcoming docs should care about "egress-specific" things:
- Auth sits in the "egress" folder only because it is useless for ingress, but it never bothers about the direction of the traffic. 
- Ratelim will most likely be the same - will sit in `guides/egress` only because it is useless for ingress. But it should still be "universal" for the direction of the traffic 

This is the only doc that actually cares about the traffic direction, and the only place where I can justify explaining. 
It makes me uncomfortable that to "set up an egress", you will need to go to the `dns-routing.md`. I'll fiddle with names and structure to make it seem more organic here :shrug: it will shift the focus of the doc a bit:
- The new name of the file is "egress-gateway.md"  
- The title changed from "DNS-Based Routing to an Egress Gateway" to "Routing Traffic Through an Egress Gateway"
- The section name changed from "Egress Gateway Setup" to "Egress Gateway Resources"

doc/user-guides/egress/egress-gateway.md
maksymvavilov: I'll show here a route that was deployed by the script and will acknowledge  it in later sections (internal hostname and dnspolicy way)

doc/user-guides/egress/dns-routing.md
maksymvavilov: sure, added!

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,518 @@
...
+```yaml
+spec:
+  dnsPolicy: None
+  dnsConfig:
+    nameservers:
+      - "<kuadrant-coredns-ip>"
+    searches:
+      - <namespace>.svc.cluster.local
+      - svc.cluster.local
+      - cluster.local
```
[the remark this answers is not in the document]
maksymvavilov: We added a note below the YAML telling users to check `/etc/resolv.conf` if their cluster uses a custom domain. Keeping `cluster.local` in the YAML itself since it's the default for the vast majority of clusters and makes the example copy-pasteable. Using placeholders like `<cluster-domain>` would hurt usability for the common case.

doc/user-guides/egress/egress-gateway.md
maksymvavilov: Actually, after sitting on it a bit and after considering the ratelimmit policy, I think I'll split the setup from the DNS routing. Otherwise, it all looks awkward  (what you were suggesting to begin with). I feel like I went in circles here a few too many times

maksymvavilov:
## Egress documentation restructure

This PR has been updated as part of the egress docs restructure ([#1896](https://github.com/Kuadrant/kuadrant-operator/pull/1896)).

**What changed:** `egress-gateway.md` renamed to `dns-routing.md`. Setup, resource YAML, and policy sections moved to the new `egress-gateway.md` in #1896. This doc now focuses purely on DNS routing approaches (internal hostname, CoreDNS, alternatives).

**Merge order:** #1896 first, then this PR, then #1893.

maksymvavilov:
@coderabbitai review

doc/user-guides/egress/dns-routing.md
mikenairn: Should "For egress gateway setup and Kuadrant policies, see the [Egress Gateway Setup](egress-gateway.md) guide." Not just be in this Prerequisites list instead of in the guide summary above?

Also, is it actually necessary to go through all that before you can start this guide, i assume its meant to link to https://github.com/Kuadrant/kuadrant-operator/blob/0d240d43162899f264352d0cbb2e0bf1260022b7/doc/user-guides/egress/egress-gateway.md which talks about rate limiting

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,332 @@
...
+| Resource | Value |
+|----------|-------|
+| Gateway namespace | `gateway-system` |
+| Gateway name | `kuadrant-egressgateway` |
+| Istio-created Service | `kuadrant-egressgateway-istio.gateway-system.svc.cluster.local` |
+| External service | `httpbin.org` |
+
+---
+
+## Recommended: Internal Hostname
```
mikenairn, line 20: Would whether its recommend or not  be determined by what the user is trying to do. Purpose of this doc is to show options right?

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,332 @@
...
+spec:
+  type: ExternalName
+  externalName: kuadrant-egressgateway-istio.gateway-system.svc.cluster.local
+```
+
+Applications in `app-namespace` can then use `http://httpbin-egress/get`.
+
+### HTTPRoute with Host Rewrite
+
+This approach replaces the HTTPRoute deployed by the setup script. Instead of matching the external hostname, it matches the gateway's internal service name and rewrites the `Host` header to the real external hostname:
```
mikenairn, line 64: Mentions a script that is never talked about before here. Maybe the prerequisite for all these guides should just state that you are expected to have ran that script instead of linking off to another guide?

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,332 @@
...
+}
+```
+
+- `Host: httpbin.org` - URLRewrite correctly rewrote the host header
+- `X-Envoy-Original-Host` - preserves the original internal hostname
+- `url: https://httpbin.org/get` - TLS origination happened (app sent HTTP, gateway upgraded to HTTPS)
+
+---
+
+## Alternative: Pod DNS with Kuadrant CoreDNS
```
mikenairn, line 131: Rather than recommend/alternative, should we not just list them as options depending on whether you want/need to work with internal or external hostnames?

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,332 @@
...
+spec:
+  dnsPolicy: None
+  dnsConfig:
+    nameservers:
+      - "<kuadrant-coredns-ip>"
+    searches:
+      - <namespace>.svc.cluster.local
+      - svc.cluster.local
+      - cluster.local
+```
```
mikenairn, line 254: Hmm, for this to work are you not essentially telling it to use this nameserver for **_all_** pod dns? If so, that means it would need to handle all internal dns, kuebernetes plugin etc.. no?

doc/user-guides/egress/dns-routing.md
```diff
@@ -0,0 +1,332 @@
...
+  - ip: "<egress-gateway-ip>"
+    hostnames:
+    - "httpbin.org"
+```
+
+This is simple and works on any platform, but the gateway IP is hardcoded in every pod spec. If the gateway service is recreated and gets a new IP, all pods need redeployment. The kuadrant CoreDNS approach avoids this problem since DNSRecord updates propagate automatically.
+
+### Sidecar-Based Routing
+
+If workloads are already enrolled in the Istio service mesh with sidecar injection, traffic can be routed through the egress gateway using a VirtualService with the `mesh` gateway. The sidecar proxy intercepts outbound connections and routes them to the egress gateway transparently - no DNS changes needed. This is a native Istio pattern but requires mesh membership, which adds resource overhead per pod and conflicts with the [RFC 0016](https://github.com/Kuadrant/architecture/blob/main/rfcs/0016-egress-gateway.md) goal of not requiring a service mesh. Consider this option only if your workloads already use sidecar injection.
```
mikenairn, line 325: "and conflicts with the [RFC 0016](https://github.com/Kuadrant/architecture/blob/main/rfcs/0016-egress-gateway.md) goal of not requiring a service mesh" seems irrelevant in this context.

doc/user-guides/egress/dns-routing.md
maksymvavilov: Yes, so all DNS for this pod goes through our instance. 
```text
httpbin.org {
    kuadrant
}
. {
    forward . <cluster-dns-ip>
}
```
This is the Corefile that is being recommended

doc/user-guides/egress/dns-routing.md
maksymvavilov: Removed this as a hard prerequisite.

doc/user-guides/egress/dns-routing.md
maksymvavilov: Now talks about the guide instead of the script

doc/user-guides/egress/dns-routing.md
maksymvavilov: Removed RFC mention

mikenairn (approved):

andrewdavidmackenzie:
The Egress Gateway Setup link in https://github.com/maksymvavilov/kuadrant-operator/blob/8b305eacf2642aa01454f92a0c6dba054ebab316/doc/user-guides/egress/dns-routing.md gave me a 404

PR #1858, opened 2026-04-01
ci: adjust nightlies cron to build earlier

averevki:
With the change to [summer time](https://en.wikipedia.org/wiki/Daylight_saving_time) a couple of days ago, and also, probably, because of the downtime issues happening with github currently, nightly kuadrant builds are now getting to the quay registry as late as 5:30 AM. With QE also having nightly pipelines and tasks that utilize nightly kuadrant image being built, these pipelines are mistimed now, and are being executed before image manages to get uploaded to quay.io. 

With this PR, I'm making nightlies build a bit earlier in the night, so we can have a working build as earlier as possible for any other existing and future nightly tasks.

## Summary by CodeRabbit

* **Chores**
  * Adjusted the nightly build schedule to run at 01:07 UTC (updated accompanying local time notes for CET/CEST).

### DIFF
--- a/.github/workflows/build-images-nightly.yaml
+++ b/.github/workflows/build-images-nightly.yaml
@@ -2,7 +2,10 @@ name: Build nightly images
 
 on:
   schedule:
-    - cron: "0 2 * * *"
+    # Runs at 01:07 UTC
+    # Local Winter (CET): 02:07 AM
+    # Local Summer (CEST): 03:07 AM
+    - cron: "7 1 * * *"
   workflow_dispatch: {}
 
 jobs:


averevki:
I've changed the building time to `7 1 * * *` now, after the problem discussion with @eguzki. So it will be scheduled to build only a 1 hour earlier. That's because usually, nightly builds in authorino and limitador operators, have enough time to build their images within 30 minutes, so 2 hours window between dependencies building start time and kuadrant-operator nightly workflow start time might not be needed.

An additional 7 minutes is a workaround for the GitHub runners delay when GitHub workflows are executed at peak hours during the night. Currently there is an hour long delay from github on kuadrant-operator nightly builds. The goal is for this workflow to be executed closer to its actual scheduled time by setting the workflow schedule to run shortly after the peak github traffic hour.

PR #1865, opened 2026-04-07
Add Istio egress gateway local dev environment

### LINKED ISSUE #1799: [Egress] Egress Gateway Test Environment
## Task 1: Egress Gateway Test Environment

**Depends on:** https://github.com/Kuadrant/architecture/issues/147 (Investigate Istio Egress Gateway Support)
**Blocks:** Tasks 2, 3, 4

## Description

Set up a reproducible Istio egress gateway environment for validation work. This is the foundation for all subsequent tasks.

## Acceptance Criteria

1. Deploy Istio with egress gateway enabled on a test cluster
2. Create ServiceEntry and DestinationRule for an external HTTP service (e.g., httpbin.org or a mock external API)
3. Create Gateway and HTTPRoute resources for the egress path
4. Configure TLS origination via DestinationRule
5. Verify basic egress connectivity: workload inside cluster can reach the external service through the egress gateway
6. Document the environment setup steps for reproducibility

maksymvavilov:
## Summary

- Add kustomize overlay in `config/dependencies/istio/egress-gateway/` with Gateway, ServiceEntry, DestinationRule, and HTTPRoute for httpbin.org
- Add `deploy-istio-egress-gateway` Makefile target to deploy egress resources on an existing Istio cluster
- Add `hack/setup-egress.sh` — a standalone script that deploys the full egress environment including a dev Vault instance with Kubernetes auth for per-workload credential injection
- Add test-client pod manifest in `examples/egress-gateway/test-client.yaml` for verifying egress connectivity
- Document the egress gateway setup in `doc/overviews/development.md`

Closes #1799

## Test plan

- [x] `make local-setup GATEWAYAPI_PROVIDER=istio` + `make deploy-istio-egress-gateway` deploys egress resources
- [x] Egress gateway reaches `Programmed` condition
- [x] test-client pod in `egress-test` namespace can reach httpbin.org through the egress gateway
- [x] `hack/setup-egress.sh` deploys full environment (egress gateway + Vault + test client) and verifies connectivity
- [x] `hack/setup-egress.sh cleanup` removes all egress resources

## Summary by CodeRabbit

* **New Features**
  * Added Istio egress gateway configuration for secure external service routing with TLS origination.
  * Included example test client for verifying egress connectivity.
  * Provided automated setup and configuration tooling for egress environments.

* **Documentation**
  * Updated development guide with egress gateway deployment procedures and verification steps.

### DIFF
--- a/config/dependencies/istio/egress-gateway/gateway.yaml
+++ b/config/dependencies/istio/egress-gateway/gateway.yaml
@@ -1,0 +1,16 @@
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: kuadrant-egressgateway
+  annotations:
+    networking.istio.io/service-type: ClusterIP
+spec:
+  gatewayClassName: istio
+  listeners:
+    - name: http
+      port: 80
+      protocol: HTTP
+      allowedRoutes:
+        namespaces:
+          from: All
--- a/config/dependencies/istio/egress-gateway/httproute.yaml
+++ b/config/dependencies/istio/egress-gateway/httproute.yaml
@@ -1,0 +1,21 @@
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: httpbin-external
+spec:
+  parentRefs:
+    - name: kuadrant-egressgateway
+      namespace: gateway-system
+  hostnames:
+    - httpbin.org
+  rules:
+    - filters:
+        - type: URLRewrite
+          urlRewrite:
+            hostname: httpbin.org
+      backendRefs:
+        - group: networking.istio.io
+          kind: Hostname
+          name: httpbin.org
+          port: 443
--- a/examples/egress-gateway/test-client.yaml
+++ b/examples/egress-gateway/test-client.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: egress-test
+---
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-client
+  namespace: egress-test
+spec:
+  containers:
+    - name: curl
+      image: curlimages/curl:latest
+      command: ["sleep", "infinity"]
+  restartPolicy: Never
--- a/hack/setup-egress.sh
+++ b/hack/setup-egress.sh
@@ -1,0 +1,158 @@
+#!/usr/bin/env bash
+# Setup an Istio egress gateway environment for Kuadrant.
+#
+# Deploys an egress Gateway, ServiceEntry, DestinationRule, HTTPRoute for
+# httpbin.org, a test-client pod, and a dev Vault instance with Kubernetes
+# auth method configured for per-workload credential injection.
+#
+# Prerequisites:
+#   - Kubernetes cluster with Kuadrant and Istio installed
+#   - kubectl and helm configured to access the cluster
+#
+# Usage:
+#   # From a cloned repo:
+#   ./hack/setup-egress.sh
+#
+#   # Without cloning:
+#   curl -sL https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/refs/heads/main/hack/setup-egress.sh | bash
+#
+#   # Cleanup:
+#   ./hack/setup-egress.sh cleanup
+#   # or:
+#   curl -sL https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/refs/heads/main/hack/setup-egress.sh | bash -s cleanup
+
+set -euo pipefail
+
+EGRESS_NS="gateway-system"
+KUADRANT_NS="kuadrant-system"
+VAULT_NS="vault"
+BASE_URL="https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/refs/heads/main"
+
+# Use local files if running from within the repo, otherwise fetch from GitHub
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")" 2>/dev/null && pwd)"
+REPO_DIR="$SCRIPT_DIR/.."
+if [ -f "$REPO_DIR/config/dependencies/istio/egress-gateway/gateway.yaml" ]; then
+    src() { echo "$REPO_DIR/$1"; }
+else
+    src() { echo "$BASE_URL/$1"; }
+fi
+
+info()  { echo "[INFO] $*"; }
+error() { echo "[ERROR] $*" >&2; }
+
+# ── Cleanup ──────────────────────────────────────────────────────────
+if [ "${1:-}" = "cleanup" ]; then
+    info "Cleaning up egress gateway resources..."
+    kubectl delete httproute httpbin-external -n "$EGRESS_NS" --ignore-not-found
+    kubectl delete destinationrule httpbin-external -n "$EGRESS_NS" --ignore-not-found
+    kubectl delete serviceentry httpbin-external -n "$EGRESS_NS" --ignore-not-found
+    kubectl delete gateway kuadrant-egressgateway -n "$EGRESS_NS" --ignore-not-found
+    # No Vault secrets to clean up — Vault K8s auth uses workload SA tokens directly
+    kubectl delete pod test-client -n egress-test --ignore-not-found
+    kubectl delete namespace egress-test --ignore-not-found
+    helm uninstall vault -n "$VAULT_NS" 2>/dev/null || true
+    kubectl delete namespace "$VAULT_NS" --ignore-not-found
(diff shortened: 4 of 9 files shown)


make/istio.mk
```diff
@@ -54,3 +54,9 @@ istio-install:
 .PHONY: deploy-istio-gateway
 deploy-istio-gateway: kustomize ## Deploy Gateway API gateway with gatewayclass set to Istio
 	$(KUSTOMIZE) build config/dependencies/istio/gateway | kubectl apply -f -
+
+.PHONY: deploy-istio-egress-gateway
+deploy-istio-egress-gateway: kustomize ## Deploy Istio egress gateway with ServiceEntry, DestinationRule and HTTPRoute for external service
+	$(KUSTOMIZE) build config/dependencies/istio/egress-gateway | kubectl apply -f -
+	kubectl apply -f config/dependencies/istio/egress-gateway/test-client.yaml
```
averevki: should it also wait for the pod to became ready?
```suggestion
	kubectl apply -f config/dependencies/istio/egress-gateway/test-client.yaml
	kubectl wait --timeout=2m -n egress-test pod/test-client --for=condition=Ready
```

config/dependencies/istio/egress-gateway/httproute.yaml
```diff
@@ -0,0 +1,21 @@
...
+  parentRefs:
+    - name: kuadrant-egressgateway
+      namespace: gateway-system
+  hostnames:
+    - httpbin.org
+  rules:
+    - filters:
+        - type: URLRewrite
+          urlRewrite:
+            hostname: httpbin.org
```
averevki, line 16: You think it would be fine if httpbin.org won't be available during some outage? This could be an issue if this make target used in the CI or during testing. But otherwise using real external service should be ok, testing a real-world scenario

maksymvavilov: It should be fine - I'd say if we get a response with "oops, we are out", it means that we still made a request as we were supposed to. We won't be able to inspect the headers, but the rest of the test should work regardless of what reply from the service is 
🤔 
what do you think?

averevki: yeah that makes sense :+1:

make/development-environments.mk
```diff
@@ -111,6 +111,17 @@ local-setup: kind ## Run local Kubernetes cluster and deploy kuadrant operator (
...
+local-setup-egress: ## Run local Kubernetes cluster with Istio egress gateway environment
+	$(MAKE) local-setup GATEWAYAPI_PROVIDER=istio
+	$(MAKE) deploy-istio-egress-gateway
+	@echo
+	@echo "Egress gateway environment is ready."
+	@echo "Test connectivity:"
+	@echo "  EGRESS_IP=\$$(kubectl get gtw kuadrant-egressgateway -n gateway-system -o jsonpath='{.status.addresses[0].value}')"
+	@echo "  kubectl exec test-client -n egress-test -- curl -s -H \"Host: httpbin.org\" http://\$$EGRESS_IP/get"
+	@echo
+
```
mikenairn: Any reason we need another make target for this? Would it not make more sense to just add it to the [istio-env-setup](https://github.com/Kuadrant/kuadrant-operator/blob/main/make/development-environments.mk#L154) make target.

examples/egress-gateway/test-client.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: Pod
+metadata:
+  name: test-client
+  namespace: egress-test
+spec:
+  containers:
+    - name: curl
+      image: curlimages/curl:latest
+      command: ["sleep", "infinity"]
+  restartPolicy: Never
```
mikenairn, line 17: What is this? Does it belong here?

make/development-environments.mk
maksymvavilov: I can make an argument for it: 
- Embedding it will deploy both every time, plus it makes the make chain even more tricky to follow. 
- A separate target is more visible, plus we will see it with the `make help` 
- Is a separate use case, hence a separate make target 

But, if you feel strongly about it, I don't mind changing

examples/egress-gateway/test-client.yaml
maksymvavilov: This is what I mentioned the other day in the call. We don't deploy any "workload" pods as part of the setup before this (more reason to have a separate target). 
I can argue that this is not an "example" or a "workload" but an infrastructure piece - without this pod, there is no source for the requests for the egress use case. 

Does it belong here? I don't know, to be honest, but this is the best idea I've had. I'll be able to sleep at night if we put it into the examples and output in the terminal after the setup: "now you need a curl pod - it is in the examples and here is a kubectl apply command for deploying it". 
Would you prefer it this way? I don't really have a preference here and will go with whatever seems more reasonable

averevki (dismissed):
LGTM

make/development-environments.mk
mikenairn: I still don't see the need for an explicit "local-setup" target for this, but i guess it's not a big issue it being here.

My point is you could just as easily apply the other stuff to a cluster created with "local-setup" instead of telling people to create an entirely new cluster(We don't have other special "local-setups" that I'm aware of and it's not something we should encourage):

```
make local-setup deploy-istio-egress-gateway onfigure-authorino-ca-trust
```

The point about adding it to [istio-env-setup](https://github.com/Kuadrant/kuadrant-operator/blob/main/make/development-environments.mk#L154) is also no longer valid since you have changed it since to alter authorino config with configure-authorino-ca-trust

examples/egress-gateway/test-client.yaml
mikenairn: OK. I mean the "source for the requests for the egress use case" are the users service(s) pods, not this. This is just a convenience to test the egress with curl, presumably in examples or guides.

make/development-environments.mk
maksymvavilov: Fair. No need for the whole new "local setup". I'll make people chain the make targets for egress.

examples/egress-gateway/test-client.yaml
maksymvavilov: moved into examples, and the make targets are no longer deploying it. The script does, however, deploy it

mikenairn:
The resources, make target and docs/readme around egress look fine to me. I would however be less sure about what `./hack/setup-egress.sh` is doing. I guess it's for the credential injection docs, so maybe @guicassolato and/or @averevki could take a look at it.

Re. testing of this, do we intend to use this same script to setup vault in the testsuite runs?

hack/setup-egress.sh
averevki, line 1: Is the main purpose of this script to help startup the egress configuration for egress docs? I'm curious why you didn't just put in-lined kubectl apply commands with these resources directly into the docs, instead of showing resources this script creates there? I think it will be more clear that way, is there anything I'm missing? I think I would prefer coping few apply commands from the doc over running some external script locally

maksymvavilov: It was like that initially. This changed it https://github.com/Kuadrant/kuadrant-operator/pull/1893#discussion_r3086251045

averevki: I forgot about this one, weren't resources also applied from a make target initially? I'd avoid commands piping to bash in docs `curl -sL https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/main/path/to/egress/example/script.sh | bash -s`, but I guess it should still be compliant if script is from a Kuadrant org project. I think other docs already use some make targets from the repo, which are basically the same as bash script

averevki (dismissed):
LGTM otherwise, lets leave using scripts in docs discussion for later 👍

hack/setup-egress.sh
mikenairn: I can see all POV here, I'm unsure what the best approach is tbh, we have a mix of ideas none of which are really wrong.

Lets merge as is and we can obviously iterate as required.

mikenairn (dismissed):

config/dependencies/istio/egress-gateway/gateway.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: kuadrant-egressgateway
+spec:
```
mikenairn, line 8: Should this not have the service type annotation? https://github.com/Kuadrant/kuadrant-operator/pull/1896/files#diff-b2949fabcdb21b16130055229e2c17db56d11f0bc052d6c8028899b79ca3fcc8R43

maksymvavilov: It is there. Also squashed all commits

mikenairn (approved):

PR #1877, opened 2026-04-09
fix: restore TLSPolicy reconciliation on issuer readiness changes

### LINKED ISSUE #1876: TLSPolicy issuer readiness changes do not trigger status reconciliation
### Problem

`TLSPolicyStatusUpdater` subscribes to `Issuer` and `ClusterIssuer` events so a policy can move between enforced and unenforced as cert-manager issuer readiness changes.

However, the cert-manager watchers registered in `internal/controller/state_of_the_world.go` currently compare the old issuer status to itself:

```go
oldStatus := e.ObjectOld.GetStatus()
newStatus := e.ObjectOld.GetStatus()
return !reflect.DeepEqual(oldStatus, newStatus)
```

That makes the update predicate always return `false`, so status-only updates on `Issuer` and `ClusterIssuer` never reach the controller.

### Impact

- `TLSPolicy` does not react when an `Issuer` or `ClusterIssuer` becomes ready or not ready after the policy already exists.
- The policy can remain stuck with a stale `Enforced` condition until some unrelated event happens.
- This breaks the readiness transition handling that the TLS status updater is explicitly wired to support.

### Expected Behaviour

The issuer watchers should compare `ObjectOld` to `ObjectNew` and trigger reconciliation whenever cert-manager status changes:

```go
oldStatus := e.ObjectOld.GetStatus()
newStatus := e.ObjectNew.GetStatus()
return !reflect.DeepEqual(oldStatus, newStatus)
```

This should be covered by a regression test so future refactors do not silently disable the watcher again.

### Additional Context

The original TLSPolicy issuer-status watch support was added in #715, so this is a regression in the current `main` branch rather than missing functionality.

officialasishkumar:
# Description
Closes: #1876

- compare cert-manager `Issuer` and `ClusterIssuer` update predicates against `ObjectNew` instead of comparing `ObjectOld` to itself
- restore TLSPolicy status reconciliation when issuer readiness changes after the policy already exists
- add regression coverage for both watcher predicates so the status-driven watch path stays protected

# Verification
- `go vet ./...`
- `go test -tags unit ./internal/controller -run 'Test(GetKuadrant|IssuerStatusChangedPredicate|ClusterIssuerStatusChangedPredicate)$'`
- `go test -race -tags unit ./api/... ./internal/... ./pkg/...`

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Fixed an issue where status updates for Issuer and ClusterIssuer were not being reliably detected, ensuring reconciliation triggers correctly when readiness status changes.

* **Tests**
  * Added unit tests covering status-change detection for both Issuer and ClusterIssuer, verifying reconciliation is triggered on readiness transitions and ignored when status is unchanged.

### DIFF
--- a/internal/controller/state_of_the_world.go
+++ b/internal/controller/state_of_the_world.go
@@ -761,25 +761,13 @@ func certManagerControllerOpts() []controller.ControllerOption {
 			&certmanagerv1.Issuer{},
 			CertManagerIssuersResource,
 			metav1.NamespaceAll,
-			controller.WithPredicates(ctrlruntimepredicate.TypedFuncs[*certmanagerv1.Issuer]{
-				UpdateFunc: func(e event.TypedUpdateEvent[*certmanagerv1.Issuer]) bool {
-					oldStatus := e.ObjectOld.GetStatus()
-					newStatus := e.ObjectOld.GetStatus()
-					return !reflect.DeepEqual(oldStatus, newStatus)
-				},
-			})),
+			controller.WithPredicates(issuerStatusChangedPredicate())),
 		),
 		controller.WithRunnable("clusterissuers watcher", controller.Watch(
 			&certmanagerv1.ClusterIssuer{},
 			CertMangerClusterIssuersResource,
 			metav1.NamespaceAll,
-			controller.WithPredicates(ctrlruntimepredicate.TypedFuncs[*certmanagerv1.ClusterIssuer]{
-				UpdateFunc: func(e event.TypedUpdateEvent[*certmanagerv1.ClusterIssuer]) bool {
-					oldStatus := e.ObjectOld.GetStatus()
-					newStatus := e.ObjectOld.GetStatus()
-					return !reflect.DeepEqual(oldStatus, newStatus)
-				},
-			})),
+			controller.WithPredicates(clusterIssuerStatusChangedPredicate())),
 		),
 		controller.WithObjectKinds(
 			CertManagerCertificateKind,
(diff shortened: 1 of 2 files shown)


internal/controller/state_of_the_world.go
```diff
@@ -761,25 +761,13 @@ func certManagerControllerOpts() []controller.ControllerOption {
 			&certmanagerv1.Issuer{},
 			CertManagerIssuersResource,
 			metav1.NamespaceAll,
-			controller.WithPredicates(ctrlruntimepredicate.TypedFuncs[*certmanagerv1.Issuer]{
-				UpdateFunc: func(e event.TypedUpdateEvent[*certmanagerv1.Issuer]) bool {
-					oldStatus := e.ObjectOld.GetStatus()
-					newStatus := e.ObjectOld.GetStatus()
```
maksymvavilov, line 767: @KevFan did we have this (old -> old) on purpose?

internal/controller/state_of_the_world.go
```diff
@@ -794,6 +782,26 @@ func certManagerControllerOpts() []controller.ControllerOption {
 	}
 }
 
+func statusChanged[S any, T interface{ GetStatus() S }](oldObj, newObj T) bool {
+	return !reflect.DeepEqual(oldObj.GetStatus(), newObj.GetStatus())
+}
```
maksymvavilov: While the other two seem necessary for testing purposes, why not replace  `tatusChanged(e.ObjectOld, e.ObjectNew)` with `!reflect.DeepEqual(oldObj.GetStatus(), newObj.GetStatus())` and get rid of this extra function?

maksymvavilov (changes requested):
I believe you might need to sign off here to have DCO passing

internal/controller/state_of_the_world.go
officialasishkumar: Removed the helper and inlined the status comparison in 2127c44. DCO is passing after the signed-off amend.

thomasmaas (approved):
LGTM — clear copy-paste bug (old vs old), clean fix, good regression tests. The predicate duplication across Issuer/ClusterIssuer is fine.

maksymvavilov (approved):
Nicw work! Thank you @officialasishkumar

Issue #1880, opened 2026-04-10
Add integration tests for sectionName targeting in policies


mikenairn:
## Summary

Add automated integration test coverage for `sectionName` targeting in AuthPolicy and RateLimitPolicy with both HTTPRoute and GRPCRoute resources.

## Context

Currently, integration tests for policies only target entire routes. The `targetRef.sectionName` field allows policies to target specific rules within a route (e.g., `rule-1`, `rule-2`), enabling method-specific or path-specific policy configuration. While manual test examples exist, there is no automated integration test coverage for this functionality.

## Motivation

- Prevent regressions in sectionName functionality
- Validate that section-targeted policies create correct AuthConfigs/Limitador limits
- Ensure method/path-specific policy enforcement works correctly
- Improve test coverage for Gateway API Policy Attachment patterns

## Proposed Solution

Add integration tests covering these scenarios:

### AuthPolicy + GRPCRoute
- [ ] Target specific rule via `sectionName: rule-2`
- [ ] Verify method-specific AuthConfigs are created
- [ ] Test that authentication rules apply only to targeted methods
- [ ] Test authorization rules per-section

### AuthPolicy + HTTPRoute  
- [ ] Target specific rule via `sectionName: rule-2`
- [ ] Verify path-specific AuthConfigs are created
- [ ] Test that authentication rules apply only to targeted paths

### RateLimitPolicy + GRPCRoute
- [ ] Target specific rule via `sectionName: rule-2`
- [ ] Verify Limitador limits are created per-section
- [ ] Test method-specific rate limiting enforcement

### RateLimitPolicy + HTTPRoute
- [ ] Target specific rule via `sectionName: rule-2`
- [ ] Verify Limitador limits are created per-section
- [ ] Test path-specific rate limiting enforcement

## Test Locations

- `tests/common/authpolicy/authpolicy_grpcroute_test.go` - add GRPCRoute sectionName tests
- `tests/common/authpolicy/authpolicy_controller_test.go` - add HTTPRoute sectionName tests
- `tests/common/ratelimitpolicy/ratelimitpolicy_grpcroute_test.go` - add GRPCRoute sectionName tests
- `tests/common/ratelimitpolicy/ratelimitpolicy_controller_test.go` - add HTTPRoute sectionName tests

## Reference

Manual test examples exist but are not committed:
- `examples/grpc-backend/authpolicy/grpcroute-sectionname.yaml`
[...]

PR #1855, opened 2026-03-31
Add GRPCRoute support to AuthPolicy

mikenairn:
## Summary

Enables AuthPolicy to target GRPCRoute resources for authentication and authorisation of gRPC services, extending Kuadrant's policy attachment capabilities beyond HTTPRoute.

**Design Proposal:** [RFC-0017: GRPCRoute Support](https://github.com/Kuadrant/architecture/blob/main/rfcs/0017-grpcroute-support.md)

### Core Functionality

- **GRPCRoute support for AuthPolicy**: Users can now attach AuthPolicy to GRPCRoute resources using the same `targetRef` mechanism as HTTPRoute
- **Type-first route detection**: Introduced `ParseTopologyPath()` with a discriminated union pattern (`RouteType`) to handle both HTTP and gRPC routes uniformly
- **Unified reconciliation**: All AuthPolicy reconcilers (authconfigs, status updater, wasm, Istio, EnvoyGateway) now use consistent type-first route detection

### Implementation Details

1. **New API Types & Utilities** (`internal/policymachinery/utils.go`):
   - `RouteType` enum (HTTP/gRPC)
   - `ParsedTopologyPath` struct with helper methods for type-agnostic access
   - `ParseTopologyPath()` function for unified topology path parsing
   - `DetectRouteType()` for automatic route type detection
   - Generic `objectsInRequestPath()` using accessor pattern

2. **CRD Changes**:
   - Updated AuthPolicy CRD to accept `GRPCRoute` in `targetRef.kind`
   - Regenerated bundle manifests and Helm charts

3. **Reconciler Updates**:
   - `authconfigs_reconciler.go`: Uses ParseTopologyPath, handles both HTTP and gRPC AuthConfigs
   - `auth_policy_status_updater.go`: Tracks both HTTPRouteRule and GRPCRouteRule separately
   - `istio_extension_reconciler.go`: Type-first route detection for wasm config
   - `envoy_gateway_extension_reconciler.go`: Type-first route detection for wasm config
   - `istio_auth_cluster_reconciler.go`: Filters gateways supporting both route types
   - `envoy_gateway_auth_cluster_reconciler.go`: Filters gateways supporting both route types

4. **GRPCRoute Utilities** (`internal/gatewayapi/`):
   - `IsGRPCRouteAccepted()` - checks GRPCRoute acceptance status
   - `IsGRPCRouteReady()` - verifies GRPCRoute is ready for policy attachment
[...]

Issue #1827, opened 2026-03-16
[GRPCRoute] Examples and documentation for GRPCRoute


mikenairn:
## Task 8: Examples & Documentation

**Parent epic:** Kuadrant/architecture#156  
**Blocked by:** #1823 (Task 2 - gRPC Backend Selection), #1826 (Task 7 - GRPCRoute Policy Discoverability Reconciler)  
**RFC:** [0017-grpcroute-support.md](https://github.com/Kuadrant/architecture/blob/main/rfcs/0017-grpcroute-support.md)

## What

Create example manifests and user guide documentation for gRPC with all supported policies, including grpcurl verification commands.

## Why

Users need working examples and clear documentation to adopt GRPCRoute support. Examples should cover all supported policies and demonstrate typical usage patterns.

## Done

- [ ] Create example GRPCRoute manifest (`examples/grpcroute/grpcroute.yaml`)
- [ ] Create example AuthPolicy targeting GRPCRoute (`examples/grpcroute/authpolicy.yaml`)
- [ ] Create example RateLimitPolicy targeting GRPCRoute (`examples/grpcroute/ratelimitpolicy.yaml`)
- [ ] Create example OIDCPolicy targeting GRPCRoute (`examples/grpcroute/oidcpolicy.yaml`)
- [ ] Create example PlanPolicy targeting GRPCRoute (`examples/grpcroute/planpolicy.yaml`)
- [ ] Create example TelemetryPolicy targeting GRPCRoute (`examples/grpcroute/telemetrypolicy.yaml`)
- [ ] Add README with grpcurl verification commands
- [ ] Update main Kuadrant documentation with GRPCRoute user guide
- [ ] Add note about DNSPolicy/TLSPolicy working on Gateways (not route-specific)

## Notes

**Deliverables location:** `examples/grpcroute/` directory

Examples should use the gRPC backend image selected in Task 2. Include grpcurl commands for manual verification of policy enforcement.

**Documentation should cover:**
- How to attach policies to GRPCRoute (same as HTTPRoute, only targetRef changes)
- GRPCRouteMatch to CEL predicate translation
- Method-specific vs service-wide policies
- Mixed HTTPRoute + GRPCRoute on same Gateway

PR #1881, opened 2026-04-10
Add GRPCRoute support to RateLimitPolicy

### LINKED ISSUE #1824: [GRPCRoute] RateLimitPolicy support for GRPCRoute
## Task 5: RateLimitPolicy Support

**Parent epic:** Kuadrant/architecture#156  
**Blocked by:** #1821 (Task 4 - AuthPolicy Support + Data Plane Wiring)  
**RFC:** [0017-grpcroute-support.md](https://github.com/Kuadrant/architecture/blob/main/rfcs/0017-grpcroute-support.md)

## What

Enable RateLimitPolicy to attach to GRPCRoutes with complete data plane integration including ratelimit cluster reconcilers and integration tests with real gRPC traffic.

## Why

RateLimitPolicy is the second data plane policy after AuthPolicy. This task follows the same pattern established in Task 4: predicates → WASM → clusters → integration tests.

## Done

- [ ] Update `RateLimitPolicy` CRD validation to accept GRPCRoute targetRef
- [ ] Update `IstioRateLimitClusterReconciler` to recognize GRPCRoute attachments
- [ ] Update `EnvoyGatewayRateLimitClusterReconciler` to recognize GRPCRoute attachments
- [ ] Unit tests for RateLimitPolicy targeting GRPCRoute
- [ ] Unit tests verifying GRPCRoute→RateLimitPolicy linkage in topology
- [ ] Integration tests for Limitador config generation
- [ ] Integration tests for cluster configuration
- [ ] Integration tests with real gRPC traffic

## Notes

**Key files:**
- `internal/cmd/leveledlogger/state_of_the_world.go` - link function implementation
- `api/v1/ratelimitpolicy_types.go` - CRD validation
- `internal/controller/istio_ratelimit_cluster_reconciler.go` - Istio clusters
- `internal/controller/envoygateway_ratelimit_cluster_reconciler.go` - EnvoyGateway clusters

**Critical:** Without the link function, RateLimitPolicies targeting GRPCRoute will not connect to limit configs in the topology graph, causing policies to appear applied but not enforce.

Predicate generation is already handled by `GRPCRouteMutator` from Task 4. This task focuses on RateLimitPolicy-specific reconcilers.

mikenairn:
## Summary

Enables RateLimitPolicy to target GRPCRoute resources for rate limiting of gRPC services, extending Kuadrant's policy attachment capabilities beyond HTTPRoute.

**Design Proposal:** [RFC-0017: GRPCRoute Support](https://github.com/Kuadrant/architecture/blob/main/rfcs/0017-grpcroute-support.md)

### Core Functionality

- **GRPCRoute support for RateLimitPolicy**: Users can now attach RateLimitPolicy to GRPCRoute resources using the same `targetRef` mechanism as HTTPRoute
- **Type-first route detection**: Uses `ParseTopologyPath()` (introduced in AuthPolicy GRPCRoute support) to handle both HTTP and gRPC routes uniformly
- **Unified reconciliation**: All RateLimitPolicy reconcilers now use consistent type-first route detection

### Implementation Details

1. **CRD Changes**:
   - Updated RateLimitPolicy CRD to accept `GRPCRoute` in `targetRef.kind`
   - Regenerated bundle manifests and Helm charts

2. **Reconciler Updates**:
   - `effective_ratelimit_policies_reconciler.go`: Processes both HTTPRouteRule and GRPCRouteRule
   - `limitador_limits_reconciler.go`: Uses ParseTopologyPath, generates limits for both route types
   - `ratelimit_policy_status_updater.go`: Tracks both HTTPRouteRule and GRPCRouteRule separately
   - `ratelimit_workflow_helpers.go`: Type-first route detection for limit namespace generation
   - `istio_ratelimit_cluster_reconciler.go`: Filters gateways supporting both route types
   - `envoy_gateway_ratelimit_cluster_reconciler.go`: Filters gateways supporting both route types

3. **gRPC Backend Examples** (`examples/grpc-backend/ratelimitpolicy/`):
   - Basic rate limiting example
   - Method-specific rate limits using CEL `when` predicates
   - No-reflection rate limiting example
   - Comprehensive documentation demonstrating all patterns

4. **Integration Tests** (`tests/common/ratelimitpolicy/ratelimitpolicy_grpcroute_test.go`):
   - Basic GRPCRoute policy attachment
   - Gateway-level policies with GRPCRoute overrides
   - Method-based rate limiting patterns
   - Multi-rule GRPCRoute scenarios
   - Policy deletion and cleanup

### Unit Tests
All existing unit tests pass.

### Integration Tests
New integration test suite covering:
- ✅ Basic GRPCRoute with RateLimitPolicy
- ✅ Gateway-level policies with GRPCRoute
- ✅ GRPCRoute method matching patterns
- ✅ Multi-rule GRPCRoute configurations
- ✅ Policy override scenarios
- ✅ Resource cleanup on policy deletion

### For Users

**No breaking changes.** This is a purely additive feature:
- Existing HTTPRoute-based RateLimitPolicies continue to work unchanged
- New capability to attach RateLimitPolicy to GRPCRoute resources

Example usage:
```yaml
apiVersion: kuadrant.io/v1
kind: RateLimitPolicy
metadata:
  name: grpc-ratelimit
spec:
  targetRef:
    kind: GRPCRoute
    name: my-grpc-route
  limits:
    "10rps":
      rates:
      - limit: 10
        window: 1s
```

### For Developers

All RateLimitPolicy reconcilers have been updated to use `ParseTopologyPath()` for consistent route handling across HTTP and gRPC.

## Manual Verification

Example gRPC backend deployment with RateLimitPolicy is available for manual testing:

📖 **[gRPC Backend Examples - RateLimitPolicy Guide](https://github.com/mikenairn/kuadrant-operator/blob/grpcroute_support_ratelimit/examples/grpc-backend/ratelimitpolicy/README.md)**

The guide includes:
- Step-by-step deployment instructions
- Basic rate limiting example
- Method-specific rate limits with CEL `when` predicates
- Manual testing commands with expected outputs

## Related Issues

Resolves https://github.com/Kuadrant/kuadrant-operator/issues/1824

## Follow-up Tasks

- [ ] #1880 - Add integration tests for sectionName targeting in policies
- [ ] #1890 - Propagate ParseTopologyPath errors in rate limit helpers
- [ ] #1891 - Optimize test helper to accept pointer parameter
- [ ] #1916 - Add type safety for rate limit namespace and scope identifiers

## Release Notes

* **New Features**
  * Added support for applying RateLimitPolicy to GRPCRoute resources in addition to HTTPRoute and Gateway targets.
  * Added example configurations demonstrating rate limiting on gRPC routes, including method-specific limits and reflection endpoint handling.

* **Documentation**
  * Added comprehensive guide for configuring RateLimitPolicy on GRPCRoute with testing instructions and best practices.

* **Tests**
  * Extended integration test coverage for RateLimitPolicy with GRPCRoute targeting.

## Summary by CodeRabbit

* **New Features**
  * RateLimitPolicy now accepts GRPCRoute as a valid target kind alongside HTTPRoute and Gateway (validation and CRD updated).

* **Documentation**
  * Added detailed gRPC backend rate-limiting examples, walkthroughs and testing guidance.

* **Tests**
  * New integration test suite exercising RateLimitPolicy with GRPCRoute targets and updated validation expectations.


mikenairn:
@coderabbitai review

mikenairn:
@coderabbitai review

internal/controller/istio_ratelimit_cluster_reconciler.go
```diff
@@ -72,8 +73,11 @@ func (r *IstioRateLimitClusterReconciler) Reconcile(ctx context.Context, _ []con
 	effectiveRateLimitPolicies, rlpOk := state.Load(StateEffectiveRateLimitPolicies)
 	if rlpOk && effectiveRateLimitPolicies != nil {
 		rlpGateways := lo.FilterMap(lo.Values(effectiveRateLimitPolicies.(EffectiveRateLimitPolicies)), func(effectivePolicy EffectiveRateLimitPolicy, _ int) (*machinery.Gateway, bool) {
-			gatewayClass, gateway, _, _, _, _ := kuadrantpolicymachinery.ObjectsInRequestPath(effectivePolicy.Path)
-			return gateway, lo.Contains(istioGatewayControllerNames, gatewayClass.Spec.ControllerName)
+			parsed, err := kuadrantpolicymachinery.ParseTopologyPath(effectivePolicy.Path)
+			if err != nil {
+				return nil, false
+			}
+			return parsed.Gateway, lo.Contains(istioGatewayControllerNames, parsed.GatewayClass.Spec.ControllerName)
```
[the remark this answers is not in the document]
mikenairn: This doesn't look like a bug, but an opportunity to improve error handling. Errors in the wasm code were always ignored before ParseTopologyPath was added. I created a follow on issue for it https://github.com/Kuadrant/kuadrant-operator/issues/1890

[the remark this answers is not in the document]
mikenairn: It will introduce a difference in behaviour between policies still using ObjectsInRequestPath and ones updated to use ParseTopologyPath. It would be better done after all policies have been updated and ObjectsInRequestPath is removed.

examples/grpc-backend/ratelimitpolicy/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+```bash
+kubectl apply -f examples/grpc-backend/authpolicy/kuadrant.yaml
+kubectl apply -f examples/grpc-backend/authpolicy/api-key-secret.yaml
+kubectl apply -f examples/grpc-backend/authpolicy/authpolicy.yaml
+kubectl apply -f examples/grpc-backend/ratelimitpolicy/ratelimitpolicy.yaml
+```
+
+Requests will need to satisfy both authentication requirements and rate limits.
+
```
guicassolato, line 161: Although this is good, WDYT about an AuthPolicy and RateLimitPolicy for an authenticated rate limit example (both in the `examples/grpc-backend/ratelimitpolicy/` directory)?

examples/grpc-backend/ratelimitpolicy/README.md
```diff
@@ -0,0 +1,203 @@
...
+Expected: Only 1 successful `grpcurl` invocation before hitting the rate limit (2 total calls consumed by reflection + method call), then `Code: Unavailable` for subsequent requests.
+
+## Combining with AuthPolicy
+
+You can combine rate limiting with authentication by applying both policies to the same GRPCRoute.
+
+**Note:** You can use either `ratelimitpolicy.yaml` (basic) or `ratelimitpolicy-no-reflection.yaml` (recommended for testing) when combining policies. The example below uses the basic policy:
+
+```bash
+kubectl apply -f examples/grpc-backend/authpolicy/kuadrant.yaml
```
guicassolato, line 154: I guess we want 3 independent examples all starting with applying a Kuadrant CR to a fresh cluster, instead of 3 interdependent steps of a single, longer example, right? Maybe name this one like other two, i.e.: `## Example 3: …`?

examples/grpc-backend/README.md
```diff
@@ -142,3 +142,19 @@ Expected output:
...
+- Method-specific rate limits using CEL expressions
+- Combining rate limiting with authentication
+
+Example:
+
+```bash
+kubectl apply -f examples/grpc-backend/ratelimitpolicy/ratelimitpolicy.yaml -n default
+```
+
+This applies a rate limit of 10 requests per 10 seconds to all traffic on the grpcbin GRPCRoute.
```
guicassolato, line 159:
```suggestion
This applies a rate limit of 10 requests per 10 seconds to all traffic on the grpcbin GRPCRoute (gRPC reflection requests included).
```

internal/controller/effective_ratelimit_policies_reconciler.go
```diff
@@ -59,14 +59,21 @@ func (r *EffectiveRateLimitPolicyReconciler) calculateEffectivePolicies(ctx cont
...
 	effectivePolicies := EffectiveRateLimitPolicies{}
 
+	// Combine HTTP and gRPC route rules
+	allRouteRules := append(httpRouteRules, grpcRouteRules...)
+
 	for _, gatewayClass := range gatewayClasses {
-		for _, httpRouteRule := range httpRouteRules {
-			paths := targetables.Paths(gatewayClass, httpRouteRule) // this may be expensive in clusters with many gateway classes - an alternative is to deep search the topology for httprouterules from each gatewayclass, keeping record of the paths
+		for _, routeRule := range allRouteRules {
+			paths := targetables.Paths(gatewayClass, routeRule) // this may be expensive in clusters with many gateway classes - an alternative is to deep search the topology for httprouterules from each gatewayclass, keeping record of the paths
```
guicassolato, line 76: nit:

```suggestion
			paths := targetables.Paths(gatewayClass, routeRule) // this may be expensive in clusters with many gateway classes - an alternative is to deep search the topology for httprouterules and grpcrouterules from each gatewayclass, keeping record of the paths
```

examples/grpc-backend/ratelimitpolicy/README.md
mikenairn: Yes, so i was thinking a bit more about what docs we need for the grpc work, and where best to add/update docs for it.

I followed a similar pattern as i did for authpolicy by just getting claude to document the necessary steps to verify rate limiting works on grpcroutes. That's what this doc is really, verification steps. I was thinking longer term this doc and the authpolicy equivalent will likely be removed, or at least reduced in size, in favour of proper docs added into the docs/ directory as part of https://github.com/Kuadrant/kuadrant-operator/issues/1827, with these being the ones that will get added to the docs site.

One of these docs i think would be a "user-guide" style doc, similar to others we have that use the toystore http service, but for a grpc service instead, adding auth and rate limiting.

mikenairn: If preferable, i could remove the README entirely from here and add verification steps into the PR description (like i did for [planpolicy](https://github.com/Kuadrant/kuadrant-operator/pull/1897)) and leave the docs to that later task. Alternatively, just get the docs into the desired shape in this PR, but I'm not sure this README is exactly what we want for that, we don't really have a Gateway or HTTPRoute equivalent.

internal/controller/ratelimit_workflow_helpers.go
```diff
@@ -376,8 +380,12 @@ func buildWasmActionsForAnyRateLimit(
 ) []wasm.Action {
 	policiesInPath := kuadrantv1.PoliciesInPath(path, policyPredicate)
 
-	_, _, _, httpRoute, _, _ := kuadrantpolicymachinery.ObjectsInRequestPath(path)
-	limitsNamespace := LimitsNamespaceFromRoute(httpRoute.HTTPRoute)
```
guicassolato, line 380: It looks like `controllers.LimitsNamespaceFromRoute` is no longer being used.

Although the func does exactly the same thing as here (i.e., return the namespaced name of the route), the idea of having a func was to try to keep consistency between what's configured in the Limitador CR and what goes in the wasm module configs.

**Any chance we can keep using the func?**

In fact, I'd even go one (small) step further and define a proper type for it:

```go
type LimitNamespace string
```

That is what [`LimitsNamespaceFromRoute`](https://github.com/Kuadrant/kuadrant-operator/blob/e32ef43577240ca4d9f8a267c9dc5000d7f8e699/internal/controller/ratelimit_workflow_helpers.go#L65) should return IMO.

While [`wasmActionFromLimit`](https://github.com/Kuadrant/kuadrant-operator/blob/e32ef43577240ca4d9f8a267c9dc5000d7f8e699/internal/controller/ratelimit_workflow_helpers.go#L113) and [`wasmActionsFromTokenLimit`](https://github.com/Kuadrant/kuadrant-operator/blob/e32ef43577240ca4d9f8a267c9dc5000d7f8e699/internal/controller/ratelimit_workflow_helpers.go#L219) should expect the same type or a castable:

```go
type ActionScope string
```

It would reduce chances of configuration mismatch in the future.

internal/controller/data_plane_policies_workflow.go
```diff
@@ -49,6 +49,7 @@ var (
 		{Kind: &machinery.GatewayClassGroupKind},
 		{Kind: &machinery.GatewayGroupKind},
 		{Kind: &machinery.HTTPRouteGroupKind},
+		{Kind: &machinery.GRPCRouteGroupKind},
```
maksymvavilov, line 52: I didn't know where to stick this comment.  Let it be here. 

Do we need to introduce a gRPC variant of this https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/extension/registry.go#L592 ? Or make it look at all routes?

internal/controller/ratelimit_workflow_helpers.go
```diff
@@ -354,7 +380,7 @@ func buildWasmActionsForTokenRateLimit(effectivePolicy EffectiveTokenRateLimitPo
 		}
 		limitIdentifier := TokenLimitNameToLimitadorIdentifier(k8stypes.NamespacedName{Name: source.GetName(), Namespace: source.GetNamespace()}, uniquePolicyRuleKey)
 		limitSpec := policyRule.GetSpec().(*kuadrantv1alpha1.TokenLimit)
-		scope := limitsNamespace
+		scope := ActionScope(limitsNamespace)
```
maksymvavilov: This seems confusing to me. Why do we have the `LimitNamespace` and `ActionScope`? They seem to hold the same data and are always (?) used together. Why the need for two of the same?

internal/controller/limitador_limits_reconciler.go
```diff
@@ -142,8 +143,12 @@ func (r *LimitadorLimitsReconciler) processEffectivePolicies(ctx context.Context
...
 func (r *LimitadorLimitsReconciler) processPolicyRules(ctx context.Context, pathID string, path []machinery.Targetable, rules map[string]kuadrantv1.MergeableRule, state *sync.Map, rateLimitIndex *ratelimit.Index) {
 	logger := controller.LoggerFromContext(ctx).WithName("LimitadorLimitsReconciler").WithName("processPolicyRules").WithValues("context", ctx)
-	_, _, _, httpRoute, _, _ := kuadrantpolicymachinery.ObjectsInRequestPath(path)
-	limitsNamespace := LimitsNamespaceFromRoute(httpRoute.HTTPRoute)
+	parsed, err := kuadrantpolicymachinery.ParseTopologyPath(path)
+	if err != nil {
+		logger.Error(err, "failed to parse request path", "pathID", pathID)
+		return
+	}
+	limitsNamespace := LimitsNamespaceFromRoute(parsed.GetRoute())
```
maksymvavilov, line 151: The `GetRoute()` can return nil. I'm not sure if `ParseTopologyPath()` will allow a nil route, but there could be an unknown one. 
If we end up with nil, this will panic

internal/controller/ratelimit_workflow_helpers.go
```diff
@@ -325,8 +347,12 @@ func buildWasmActionsForTokenRateLimit(effectivePolicy EffectiveTokenRateLimitPo
 	rules := effectivePolicy.Spec.Rules()
 	policiesInPath := kuadrantv1.PoliciesInPath(path, policyPredicate)
 
-	_, _, _, httpRoute, _, _ := kuadrantpolicymachinery.ObjectsInRequestPath(path)
-	limitsNamespace := LimitsNamespaceFromRoute(httpRoute.HTTPRoute)
+	parsed, err := kuadrantpolicymachinery.ParseTopologyPath(path)
+	if err != nil {
+		// If the path is invalid, return empty actions
+		return []wasm.Action{}
```
maksymvavilov, line 330: Why don't we return an error here (and in the (and in the buildWasmActionsForAnyRateLimit)?

internal/controller/ratelimit_workflow_helpers.go
mikenairn: I had fully implemented this as suggested, but backed it out to the minimum for now [see](https://github.com/Kuadrant/kuadrant-operator/pull/1881/commits/f08934d570b14cce67ff7e54a1a8400b6e0f954e), only adding back the use of `LimitsNamespaceFromRoute`.

A lot of additional code, test etc... needed updated for the the rest of the changes, expanding the scope of this PR. 

I added a follow on issue for the rest https://github.com/Kuadrant/kuadrant-operator/issues/1916

internal/controller/effective_ratelimit_policies_reconciler.go
```diff
@@ -59,14 +59,21 @@ func (r *EffectiveRateLimitPolicyReconciler) calculateEffectivePolicies(ctx cont
...
+		return ok
+	})
 
-	logger.V(1).Info("calculating effective rate limit policies", "httpRouteRules", len(httpRouteRules))
+	logger.V(1).Info("calculating effective rate limit policies", "httpRouteRules", len(httpRouteRules), "grpcRouteRules", len(grpcRouteRules))
 
 	effectivePolicies := EffectiveRateLimitPolicies{}
 
+	// Combine HTTP and gRPC route rules
+	allRouteRules := append(httpRouteRules, grpcRouteRules...)
```
maksymvavilov: This was highlighted to me as a potential mutation of the grpcRouteRules/httpRouteRules. If the http slice will have enough capacity to hold everything, there will be no new slice, and the two will share the memory for some elements. 
```
  allRouteRules := make([]machinery.Object, 0, len(httpRouteRules)+len(grpcRouteRules))                                                                                                                                                                                                                             
  allRouteRules = append(allRouteRules, httpRouteRules...)                                                                                                                                                                                                                                                          
  allRouteRules = append(allRouteRules, grpcRouteRules...)  
```
Will something like this work? Or the ` slices.Concat()`.

examples/grpc-backend/README.md
mikenairn: I actually removed this entire part from this README

examples/grpc-backend/ratelimitpolicy/README.md
mikenairn: Removed this authpolicy part, it was never my intention for this to become a guide or anything that should be talking about cross policy examples.

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -225,7 +225,7 @@ metadata:
     categories: Integration & Delivery
     console.openshift.io/plugins: '["kuadrant-console-plugin"]'
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2026-04-13T09:09:53Z"
+    createdAt: "2026-04-07T20:34:17Z"
```
maksymvavilov: Did the date go backwards?

internal/controller/ratelimit_workflow_helpers.go
maksymvavilov: I was so slow that this is gone now, never mind it

internal/controller/data_plane_policies_workflow.go
mikenairn: Probably but not in this PR, it should be part of the next one that is adding support to extensions https://github.com/Kuadrant/kuadrant-operator/pull/1897/commits/09f8dbf91036cbc17365e3f56ab301122e671fce

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
mikenairn: Meh, it's generated. I do recall seeing this and trying to get it to re-generate after a rebase, but in reality it doesn't make any difference to anything as long its moved in the correct direction between releases.

Will see if i can generate it again.

internal/controller/effective_ratelimit_policies_reconciler.go
mikenairn: Will update, as well as authpolicy since its the same and I'd prefer they were kept consistent.

internal/controller/limitador_limits_reconciler.go
mikenairn: This is really a trade off with the pattern chosen when initially adding the GRPCRoute support, it kept the blast radius small versus an interface-based approach that would have touched far more call sites.

ParseTopologyPath either returns an error or returns a fully populated ParsedTopologyPath with one of HTTPRoute/GRPCRoute always set. There's no code path that returns a non-error result with both nil. The nil return in GetRoute() is only reachable if someone manually constructs a ParsedTopologyPath{} without going through the parser. 

I'd be in favour of exploring an interface-based approach as a refactoring exercise once all GRPCRoute support is complete and the interface surface is better understood from real usage.

internal/controller/data_plane_policies_workflow.go
mikenairn: Will be added in the next PR.

internal/controller/ratelimit_workflow_helpers.go
mikenairn: Maybe, it currently doesn't so i would probably prefer to not change the signature in this PR and deal with all the knock on effect s of that here.  

Previously it just completely ignored any errors coming from ObjectsInRequestPath, this seems slightly better, even though we are still just ignoring errors. They should be logged at the very least, but that would involve getting a logger to this function

examples/grpc-backend/README.md
mikenairn: resolving, re-open if you think its still an issue. Docs wise though i intend to update them all after this PR merges. For now, these README scoped to the contents of the current directory.

examples/grpc-backend/ratelimitpolicy/README.md
mikenairn: re-open if you think this is still an issue. I updated both the authpolicy and ratelimitpolicy README to have a prerequisite of adding the kuadrant CR at the start.

examples/grpc-backend/ratelimitpolicy/README.md
mikenairn: I have left this README in this PR for now. Each README (authpolicy and ratelimitpolicy), is currently scoped only to itself and doesn't mention the other.

Longer term (after this PR and PlanPolicy PR is merged), i will be adding docs as you described, until then these narrower scoped READMEs can serve as reference.

Resolving this, re-open if you think we should address these docs differently in this PR and we can discuss.

internal/controller/limitador_limits_reconciler.go
maksymvavilov: Is there a follow-up issue for this? Either way, I'm happy to resolve this conversation

internal/controller/ratelimit_workflow_helpers.go
maksymvavilov: Is there a follow-up task for it then?

mikenairn: Tracked in #1890 as an enhancement opportunity, not a bug fix.

internal/controller/limitador_limits_reconciler.go
mikenairn: This isn't really a bug/issue, so there is no follow up specifically. ParseTopologyPath guarantees non-nil on success. We have #1916 is about type safety (a different concern), not nil safety. 

Regarding changing to an interface-based approach, I'd rather create that once we have all the code changes in, and we know better about how the interface would need to look.

internal/controller/ratelimit_workflow_helpers.go
mikenairn: Marking as resolved.

internal/controller/limitador_limits_reconciler.go
mikenairn: Marking as resolved, re-open if you have further thoughts/concerns.

examples/grpc-backend/ratelimitpolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+```bash
+for i in {1..5}; do
+  echo "Request $i:"
+  grpcurl -plaintext -authority grpcbin.local \
+    $GATEWAY_IP:80 grpcbin.GRPCBin/DummyServerStream
+  sleep 1
+done
+```
+
+Expected: Only 1 successful `grpcurl` invocation before hitting the rate limit (2 total calls consumed by reflection + method call), then `Code: Unavailable` for subsequent requests.
```
guicassolato: Does this example work as expected?

I believe the `DummyServerStream` method will take a few seconds per request streaming back. It's long enough to reset the rate limit counter before the next request starts. Unless we perform those requests in parallel, I think we won't actually see a 429.

While trying this locally, I changed the spec of the `streaming-method` limit definition in the `grpcbin-method-rlp` RLP to 2rp30s. It did the trick for me. I was able to send **2** requests to the `DummyServerStream` method, and then started getting a 429.

maksymvavilov (dismissed):

examples/grpc-backend/ratelimitpolicy/README.md
mikenairn: Using "DummyServerStream" for any of this was possibly a mistake, really it just needs a different method, maybe using one of the other available methods would be a better option for a more reliable test.

mikenairn: I updated the example to use '/grpcbin.GRPCBin/HeadersUnary' and '/grpcbin.GRPCBin/DummyUnary' as both will return immediately.

Also updated the docs a bit to make it clearer that reflection is not an issue with the method specific RLP targeting since it only applies to that grpc endpoint and not the reflection one.

Also added explicit steps before each example to cleanup previous examples RLPs.

guicassolato (approved):
Nice work!

The example is easy to follow and straightforward for demonstrating attaching RLPs to GRPCRoutes.

Perhaps for the method-based RL use case, not in this PR but in the future, maybe at the level of documentation/user-guide or future demo/example provided, we could highlight attaching the policy to a specific section of the route. Similarly to discussed at https://github.com/Kuadrant/kuadrant-operator/pull/1855#discussion_r3050723217.

mikenairn:
> Perhaps for the method-based RL use case, not in this PR but in the future, maybe at the level of documentation/user-guide or future demo/example provided, we could highlight attaching the policy to a specific section of the route. Similarly to discussed at [#1855 (comment)](https://github.com/Kuadrant/kuadrant-operator/pull/1855#discussion_r3050723217).

Agreed, docs will be created/updated to include examples demonstrating some of the gateway API ways to apply policies to particular sections.

PR #1882, opened 2026-04-10
ci: add verify-manifests

mikenairn:
## Summary
- Adds a `verify-manifests` CI job to the test workflow
- Ensures CRD bases (`./config`), bundle manifests (`./bundle`), and helm charts (`./charts`) are verified against generated output on every PR
- Wired into `required-checks` so PRs cannot merge with stale manifests

## Context
The `verify-manifests` make target already existed but was only available locally via `make pre-commit`. This gap allowed the ThreatPolicy extension rename (#1846) to merge without its generated bundle manifest (`bundle/manifests/extensions.kuadrant.io_threatpolicies.yaml`), which went undetected.

## Notes
- This PR will fail `verify-manifests` until #1883 is merged, which adds the missing ThreatPolicy bundle manifest
- This is expected and demonstrates the check working correctly — detecting that manifests are out of sync with the codebase

## Summary by CodeRabbit

* **Chores**
  * Introduced automated manifest verification as a new quality assurance checkpoint in the continuous integration pipeline, ensuring enhanced build reliability.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -361,12 +361,29 @@ jobs:
         run: |
           make verify-go-mod
 
+  verify-manifests:
+    name: Verify manifests
+    if: needs.pre-job.outputs.should_skip != 'true'
+    needs: pre-job
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+        id: go
+      - name: Verify manifests
+        run: |
+          make verify-manifests
+
   required-checks:
     name: Test Required Checks
     # This check adds a list of checks to one job to simplify adding settings to the repo.
     # If a new check is added in this file, and it should be retested on entry to the merge queue,
     # it needs to be added to the list below aka needs: [ existing check 1, existing check 2, new check ].
-    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-fmt, test-scripts, verify-generate, verify-go-mod ]
+    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-fmt, test-scripts, verify-generate, verify-go-mod, verify-manifests ]
     if: always()
     runs-on: ubuntu-latest
     steps:


guicassolato (commented):
I think there may have been a misunderstanding [here](https://github.com/Kuadrant/kuadrant-operator/pull/1181#pullrequestreview-2664684293), when calling `verify-manifests` was replaced with `verify-prepare-release`.

These two targets are not equivalent regarding checking whether the manifests (config, bundle and chart) are healthy in a PR. This is because the pattern of `verify-manifests` goes:
1. Generate the manifests (config, bundle or chart)
2. `git diff --exit-code ./(config|bundle|chart)`
3. `[ -z "$(git ls-files --other --exclude-standard --directory --no-empty-directory ./(config|bundle|chart))" ]`

whereas `verify-prepare-release` performs 1 and 2, but misses 3.

Maybe a simpler fix could consist of just adding step 3 to `verify-prepare-release`. I.e.:

```diff
diff --git a/make/verify.mk b/make/verify.mk
index 6386dbc0..44855bf7 100644
--- a/make/verify.mk
+++ b/make/verify.mk
@@ -42,4 +42,4 @@ verify-manifests: ## Verify manifests update.
 .PHONY: verify-prepare-release ## Verify set of manifests based on release.yaml file.
 verify-prepare-release: prepare-release
 	git diff --exit-code .
-
+	[ -z "$$(git ls-files --other --exclude-standard --directory --no-empty-directory .)" ]
```

However, I'm not very familiar with everything `prepare-release` (a dependency of `verify-prepare-release`) does. So perhaps @Boomatang and @didierofrivia can help us figure what's the best approach here to ensure config, bundle and chart manifests are verified at every PR _and_ preparing for a release works as expected.

bundle/manifests/extensions.kuadrant.io_threatpolicies.yaml
```diff
@@ -0,0 +1,180 @@
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.19.0
+  creationTimestamp: null
+  labels:
+    app: kuadrant
+  name: threatpolicies.extensions.kuadrant.io
+spec:
+  group: extensions.kuadrant.io
+  names:
+    kind: ThreatPolicy
```
adam-cattermole: For what it's worth, this CRD is a new example custom policy that won't be packaged with the kuadrant operator like our existing extensions. It was mistakenly included in the base Dockerfile but ~has been~ is in the process of being removed. I don't think this CRD should be present in the default bundle generation

guicassolato: In this case, maybe [this line](https://github.com/Kuadrant/kuadrant-operator/blob/65a059bebbd282836a6831ad98ab318769aff196/config/crd/kustomization.yaml#L14) should be removed.

mikenairn: Yes, sounds like that is the better option.

guicassolato: https://github.com/Kuadrant/kuadrant-operator/pull/1883

mikenairn: Created https://github.com/Kuadrant/kuadrant-operator/pull/1884

mikenairn: Closed mine. We can use https://github.com/Kuadrant/kuadrant-operator/pull/1883

guicassolato (approved):
If this ends being a redundant step after fixing `verify-prepare-release` (#1888), we can remove it. Meanwhile, we need this check in place.

PR #1893, opened 2026-04-14
docs: credential injection guide for egress gateway

### LINKED ISSUE #1800: [Egress] AuthPolicy Validation on Egress
## Task 2: AuthPolicy Validation on Egress

**Depends on:** https://github.com/Kuadrant/architecture/issues/148 (Investigate Credential Injection), Task 1 (Egress Gateway Test Environment)

## Description

Validate that AuthPolicy works when attached to an egress Gateway or HTTPRoute, including credential injection via the `response` section. Identify any gaps in the Kuadrant operator or wasm-shim.

**Current sprint priority** — per CONNLINK-898, credential injection validation is a key deliverable this sprint.

## Acceptance Criteria

1. Attach AuthPolicy to the egress HTTPRoute with a simple authentication rule
2. Verify Kuadrant operator creates the expected WasmPlugin and EnvoyFilter resources targeting the egress gateway
3. Verify Authorino receives ext_authz requests with correct attributes (headers, path, source identity)
4. Test authorization: workload with valid credentials can reach external service, workload without credentials is denied
5. Test namespace-scoped authorization: only workloads in allowed namespaces can access the external service
6. Test credential injection: configure AuthPolicy `response` section to read a Secret and inject it as an Authorization header into the outbound request
7. Test N-to-1 credential mapping: multiple workloads sharing one external API key
8. Test credential selection based on destination (different credentials for different external services)
9. Document any differences in request attributes between ingress and egress (e.g., source IP, headers)
10. Document credential injection limitations: what patterns work via AuthPolicy vs what would require a dedicated AuthTokenManagementPolicy
11. Create a worked example for the AI inference use case (e.g., injecting an OpenAI API key)
12. Document gaps found and file issues if code changes are needed

maksymvavilov:
## Summary

- Adds `doc/user-guides/egress/credential-injection.md` — user guide for injecting credentials into outbound requests via AuthPolicy on an Istio egress gateway
- Adds `examples/egress-credential-injection/` — per-destination example manifests (OpenAI + Anthropic)
- Updates `doc/overviews/auth.md` — adds egress section with link to new guide

### What the guide covers

- API key authentication to gate egress access (references existing [auth-for-app-devs](doc/user-guides/auth/auth-for-app-devs-and-platform-engineers.md) guide)
- Credential injection via `metadata.http` + `response.success.headers` — fetching a K8s Secret and injecting it as a header into the outbound request
- N-to-1 credential mapping — multiple teams with separate internal API keys sharing one external credential
- Per-destination credential selection — different credentials for different external services
- Security considerations and Vault recommendation for production

### Key design decisions

- Uses `authorizationHeader` (not `customHeader`) for internal API keys — credential injection overwrites the `Authorization` header, cleanly replacing the internal token. `customHeader` leaks to the external service (experimentally verified)
- References `make local-setup-egress` from PR #1865 for infrastructure setup instead of duplicating it
- Links to existing auth guides for deny-all and API key patterns instead of repeating them

Closes #1800

## Test plan

- [x] API key auth on egress: 401 without key, 200 with key (experimentally verified on local cluster)
- [x] Credential injection: `Authorization: Bearer <external-key>` appears in httpbin.org response (experimentally verified)
- [x] Internal token not leaked: `Authorization: APIKEY <token>` is overwritten, not forwarded (experimentally verified)
- [x] All YAML matches validated patterns from [architecture#148](https://github.com/Kuadrant/architecture/issues/148) investigation
- [ ] Reviewer: follow guide steps on `make local-setup-egress` environment

## Release Notes

* **Documentation**
  * Added egress gateway coverage to AuthPolicy docs, describing credential-injection scenarios for outbound requests.
  * Published an Egress Gateway Credential Injection walkthrough demonstrating end-to-end setup with Vault-backed credential retrieval, per-identity credential behaviour, test steps (401/403 cases), a two-layer security model and cleanup instructions.

### DIFF
--- a/doc/overviews/auth.md
+++ b/doc/overviews/auth.md
@@ -220,11 +220,18 @@ The selectors within the `when` conditions of an AuthPolicy are a subset of Kuad
 
 Authorino [JSON path string modifiers](https://docs.kuadrant.io/latest/authorino/docs/features/#string-modifiers) can also be applied to the selectors within the `when` conditions of an AuthPolicy.
 
+## Egress
+
+AuthPolicy works on egress gateways with the same attachment model described above. On egress, the primary use case is **credential injection** — authenticating workloads via Kubernetes service account tokens and injecting external API credentials (from Vault) into outbound requests via `metadata.http` + `response.success.headers`.
+
+The Gateway resource and policy attachment model are identical between ingress and egress — the Kuadrant operator does not distinguish between them. See the [Egress Gateway Credential Injection](../user-guides/egress/credential-injection.md) guide for a full walkthrough.
+
 ## Examples
 
 Check out the following user guides for examples of protecting services with Kuadrant:
 
 - [Enforcing authentication & authorization with Kuadrant AuthPolicy, for app developers and platform engineers](../user-guides/auth/auth-for-app-devs-and-platform-engineers.md)
+- [Egress Gateway Credential Injection](../user-guides/egress/credential-injection.md)
 - [Authenticated Rate Limiting for Application Developers](../user-guides/ratelimiting/authenticated-rl-for-app-developers.md)
 - [Authenticated Rate Limiting with JWTs and Kubernetes RBAC](../user-guides/ratelimiting/authenticated-rl-with-jwt-and-k8s-authnz.md)
 
--- a/doc/overviews/auth.md
+++ b/doc/overviews/auth.md
@@ -222,7 +222,7 @@ Authorino [JSON path string modifiers](https://docs.kuadrant.io/latest/authorino
 
 ## Egress
 
-AuthPolicy works on egress gateways with the same attachment model described above. On egress, the primary use case is **credential injection** — authenticating workloads via Kubernetes service account tokens and injecting external API credentials (from Vault) into outbound requests via `metadata.http` + `response.success.headers`.
+AuthPolicy works on egress gateways with the same attachment model described above. On egress, the primary use case is **credential injection** — i.e., authenticating workloads using a common authentication mechanism at the gateway (e.g. Kubernetes service account tokens) and injecting specific external API credentials (e.g. API key fetched from Vault) into the outbound requests, typically replacing the `Authorization` header.
 
 The Gateway resource and policy attachment model are identical between ingress and egress — the Kuadrant operator does not distinguish between them. See the [Egress Gateway Credential Injection](../user-guides/egress/credential-injection.md) guide for a full walkthrough.
 
--- a/doc/user-guides/egress/credential-injection.md
+++ b/doc/user-guides/egress/credential-injection.md
@@ -1,0 +1,237 @@
+# Egress Gateway Credential Injection
+
+This guide demonstrates how to use AuthPolicy with an Istio egress gateway to inject credentials into requests to external services using [HashiCorp Vault](https://developer.hashicorp.com/vault) as the credential source.
+
+## Overview
+
+This guide builds on the [workload identity](egress-gateway.md#workload-identity) pattern from the egress gateway setup guide. That pattern uses `kubernetesTokenReview` to validate workload SA tokens. Credential injection extends it by adding Vault integration - using the validated identity to fetch per-workload credentials and inject them into outbound requests.
+
+### Request Flow
+
+1. Internal workload sends an HTTP request to the egress gateway with `Authorization: Bearer <sa-token>`
+2. The gateway triggers an ext_authz check with Authorino
+3. Authorino validates the workload's Kubernetes service account token via [TokenReview](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/)
+4. Authorino authenticates to Vault using the workload's SA token (Vault [Kubernetes auth method](https://developer.hashicorp.com/vault/docs/auth/kubernetes))
+5. Authorino reads the external credential from Vault and overwrites the `Authorization` header via `response.success.headers`
+6. The egress gateway originates TLS and forwards the request with the external credential to the external service
+
+### Topology
+
+```mermaid
+graph TB
+    WL([Internal<br/>Workload])
+
+    subgraph "gateway-system namespace"
+        GW[Gateway<br/>kuadrant-egressgateway]
+        HR[HTTPRoute<br/>httpbin-external]
+        SE[ServiceEntry<br/>httpbin-external]
+        DR[DestinationRule<br/>httpbin-external]
+        AP[AuthPolicy<br/>httpbin-external-auth]
+    end
+
+    subgraph "kuadrant-system namespace"
+        AUTH[Authorino]
+    end
+
+    subgraph "vault namespace"
+        VAULT[Vault<br/>K8s auth]
+    end
+
+    WL -->|"1. Authorization:<br/>Bearer sa-token"| GW
+    HR -.->|routes| GW
+    AP -.->|attached to| HR
+    GW -->|"2. ext_authz"| AUTH
+    AUTH -->|"3. Vault login<br/>(SA token)"| VAULT
+    AUTH -->|"4. read secret"| VAULT
+    GW -->|"5. HTTPS +<br/>Bearer external-key"| EXT([External API<br/>e.g. httpbin.org])
+    SE -.->|registers| EXT
+    DR -.->|TLS config| EXT
+```
+
+## Prerequisites
+
+- Egress gateway infrastructure deployed. See the [Egress Gateway Setup](egress-gateway.md) guide.
+- [HashiCorp Vault](https://developer.hashicorp.com/vault) with [Kubernetes auth method](https://developer.hashicorp.com/vault/docs/auth/kubernetes) configured. The setup script deploys a dev Vault instance automatically.
+
+```sh
+export EGRESS_IP=$(kubectl get gtw kuadrant-egressgateway -n gateway-system \
+    -o jsonpath='{.status.addresses[0].value}')
+```
+
+## Credential Injection
+
+The AuthPolicy below authenticates the workload via its Kubernetes service account token, fetches a per-identity credential from Vault, and injects it into the outbound request. The Vault path is constructed dynamically from the workload's namespace and service account name, so different workloads get different credentials. The workload sends `Authorization: Bearer <sa-token>`; the gateway replaces it with `Authorization: Bearer <external-key>`.
+
+### Step 1: Apply the AuthPolicy
+
+```sh
+kubectl apply -f - <<'EOF'
+apiVersion: kuadrant.io/v1
+kind: AuthPolicy
+metadata:
+  name: httpbin-external-auth
+  namespace: gateway-system
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: httpbin-external
+  rules:
+    authentication:
+      "workload-identity":
+        kubernetesTokenReview:
+          audiences:
+            - "https://kubernetes.default.svc.cluster.local"
+    metadata:
+      vault_login:
+        http:
+          url: "http://vault.vault.svc.cluster.local:8200/v1/auth/kubernetes/login"
+          method: POST
+          contentType: application/json
+          body:
+            expression: '"{\"jwt\": \"" + request.headers.authorization.substring(7) + "\", \"role\": \"egress-workload\"}"'
+        priority: 0
+      vault_secret:
+        http:
+          urlExpression: '"http://vault.vault.svc.cluster.local:8200/v1/secret/data/egress/" + auth.metadata.vault_login.auth.metadata.service_account_namespace + "/" + auth.metadata.vault_login.auth.metadata.service_account_name'
+          method: GET
+          headers:
+            X-Vault-Token:
+              expression: 'auth.metadata.vault_login.auth.client_token'
+        priority: 1
+    authorization:
+      vault_credential_check:
+        patternMatching:
+          patterns:
+          - predicate: 'has(auth.metadata.vault_secret.data)'
+    response:
+      success:
+        headers:
+          authorization:
+            plain:
+              expression: '"Bearer " + auth.metadata.vault_secret.data.data.api_key'
+EOF
+```
+
+**How it works:**
+- `authentication.workload-identity` - validates the workload's Kubernetes SA token via [TokenReview](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/). No API keys to distribute - every pod already has an SA token mounted.
+- `metadata.vault_login` - POSTs the SA token to Vault's Kubernetes auth endpoint. Vault validates the token and returns a scoped client token based on the workload's namespace and service account.
+- `metadata.vault_secret` - reads the external credential from Vault at a path derived from the workload's namespace and service account (e.g., `secret/egress/egress-test/default`). Each workload identity gets its own credential.
+- `authorization.vault_credential_check` - verifies the Vault credential fetch succeeded. If the workload's SA isn't authorized by Vault's policy, the request is denied (403).
+- `response.success.headers.authorization` - overwrites the `Authorization` header with the external credential. The workload's SA token never reaches the external service.
+
+### Step 2: Test
+
+The setup script stored a test credential at `secret/egress/egress-test/default` for the test client. Test the three access scenarios:
+
+```sh
+# No token - denied (401)
+kubectl exec test-client -n egress-test -- curl -s -o /dev/null -w "%{http_code}" \
+    -H "Host: httpbin.org" http://${EGRESS_IP}/get
+# 401
+
+# SA token from authorized namespace - credential injected
+kubectl exec test-client -n egress-test -- sh -c '
+curl -s -H "Host: httpbin.org" \
+    -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+    http://'"${EGRESS_IP}"'/get
+'
+# Response shows: Authorization: Bearer sk-test-openai-key-for-egress
+
+# SA token from unauthorized namespace - denied (403)
+kubectl run bad-client --image=curlimages/curl:latest -n default --restart=Never \
+    --command -- sleep infinity
+kubectl wait --for=condition=Ready pod/bad-client -n default --timeout=30s
+kubectl exec bad-client -n default -- sh -c '
+curl -s -o /dev/null -w "%{http_code}" -H "Host: httpbin.org" \
+    -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+    http://'"${EGRESS_IP}"'/get
+'
+# 403
+kubectl delete pod bad-client -n default
+```
+
+### Per-Identity Credentials
+
+Different workload identities automatically get different credentials from Vault. The `urlExpression` constructs the Vault path from the workload's namespace and service account name.
+
+To verify, store a different credential for another namespace and test:
+
+```sh
+# Add workload-b namespace to the Vault role
+kubectl exec vault-0 -n vault -- vault write auth/kubernetes/role/egress-workload \
+    bound_service_account_names=default \
+    bound_service_account_namespaces=egress-test,workload-b \
+    policies=egress-read \
+    ttl=1h
+
+# Store a different credential for workload-b
+kubectl exec vault-0 -n vault -- vault kv put secret/egress/workload-b/default \
+    api_key=sk-different-key-for-workload-b
+
+# Deploy a workload in workload-b namespace
+kubectl create namespace workload-b
+kubectl apply -f - <<EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-client-b
+  namespace: workload-b
+spec:
+  containers:
+    - name: curl
+      image: curlimages/curl:latest
+      command: ["sleep", "infinity"]
+  restartPolicy: Never
+EOF
+kubectl wait --for=condition=Ready pod/test-client-b -n workload-b --timeout=60s
+
+# Each workload gets its own credential
+kubectl exec test-client -n egress-test -- sh -c '
+curl -s -H "Host: httpbin.org" \
+    -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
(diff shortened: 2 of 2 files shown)


doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+1. **Access control** — restrict which workloads can reach external services through the egress gateway
+2. **Credential injection** — inject API keys or tokens into outbound requests so application code never handles secrets directly
+
+### Request Flow
+
+1. Internal workload sends an HTTP request to the egress gateway with `Authorization: APIKEY <token>`
+2. The gateway triggers an ext_authz check with Authorino
+3. Authorino validates the workload's API key
+4. Authorino fetches the external credential via `metadata.http` and overwrites the `Authorization` header via `response.success.headers`
+5. The egress gateway originates TLS and forwards the request with the external credential to the external service
```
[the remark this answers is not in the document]
maksymvavilov: Intentionally omitted. The AuthPolicy patterns (selectors, metadata.http, response.success.headers) work identically regardless of traffic direction — the operator reconciles all Gateways uniformly. Differences in Envoy-level attributes (source IP, peer metadata) are outside AuthPolicy's scope and would belong in Istio documentation. We had a comparison table earlier and removed it because it was saying "everything is the same" in seven rows.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
+# Egress Gateway Authorization and Credential Injection
+
+This guide demonstrates how to use AuthPolicy with an Istio egress gateway to control outbound traffic and inject credentials into requests to external services.
+
+## Overview
+
+[AuthPolicy](../../overviews/auth.md) works on egress gateways with the same policy attachment model used for ingress — the Kuadrant operator does not distinguish between them. The patterns in this guide are direction-agnostic, but are demonstrated here in an egress context where credential injection is most useful. On egress, AuthPolicy enables two capabilities:
+
+1. **Access control** — restrict which workloads can reach external services through the egress gateway
+2. **Credential injection** — inject API keys or tokens into outbound requests so application code never handles secrets directly
```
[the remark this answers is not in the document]
maksymvavilov: Fixed in 7ec7211e. Added commas in both locations.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Security Considerations
+
+The credential injection pattern shown here reads Kubernetes Secrets via `metadata.http` and loads values into the request pipeline. Keep the following in mind:
+
+- **Scope RBAC tightly** — use `resourceNames` in the Role so Authorino can only read the specific Secrets it needs, not all Secrets in the namespace
+- **Credentials flow upstream only** — use `response.success.headers` (injected into the outbound request), never `response.success.dynamicMetadata` (exposed to Envoy's filter chain and could be logged)
+- **Restrict AuthPolicy creation** — anyone who can create an AuthPolicy in the egress namespace can configure which Secrets are read. Limit access to that namespace with Kubernetes RBAC
+- **Consider Vault for production** — for production deployments, use [HashiCorp Vault](https://github.com/Kuadrant/mcp-gateway/blob/main/docs/guides/vault-token-exchange.md) as the credential source instead of raw Kubernetes Secrets. The MCP Gateway project documents this using the same `metadata.http` → `response.success.headers` pattern
+
```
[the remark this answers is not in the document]
maksymvavilov: Intentionally omitted. AuthTokenManagementPolicy does not exist yet — it is a future concept from the RFC. Documenting boundaries against an unbuilt API would make this guide speculative rather than practical. When AuthTokenManagementPolicy ships, its own documentation will define what it handles. This guide documents what AuthPolicy can do today.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+    SE -.->|registers| EXT
+    DR -.->|TLS config| EXT
+```
+
+## Prerequisites
+
+Set up a local environment with the egress gateway infrastructure, Authorino CA trust, and a test client:
+
+```sh
+make local-setup-egress
```
guicassolato: I think having a script that helps with the setup is not necessarily a bad thing. However, in this case, not only the user is required to clone the Kuadrant Operator repo, but the script will also build and deploy a new image of the Operator locally.

This feels all a bit cumbersome for one who only wants to try the egress use case.

Perhaps the tooling introduced in https://github.com/Kuadrant/kuadrant-operator/pull/1865 (focused on the _developer workflow_, I reckon) could split into using `local-setup` and running a shell script on top to take care of the rest of the sample setup (deploy the egress gateway, mount the CA bundle into the Authorino pod, test the connectivity, etc). Then, maybe here the steps could focus more on _user workflow_. I.e:
1. `kind create cluster`
2. `helm install <deps>`
3. `helm install kuadrant`
4. `curl -sL https://raw.githubusercontent.com/Kuadrant/kuadrant-operator/main/path/to/egress/example/script.sh | bash -s`

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+  name: authorino-secret-reader
+  namespace: gateway-system
+subjects:
+  - kind: ServiceAccount
+    name: authorino-authorino
+    namespace: kuadrant-system
+roleRef:
+  kind: Role
+  name: authorino-secret-reader
+  apiGroup: rbac.authorization.k8s.io
```
guicassolato: Are you sure the RBAC rules are needed?

With a cluster-wide Authorino instance deployed, Authorino Operator will bind the instance's Service Account to the `authorino-manager-role` ClusterRole, which already grants read access to secrets.

```
❯ kubectl auth can-i get secret --as system:serviceaccount:kuadrant-system:authorino-authorino -A
yes
```

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+            matchLabels:
+              app: egress-access
+        credentials:
+          authorizationHeader:
+            prefix: APIKEY
+    metadata:
+      # Use underscores in metadata key names — hyphens cause issues in CEL expressions
+      credential_fetch:
+        http:
+          url: "https://kubernetes.default.svc/api/v1/namespaces/gateway-system/secrets/openai-api-key"
```
guicassolato: > [!WARNING]
> **This opens a dangerous precedent. Not sure if we should be risking advertising it as a recommended approach.**

The two main problems I see with it are:
1. It demonstrates how one can use Authorino's RBAC permissions to elevate privileges for reading secrets cluster-wide.
2. It models a pattern of exchanging individual workload identities for a single shared external authentication key. As easy as this may sound, it should not be encouraged as good practice IMO.

A better alternative would be using a secret management solution such as Hashicorp Vault to manage the secrets to authenticate with the external service. Authorino would request the secret as an authenticated client with no risk to disclose any other. This doesn't have to be for production environment only.

Secrets should also be namespaced by user and avoid sharing them across.

This would be similar to https://github.com/guicassolato/mcp-gateway/blob/main/docs/guides/vault-integration.md

Alternatively, if we only want to demonstrate exchanging Kuadrant API key secret for a secrets to authenticate with an external service, without caring much about good practices of secret management, then:
1. Let's put huge disclaimer message at the top, instead of a bottom "Security considerations" section. (Tip: Consider using `[!WARNING]` markers like I did in this very comment.)
2. Store the external service authentication secret in an annotation of the API key secret. It's far from ideal, but at least it will save all the burden of configuring Authorino's generic HTTP metadata fetcher for consuming from the cluster's API server, and the secrets would be namespaced by workload identity as a bonus.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+
+```sh
+# Without API key — denied (401)
+kubectl exec test-client -n egress-test -- curl -s -o /dev/null -w "%{http_code}" \
+    -H "Host: httpbin.org" http://${EGRESS_IP}/get
+# 401
+
+# With API key — allowed, credential injected
+kubectl exec test-client -n egress-test -- curl -s \
+    -H "Host: httpbin.org" -H "Authorization: APIKEY team-a-egress-token" \
```
averevki: There is no yaml for secret with this api key in this guide. I suppose this is the API key that should be created during the prerequisites, but I'd still prefer having yaml here also for team-a API key to avoid ambiguity, since this exact `team-a-egress-token` API key value isn't mentioned anywhere before the verification step.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,376 @@
...
+
+httpbin.org should return a JSON response echoing request headers — no credentials present.
+
+## Securing Egress Access
+
+Before adding credential injection, secure the egress gateway using the same [deny-all + API key authentication](../auth/auth-for-app-devs-and-platform-engineers.md) pattern used for ingress:
+
+1. Apply a [gateway-level deny-all](../auth/auth-for-app-devs-and-platform-engineers.md#create-a-default-deny-all-policy-at-the-level-of-the-gateway-persona-platform-engineer) AuthPolicy targeting `kuadrant-egressgateway` — this blocks all egress routes by default
+2. Create an [API key Secret](../auth/auth-for-app-devs-and-platform-engineers.md#protect-the-toy-store-application-persona-app-developer) in `kuadrant-system` with label `app: egress-access`
+3. Apply a route-level AuthPolicy on `httpbin-external` with `apiKey` authentication using `authorizationHeader` with prefix `APIKEY`
```
averevki: I think it would be better if there is yaml here instead for at least these 2 last steps. If in every step here only a single manifest applied, why not have direct kubectl apply commands? Links to the other guide are useful, but I think they don't have the exact manifests that would work for the setup in this guide

doc/user-guides/egress/credential-injection.md
maksymvavilov: Now there is a script that will deploy everything. As a note, the script lives in the #1865

doc/user-guides/egress/credential-injection.md
maksymvavilov: No longer recomend to use k8s secrets or rbac

doc/user-guides/egress/credential-injection.md
maksymvavilov: Pivited to Vault with each combination of namespace/sa gets its own token. This should solve the problems

doc/user-guides/egress/credential-injection.md
maksymvavilov: No longer using API keys.

doc/user-guides/egress/credential-injection.md
maksymvavilov: No longer applicable

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Per-Destination Credential Selection
+
+For different external services requiring different credentials, create separate HTTPRoutes with separate AuthPolicies. Each AuthPolicy points `vault_secret` at a different Vault path.
+
+The setup script stored test credentials at `secret/egress/openai` and `secret/egress/anthropic`. Apply the per-destination examples:
+
+```sh
+kubectl delete authpolicy httpbin-egress -n gateway-system
+kubectl delete httproute httpbin-external -n gateway-system
```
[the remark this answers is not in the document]
maksymvavilov: No longer applicable — those delete commands were removed when the per-destination section was replaced by per-identity urlExpression. All remaining delete commands in the cleanup section already use --ignore-not-found.

maksymvavilov:
## Egress documentation restructure

This PR has been updated as part of the egress docs restructure ([#1896](https://github.com/Kuadrant/kuadrant-operator/pull/1896)).

**What changed:** Slimmed the overview and prerequisites to reference `egress-gateway.md` for setup and workload identity basics. The guide now focuses on what's unique to credential injection: Vault integration, per-identity credential paths, and the two-layer security model. All Vault-specific content is unchanged.

**Merge order:** #1896 first, then #1856, then this PR.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,237 @@
...
+          authorization:
+            plain:
+              expression: '"Bearer " + auth.metadata.vault_secret.data.data.api_key'
+EOF
+```
+
+**How it works:**
+- `authentication.workload-identity` - validates the workload's Kubernetes SA token via [TokenReview](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-review-v1/). No API keys to distribute - every pod already has an SA token mounted.
+- `metadata.vault_login` - POSTs the SA token to Vault's Kubernetes auth endpoint. Vault validates the token and returns a scoped client token based on the workload's namespace and service account.
+- `metadata.vault_secret` - reads the external credential from Vault at a path derived from the workload's namespace and service account (e.g., `secret/egress/egress-test/default`). Each workload identity gets its own credential.
```
averevki, line 119:
```suggestion
- `metadata.vault_login` (`priority: 0`) - POSTs the SA token to Vault's Kubernetes auth endpoint. Vault validates the token and returns a scoped client token based on the workload's namespace and service account.
- `metadata.vault_secret` (`priority: 1`) - reads the external credential from Vault at a path derived from the workload's namespace and service account (e.g., `secret/egress/egress-test/default`). Each workload identity gets its own credential.
```
Or some separate note about the dependency between these two might be helpful

doc/user-guides/egress/credential-injection.md
averevki, line 1: I see a few similarities between this guide and the guide for Vault integration with MCP Gateway https://docs.kuadrant.io/1.4.x/mcp-gateway/docs/guides/vault-integration/. You think we could abstract the Vault integration part somehow and reuse it in both, egress and mcp, guides so there is less duplication in docs?

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,237 @@
+# Egress Gateway Credential Injection
```
averevki, line 1: Do we want to mention any alternatives to Vault for credentials injection, if Vault is not an option for some reason? Like a prior approach with kubernetes secret and kubernetes API if we still consider it valid

doc/user-guides/egress/credential-injection.md
maksymvavilov: What I have here was "borrowed" from the guide you are linking to. The overall structure is the same (authenticate -> fetch from Vault -> gate on success -> inject credential). But over time it drifted far enough imo to deserve its own doc:

| Aspect | MCP Gateway | Egress Guide |
|---|---|---|
| Authentication | OIDC JWT (external users) | `kubernetesTokenReview` (internal pods) |
| Vault auth method | JWT auth (`/auth/jwt/login`) | K8s auth (`/auth/kubernetes/login`) |
| Metadata steps | 3 (oauth-token → vault-login → vault) | 2 (vault_login → vault_secret) |
| Vault login body | OAuth access token from prior step | Raw SA token from request header |
| Secret path | `mcp-gateway/{auth.identity.sub}` | `egress/{namespace}/{sa-name}` |
| Caching | Yes (TTLs on metadata) | No |

The abstractable part is just the `response.success.headers` line and the authorisation check. The Vault auth setup, metadata chain, and identity model changed. I don't think there's enough shared content to abstract without making both guides harder to follow. I was trying to reference existing guides as much as possible, but this one is an awkward bit.

doc/user-guides/egress/credential-injection.md
maksymvavilov: This is another point of debate. Just like with the DNS routing. I had multiple ways in previous versions, but with the injection, it is a bit simpler - the Vault solves all the problems (k8s secrets have two big security [concerns](https://github.com/Kuadrant/kuadrant-operator/pull/1893#discussion_r3086523282)). There is one use case that requires something else, but the solution they have is very messy to recommend it with a straight face. For now, there are no "close second" alternatives. Or rather, there is no single "close second" one. 

What we are proposing here is an approach that uses Vault as an external entity. Choose Vault because it is well-documented in Kuadrant. It can be anything that will do the `metadata.http + response.success.headers ` pattern. But the approach stays the same. 

There can be another approach, but not with the AuthPolicy. You'll need to use something else, and in several sprints, we will have a dedicated policy for this - this will be another approach. 

For now, I could add WS Secrets Manager / Azure Key Vault / GCP Secret Manager / CyberArk Conjur, but I don't think it will add value to the guide :shrug:

doc/overviews/auth.md
```diff
@@ -220,11 +220,18 @@ The selectors within the `when` conditions of an AuthPolicy are a subset of Kuad
 
 Authorino [JSON path string modifiers](https://docs.kuadrant.io/latest/authorino/docs/features/#string-modifiers) can also be applied to the selectors within the `when` conditions of an AuthPolicy.
 
+## Egress
+
+AuthPolicy works on egress gateways with the same attachment model described above. On egress, the primary use case is **credential injection** — authenticating workloads via Kubernetes service account tokens and injecting external API credentials (from Vault) into outbound requests via `metadata.http` + `response.success.headers`.
```
guicassolato:
```suggestion
AuthPolicy works on egress gateways with the same attachment model described above. On egress, the primary use case is **credential injection** — i.e., authenticating workloads using a common authentication mechanism at the gateway (e.g. Kubernetes service account tokens) and injecting specific external API credentials (e.g. API key fetched from Vault) into the outbound requests, typically replacing the `Authorization` header.
```

doc/user-guides/egress/credential-injection.md
guicassolato: Good call on using K8s auth to authenticate to Vault!

If we ever want to simplify and reuse resources across examples, we should use this one as model IMO.

doc/user-guides/egress/credential-injection.md
guicassolato: If we want to revisit this in the future, reading Secrets from the API server can be an alternative to using Vault, as long as we reuse the requestor's service account token for authenticating with the API server (like it was done in the Vault example here), and not a shared token with permissions granted to Authorino's Service Account.

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,237 @@
...
+    metadata:
+      vault_login:
+        http:
+          url: "http://vault.vault.svc.cluster.local:8200/v1/auth/kubernetes/login"
+          method: POST
+          contentType: application/json
+          body:
+            expression: '"{\"jwt\": \"" + request.headers.authorization.substring(7) + "\", \"role\": \"egress-workload\"}"'
+        priority: 0
+      vault_secret:
```
guicassolato, line 94: There are ways to avoid the extra hop to Vault if the login fails and short-circuit authorisation to deny access immediately in such cases. It adds complexity to the example though. Not sure if it's worth including.

cc @maleck13

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,237 @@
...
+    AUTH -->|"3. Vault login<br/>(SA token)"| VAULT
+    AUTH -->|"4. read secret"| VAULT
+    GW -->|"5. HTTPS +<br/>Bearer external-key"| EXT([External API<br/>e.g. httpbin.org])
+    SE -.->|registers| EXT
+    DR -.->|TLS config| EXT
+```
+
+## Prerequisites
+
+- Egress gateway infrastructure deployed. See the [Egress Gateway Setup](egress-gateway.md) guide.
```
guicassolato, line 53: Added in #1896

doc/user-guides/egress/credential-injection.md
maleck13: I would side with reducing complexity in the guide

doc/user-guides/egress/credential-injection.md
```diff
@@ -0,0 +1,237 @@
...
+
+Different workload identities automatically get different credentials from Vault. The `urlExpression` constructs the Vault path from the workload's namespace and service account name.
+
+To verify, store a different credential for another namespace and test:
+
+```sh
+# Add workload-b namespace to the Vault role
+kubectl exec vault-0 -n vault -- vault write auth/kubernetes/role/egress-workload \
+    bound_service_account_names=default \
+    bound_service_account_namespaces=egress-test,workload-b \
```
guicassolato, line 164: I wonder how binding 2 service accounts (with different names) from 2 different namespaces looks like 🤔

doc/user-guides/egress/credential-injection.md
guicassolato: https://github.com/Kuadrant/kuadrant-operator/pull/1893/changes#r3117097575

doc/user-guides/egress/credential-injection.md
maksymvavilov: Done

doc/user-guides/egress/credential-injection.md
maksymvavilov: It should be a cross product. 
We would 
```
bound_service_account_names=app-a,app-b                                                                                                                                                                                                                                                                           
bound_service_account_namespaces=namespace-1,namespace-2
```

and will allow: `app-a`+`ns`, `app-b`+`ns-1`, `app-a`+`ns-2` and `app-b`+`ns-2`. 

As an alternative you can have two separate roles 
```
  # Role for app-a in namespace-1 only                                                                                                                                                                                                                                                                              
  vault write auth/kubernetes/role/egress-ns1-app-a \                                                                                                                                                                                                                                                               
      bound_service_account_names=app-a \                                                                                                                                                                                                                                                                           
      bound_service_account_namespaces=namespace-1 \
      policies=egress-read \                                                                                                                                                                                                                                                                                        
      ttl=1h                 
                                                                                                                                                                                                                                                                                                                    
  # Role for app-b in namespace-2 only
  vault write auth/kubernetes/role/egress-ns2-app-b \
      bound_service_account_names=app-b \                                                                                                                                                                                                                                                                           
      bound_service_account_namespaces=namespace-2 \
      policies=egress-read \                                                                                                                                                                                                                                                                                        
      ttl=1h     
```
and have something like this `  body: expression: '"{\"jwt\": \"" + request.headers.authorization.substring(7) + "\", \"role\": \"egress-" + auth.identity.username + "\"}"'` but that raises even more questions and requires a lot of "warning" notes

doc/user-guides/egress/credential-injection.md
averevki: The new dedicated policy would be a good alternative (maybe even change to the main method). We can revisit this when we will be adding new user guides for it :+1:

guicassolato (approved):
Nice work!

Issue #1889, opened 2026-04-13
SDK API to orchestrate custom policies' `Action`s


alexsnaps:
Define an abstract `Action` type with implementations available to the kuadrant policy author to:

 - Call into an gRPC service
 - Allow request based off some `request` attributes
 - An `Action` either allows the chain to continue or denies it
 
Open questions:
 - How do we allow for "branching" in the logic, e.g. add headers, and then deny.

Maybe don't overthink it, and just have the headers and other props of the response be owned by the caller.

<details>

```golang

policyPipeline := ctx.PolicyPipeline() // ?
policyPipeline.onHttpRequest(actions...) // cel `response` is not valid, but `request` is 
policyPipeline.onHttpResponse(actions...) // cel `response` is now valid, but should `request` be?
```

</details>

Thoughts on user API:

All names of objects and concepts below can and should be improved.

- `RegisterUpstreamMethod` becomes `RegisterActionMethod`.
- `RegisterActionMethod` requires an identifier for the ActionMethod plus a definition of any expected arguments, this must be unique.
- A new pipeline exists which will be used to chain actions on the request (`OnRequest`) or response (`OnResponse`) to ensure they run sequentially.
- `OnRequest` and `OnResponse` will accept action objects which will define logic to be executed by the wasm-shim, these action objects may or may not implement the same interface or abstract, this will be decided during investigation.
- `OnRequest` will expect an action object that will not always define the same properties, two examples:
  -  `GRPCMethodAction` (Decide to permit the request or deny it based on the response of a GRPC method) might expect: Predicate, Intention and Method
     - Predicate: Test whether or not to run the method
     - Method: The method to run if predicate is true
     - Intention: How to test the result of the method to decide whether to continue with the request or not
  - `AllowAction` (Permit or deny request based on well known attributes - no GRPC method) might expect: Predicate and Intention
     - Predicate: Test whether to apply this action or not
     - Intention: How to test whether to decide to continue with the request or not
- `OnResponse` will also expect an action object - we will need to decide if this will be the same as OnRequest or unique,  two examples:
  - `EnsureHeadersAction` (ensure the response has specific headers) might expect: Predicate and Headers
    - Predicate: Test whether to apply the modification or not
    - Headers: The headers to ensure exist
  - `WithResponseCodeAction` (modify the response code of the response) might expect: Predicate and NewResponseCode
    - Predicate: Test whether to apply the modification or not
    - NewResponseCode: The code to set on the response
[...]

PR #1895, opened 2026-04-16
Proposal: Extensions SDK API to orchestrate actions in a pipeline.

### LINKED ISSUE #1838: Extension SDK: API for Upstream Request Orchestration
# Extension SDK: API for Upstream Request Orchestration

## Summary

Design an Extension SDK API that allows extensions to initiate gRPC calls to registered upstream services and define behavior based on the response. This will enable policies like ThreatPolicy to make dynamic decisions (Allow/Deny) based on external service responses.

## Problem Statement

**What already exists:**
- `RegisterUpstreamMethod(ctx, policy, UpstreamConfig)` - Registers services and creates Envoy clusters
- Upstream registry and cluster generation
- gRPC reachability checking

**What's missing:**
- **How to build a gRPC request message** from request context
- **When to send the message** (during request processing)
- **What to do with the response** (Allow/Deny based on response data)
- **How to translate this into wasm-shim config** that executes at runtime

This issue is for **designing the API**, not implementing it. We need to decide on the abstraction before writing code.

## Requirements for the API

The API must enable extensions to:

### 1. Build gRPC Request Messages Using CEL

**What it needs to do:**
- Accept a CEL expression that constructs a protobuf message
- Extract values from request context (URI, headers, auth state, source IP)
- Build a message like: `ThreatRequest{uri: request.path, is_authenticated: has(auth.identity), source_ip: request.source.address}`

**Example (conceptual):**
```go
// Extension code needs to specify:
// - WHAT message to build
// - FROM WHAT data (CEL expressions)
Message: `ThreatRequest{
    uri: request.path,
    is_authenticated: has(auth.identity),
    source_ip: request.source.address
}`
```

**Open questions:**
- How do we validate the CEL syntax at reconcile time?
- How do we know what message types are available? (via proto registration/reflection)
- Should message construction be a single CEL expression or structured fields?

### 2. Define Behavior Based on Response

**What it needs to do:**
- Receive the gRPC response from the upstream
- Evaluate response fields (e.g., `response.threat_level`)
- Make a decision: Allow the request OR Deny with status code and message

**Example (conceptual):**
```go
// Extension code needs to specify:
// - WHAT to check in the response
// - WHAT to do (Allow/Deny/Modify)
Callback: `
    response.threat_level >= 5
        ? Deny(403, "Request blocked: threat level " + string(response.threat_level))
        : Allow()
`
```

**Open questions:**
- What actions are available beyond Allow/Deny? (Modify headers? Redirect?)
- How do we handle errors (service timeout, gRPC errors)?
- Should there be a default behavior if callback fails?

### 3. Orchestrate the Call at the Right Time

**What it needs to do:**
- Execute the gRPC call **during request processing** (before forwarding to upstream)
- Wait for response (with timeout)
- Apply the callback decision

**Example flow:**
```
1. HTTP request arrives at gateway
2. Wasm-shim matches policy rules
3. Wasm-shim builds ThreatRequest message from CEL
4. Wasm-shim sends gRPC call to registered upstream
5. Upstream responds with ThreatResponse
6. Wasm-shim evaluates callback CEL
7. Decision: Allow (200 to upstream) OR Deny (403 to client)
```

### 4. Translate to Wasm-Shim Configuration

**What it needs to do:**
- Generate wasm config JSON that wasm-shim can execute
- Include service definition, message construction CEL, callback CEL
- Ensure wasm-shim can build protobuf messages dynamically (uses proto reflection)

**Example wasm config (what the API should generate):**
```json
{
  "services": {
    "ext_threat": {
      "type": "extension",
      "endpoint": "ext-threat-service-8080",
      "failureMode": "allow"
    }
  },
  "actionSets": [
    {
      "name": "default/api-threat-protection",
      "actions": [
        {
          "service": "ext_threat",
          "dispatch": "ThreatRequest{uri: request.path, is_authenticated: has(auth.identity), source_ip: request.source.address}",
          "callback": "response.threat_level >= 5 ? Deny(403, 'Blocked') : Allow()",
          "onError": "Allow()"
        }
      ]
    }
  ]
}
```

**Open questions:**
- Does this need new wasm-shim features or can we use existing action system?
- How does wasm-shim build protobuf from CEL without hardcoded types?
- What's the error handling strategy in wasm-shim?

## ThreatPolicy Use Case (Concrete Example)

**What ThreatPolicy needs to do in its reconciler:**

```go
const threatServiceURL = "grpc://threat-service.security.svc.cluster.local:8080"

func (r *ThreatPolicyReconciler) Reconcile(ctx context.Context, request reconcile.Request, kuadrantCtx types.KuadrantCtx) (reconcile.Result, error) {
    policy := &v1alpha1.ThreatPolicy{}
    if err := r.Client.Get(ctx, request.NamespacedName, policy); err != nil {
        return reconcile.Result{}, err
    }

    // Step 1: Register the upstream (ALREADY WORKS)
    if err := kuadrantCtx.RegisterUpstreamMethod(ctx, policy, types.UpstreamConfig{
        URL: threatServiceURL,
    }); err != nil {
        return reconcile.Result{}, err
    }

    // Step 2: Define what message to send and what to do with response (NEEDS API)
    //
    // We need to tell the data plane:
    // - Build this message: ThreatRequest{uri: request.path, is_authenticated: has(auth.identity), source_ip: request.source.address}
    // - Send it to threatServiceURL during request processing
    // - Evaluate response: if threat_level >= threshold, deny; otherwise allow
    //
    // ??? What API call goes here ???

    return reconcile.Result{}, nil
}
```

**Expected data plane behavior:**
```
HTTP Request → Gateway
    ↓
Wasm-shim matches ThreatPolicy rules
    ↓
Wasm-shim builds: ThreatRequest{uri: "/api/users", is_authenticated: false, source_ip: "10.0.0.1"}
    ↓
Wasm-shim sends gRPC to ext-threat-service-8080
    ↓
ThreatService responds: ThreatResponse{threat_level: 6, reasons: ["unauthenticated", "blacklisted_ip"]}
    ↓
Wasm-shim evaluates: threat_level (6) >= threshold (5) → Deny
    ↓
Client receives: 403 Forbidden "Request blocked: threat level 6"
```

### 1. Envoy Cluster (EnvoyFilter for Istio)

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: ext-threat-default-api-protection
  namespace: istio-system
  ownerReferences:
    - apiVersion: kuadrant.io/v1alpha1
      kind: ThreatPolicy
      name: api-protection
spec:
  targetRefs:
      kind: Gateway
      name: my-gateway
  configPatches:
    - applyTo: CLUSTER
      patch:
        operation: ADD
        value:
          name: ext_threat_default_api-protection
          type: STRICT_DNS
          connect_timeout: 1s
          lb_policy: ROUND_ROBIN
          http2_protocol_options: {}
          load_assignment:
            cluster_name: ext_threat_default_api-protection
            endpoints:
              - lb_endpoints:
                  - endpoint:
                      address:
                        socket_address:
                          address: threat-service.security.svc.cluster.local
                          port_value: 8080
```

### 2. Wasm Config JSON

```json
{
  "services": {
    "ext_threat_default_api-protection": {
      "type": "extension",
      "endpoint": "ext_threat_default_api-protection",
      "failureMode": "allow",
      "timeout": "100ms"
    }
  },
  "actionSets": [
    {
      "name": "default/api-threat-protection",
      "routeRuleConditions": {
        "hostnames": ["*.example.com"]
      },
      "actions": [
        {
          "service": "ext_threat_default_api-protection",
          "scope": "threat-assessment",
          "dispatch": "ThreatRequest{uri: request.path, is_authenticated: has(auth.identity), source_ip: request.source.address}",
          "callback": "response.threat_level >= 5 ? Deny(403, 'Blocked: threat level ' + string(response.threat_level)) : Allow()",
          "onError": "Allow()"
        }
      ]
    }
  ]
}
```
## Acceptance Criteria

- [ ] Team has reviewed the API design options
- [ ] A specific API approach has been chosen
- [ ] API signatures are defined in Go (interfaces, structs, methods)
- [ ] ThreatPolicy example code is written using the chosen API
- [ ] Wasm config generation strategy is documented
- [ ] CEL validation strategy is decided (reconcile-time vs runtime)
- [ ] Error handling approach is defined
- [ ] Design document approved by team

## References

- Existing service registration: `internal/controller/istio_auth_cluster_reconciler.go`
- Wasm config types: `internal/wasm/types.go`
- Extension registry: `internal/extension/registry.go`
- Current wasm action builder: `internal/controller/ratelimit_workflow_helpers.go:wasmActionFromLimit`

philbrookes:
## Summary

- Design proposal for SDK API to orchestrate custom policies' Actions (#1889)
- Introduces `RegisterActionMethod` (replaces `RegisterUpstreamMethod`), `Pipeline` API with `OnRequest`/`OnResponse` phases, and explicit `ActionType` discriminator for wasm-shim dispatch
- Builds on Phase 1 (RegisterUpstreamMethod) and Phase 2 (gRPC reflection / ProtoCache)

## Test plan

- [ ] Review design document for completeness and accuracy
- [ ] Verify deviations from issue metacode are justified
- [ ] Confirm action method name scoping analysis is sound
- [ ] Validate wasm Action struct changes are backwards compatible

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design spec for an Extension SDK action pipeline, describing ordered request/response pipelines, typed action variants (method invocation, allow, header injection, response-code) and runtime behaviours.
  * Documents registration, validation, storage and controller responsibilities, compatibility considerations and testing guidance for composing multiple actions per call.

### DIFF
--- a/doc/proposals/extensions-SDK-action-pipeline-design.md
+++ b/doc/proposals/extensions-SDK-action-pipeline-design.md
@@ -8,7 +8,7 @@ Reference: [Kuadrant/kuadrant-operator#1889](https://github.com/Kuadrant/kuadran
 
 ## Goals
 
-- Extension authors can register a named gRPC upstream method and invoke it from the data plane during request processing
+- Extension authors can register a named gRPC action method and invoke it from the data plane during request processing
 - Extension authors can compose ordered pipelines of actions that run on inbound requests and outbound responses
 - Request-phase actions can call a gRPC service and allow or deny the request based on the response, or allow/deny based on request attributes alone
 - Response-phase actions can add headers or override the HTTP status code
--- a/doc/proposals/extensions-SDK-action-pipeline-design.md
+++ b/doc/proposals/extensions-SDK-action-pipeline-design.md
@@ -30,7 +30,7 @@ The [issue metacode](https://github.com/Kuadrant/kuadrant-operator/issues/1889) 
 #### 1. `NewPipeline()` becomes `NewPipeline(policy)`
 
 The issue uses a no-arg constructor:
-```
+```text
 pipeline = kCtx.NewPipeline()
 ```
 
@@ -44,7 +44,7 @@ pipeline := kCtx.NewPipeline(policy)
 #### 2. `OnRequest` / `OnResponse` return `error` and take `ctx`
 
 The issue uses fire-and-forget calls:
-```
+```text
 pipeline.OnRequest(AllowAction{...})
 pipeline.OnResponse(AddHeadersAction{...})
 ```
@@ -64,7 +64,7 @@ if err := pipeline.OnRequest(ctx, GRPCMethodAction{...}); err != nil {
 
 ### Architecture Changes
 
-```
+```text
 Extension Controller                    Operator (gRPC server)
    │                                       │
    │── RegisterActionMethod ──────────────►│── Validate name uniqueness (per-policy)
@@ -189,8 +189,8 @@ type AllowAction struct {
 
 // AddHeadersAction adds headers to the response. Implements ResponseAction.
 type AddHeadersAction struct {
-    Predicate    []string          // CEL predicates — if any is false, skip this action
-    HeadersToAdd map[string]string // Header name → value (or CEL expression)
+    Predicate    []string // CEL predicates — if any is false, skip this action
+    HeadersToAdd string   // CEL expression evaluating to a map of headers to add
 }
 
 // WithResponseCodeAction modifies the HTTP response code. Implements ResponseAction.
@@ -241,7 +241,7 @@ message PipelineOnResponseRequest {
   kuadrant.v1.Policy policy = 1;
   ActionType action_type = 2;
   repeated string predicates = 3;
-  map<string, string> headers_to_add = 4;  // AddHeadersAction only
+  string headers_to_add = 4;               // CEL expression evaluating to a map of headers (AddHeadersAction only)
   int32 new_response_code = 5;             // WithResponseCodeAction only
 }
 ```
@@ -290,7 +290,7 @@ type Action struct {
     // New fields for pipeline actions
     Intention       string            `json:"intention,omitempty"`        // CEL expression
     ActionMethod    string            `json:"actionMethod,omitempty"`     // registered method name
-    HeadersToAdd    map[string]string `json:"headersToAdd,omitempty"`     // for add_headers
+    HeadersToAdd    string            `json:"headersToAdd,omitempty"`     // CEL expression → map of headers
     NewResponseCode *int              `json:"newResponseCode,omitempty"`  // for with_response_code
 
     SourcePolicyLocators []string `json:"sources,omitempty"`
@@ -384,8 +384,12 @@ New store for pipeline actions, ordered per-policy:
 type PipelineActionKey struct {
     Policy ResourceID
     Phase  PipelinePhase // "request" or "response"
-    Index  int           // insertion order
+    Index  int           // insertion order — assigned atomically per (Policy, Phase)
 }
+
+// Index allocation is atomic per (Policy, Phase) pair. The store maintains
+// a mutex-protected counter map keyed by (Policy, Phase) so that concurrent
+// OnRequest/OnResponse calls cannot produce duplicate or out-of-order indices.
 
 type PipelinePhase string
 
@@ -399,7 +403,7 @@ type PipelineActionEntry struct {
     Predicates      []string
     Intention       string            // GRPCMethodAction, AllowAction
     ActionMethod    string            // GRPCMethodAction only
-    HeadersToAdd    map[string]string // AddHeadersAction only
+    HeadersToAdd    string            // CEL expression → map of headers (AddHeadersAction only)
     NewResponseCode *int              // WithResponseCodeAction only
 }
 ```
@@ -502,7 +506,7 @@ func (p *PipelineImpl) OnResponse(ctx context.Context, action types.ResponseActi
             Policy:       convertPolicyToProtobuf(p.policy),
             ActionType:   extpb.ACTION_TYPE_ADD_HEADERS,
             Predicates:   a.Predicate,
-            HeadersToAdd: a.HeadersToAdd,
+            HeadersToAdd: a.HeadersToAdd, // CEL expression string
         })
         return err
     case types.WithResponseCodeAction:
@@ -563,7 +567,7 @@ func (r *ThreatPolicyReconciler) reconcileSpec(ctx context.Context, pol *v1alpha
     // cross-action data flow is future work)
     if err := pipeline.OnResponse(ctx, types.AddHeadersAction{
         Predicate:    []string{"response.headers['check_threat'] == ''"},
-        HeadersToAdd: map[string]string{"x-threat-checked": "true"},
+        HeadersToAdd: "{'x-threat-checked': 'true'}",
     }); err != nil {
         return calculateErrorStatus(pol, err), err
     }
@@ -576,7 +580,7 @@ func (r *ThreatPolicyReconciler) reconcileSpec(ctx context.Context, pol *v1alpha
 
 - **Name validation**: Action method names are validated to contain only alphanumeric characters, hyphens, and underscores, preventing injection of crafted identifiers into wasm config
 - **CEL validation at registration time**: Intention expressions are validated against proto schemas before being stored, preventing malformed CEL from reaching the wasm-shim
-- **Policy-scoped lifecycle**: Action methods and pipeline actions are tied to their owning policy and cleaned up via `ClearPolicy`
+- **Policy-scoped lifecycle**: Action methods and pipeline actions are tied to their owning policy and cleaned up via `ClearPolicyData`
 - **No user-controlled names in Envoy/wasm service config**: Cluster names and wasm service keys remain operator-generated (Phase 1 design)
 - **ActionType validation**: The operator rejects unknown action types; the wasm-shim ignores actions with unrecognised types
 
--- a/doc/proposals/extensions-SDK-action-pipeline-design.md
+++ b/doc/proposals/extensions-SDK-action-pipeline-design.md
@@ -8,11 +8,12 @@ Reference: [Kuadrant/kuadrant-operator#1889](https://github.com/Kuadrant/kuadran
 
 ## Goals
 
-- Rename `RegisterUpstreamMethod` to `RegisterActionMethod` with an explicit name and message template
-- Provide a `Pipeline` API with `OnRequest` and `OnResponse` phases that send actions immediately via gRPC
-- Support multiple action types: `GRPCMethodAction`, `AllowAction`, `AddHeadersAction`, `WithResponseCodeAction`
-- Translate pipeline actions into wasm `Action` entries with an explicit `ActionType` discriminator field
-- Validate CEL intention expressions against gRPC response proto schemas at registration time using Phase 2 ProtoCache
+- Extension authors can register a named gRPC upstream method and invoke it from the data plane during request processing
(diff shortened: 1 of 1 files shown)


alexsnaps (approved):

doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -0,0 +1,485 @@
...
+  - name: "abc123-hash"
+    routeRuleConditions:
+      hostnames: ["api.example.com"]
+    actions:
+      # GRPCMethodAction from ThreatPolicy
+      - service: ext-a1b2c3d4
+        scope: request
+        actionType: grpc_method
+        predicates: ["request.headers['check_threat'] == '1'"]
+        actionMethod: checkThreatLevel
```
crstrn13, line 266: I'd like to see how some arguments to this method are being sent. Also I am thinking if this is not duplicating the field in the service above?

doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -0,0 +1,485 @@
...
+        intention: "checkThreatLevelResponse.HeatLevel == 5"
```
crstrn13, line 267: I can't say I am sold on the `intention` name here, but I think that more important than that is how we end up doing the bindings in cel here. From this example, I get the assumption that there is a 'convention' that we bind the variable based on the actionMethod + "Response". Wondering if it wouldn't be better to be just response in this case? Much easier to change if an extension developer decides to reuse an action on a separate method.

philbrookes: With the future work of making these responses available to all actions (rather than just the one which has it as it's method), this naming is required to ensure: 1) The name is unique and 2) the correct response can be found.

doc/proposals/extensions-SDK-action-pipeline-design.md
philbrookes: good question, I've a PR to this doc [here](https://github.com/Kuadrant/kuadrant-operator/pull/1915/changes).

doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -0,0 +1,485 @@
...
+if err := pipeline.OnRequest(ctx, GRPCMethodAction{...}, AllowAction{...}); err != nil {
+    return reconcile.Result{}, err
+}
+```
+
+**Why:** Each `OnRequest`/`OnResponse` call sends a gRPC message to the operator immediately with all supplied actions. The operator may reject the batch (e.g. invalid CEL expression, unknown action method name). Go idiom requires surfacing these errors to the caller rather than silently swallowing them. The `ctx` parameter carries cancellation and deadline semantics for the gRPC call. There is no separate `Register()` step — each call is immediately effective.
+
+### Backwards Compatibility
+
+**Breaking change**: `RegisterUpstreamMethod` is renamed to `RegisterActionMethod` and its config type changes from `UpstreamConfig` to `ActionMethodConfig`. Extensions using the Phase 1 API must update their calls. Since the extension SDK is pre-1.0, this is acceptable. The gRPC proto RPC is also renamed; the old RPC is removed.
```
adam-cattermole, line 64: Not sure I'm the biggest fan of `RegisterActionMethod` as a name but I can live with it if others prefer it - I find it feels ambiguous and I think as a user I wouldn't know what this is for

doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -0,0 +1,485 @@
...
+
+Index allocation is atomic per `(Policy, Phase)` pair — the store maintains a mutex-protected counter so that concurrent `OnRequest`/`OnResponse` calls cannot produce duplicate or out-of-order indices.
+
+#### 3. Server-Side Handlers (internal/extension/manager.go)
+
+**RegisterActionMethod handler:**
+- Validates `Name` is non-empty and unique for this policy
+- Validates `URL`, `Service`, `Method` are set
+- Performs gRPC dial reachability check (from Phase 1)
+- Triggers gRPC reflection and caches descriptors (from Phase 2)
```
adam-cattermole, line 310: Just for ref, the dial has been removed as part of phase 2 - the reachability is 'confirmed' by the reflection

doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -0,0 +1,485 @@
...
+- Validates predicate CEL expressions
+- If any action fails validation, the entire request is rejected (no partial writes)
+- Appends all actions to `PipelineActionStore` with sequential indices for this policy's response phase
+- Triggers reconciliation
+
+#### 4. Client-Side Implementation (pkg/extension/controller/controller.go)
+
+- **`RegisterActionMethod`** on `ExtensionController` — converts `ActionMethodConfig` to the proto request, sends the RPC, and translates gRPC `Unavailable` status to `types.ErrUpstreamUnreachable`
+- **`NewPipeline`** on `ExtensionController` — creates a `PipelineImpl` bound to the policy and gRPC client. Performs no I/O.
+- **`PipelineImpl`** — `OnRequest` converts each `RequestAction` to a proto `RequestActionEntry`, sends them all in a single `PipelineOnRequestRequest` RPC. `OnResponse` does the same with `ResponseActionEntry` and `PipelineOnResponseRequest`. Unknown action types return an error before the RPC is sent.
```
adam-cattermole, line 336: This is interesting, I wonder if there'd have been an alternative where you 'finalise' the pipeline that makes a single gRPC call to provide the actions/pipeline for both request and response. Perhaps that'd be too confusing for a user and too error prone.

Also just wondering on this, if I call OnRequest twice in one reconcile with different actions what happens.. does last one win or are they chained based on the calls?

philbrookes: Calling pipeline.OnRequest twice would append both batches in order. For example:
```
pipeline.OnRequest(ctx, actionA, actionB)  // stored at indices 0, 1
pipeline.OnRequest(ctx, actionC)           // stored at index 2
```
All three actions end up in the request phase for that policy, in order. Each call is a separate gRPC message that gets validated and stored independently — if the second call fails validation, the first batch is already stored.

doc/proposals/extensions-SDK-action-pipeline-design.md
philbrookes: Yeah, the naming of this method is proving difficult, as it is a GRPC method that is called from Actions, I went with this option. Definitely open to hearing other options that might be more clear.

doc/proposals/extensions-SDK-action-pipeline-design.md
adam-cattermole: Okay, so is there some kind of identifier as part of the action so we know which calls correspond to the next reconcile loop?

i.e.
```
// reconcile 1
pipeline.OnRequest(ctx, actionA, actionB)  // stored at indices 0, 1
pipeline.OnRequest(ctx, actionC)           // stored at index 2
// reconcile 2
pipeline.OnRequest(ctx, actionA, actionB)  // stored/update at indices 0, 1??
pipeline.OnRequest(ctx, actionC)           // stored/update at index 2??
```

I assume we need some way of knowing that this `OnRequest` is the same `OnRequest` as before in a new reconcile loop and not a new call in the same reconcile loop..

I'm slightly concerned by this but it's not a huge problem; I do think from at least our point of view if we asked the author to call 'finalise'/'build'/'complete' or something on a pipeline where we make one gRPC call to the server with 'all of the actions this policy wants to include' it makes our lives a little easier but maybe there's a different option? I just think this might be brittle

philbrookes: yes, I see your point. This might need some rethinking when we get to implementing it.

Issue #946, opened 2024-10-18
[state-of-the-world reconciler] Error handling for final API server errors


guicassolato:
Handle API server errors that does not allow assuming presence of new events pending to be processed in the. E.g. `429 Too many requests`.

These kinds of errors may require re-triggering the reconciliation workflow if not handled at the source – i.e. in the reconciliation task where the error first originated.

Boomatang:
This is a know problem and it was agreed that for now it will exist. Planned to be addressed in 1.1.

Boomatang:
There are two generators of errors that we need to allow for, internal and external.
The external generator can create errors that we can or can not recover from automatically. 
An example of an error that we can automatically recover from is trying to updated resources that are out of date.
Out of date resource error is possibly the most common error we will hit.
This is automatically recoverable as the operator would have gotten the event for that resources update, and would be going to process that event on the next reconcile. 
An example of an external error we can not auto recover from is the API server is unreachable, or we are being rate limited by the API server.

Internal errors I don't see as have any automatically recoverable errors.
An example of an internal error would the failure to parse a json blob.

# External interface

External resources there can be a queue that is consumed by a process outside the workflows of the policy machinery.
This has a number of advantages.
- A workflow can assume the resources will be applied to the cluster, and do not have to deal with the error handling. 
- There is one place to handle processes responsible for handling external errors. 
- Workflows running in parallel don't need to get events from other workflows that are having issues with the API server.
- It allows for future improvements, where a batch of resource can be processed at the same time, consolidating updates to the same resources.

This would also be more inline with the structure that the operator currently uses for inbound events, where there is only one process responsible. 

The process acting against the external service can make a judgement calls in a better way. 
For example, if the operator is being rate limited, re-queuing the workflow may solve the issue, or create an ever ending loop of being rate limited.
The external service process can see that it is being rate limited, and rate.
The workflows can still process events, with adding any writes to the cluster to the external service process queue.

# Internal Errors

Internal errors are interesting as they may never be recoverable.
Poorly formatted json will never be parsed no matter how many times it is tried.

What should happen with an internal error is all children / dependent process need to stop or not be triggered.
Good logging around an internal error should be in place.

Workflows should only be rescheduled if there has not being any event post the triggering of the workflow.
For example if there is an internal error on a policy which would cause a rescheduling event, and that policy had already being updated on the cluster creating a new event.
The workflow that error should be dropped in favour of the new event.

[...]

PR #1913, opened 2026-04-19
internal/controller: return stale AuthConfig deletion errors so the controller retries

### LINKED ISSUE #1867: AuthConfigs reconciler doesn't handle deletion errors, causing potential resource leaks
### Problem

The AuthConfigs reconciler logs deletion errors but doesn't return them, preventing the controller from retrying failed deletions. This can lead to resource leaks over time.

**Location:** `internal/controller/authconfigs_reconciler.go:97-100`

```go
for _, authConfig := range staleAuthConfigs {
    if err := r.client.Resource(...).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
        logger.Error(err, "failed to delete authconfig object", "authconfig", ...)
        // TODO: handle error
    }
}
return nil  // Error is lost
```

### Impact

- Failed deletions won't trigger controller retry
- Stale AuthConfigs may persist in the cluster
- Resource leaks accumulate over time
- Affects both HTTP and gRPC route policies

### Expected Behaviour

Deletion errors should be returned to trigger controller reconciliation retry:

```go
var deleteErrors []error
for _, authConfig := range staleAuthConfigs {
    if err := r.client.Resource(...).Delete(...); err != nil {
        logger.Error(err, "failed to delete authconfig object", ...)
        deleteErrors = append(deleteErrors, fmt.Errorf("failed to delete stale authconfig %s/%s: %w",
            authConfig.GetNamespace(), authConfig.GetName(), err))
    }
}
return errors.Join(deleteErrors...)
```

### Additional Context

Discovered during code review of #1855, but this is a pre-existing issue affecting both HTTP and gRPC routes.

SAY-5:
## What

`AuthConfigsReconciler.Reconcile` in `internal/controller/authconfigs_reconciler.go` sweeps out stale `AuthConfig` objects but swallows the error:

```go
for _, authConfig := range staleAuthConfigs {
    if err := r.client.Resource(...).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
        logger.Error(err, "failed to delete authconfig object", "authconfig", ...)
        // TODO: handle error
    }
}
return nil
```

The controller runtime only re-queues a reconcile when `Reconcile` returns a non-nil error. So a failed delete was a one-shot log line and the stale resource was left in-cluster until some *other* event happened to trigger another reconcile pass. Under API-server pressure, a transient permission glitch, or a brief conflict, that's long enough for stale AuthConfigs to accumulate, and for Authorino to keep enforcing policy fragments the operator already tried to remove.

Per #1867.

## Fix

Collect per-delete failures and return them through `errors.Join` at the end of the function:

```go
var deleteErrors []error
for _, authConfig := range staleAuthConfigs {
    if err := r.client.Resource(...).Delete(...); err != nil {
        logger.Error(err, "failed to delete authconfig object", "authconfig", ...)
        deleteErrors = append(deleteErrors, fmt.Errorf("failed to delete stale authconfig %s/%s: %w",
            authConfig.GetNamespace(), authConfig.GetName(), err))
    }
}

return errors.Join(deleteErrors...)
```

- Successful deletes in the same loop still complete; only the failed ones are surfaced.
- The existing per-object `logger.Error(...)` call is kept so the operator log still has the structured context.
- `errors.Is`/`errors.As` in any upstream error handlers continue to work because we wrap the original error with `%w`.

`errors.Join(nil...)` returns `nil`, so the success path is unchanged from the previous `return nil`.

Fixes #1867

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during stale configuration cleanup: deletion errors encountered while removing obsolete AuthConfig resources are now accumulated and reported collectively, while not-found cases are ignored. As a result, cleanup failures are surfaced instead of being silently skipped, and the reconciler may now return an error when stale configuration deletions fail for reasons other than not-found.

### DIFF
--- a/internal/controller/authconfigs_reconciler.go
+++ b/internal/controller/authconfigs_reconciler.go
@@ -93,14 +93,16 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 		_, desired := desiredAuthConfigs[k8stypes.NamespacedName{Name: o.GetName(), Namespace: o.GetNamespace()}]
 		return o.GroupVersionKind().GroupKind() == kuadrantauthorino.AuthConfigGroupKind && labels.Set(o.(*controller.RuntimeObject).GetLabels()).AsSelector().Matches(AuthObjectLabels()) && !desired
 	})
+	var deleteErrors []error
 	for _, authConfig := range staleAuthConfigs {
 		if err := r.client.Resource(kuadrantauthorino.AuthConfigsResource).Namespace(authConfig.GetNamespace()).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
 			logger.Error(err, "failed to delete authconfig object", "authconfig", fmt.Sprintf("%s/%s", authConfig.GetNamespace(), authConfig.GetName()))
-			// TODO: handle error
+			deleteErrors = append(deleteErrors, fmt.Errorf("failed to delete stale authconfig %s/%s: %w",
+				authConfig.GetNamespace(), authConfig.GetName(), err))
 		}
 	}
 
-	return nil
+	return errors.Join(deleteErrors...)
 }
 
 // reconcileAuthConfigForPath reconciles the AuthConfig for a single effective policy path.
--- a/internal/controller/authconfigs_reconciler.go
+++ b/internal/controller/authconfigs_reconciler.go
@@ -15,6 +15,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/propagation"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	k8stypes "k8s.io/apimachinery/pkg/types"
@@ -96,6 +97,12 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	var deleteErrors []error
 	for _, authConfig := range staleAuthConfigs {
 		if err := r.client.Resource(kuadrantauthorino.AuthConfigsResource).Namespace(authConfig.GetNamespace()).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				// Already gone - idempotent success, matching the pattern
+				// used in effective_dnspolicies_reconciler.go:311 and
+				// effective_tls_policies_reconciler.go:131.
+				continue
+			}
 			logger.Error(err, "failed to delete authconfig object", "authconfig", fmt.Sprintf("%s/%s", authConfig.GetNamespace(), authConfig.GetName()))
 			deleteErrors = append(deleteErrors, fmt.Errorf("failed to delete stale authconfig %s/%s: %w",
 				authConfig.GetNamespace(), authConfig.GetName(), err))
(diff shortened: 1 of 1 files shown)


internal/controller/authconfigs_reconciler.go
```diff
@@ -93,14 +94,22 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 		_, desired := desiredAuthConfigs[k8stypes.NamespacedName{Name: o.GetName(), Namespace: o.GetNamespace()}]
 		return o.GroupVersionKind().GroupKind() == kuadrantauthorino.AuthConfigGroupKind && labels.Set(o.(*controller.RuntimeObject).GetLabels()).AsSelector().Matches(AuthObjectLabels()) && !desired
 	})
+	var deleteErrors []error
 	for _, authConfig := range staleAuthConfigs {
 		if err := r.client.Resource(kuadrantauthorino.AuthConfigsResource).Namespace(authConfig.GetNamespace()).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
+			if apierrors.IsNotFound(err) {
+				// Already gone - idempotent success, matching the pattern
+				// used in effective_dnspolicies_reconciler.go:311 and
+				// effective_tls_policies_reconciler.go:131.
```
maksymvavilov: Nit: a grumpy side of mine sees line numbers and yells: "This will go out of date in a month!"; It is a fairly well-known pattern, and I see no added benefit of the comments. If this were me, I'd remove them, but no engineer died of having more comments afaik - this should not hold the merger.

maksymvavilov (dismissed):
@SAY-5 up to you to act or ignore the nit. In any case this is good

maksymvavilov (approved):
Good work! Thank you

guicassolato (changes requested):
The solution proposed here is consistent with the hints left in #1867. However, I think I would favour a more holistic approach to tackle the problem of non-blocking, final errors raised by the API server (#946).

The current State of the World reconciliation has several locations where API server errors are logged but not propagated (marked with `// TODO: handle error`). The challenge is:
- **Returning errors aborts the workflow** - If you return an error from a reconciler, the entire workflow stops, preventing other reconcilers from running
- **Ignoring errors causes resource leaks** - Stale resources accumulate when deletion/update failures are swallowed
- **Standard controller retry waits for new events** - Without returning an error, there's no automatic retry until another API server event arrives
- **Risk of infinite loops** - Naive retry mechanisms could create reconciliation loops

I think we should introduce a reconciliation error registry that collects all non-blocking errors during workflow execution and triggers a single deferred reconciliation after the workflow completes successfully.

It could be added to the [final steps of the workflow](https://github.com/Kuadrant/kuadrant-operator/blob/b727d75f2d28730076ddde2a1cef5c4487b73751/internal/controller/state_of_the_world.go#L806-L826):

```go
func (b *BootOptionsBuilder) finalStepsWorkflow() *controller.Workflow {
	workflow := &controller.Workflow{
			// ... existing tasks ...
		},
		Postcondition: // Add error retry reconciler as the final task
			traceReconcileFunc("finalize.error_retry", NewErrorRetryReconciler().Reconcile),
	}
    
	// ... rest of the function
}
```

Then, in the AuthConfig reconciler (and similar):

```go
func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
	// ... 

	// Get or create error registry from state
	errorRegistry := GetOrCreateErrorRegistry(state)

	for _, authConfig := range staleAuthConfigs {
		if err := r.client.Resource(kuadrantauthorino.AuthConfigsResource).Namespace(authConfig.GetNamespace()).Delete(ctx, authConfig.GetName(), metav1.DeleteOptions{}); err != nil {
			logger.Error(…)

			// Record the error for deferred retry instead of returning it
			errorRegistry.Record(…)
	}

	// ... 
}
```

Finally, the implementation of a hypothetical `ErrorRetryReconciler` can get complex, but it shoould:
1. Check if the registry contains any error
2. Merges the tracked errors into an error tracker struct that lives outside the workflow

The external error tracker can be a wrapper for Policy Machinery's controller that implements controller-runtime's `Reconcile`. It:
1. Keeps count of the accrued number of occurrences for each type of error
2. Triggers a retry (by returning a controller-runtime's `Result`)

PR #1914, opened 2026-04-20
oidc-policy: compile claim predicates as equality on scalar JWT claims

### LINKED ISSUE #1625: OIDCPolicy Authorization not working with string/boolean claims
**Describe the bug**
When adding in the `auth` section of the _OIDCPolicy_ a claim that requires an equality check, such as in :

```yaml
...
auth:
  claims:
    email_verified: 'true'
```

In the authorino logs we can see that the request was Unauthorized and the Denied reason was `no such overload`

**To Reproduce**
1. Follow the OIDCPolicy & Keycloak guide (pending link coming soon)
2. Change the _OIDCPolicy_  adding the auth secition as in:
```yaml
auth:
  claims:
    email_verified: 'true'
```

**Expected behavior**
The call was authorized, I could successfully authorize my app request.

**Notes**
* The predicate that results from the OIDCPolicy is `'"true" in auth.identity.email_verified'` when it should be `auth.identity.email_verified == "true"`
* The way the OIDCPolicy reconciler is building the cel expressions is wrong, it only works with validating presence of a value in a list/map

SAY-5:
## What

`oidcpolicy_reconciler.go` lowers every `auth.claims` entry into a CEL predicate via:

```go
Predicate: fmt.Sprintf(`"%s" in auth.identity.%s`, v, k),
```

which expands to e.g. `"true" in auth.identity.email_verified`. That shape only type-checks when the claim is a list/map value; on *scalar* claims (`sub`, `iss`, `email_verified`, `azp`, `name`, the majority of OIDC standard claims) authorino denies the request with:

```
{"reason": "no such overload"}
```

So any `OIDCPolicy` whose `auth.claims` references a scalar claim silently denies every request. Per #1625.

## Fix

Claims are declared as `map[string]string`, so the caller's intent is unambiguous: "claim X equals string V". Compile that directly as equality instead of set containment:

```go
Predicate: fmt.Sprintf(`auth.identity.%s == "%s"`, k, v),
```

- Scalar claims now match (`auth.identity.email_verified == "true"`).
- The reporter in #1625 explicitly calls out this as the expected predicate shape.
- The old `in` form never worked end-to-end against a scalar, so this isn't breaking a behaviour users could have been relying on, it's fixing the code path to match the declared API.

## Not in this PR

If/when list-valued claims (e.g. `groups`, `roles`) need their own predicate style, that's a separate API shape, either widening `Claims` to `map[string][]string` or adding a dedicated `containsClaims` field. Either will sit alongside this equality form rather than replacing it.

Fixes #1625

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed JWT claim matching in OpenID Connect authorisation policies so scalar (single-value) claims are evaluated by direct equality, while list-valued claims are matched by membership with proper type handling to avoid false negatives.

* **Tests**
  * Added unit tests covering scalar and list claim matching, boolean-as-string handling and membership edge cases.

### DIFF
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -303,7 +303,7 @@ func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*ku
 		for k, v := range claims {
 			authPatterns = append(authPatterns, authorinov1beta3.PatternExpressionOrRef{
 				CelPredicate: authorinov1beta3.CelPredicate{
-					Predicate: fmt.Sprintf(`"%s" in auth.identity.%s`, v, k),
+					Predicate: fmt.Sprintf(`auth.identity.%s == "%s"`, k, v),
 				},
 			})
 		}
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -288,8 +288,13 @@ func (r *OIDCPolicyReconciler) reconcileHTTPRoute(ctx context.Context, desired *
 // membership. The combined expression covers both because CEL `==` between mismatched
 // types yields false rather than erroring, so the `in` branch only applies when the
 // claim is actually a list.
+//
+// Accessing a missing field with dot notation raises a no_such_field error in CEL
+// rather than yielding false, so the predicate is guarded with has() on the identity
+// and the claim key; a token that lacks the claim then evaluates to false instead of
+// failing the whole authorization expression.
 func claimPredicate(k, v string) string {
-	return fmt.Sprintf(`auth.identity.%s == "%s" || (type(auth.identity.%s) == list && "%s" in auth.identity.%s)`, k, v, k, v, k)
+	return fmt.Sprintf(`has(auth.identity) && has(auth.identity.%s) && (auth.identity.%s == "%s" || (type(auth.identity.%s) == list && "%s" in auth.identity.%s))`, k, k, v, k, v, k)
 }
 
 func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*kuadrantv1.AuthPolicy, error) {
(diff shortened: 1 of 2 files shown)


cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -301,9 +301,15 @@ func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*ku
 
 	if len(claims) > 0 {
 		for k, v := range claims {
+			// Build an equality predicate against the JWT claim. The prior
+			// shape (`"v" in auth.identity.k`) required the claim to be a
+			// list/map and failed outright on scalar claims such as `sub`,
+			// `iss`, or `email_verified` with authorino reporting "no such
+			// overload", so the OIDCPolicy auth check never matched anything
+			// for configurations that named a scalar claim (#1625).
```
maksymvavilov: I'd rather have this comment in a PR or in the PR description. Once merged, it seems a bit out of place.  Would you be able to remove it from the commit?

maksymvavilov (changes requested):
looks good. @didierofrivia for a sanity check

didierofrivia (changes requested):
Hi @SAY-5, thanks for the contribution! The fix should also work with list values in claims. Could you write some tests when the final fix is in place? thanks!

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -282,6 +282,16 @@ func (r *OIDCPolicyReconciler) reconcileHTTPRoute(ctx context.Context, desired *
...
 }
 
+// claimPredicate builds the CEL predicate that matches a configured claim against
+// the JWT identity. A scalar claim (the common case, e.g. email, sub, email_verified)
+// is matched by equality; a list-valued claim (e.g. groups, roles) is matched by
+// membership. The combined expression covers both because CEL `==` between mismatched
+// types yields false rather than erroring, so the `in` branch only applies when the
+// claim is actually a list.
+func claimPredicate(k, v string) string {
+	return fmt.Sprintf(`auth.identity.%s == "%s" || (type(auth.identity.%s) == list && "%s" in auth.identity.%s)`, k, v, k, v, k)
```
didierofrivia: `ClaimPredicate` function [should check existence](https://github.com/Kuadrant/kuadrant-operator/pull/1914/changes#r3398464053) for `claim.identity` and `claim.identity.%s` before doing the checks

didierofrivia (approved):
Thanks for the contribution, great work 🙏🏼

maksymvavilov (approved):

PR #1915, opened 2026-04-20
doc: document MessageTemplate processing in action pipeline design

philbrookes:
## Summary
- Adds a **Message Template Processing** section to the action pipeline design doc, explaining how `MessageTemplate` flows from registration through the operator to the wasm-shim at request time
- Updates the wasm config example to include `messageTemplate` on the service definition
- Updates the `RegisterActionMethod` handler description to include template validation steps
- Adds a note to the Wasm Changes section linking the new `messageTemplate` service field to the processing section

## Test plan
- [x] Doc-only change, no code affected

## Summary by CodeRabbit

* **Documentation**
  * Added specification for a new message template configuration to construct upstream gRPC requests from request context.
  * Clarified template processing: parsing, CEL expression validation, template-to-proto validation, storage and propagation to runtime configuration.
  * Updated action model to include an explicit action-type discriminator and streamlined example payloads.

### DIFF
--- a/doc/proposals/extensions-SDK-action-pipeline-design.md
+++ b/doc/proposals/extensions-SDK-action-pipeline-design.md
@@ -254,7 +254,9 @@ services:
     grpcService: "threat.ThreatService"
     grpcMethod: "CheckThreatLevel"
     messageTemplate:
-      values: "request.headers['phil']"
+      uri: "request.path"
+      is_authenticated: "has(auth.identity)"
+      source_ip: "source.address"
 
 actionSets:
   - name: "abc123-hash"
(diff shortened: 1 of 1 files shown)


doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -252,6 +253,8 @@ services:
     timeout: 100ms
     grpcService: "threat.ThreatService"
     grpcMethod: "CheckThreatLevel"
+    messageTemplate:
+      values: "request.headers['phil']"
```
adam-cattermole, line 257: I'm not sure I understand what this `values` is? I thought we were going to use CEL to construct based on the type?

i.e something like this using our [threat service proto](https://github.com/Kuadrant/threat-assessment-service/blob/9c9940e66a0895ff9aa4d3fa801c636d5c2ea555/api/threat/v1/threat.proto#L11-L15) example:

```
ThreatRequest { uri: request.path, is_authenticated: has(auth.identity), source_ip: source.address }
```

Where the start is the type of the struct, the fields are the exact named fields in the protobuf def. I'm not sure how in the `messageTemplate` above we'd map this to fields? is that for the first field, what if you want to omit field 1 and have that be the default from the protobuf?

philbrookes: The values here is derived from [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/proposals/extensions-SDK-action-pipeline-design.md?plain=1#L357) `RegisterActionMethod`. 

But I think it'd make more sense for the example usage, and this example config to use the real threat service values. I'll update it and hopefully that will make this more obvious.

philbrookes:
closed in favour of: https://github.com/Kuadrant/kuadrant-operator/pull/1917

doc/proposals/extensions-SDK-action-pipeline-design.md
adam-cattermole: Right I see, so we'd be kinda 'extracting' the fields from the struct in a way that the user can define in this `messageTemplate` and then rebuilding it in a CEL expression. I'm not against it - Is there any downsides you can think of just having them define the message themselves directly like the above in CEL? or do you think it's better for them to define each field and then we build the CEL with the struct type

PR #1917, opened 2026-04-21
doc: document MessageTemplate processing in action pipeline design

philbrookes:
## Summary
- Adds a **Message Template Processing** section to the action pipeline design doc, explaining how `MessageTemplate` flows from registration through the operator to the wasm-shim at request time
- Updates the wasm config example to include `messageTemplate` on the service definition
- Updates the `RegisterActionMethod` handler description to include template validation steps
- Adds a note to the Wasm Changes section linking the new `messageTemplate` service field to the processing section

## Test plan
- [x] Doc-only change, no code affected

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation on the new messageTemplate capability for wasm service definitions, describing the mapping of protocol buffer field names to CEL expressions.
  * Included detailed sections on validation and evaluation processes across operator registration, reconciliation, and request runtime stages.
  * Updated configuration examples to reflect the current messageTemplate field mapping structure and usage.

PR #1922, opened 2026-04-22
doc: messageTemplate as opaque string in wasm config

philbrookes:
## Summary
- Changed `messageTemplate` in wasm config from a structured map (field → CEL pairs) to an opaque JSON string passed through from the SDK
- Operator validates at registration time but passes the string through as-is — the wasm-shim parses it at request time
- Fewer transformations, simpler config serialization, and template format changes only affect SDK + wasm-shim

Addresses review feedback from @adam-cattermole on #1917.

## Test plan
- [ ] Verify design doc renders correctly on GitHub

## Summary by CodeRabbit

* **Documentation**
  * Updated extension SDK documentation to clarify message template representation and processing across the action pipeline. Documentation now includes clearer guidance on configuration, storage semantics, and request-time behaviour, along with updated examples and changelog entries.

### DIFF
--- a/doc/proposals/extensions-SDK-action-pipeline-design.md
+++ b/doc/proposals/extensions-SDK-action-pipeline-design.md
@@ -229,7 +229,7 @@ Two `ThreatPolicy` instances targeting the same Gateway both register `"checkThr
 
 #### New ActionType Field
 
-The wasm service definition gains a `messageTemplate` field — a JSON object where keys are proto field names and values are CEL expressions. The wasm-shim evaluates these expressions against the request context at request time to construct the gRPC request message (see [Message Template Processing](#message-template-processing) below).
+The wasm service definition gains a `messageTemplate` field — a JSON string where keys are proto field names and values are CEL expressions. The operator passes this string through to the wasm config as-is (after validation). The wasm-shim parses the JSON, evaluates the CEL expressions against the request context at request time, and uses the results to construct the gRPC request message (see [Message Template Processing](#message-template-processing) below).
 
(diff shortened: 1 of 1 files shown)


doc/proposals/extensions-SDK-action-pipeline-design.md
```diff
@@ -253,10 +253,7 @@ services:
     timeout: 100ms
     grpcService: "threat.ThreatService"
     grpcMethod: "CheckThreatLevel"
-    messageTemplate:
-      uri: "request.path"
-      is_authenticated: "has(auth.identity)"
-      source_ip: "source.address"
+    messageTemplate: '{"uri": "request.path", "is_authenticated": "has(auth.identity)", "source_ip": "source.address"}'
```
adam-cattermole, line 256: I think this needs the struct type and fixed quotation marks to be valid CEL:

```suggestion
    messageTemplate: "ThreatRequest { uri: request.path, is_authenticated: has(auth.identity), source_ip: source.address }"
```

philbrookes: Is this how we'd expect that to look in the call to RegisterActionMethod as well?

e.g.
```go
    err := kCtx.RegisterActionMethod(ctx, pol, types.ActionMethodConfig{
        Name:            "checkThreatLevel",
        URL:             threatServiceURL,
        Service:         "threat.ThreatService",
        Method:          "CheckThreatLevel",
        MessageTemplate: "ThreatRequest { uri: request.path, is_authenticated: has(auth.identity), source_ip: source.address }",
    })
```

adam-cattermole: I feel like that's potentially reasonable.. but I also am not 100% sure where these bits will end up - i.e. whether it would be in `RegisterActionMethod` vs the action itself, so I guess I tend to think we should just see how these bits end up and be ready to change.

Issue #1919, opened 2026-04-21
Status Enforced Scaling Issue


Boomatang:
The [enforcedCondition](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/controller/auth_policy_status_updater.go#L108) function call has poor scaling in relation to the number of resources in the topology. In a recent investigation I found the following timings for this function call.

Timing collected in seconds, with the operator using 3.5 CPU.
| No. Resources | Enforced Function Call | Update to API Server | Total time of status update|
|---|---|---|---|
| 1   | 0.00027 | 0.01121 | 0.01186 |
| 3   | 0.00057 | 0.00582 | 0.00665 |
| 30  | 0.00724 | 0.00493 | 0.01236 |
| 300 | 0.38046 | 0.00419 | 0.38484 |

While the timing in the above are the average of a small sample set, it does show how the time require is scaling exponentially.

In the setup a resource is a HTTPRoute with 1 AuthPolicy and 1 RateLimitPolicy attached. The status updating for the RateLimitPolicy show a similar problem, but at the time I was investigating something else, and did not make the required time recordings.

Boomatang:
I have done some initial set up for checking the improvements by introducing the benchmark tests. As there is no existing benchmark test in the repo I am not going to open a PR to add them to the main, but they can work as a starting point. The commit with the change can be found [here](https://github.com/Boomatang/kuadrant-operator/tree/perf_status_enforced).

A lot of this was generated, but does seem to be working as expected. The [BENCHMARKING-README.md](https://github.com/Boomatang/kuadrant-operator/blob/perf_status_enforced/BENCHMARKING-README.md) give good steps for setting up the test. The doc also outlines how to get a flame graph of the test configured.

These are the result that I got.

```sh
SNAPSHOTS_DIR=$PWD/snapshots go test -tags=prof -bench=BenchmarkEnforcedCondition ./internal/controller/ -benchmem
goos: linux
goarch: amd64
pkg: github.com/kuadrant/kuadrant-operator/internal/controller
cpu: AMD Ryzen 7 PRO 7840HS w/ Radeon 780M Graphics 
BenchmarkEnforcedCondition_RealCluster/10_authpolicies-16         	    259	  4589858 ns/op	4458813 B/op	  94428 allocs/op
BenchmarkEnforcedCondition_RealCluster/150_authpolicies-16        	      1	9835236231 ns/op	10562762880 B/op	209728698 allocs/op
[...]

PR #1923, opened 2026-04-22
Fix WasmPlugin comparison to allow order-insensitive equality checks

Boomatang:
# Summary

This PR fixes a critical bug in the WasmPlugin comparison logic that was causing reconciliation instability when a "large" number of policies were applied to the cluster. With 20 or more RateLimitPolicies and AuthPolicies, the reconcile loop could enter an unstable state and never converge.

# Problem

The WasmPlugin equality comparison was performing strict order-sensitive comparisons on the `ConditionalData` slice within `Action` objects. However, the order of `ConditionalData` items is not semantically significant - they represent independent conditions that can be evaluated in any order without affecting the outcome.

When multiple policies are merged into a single WasmPlugin configuration, the order in which conditional data appears can vary between reconciliation cycles due to Go's map iteration order being non-deterministic. This caused the operator to detect phantom "changes" in the WasmPlugin spec, triggering unnecessary updates and preventing the system from reaching a stable state.

# Solution

Changed the `Action.EqualTo()` method to use order-insensitive comparison for the `ConditionalData` slice using `slices.ContainsFunc()`. This ensures that two Actions are considered equal if they contain the same conditional data, regardless of ordering.

The fix includes:
- Updated `Action.EqualTo()` to use `slices.ContainsFunc()` instead of index-based comparison for `ConditionalData`
- Added comprehensive documentation explaining the semantic equality semantics for all WASM types
- Added package-level documentation in `internal/wasm/doc.go` explaining the semantic equality concept
- Added extensive test coverage (932+ new test cases) covering various comparison scenarios

# Testing

- Added comprehensive unit tests in `internal/wasm/types_test.go` covering all equality comparison edge cases
- Added integration tests in `internal/controller/istio_extension_reconciler_test.go` verifying WasmPlugin comparison behavior
- Added tests in `internal/istio/utils_test.go` for order-independent comparisons

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for WASM plugin equality logic and target reference matching.
  * Added comprehensive unit tests for configuration comparison semantics.

* **Documentation**
  * Added package-level documentation for internal WASM configuration handling.

* **Chores**
  * Minor type refinements in configuration structures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-operator/pull/1923)

### DIFF
--- a/internal/wasm/types.go
+++ b/internal/wasm/types.go
@@ -42,7 +42,24 @@ func (c *Config) ToJSON() (*apiextensionsv1.JSON, error) {
 	return &apiextensionsv1.JSON{Raw: configJSON}, nil
 }
 
+// EqualTo performs semantic equality comparison between two Config instances.
+// This is not a strict equality check - it considers two configs equal if they are
+// functionally equivalent, even if some collection orderings differ.
+//
+// Order-sensitive fields:
+//   - ActionSets: Order matters - compared by index position
+//   - RequestData: Map comparison (order doesn't apply)
+//   - Services: Map comparison (order doesn't apply)
+//   - DescriptorService: String comparison
+//   - Observability: Strict equality via nested EqualTo
+//
+// Note: ActionSets order is significant because it affects the evaluation order
+// in the data plane.
 func (c *Config) EqualTo(other *Config) bool {
+	if other == nil {
+		return false
+	}
+
 	if len(c.RequestData) != len(other.RequestData) || len(c.Services) != len(other.Services) || len(c.ActionSets) != len(other.ActionSets) {
 		return false
 	}
@@ -207,6 +224,15 @@ func (s *ActionSet) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// EqualTo performs semantic equality comparison between two ActionSet instances.
+//
+// Order-sensitive fields:
+//   - Name: String comparison
+//   - RouteRuleConditions: Uses RouteRuleConditions.EqualTo (which has its own ordering rules)
+//   - Actions: Order matters - compared by index position
+//
+// Note: Actions order is significant because it affects the evaluation order
+// in the data plane.
 func (s *ActionSet) EqualTo(other ActionSet) bool {
 	if s.Name != other.Name || !s.RouteRuleConditions.EqualTo(other.RouteRuleConditions) || len(s.Actions) != len(other.Actions) || len(s.TypedActions) != len(other.TypedActions) {
 		return false
@@ -234,6 +260,14 @@ type RouteRuleConditions struct {
 	Predicates []string `json:"predicates,omitempty"`
 }
 
+// EqualTo performs semantic equality comparison between two RouteRuleConditions instances.
+//
+// Order-sensitive fields:
+//   - Hostnames: Order matters - compared by index position
+//   - Predicates: Order matters - compared by index position
+//
+// Note: Hostname order is preserved as it may reflect priority or specificity,
+// while predicates are order-insensitive as they are evaluated independently.
 func (r *RouteRuleConditions) EqualTo(other RouteRuleConditions) bool {
 	if len(r.Hostnames) != len(other.Hostnames) || len(r.Predicates) != len(other.Predicates) {
 		return false
@@ -304,6 +338,15 @@ func (a *Action) HasAuthAccess() bool {
 	return false
 }
 
+// EqualTo performs semantic equality comparison between two ConditionalData instances.
+// Note: This has mixed ordering semantics for different fields.
+//
+// Order-sensitive fields:
+//   - Predicates: Order matters - strict slice equality
+//   - Data: Order matters - compared by index position
+//
+// Note: Both predicates and data order are significant as they may affect
+// evaluation order in the data plane.
 func (c *ConditionalData) EqualTo(other ConditionalData) bool {
 	if len(c.Predicates) != len(other.Predicates) || len(c.Data) != len(other.Data) {
 		return false
@@ -322,6 +365,17 @@ func (c *ConditionalData) EqualTo(other ConditionalData) bool {
 	return true
 }
 
+// EqualTo performs semantic equality comparison between two Action instances.
+// Note: This has mixed ordering semantics for different fields.
+//
+// Order-insensitive fields:
+//   - ConditionalData: Checks that the same conditional data exists, regardless of order
+//
+// Order-sensitive fields (strict equality):
+//   - Scope: String comparison
+//   - ServiceName: String comparison
+//   - Predicates: Strict slice equality - order matters
+//   - SourcePolicyLocators: Strict slice equality - order matters
 func (a *Action) EqualTo(other Action) bool {
 	if a.Scope != other.Scope ||
 		a.ServiceName != other.ServiceName ||
@@ -338,7 +392,7 @@ func (a *Action) EqualTo(other Action) bool {
 	}
 
 	for i := range a.ConditionalData {
-		if !a.ConditionalData[i].EqualTo(other.ConditionalData[i]) {
+		if !slices.ContainsFunc(other.ConditionalData, a.ConditionalData[i].EqualTo) {
 			return false
 		}
 	}
@@ -347,12 +401,12 @@ func (a *Action) EqualTo(other Action) bool {
 }
 
 type DataType struct {
-	Value interface{}
+	Value any
 }
 
 func (d *DataType) UnmarshalJSON(data []byte) error {
 	// Precisely one of "static", "selector" must be set.
-	types := []interface{}{
+	types := []any{
 		&Static{},
 		&Expression{},
 	}
--- a/internal/controller/istio_extension_reconciler_test.go
+++ b/internal/controller/istio_extension_reconciler_test.go
@@ -6,7 +6,11 @@ import (
 	"context"
 	"testing"
 
+	_struct "google.golang.org/protobuf/types/known/structpb"
 	"gotest.tools/assert"
+	istioextensionsv1alpha1 "istio.io/api/extensions/v1alpha1"
+	istiov1beta1 "istio.io/api/type/v1beta1"
+	istioclientgoextensionv1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
 
 	"github.com/kuadrant/kuadrant-operator/internal/wasm"
 )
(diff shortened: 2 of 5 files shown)


adam-cattermole:
I'm slightly concerned by this change, unless something has changed, the ordering of ActionSets in the wasm-shim does matter by design. The operator ensures they're in the correct order and the 'first matching one wins' - this is to support defaults/overrides. Actions are also executed in order sequentially. Just highlighting this here in case this wasn't taken into account

Boomatang:
@adam-cattermole I find in concerning also. I can share two version of the wasmPlugin  CR that were generated before this fix. The version are one generation apart and contain all the data from the cluster. The run had a 100 routes, AuthPolicy and RateLimitPolicy. The reconcile never becomes stable due the wasmPlugin generation is not deterministic. When I was running a test with 12 sets of each resource the generation of the wasmPlugin reached 5244, but did become stable. 

[a.json](https://github.com/user-attachments/files/26966193/a.json)
[b.json](https://github.com/user-attachments/files/26966194/b.json)

eguzki:
For some lists, the order matters, while for other lists the order does not matter.

Some examples of lists where order matters: actionsets, actions within the actionset.

Some examples of lists where order does not matter: conditional data objects, data of the conditional data objects, predicates of the conditional data objects.

The Kuadrant operator must ensure the correct order for those list where order matters to comply with the expectations of the user who owns the policies. Failing to do so is a clear bug in the operator.

The examples shared by Jim are very illustrative and clearly show that the list of conditional data objects within an action is rendered in different orders in two different rendering processes. This, together with the implemented strict, order-sensitive equality check, leads to the observed issue that must be fixed.

For lists where the order does not matter, there are two approaches we can implement:

a) Have the Kuadrant operator sort them in a deterministic way when computing the WASM configuration. Then, during reconciliation and comparison, the operator can use strict equality or, alternatively, flexible equality, meaning that equality can be performed in an order-insensitive way.

b) Have the Kuadrant operator render them in arbitrary order. Then, the only valid equality check for those lists would be order-insensitive.

Both are valid unless proven otherwise. I would prefer a deterministic way of rendering the WASM configuration for those lists.

adam-cattermole:
@eguzki yep I'm in agreement, I think `a)` sounds like the best approach to address

guicassolato:
It looks like the problem we're trying to solve here is that, by generating non-deterministic configs, the operator oftentimes runs into long and hard-to-predict reconciliation looping.

I agree with @eguzki the correct way to approach this is by finding the source of the non-determinism and fixing there. This is usually a Go `map` used to enforces uniqueness, an iteration over a list of resources returned by an API that itself does not guarantee elements are sorted, etc.

By loosening the comparison of a few collections of items into set equality (or "semantic equality", as called), we may introduce bugs that are very hard to trace back later on. In contrast, if all collections are treated as order-sensitive – even the ones that do not have to be –, the overhead of keeping only sorted lists tends to pay off against the burden of debugging a misbehaviour caused by having two kinds of collections (order-sensitive and order-insensitive) and one memory state that is not guaranteed to match the cluster state.

Boomatang:
I may have being heavy handed in my initial approach. I have refined it again and found a smaller set of changes I’ve streamlined this PR to provide a faster, less invasive fix for the issue.

While I agree with @eguzki that addressing the root cause (determinism in wasmplugin generation) is the ideal path, we need an immediate solution for our RHCL users. I've focused this PR on that 'fast fix.' If anyone is interested in tackling the root cause in a separate PR, that would be a great improvement, but I need to focus my current bandwidth on other pressing issues.required to do the fix. 

On what the correct approach maybe layout by @eguzki, yes the generation of the wasmplugin should have being determinism, but it is not, and yes fixing that issue is possible the correct approach. The change I have done here how is the approach I have taken, as it is the fast fix. This affects a RHCL user, and is one of many issues that have being found. 

If some one else wants to go fix this at the root in the generation of the wasmshim, please do, but I will be putting my effort else where.

internal/wasm/types.go
```diff
@@ -155,6 +181,16 @@ type RouteRuleConditions struct {
 	Predicates []string `json:"predicates,omitempty"`
 }
 
+// EqualTo performs semantic equality comparison between two RouteRuleConditions instances.
+//
+// Order-sensitive fields:
+//   - Hostnames: Order matters - compared by index position
+//
+// Order-insensitive fields:
+//   - Predicates: Checks that the same predicates exist, regardless of order
```
guicassolato: Predicates should also be order-sensitive as they may affect performance. Sometimes we may deliberately configure a given predicate to be first to evaluate as a way short-circuit the condition and avoid more expensive computation when it's not needed.

guicassolato:
I think we're taking this the wrong way. Unless we can absolutely ensure that loosening the ordering constraint of a collection will not introduce any bugs, the fast fix approach is the problem.

eguzki:
@guicassolato  Can you elaborate on what's wrong? If the list order doesn't matter and the kuadrant operator renders it non-deterministically, then an order-insensitive equality check seems right to me. Deterministic rendering isn't the goal here. The goal is ensuring the operator works correctly and doesn't trigger updates when two versions are considered equal.

internal/wasm/types.go
Boomatang: I messed this one when I was rechecking. In the test setup I used this was stable.

guicassolato:
> Can you elaborate on what's wrong?

The fast fix approach with something that is as complex as policy reconciliation. Let's just not simply add order-insensitive equality check wherever we think there may be a performance bottleneck.

> If the list order doesn't matter and the kuadrant operator renders it non-deterministically, then an order-insensitive equality check seems right to me. 

Wherever the operator indeed generates a collection with no specific ordering of the elements, I see no problem with switching to order-insensitive equality check. I.e. these collections can be treated as true sets.

We need to be certain though. That's all.

> Deterministic rendering isn't the goal here. The goal is ensuring the operator works correctly and doesn't trigger updates when two versions are considered equal.

The goal is 100% legit. If we can achieve this goal without creating an additional problem for debugging, that's better.

Deterministic lists yielded by reconciliation tend to be easier to debug, but I do understand the trade-off. If we can avoid an unnecessary update, that wins.

Boomatang:
To be clear this is not trying to fix a performance bottleneck, it is fixing a bug. With the operator in a stable state, CPU usage below 10m, and 14 httproutes, authpolicy, & ratelimitpolicy, the generation version of the WasmPlugin CR was above 5200. It also took about half an hour to get to this stable state while consuming 3.5 CPU. 

With the fix the generation version I expect to see around 5 to 10. And stable in a <10 seconds. Of course with the same CPU configuration. If you were to have more policies the odds of ever becoming stable trend to zero. 

Again this is not about a performance bottleneck, if you want to look at one of them see #1919. This is a bug.

Boomatang:
Can I get a status update from everyone that is reviewing this? There does not seem to be any clear next steps from the conversions. 

@adam-cattermole @eguzki @guicassolato

eguzki (commented):
I'm on board with these changes.

The description doesn't accurately reflect the changes, though. The only order-insensitive check is done on the Action.ConditionalData lists. That is exactly the list that we spotted not being deterministically generated. So, at least, we cover that non-deterministically generated list. Note that others may exist, so we should watch out for this with additional tests covering a wide variety of scenarios involving policies with various configurations.

Boomatang:
I have updated the description to align more with the current state of the PR.

adam-cattermole:
Yep these changes also look reasonable to me

internal/wasm/types.go
```diff
@@ -338,7 +392,7 @@ func (a *Action) EqualTo(other Action) bool {
 	}
 
 	for i := range a.ConditionalData {
-		if !a.ConditionalData[i].EqualTo(other.ConditionalData[i]) {
+		if !slices.ContainsFunc(other.ConditionalData, a.ConditionalData[i].EqualTo) {
```
guicassolato, line 395: IIUC, in the wasm-shim, the order of `actionSets.actions.conditionalData` affects:
- The order of descriptor entries sent to Limitador – even if Limitador ends up evaluating all descriptors
- Which value is used for WKAs such as `ratelimit.domain` and `ratelimit.hits_addend` (if duplicated) – I believe [later values overwrite earlier ones](https://github.com/Kuadrant/wasm-shim/blob/09be48e34e8374ddcad01e9f6aca697c8366c385/src/kuadrant/pipeline/tasks/ratelimit.rs#L203-L230)
- Whether later elements are evaluated (if earlier ones are "pending")

So, _ceteris paribus_, if the order of two elements in `spec.when` or equivalently in `spec.limits.when` in an RateLimitPolicy or TokenRateLimitPolicy CR swaps, we will consider these two states of the CR equivalent. Are we OK with this?

Boomatang: I don't know the answer to this question. If this is true the system never worked as expected, while this PR won't address that is it does address the thousand of reconciles that is required to get a the operator stable when the number of resources is still small above 20. With large numbers it would never become stable.

Specking with @adam-cattermole the other day, he mention there is other work that is planned of the shim which will be changing the configuration layout.

guicassolato: IMO, the Kuadrant Operator should not treat 2 wasm configs that today will be evaluated differently by the wasm-shim as if they were the same. IOW, if my assumptions here are correct, `actionSets.actions.conditionalData` should be **order-sensitive**.

To be clear, I'm not denying the performance improvement experienced after the proposed change. I'm saying that treating `conditionalData` as order-insensitive – as well as many other fields in the derivative resources – may lead to or intensify an unpredictable behaviour in the dataplane. Aka a bug.

Differently put, if the order of `conditionalData` indeed didn't matter, then this PR would be fixing a bug. While it does matter, then this PR intensifies a bug and/or introduce new ones.

So, fundamentally, enforcing set equality semantics on collections that, formally speaking, are not sets is not the right fix for the performance issue, even if our intuition seems to point in the other direction after witnessing significantly less reconciliation events with this change.

IMO, the correct fix to this problem is controlling the deterministic order of how the configuration is generated in the first place, with only rare exceptions where the Operator is exempted to update the resources due knowing, without a doubt, that the new produced state would result in the same behaviour in the data plane as the previous state. These rare exceptions are still to be pointed, but it doesn't seem to be the case of wasm's `actionSets.actions.conditionalData`.

Possibly, the root cause of the indeterminist we're dealing with is the very API design mistake we made of using maps to specify policy rules instead of lists – along with how those are deserialised in Go. And there may be several other Go maps used internally in the code to pass data over the stack, breaking determinism as well. Either way, there are ways to solve that, such as ordering policy rules alphabetically within the effective policies (easier to implement), or reading the raw resources from the cluster and, before parsing as Go structs, respecting the order of the rules as stored in etcd (harder).

alexsnaps: I'm about to come out of left field on this, but I'm slightly lost in the rational here.

- If the order matters, then _it has to be stable always_. 
- If the order _is indeed stable_, then how would it fail that equality test?
- If it isn't stable, and as such require us to treat this `list` as a `set`, then... it's broken already, isn't it?
- If the order doesn't matter _and_ it is being reshuffled across reads/writes then... this change is fine, no?

alexsnaps: afaict, and relooking through the code quickly on GH, the order does indeed not matter. So yeah, this probably _should_ be a `set` and not a `list`. But as @Boomatang said, @adam-cattermole is currently working on rewriting that config anyways.

guicassolato: > * If the order matters, then _it has to be stable always_.

I agree. Yet, it isn't stable always today (and neither it will be after this PR). This is a bug that exists today and it's not due to this particular list treated as order-sensitive or order-insensitive. There are other ways this preexisting bug can bite us.

> * If the order _is indeed stable_, then how would it fail that equality test?

It fails because it isn't stable.

> * If it isn't stable, and as such require us to treat this `list` as a `set`, then... it's broken already, isn't it?

Yes, except that "requiring us to treat as this `list` as a `set`" is not the fix to _the_ bug, whose root cause is certainly other.

We should treat a `list` as a `set` when the list is a set, period. This should be the only driver, hence me insisting on the question: Does the order for `conditionalData` matter or not? If it doesn't matter, I don't care about the bug, we should do this regardless. Attenuating the bug is a bonus.

> * If the order doesn't matter _and_ it is being reshuffled across reads/writes then... this change is fine, no?

No, if order does matter then this only partially hides ("attenuates") the original bug (order being reshuffled across reads/writes), by introducing another bug (treating as if the order didn't matter).

guicassolato: > afaict, and relooking through the code quickly on GH, the order does indeed not matter. So yeah, this probably _should_ be a `set` and not a `list`. But as @Boomatang said, @adam-cattermole is currently working on rewriting that config anyways.

This is key.

The use case I was investigating revolves around `ratelimit.hits_addend` in the response phase of TokenRateLimitPolicy, which I understand has value set by the operator to `responseBodyJSON("/usage/total_tokens")`. It's essentially backed by my second assumption [here](https://github.com/Kuadrant/kuadrant-operator/pull/1923#discussion_r3227572108). If you dismiss the assumption, then it's just me overthinking the problem.

1. User creates 2 TRLPs that affect the same route – I think the name of the route dictates the "scope" of the RL action (TBC), this is important for the next step
2. Kuadrant Operator generates actions for these two policies, then [merges](https://github.com/Kuadrant/kuadrant-operator/blob/4b765a83d1aed8a080f111202496da75903e96eb/internal/controller/data_plane_policies_workflow.go#L183) them into one action with a single `conditionData`, and creates a WasmPlugin with `conditionData` elements in a given random order _X_ **<== BUG-1 (already exists today - order of the elements in conditionalData is non-deterministic)**
3. Data plane starts behaving according to order _X_
4. User performs a change that triggers reconciliation without changing anything in the WasmPlugin _but_ the order of the elements in `conditionData` – say the computed new order is _Y_
5. However, Kuadrant Operator skips updating the WasmPlugin because now _X == Y_ **<== BUG-2 (introduced with the PR - `conditionData` treated as order-insensitive)** 
6. Therefore, data plane continues to behave according to order _X_
7. User creates a completely unrelated AuthPolicy
8. Kuadrant Operator generates a new WasmPlugin state that: a) adds the required configuration for the AuthPolicy and b) calculates `conditionData` in a new random order _Z_
9. Data plane starts behaving according to order _Z_

Between 3 and 9, user has no idea why, but the data plane now behaves differently for the user's TRLPs. All the user did was creating an unrelated AuthPolicy.

I added steps 4 and 5 to highlight the difference between with and without the "patch". Without the patch, at step 5, the Operator would perform the update, and therefore, at step 6, the data plane would start behaving according to order _Y_. User would then see 3 different behaviours along the entire journey (_X_, _Y_ and _Z_) instead of 2 (_X_ and _Z_). So one could argue A) the bug is already there, which I'd agree (this is bug the BUG-1), and B) the introduction of BUG-2 attenuates significantly the incidence of BUG-1 – also true.

The main problem is that now we have 2 bugs to fix instead of 1, and BUG-2 looks like it was by design and therefore even easier to forget/harder to fix.

The discussion may sound philosophical, but it is in fact for the principle it encompasses. It's about how we do things and why we do things.

Boomatang: In your scenario of 2 TRLPs affecting the same route, how is this configured under policy attachment? If both policies target the same route and no strategy is defined, the oldest has priority and the challenger is marked Overridden — its limits never reach the WasmPlugin. The only way I can see this behaviour occurring is with one policy using a "defaults merge" strategy. Even then, I don't think the user can control what order the conditionalData is in.
                                                                                                                                                                                                                                                            
Can you create a set of resources that would produce the scenario you are describing? Ideally with a large number of resources to show the non-determinism.

guicassolato: > In your scenario of 2 TRLPs affecting the same route, how is this configured under policy attachment?

One TRLP can target the route directly while the other targets a Gateway the route is attached to, yes, using `merge` strategy.

> Even then, I don't think the user can control what order the conditionalData is in.

If the user can or cannot control the order is irrelevant. What matters is whether different orders yield different behaviours or not. If it does, the Operator _can_ control the order – and it should! If it doesn't, then go ahead with the PR.

alexsnaps: Ok, I probably should look at this PR before commenting now... and I have not done that. 
_But_ I'm slightly getting lost in the arguments being made and _most importantly_, in what each party expects here. 

Wrt the `set` vs `list` (and this is where I probably should have read the PR first): if the argument is that we should have predictable configs being pushed to the wasm-shim, then I agree with that. As far as I cant tell, the order of the `conditionalData` elements does indeed _not_ matter. I'd still prefer tho that the wasm-shim gets a fixed order for these nonetheless (e.g. alphabetically ordered), i.e. that the ordering happens _always_ and as such "two otherwise different configs" not only compare the same, but also _always produce the same output for the wasm-shim to consume_. Which, if I understand this right, addresses the "uncertainty" aspect that @guicassolato is eluding to.

Now, in any case, nothing should be a list, if indeed the order doesn't matter. If the argument is that "we should fix this properly", then I agree with that. Whether this should be part of _this PR_, or if this PR should even exist, is a different matter tho. On the latter, I'd say: work is done, let's merge this and move on. If then tests start failing, then... well... we had an assumption wrong! On the former, i.e. fixing this properly, is why I pinged @adam-cattermole on this. We are currently revamping the entirety of the wasm-config. I _think_ `conditionalData` goes away (it was always sort of a hack). If it does... problem solved. If it doesn't, let use this opportunity to fix the data model, so it mirrors reality.

With all that being said:
@guicassolato what are you expecting out of _this_ PR? Should we make it so we massage the order to be "stable" prior outputting it to the wasm-shim - would that address your concern? What is _your ask_, as it isn't entire clear to me.

guicassolato: I'd prefer predictable, deterministic outputs always, but that is **not** what I am asking from this PR.

I can live with slightly unpredictable outputs if:
1. It doesn't affect the behaviour – as in unpredictable _output_ is not the same as unpredictable _behaviour_, multiple different outputs can yield the same behaviour.
2. It improves performance – which I understand was the motivation behind this PR from start.
3. We commit to looking into ways of making the output more predictable in the future – it doesn't have to be now, and I mean a collective commitment (as opposed to being a single person's responsibility.)

> the order of the `conditionalData` elements does indeed _not_ matter

This is what matters for now. (Pun intended.) It should be enough to unblock this PR (if it was even blocked in the first place!)

This debunks the hypothesis I raised at the beginning of this thread as false.  I.e., response phase `ratelimit.hits_addend` will NOT change value because the Operator produced two wasm configs where the `conditionalData` entries generated for 2+ TRLPs were differently ordered between reconciliation event 1 and reconciliation event 2.

I won't simply call this hypothesis a hallucination from Claude, because I too went to the code and read it in the same way. So it's good to have confirmation this is **not** a real problem in the end.

@alexsnaps, can you please approve the PR since you have confirmed ordering not to be an issue for the wasm-shim?

---

In time:

> let's merge this and move on. If then tests start failing, then... well... we had an assumption wrong!

While we do not control/cannot predict the order of elements, I don't think we have tests for the case I described. I don't think it's even possible to have tests for that case. They would have to be probabilistic and therefore with a non-zero chance of issuing false negatives. Hence the importance of checking the code.

> I _think_ `conditionalData` goes away (it was always sort of a hack).

I prefer not factoring this into the decision of whether this PR should be merged or not. Otherwise, we'd be discussing the order things should happen, in which case this PR probably wouldn't even exist and we'd be removing code instead patching one. So, yes, let's please stick with the settled point of "order of `conditionalData` does not matter" as the common ground and move on.

alexsnaps (approved):
wcgw?

PR #1929, opened 2026-04-23
docs: clarify supported OTLP protocols for Kuadrant CR tracing endpoint

silvi-t:
## Summary                                                                                                                                                                                                                                                    
  - Clarify that `defaultEndpoint` in the Kuadrant CR must use `rpc://` (gRPC OTLP, port 4317)     
  - Document that Authorino is the only component supporting `http://` (HTTP OTLP, port 4318), and this must be configured directly in the Authorino CR

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Documentation**
  * Clarified tracing endpoint guidance: recommend using rpc:// for broad compatibility across components.
  * Clarified that http:// OTLP is only supported for Authorino and requires direct Authorino configuration, which overrides the central default endpoint.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/doc/observability/tracing.md
+++ b/doc/observability/tracing.md
@@ -83,9 +83,7 @@ spec:
 
 **Tracing Configuration Fields:**
 
-- `defaultEndpoint`: The URL of the tracing collector backend (OTLP endpoint). Supported protocols:
-  - `rpc://` for gRPC OTLP (port 4317)
-  - `http://` for HTTP OTLP (port 4318)
+- `defaultEndpoint`: The URL of the tracing collector backend (OTLP endpoint). Use `rpc://` (gRPC OTLP, port 4317) for full compatibility across all components.
 - `insecure`: Set to `true` to skip TLS certificate verification (useful for development environments).
 
 **Important:** Point to the **collector** service (e.g., `jaeger-collector`), not the query service. The collector receives traces from your applications, while the query service is only for viewing traces in the UI.


doc/observability/tracing.md
```diff
@@ -83,9 +83,7 @@ spec:
 
 **Tracing Configuration Fields:**
 
-- `defaultEndpoint`: The URL of the tracing collector backend (OTLP endpoint). Supported protocols:
-  - `rpc://` for gRPC OTLP (port 4317)
-  - `http://` for HTTP OTLP (port 4318)
+- `defaultEndpoint`: The URL of the tracing collector backend (OTLP endpoint). Must use `rpc://` (gRPC OTLP, port 4317). Authorino is the only component that also supports `http://` (HTTP OTLP, port 4318). To use HTTP OTLP for Authorino, configure it directly in the Authorino CR (see [Direct Configuration](#direct-configuration-advanced)), which will override this value.
```
crstrn13: Is this necessarily a __must__? From what I am aware is not like the kuadrant operator will fail reconciling if the schema is http or https. IIRC only the wasm-shim cannot export traces if the schema is http/https. Maybe that should be better reflected in the docs, rather than redirecting everything through rpc, claiming we don't __support__ http/https.

crstrn13 (changes requested):

doc/observability/tracing.md
silvi-t: You're right, I've updated the wording to recommend `rpc://` for full compatibility.

adam-cattermole: Fwiw limitador is also gRPC only - our rust based components do not currently support http/https

silvi-t: Thanks @adam-cattermole , do you think the current wording ("Use rpc:// for full compatibility across all components") is clear enough, or should I explicitly mention that Limitador and wasm-shim only support gRPC?

adam-cattermole: Yeah I think your wording makes sense - I think calling out limitador and wasm-shim is _true_ but it also has consequences regarding parent traces that affects linking of all of the traces; your wording is clearer that the e2e tracing won't work well without rpc

adam-cattermole: The bit about authorino though isn't entirely true - authorino is the only 'dataplane' component that supports http, but all of the control plane ones (operators) support http too (I think) - I probably wouldn't specify which ones and just say that some of the pieces support http but not all...

There is the argument though that we should just say gRPC in the docs - why would a user want to enable tracing 'partially'? I feel like if you enabled tracing across both the control plane and dataplane using http you're seeing 'unexpected behaviour'.  If we eventually decide to support http in all of the components then we could implement and update

silvi-t: Should I remove the Authorino/HTTP detail entirely and just say that gRPC (rpc://) is required for full end-to-end tracing?

adam-cattermole: Personally I'd be tempted to do that - I think giving the extra 'not fully supported' option just confuses things, but I know @crstrn13 had a differing opinion.

Perhaps just putting this is good enough? if they find the http and use it then they're on 'undocumented' ground:

`defaultEndpoint`: The URL of the tracing collector backend (OTLP endpoint). Use `rpc://` (gRPC OTLP, port 4317) for full compatibility across all components.

crstrn13: I reckon it is good like this. I was focusing too much maybe on the __must__, but now that the wording has changed, I fancy the simplicity.

crstrn13 (dismissed):

doc/observability/tracing.md
silvi-t: @crstrn13 I'm a bit confused now. Should I keep the Authorino detail but clarify it as "data plane component" as @adam-cattermole  pointed out, or remove the Authorino/HTTP part entirely?

adam-cattermole: I'd drop the authorino bit please, it's not specific for authorino that it's the only one that supports http so we could just go for the simpler if everyone agrees?

adam-cattermole (approved):
Thanks!

doc/observability/tracing.md
silvi-t: Agreed, I've updated the doc with the simpler wording.

crstrn13 (approved):
LGTM! 🎉 :shipit:

PR #1935, opened 2026-04-27
draft for umbrella approach to OLMv1

maleck13:
@jasonmadigan This covers the problem and a proposed approach that was recommended by the OLM team. This is a very early very generated doc at this point and needs digging into further and the details refining

## Documentation
* Added RFC proposal defining a transition to an OLMv1 umbrella operator model for dependency management, including user install workflows, dependency-aware upgrade flows, manifest deployment strategies, and failure handling guidance.

### DIFF
--- a/doc/proposals/olm-dependency-model-transition.md
+++ b/doc/proposals/olm-dependency-model-transition.md
@@ -3,179 +3,164 @@
 - Feature Name: `olm_dependency_model_transition`
 - Start Date: 2026-03-30
 - RFC PR: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/pull/0000)
-- Issue tracking: [Kuadrant/architecture#0000](https://github.com/Kuadrant/architecture/issues/0000)
+- Issue tracking: [Kuadrant/architecture#164](https://github.com/Kuadrant/architecture/issues/164)
 
 # Summary
 [summary]: #summary
 
-Replace the OLMv0 dependency model (which automatically installs Authorino, Limitador, and DNS operators) with an umbrella operator that directly deploys and manages the lifecycle of all Kuadrant components. This dedicated installer operator preserves the single-action install experience while aligning with OLMv1's design, which explicitly removes automatic dependency installation. The umbrella operator pattern follows the approach recommended by the OLM team and used by OpenShift itself (e.g., [cluster-olm-operator](https://github.com/openshift/cluster-olm-operator)).
+Replace OLMv0's automatic dependency installation with a separate umbrella operator that deploys and manages all Kuadrant components. Helm charts become the single manifest source — consumed directly via `helm install` and by the umbrella operator via client-only rendering. The kuadrant-operator is simplified to focus exclusively on policy reconciliation; operator installation and operand creation move to the umbrella operator.
+
+- OpenShift: Umbrella Operator domain
+- Kubernetes: Helm domain
+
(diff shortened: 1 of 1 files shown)


maleck13:
@jasonmadigan actually created in the kuadrant operator repo as that was useful for context but prob should move to arch repo tbh

eguzki (commented):
This looks promising. I like the umbrella approach idea. A couple of unanswered questions after reading the proposal:

* What will the new umbrella operator be named? I think kuadrant operator is the best candidate, with a kuadrant-controller component managing all policy reconciliation work.

* What's being replaced? Will the current kuadrant operator's new OLM release become the OLMv1-compatible umbrella operator, or are we looking at a brand new operator release?  Will the current operators continue to be released to the OLM public catalog?

maleck13:
> What will the new umbrella operator be named? I think kuadrant operator is the best candidate, with a kuadrant-controller component managing all policy reconciliation work.

+1

> What's being replaced? Will the current kuadrant operator's new OLM release become the OLMv1-compatible umbrella operator. or are we looking at a brand new operator release?

Still to be decided. In an ideal world we would replace and everything would "seamlessly" migrate to v1 compatible but I am not sure right now how feasible that is.

doc/proposals/olm-dependency-model-transition.md
```diff
@@ -0,0 +1,181 @@
...
+2. **Deployment Controllers** - manage component Deployments including image references, resource configuration, and rollout monitoring.
+
+3. **Status Controller** - aggregates health from all managed Deployments and reports overall readiness.
+
+### Version Pinning
+
+Each umbrella operator release embeds component image references as environment variables or build-time constants:
+
+```
+KUADRANT_OPERATOR_IMAGE=quay.io/kuadrant/kuadrant-operator:v1.2.0
```
eguzki: That would be the bundle image actually, right? `quay.io/kuadrant/kuadrant-operator-bundle` ?

doc/proposals/olm-dependency-model-transition.md
```diff
@@ -0,0 +1,181 @@
...
+- Create/manage Namespaces, ServiceAccounts, Deployments, Services
+- Create/manage ClusterRoles, ClusterRoleBindings, Roles, RoleBindings
+- Create/manage CRDs
+- Read Deployment/Pod status for health monitoring
+
+These permissions are granted via the ClusterExtension's ServiceAccount at install time. Each component operator's RBAC is created by the umbrella operator as part of that component's manifest set.
+
+### Upgrade Flow
+
+1. OLMv1 deploys the new umbrella operator image containing updated component image references.
```
eguzki: This would be my only concern here. The umbrella operator would be doing OLM operator's job. We need to implement it and maintain it.

eguzki (commented):
Leaving some further thoughts about it

doc/proposals/olm-dependency-model-transition.md
maleck13: Yes but we have to take on that role in some form given OLM dependencies are going away. We either build a big CSV with all the dependencies in it or we do something more controllable such as the umbrella operator.

maleck13:
@eguzki @jasonmadigan @didierofrivia I have made some updates to the proposal. Mainly it focuses on leveraging Helm at the core of the new operator. 

@didierofrivia Interested in any thoughts you may have on this proposal

@Boomatang you may be interested in this also

maleck13:
Will  move this to the arch repo instead of keeping it here actually

maleck13:
Closing in favour of https://github.com/Kuadrant/architecture/pull/179

PR #1941, opened 2026-04-30
MULTIARCH-5995: add ppc64le support to the build

raja-0940:


maksymvavilov (approved):
@raja-0940 similar to the dns-operator one

maksymvavilov:
i'll close and reopen it - there seems to be something odd with workflow runs

PR #1948, opened 2026-05-06
[WIP] Add disconnected cluster support for wasm-shim image resolution

### LINKED ISSUE #1894: Support disconnected (air-gapped) clusters
## Support disconnected (air-gapped) clusters

**Related Jira:** <https://redhat.atlassian.net/browse/CONNLINK-805>

### Problem Statement

Currently, the wasm-shim image download process is not aware of disconnected (air-gapped) cluster configurations. When installing Kuadrant in these environments, the Gateway fails to pull the wasm-shim image because it attempts to reach external registries instead of utilizing the internal mirror registry.

### Current Behavior

- The installation ignores cluster-wide mirror settings (e.g., `ImageContentSourcePolicy` or `ImageDigestMirrorSet`)
- The process fails unless specific manual overrides are applied
- It does not automatically handle TLS configurations (CA bundles) for custom internal registries

### Desired Behavior

The process should be transparent and "disconnected-aware." Kuadrant should automatically:

- Detect the disconnected environment settings
- Map the wasm-shim image to the correct internal mirror registry
- Respect TLS settings (CA bundles) without requiring insecure skip verification

### Current Workaround

Users must manually intervene with multiple configuration steps:

1. **Patch Kuadrant operator subscription** with environment variables:
   - `RELATED_IMAGE_WASMSHIM`: Configure the mirror registry URL (requires prior knowledge of the default image SHA to construct the mirror URL)
   - `PROTECTED_REGISTRY`: Set to the mirror registry hostname to tell Kuadrant operator to inject pull secrets into the WasmPlugin configuration, enabling the gateway to access the registry with proper credentials

2. **Inject environment variable into Istio gateway deployment**:
   - `WASM_INSECURE_REGISTRIES`: Skip TLS verification for mirror registries with self-signed certificates (introduced in [istio/istio#36903](https://github.com/istio/istio/pull/36903))

**Note:** Skipping TLS verification with `WASM_INSECURE_REGISTRIES` is bad practice and should be avoided. Custom CA certificate bundles should be used instead, but there is currently no documented way to configure Istio gateway pods to honor custom CA bundles when downloading WASM modules from OCI registries.

See attached deployment script for detailed workaround implementation.

### Acceptance Criteria

- [ ] Kuadrant installation succeeds in a fully disconnected environment without setting `RELATED_IMAGE_WASMSHIM`
- [ ] The operator correctly resolves the wasm-shim image from the configured internal mirror registry
- [ ] **Custom CA bundle support for WASM OCI downloads**: This may require changes in Istio to support configuring custom CA certificates for the WASM module download process. Currently, the only option is to disable TLS verification using `WASM_INSECURE_REGISTRIES`, which is not suitable for production environments. This is a potential Istio issue and may not be fixable by kuadrant itself.

### Additional Context

**Related PRs:**

- #1083 - Added `PROTECTED_REGISTRY` support for imagePullSecret injection

### Technical Details

The issue involves two separate problems:

1. **Image resolution**: Consider the approach where the operator respects OpenShift's `ImageContentSourcePolicy`/`ImageDigestMirrorSet` to automatically resolve wasm-shim images from mirror registries when reconciling wasmplugin resources.
2. **TLS trust**: Istio gateway pods need to trust custom CA certificates when downloading WASM modules from internal registries with self-signed certificates

pierDipi:
On disconnected (air-gapped) OpenShift clusters, the wasm-shim container image is served from an internal mirror registry instead of its original source (e.g., `registry.redhat.io`). Gateway providers like Istio and Envoy Gateway pull wasm images directly and do not honor node-level mirror configuration (`registries.conf` / CRI-O), so the operator must resolve the correct mirror URL and supply registry credentials on their behalf.

## Goals

- Wasm-shim images are pulled from mirror registries on disconnected OpenShift clusters without manual operator configuration
- Registry credentials are automatically discovered from OpenShift cluster pull secrets
- Pull secrets are managed per gateway namespace with proper lifecycle (create, update, cleanup)
- User-created pull secrets are never modified or overwritten
- The feature is transparent on non-OpenShift clusters (no errors, no API calls to missing CRDs)
- A kill-switch (`DISABLE_IMAGE_MIRROR_RESOLUTION=true`) disables the entire feature
- Istio and Envoy Gateway reconcilers follow the same pattern

## Non-Goals

- Mirroring images (this feature only resolves URLs to existing mirrors)
- Supporting mirror configurations outside of OpenShift CRDs (e.g., `registries.conf` files)
- Multi-cluster mirror federation
- Pull secret rotation or expiry handling

### Overview

The feature adds two steps to the existing data-plane reconciliation loop:

1. **Mirror URL resolution** -- The operator reads OpenShift mirror CRDs (`ImageDigestMirrorSet`, `ImageTagMirrorSet`, `ImageContentPolicy`) and rewrites the wasm-shim image URL to point at the configured mirror registry, following OpenShift's prefix-matching semantics (longest match wins, path-boundary enforcement, wildcard subdomain support, digest-vs-tag filtering).

2. **Pull secret reconciliation** -- For each gateway namespace, the operator discovers registry credentials from the cluster's pull secrets, then creates or updates a managed secret so the gateway provider can authenticate to the mirror registry. When no policies target a gateway, the managed secret is cleaned up.

Both steps are no-ops on non-OpenShift clusters or when the kill-switch is active.

### High-Level Flow

```
Reconcile cycle (per gateway provider)
│
├─ Resolve image URL
│    RELATED_IMAGE_WASMSHIM → mirror CRDs → resolved URL
│    (returns original URL if no mirrors match or not on OpenShift)
│
└─ For each gateway:
     ├─ Build wasm config
     ├─ Reconcile pull secret (single entry point)
     │    ├─ Feature disabled? → no-op
     │    ├─ No active policies? → clean up managed secret
     │    └─ Active? → discover credentials → create/update managed secret
     ├─ Fallback: PROTECTED_REGISTRY check (backward compat)
     └─ Build WasmPlugin / EnvoyExtensionPolicy with resolved URL and pull secret ref
```

Fixes https://github.com/Kuadrant/kuadrant-operator/issues/1894

## Release Notes

* **New Features**
  * Added support for Wasm plugins in disconnected OpenShift clusters with automatic image URL resolution through OpenShift image mirror configurations.
  * Implemented automatic pull secret reconciliation for Wasm plugin authentication across gateway namespaces.

* **Permissions**
  * Expanded operator RBAC permissions to access OpenShift image mirror resources and manage pull secrets.

* **Documentation**
  * Added comprehensive design documentation for disconnected cluster Wasm support.


eguzki:
Interesting approach. 

Are we ok copying a pull secret from openshift namespace to user namespaces? How risky is this method?

> Istio and Envoy Gateway reconcilers follow the same pattern

Do you have evidences of this? I would like to learn.

pierDipi:
> Are we ok copying a pull secret from openshift namespace to user namespaces? How risky is this method?

The cluster pull secrets in openshift-config are already globally available to every node via CRI-O's registries.conf. Any pod scheduled on any node already pulls images using those credentials. We're extracting only the credentials for the specific mirror registry host and placing them where the gateway provider can use them. The managed secret contains a filtered subset, not the full cluster pull secret.

> > Istio and Envoy Gateway reconcilers follow the same pattern
> 
> Do you have evidences of this? I would like to learn.

This is in the context of Kuadrant controller (meaning this PR) both the Istio and Envoy Gateway reconcilers in this PR use the same `ReconcileWasmPluginPullSecret` entry point.

eguzki:
> Istio and Envoy Gateway reconcilers follow the same pattern

Just to be clear, this approach of copying the secret is not a pattern used anywhere else, right?

> The cluster pull secrets in openshift-config are already globally available to every node via CRI-O's registries.conf. Any pod scheduled on any node already pulls images using those credentials. We're extracting only the credentials for the specific mirror registry host and placing them where the gateway provider can use them. The managed secret contains a filtered subset, not the full cluster pull secret.

I have my concerns here (not discarding it). Some user can have secret reading permissions only in namespace `my-gateway` and now you are exposing (partially) a secret to this user. 

I would like a second opinion, maybe @maleck13  @guicassolato @mikenairn ?

pierDipi:
We could have that particular sub-logic disabled by default (Kuadrant operator env var)?

internal/openshift/pullsecret.go
```diff
@@ -0,0 +1,313 @@
...
+//
+// User-created secrets (those without the kuadrant.io/managed label) are never modified.
+// If a user-created secret exists, this function returns true so that the caller still
+// sets the imagePullSecret reference on the WasmPlugin / EnvoyExtensionPolicy.
+//
+// Returns true if the imagePullSecret reference should be set.
+func ensureWasmPluginPullSecret(ctx context.Context, client dynamic.Interface, namespace, secretName string, dockerConfigData []byte, owner GatewayOwnerRef, logger logr.Logger) (bool, error) {
+	resource := client.Resource(secretsResource).Namespace(namespace)
+
+	existing, err := resource.Get(ctx, secretName, metav1.GetOptions{})
```
guicassolato, line 174: We moved away from this pattern of reconciling resources by reading the state of the cluster from the API server at multiple points throughout the workflow, which often leads to anachronistic decisions, to favour a pattern where the state is composed of a single snapshot taken at the beginning of the process and passed along into the stack. This is what we call "state of the world reconciliation".

To achieve that, all resources of interest – which here seem to include this Kubernetes Secret – should be present in the snapshot, properly linked in the topology DAG to the other objects of interest they relate to. The concept of `existing` then comes from reading the Secret from the topology – and trusting that any concurrent event that may have affected its state in between will be handled by a consecutive triggering of the reconciliation workflow.

I know there's precedence of mixing the old pattern into the new one. However, the more we do it, the closer it brings us back to the reconciliation problems that have motivated us into this architectural redesign in the first place.

pierDipi: isn't this a necessary lookup given that the secret is shared to handle both [1] and [2]?

Perhaps you're suggesting to have one secret per Gateway so that we don't have a shared secret?
 
[1] the case where the secret is not managed by Kuadrant, for example, existing secrets created before upgrading Kuadrant, we don't really want to touch user-created secrets as we might end up fighting with other controllers.

[2] the case where the same secret is shared across multiple Gateway(s) and we need to set the Gateway as owner reference (updating the resource which requires reading resource version)

guicassolato: Likely, this will all be superseded by https://github.com/Kuadrant/kuadrant-operator/pull/1953 and it won't matter, but here's what I meant when I referred to Kuadrant's State of the World reconciliation pattern before.

Every object of interest for the reconciliation process should be watched by the Kuadrant Operator and linked to the other objects it relates to in the topology structure. Then, specific tasks of the reconciliation workflow that care about the state of this object kind can subscribe to it.

This ensures two things:
1. When the secret changes, we reconcile all objects that depend on it.
2. The Operator is not making anachronistic decisions throughout the reconciliation workflow due to concurrent operations changing the state of the cluster in between (including operations performed the Operator itself). I.e., the state of the secret, as of when the reconciliation task that handles it kicks in, is assumed to be the same as of when the overall reconciliation workflow was triggered in the first place – just like for all other objects that are handled by the reconciliation workflow and shaping the logical decisions taken along the way.

In the case of secrets, watching for all events cluster-wide can get very noisy – let alone being a security concern. So, we typically setup additional predicates to filter down the events – e.g. with label selectors. I guess this relates to your concern when you mention preexisting secrets "not managed by Kuadrant" and preferring not to ask users to have to annotate/label them in order to put the objects in scope for Kuadrant. I'm afraid, without this step, we lose those two important things from our way of doing reconciliation.

Regarding updating an object whose state has mutated compared to when that one call to the reconciliation workflow happened, don't worry about it. If the generation of the object has changed, the API server will reject the update and the Operator will handle it gracefully, taking care of it with the next snapshot of the topology of resources of interest that it has already accumulated in the background for a subsequent call to the reconciliation workflow. At this point, this call should be already waiting in the queue to be processed. By checking the state of the resource again from the cluster before the update, you are jumping ahead, trying to avoid the API server error and speed things up. However, that breaks rule # 2 above.

pierDipi:
Superseded by https://github.com/Kuadrant/kuadrant-operator/pull/1953

PR #1950, opened 2026-05-07
fix(gatewayapi): validate listener fields in IsHTTPRouteReady

### LINKED ISSUE #1868: IsHTTPRouteReady doesn't check listener-specific ParentRef fields (SectionName, Port)
### Problem

Gateway API ParentReferences can target specific listeners using `sectionName` or `port` fields. `IsGRPCRouteReady` correctly validates these fields, but `IsHTTPRouteReady` doesn't, creating an API inconsistency.

**Location:** `internal/gatewayapi/utils.go`

```go
// HTTP version - no listener parameter, doesn't check SectionName/Port
func IsHTTPRouteReady(httpRoute *gatewayapiv1.HTTPRoute, gateway *gatewayapiv1.Gateway,
                      controllerName gatewayapiv1.GatewayController) bool

// gRPC version - has listener parameter, checks SectionName/Port
func IsGRPCRouteReady(grpcRoute *gatewayapiv1.GRPCRoute, gateway *gatewayapiv1.Gateway,
                      listener *gatewayapiv1.Listener, controllerName gatewayapiv1.GatewayController) bool
```

### Impact

- API inconsistency between HTTP and gRPC route helpers
- `IsHTTPRouteReady` may incorrectly return `true` for routes attached to a different listener on the same gateway
- Makes code harder to maintain and understand

### Example Scenario

```yaml
Gateway has:
  - Listener "http" (port 80)
  - Listener "https" (port 443)

HTTPRoute has ParentRef:
  - name: my-gateway
    sectionName: "https"  # Targets only HTTPS listener

When calling IsHTTPRouteReady(route, gateway, httpListener, ...):
  - Current behaviour: Returns true if route is accepted by ANY listener
  - Expected behaviour: Returns false (route doesn't target this specific listener)
```

### Proposed Solution

Update `IsHTTPRouteReady` to match `IsGRPCRouteReady`:

1. Add `listener *gatewayapiv1.Listener` parameter
2. Add SectionName and Port matching logic:
   ```go
   if ref.SectionName != nil && *ref.SectionName != listener.Name {
       return false
   }
   if ref.Port != nil && *ref.Port != listener.Port {
       return false
   }
   ```
3. Update all callers (primarily in `auth_policy_status_updater.go`)

### Additional Context

This inconsistency was discovered and fixed for gRPC routes in #1855 (commit 55be79ac). The same fix should be applied to HTTP routes for consistency.

vibhor-5:
Align IsHTTPRouteReady with IsGRPCRouteReady by checking SectionName and Port in ParentReferences. Update status updaters to pass listener.

Closes #1868

## Release Notes

**Bug Fixes**
- Improved HTTP route readiness checks used by AuthPolicy, RateLimitPolicy, and TokenRateLimitPolicy by incorporating listener-specific matching and tighter parent-status identity comparisons.

**Tests**
- Expanded unit test coverage for HTTP route readiness, including controller-name matching, listener SectionName/port handling, and scenarios where SectionName/port are omitted.


thomasmaas (commented):
## Review

Clean, well-scoped fix that aligns `IsHTTPRouteReady` with `IsGRPCRouteReady` by adding listener-specific `SectionName` and `Port` validation. The implementation now mirrors the GRPC variant exactly, all three call sites are updated consistently, and table-driven tests cover the core matching scenarios. Ready to merge with minor test coverage suggestions below.

### Suggestions (non-blocking)

1. **Removed test coverage for `GetHTTPRouteAcceptedParentRefs` with SectionName** — The old `TestGetHTTPRouteAcceptedParentRefs_WithSectionName` (from #1855) was replaced by `TestIsHTTPRouteReady`, which tests a different function. While SectionName matching is indirectly covered by `TestParentReferencesMatch`, consider restoring the old test alongside the new one, or adding a SectionName-aware case to the existing `TestGetHTTPRouteAcceptedParentRefs` table.

2. **Missing test cases for the `!found` and `!Accepted` paths** — All test cases have `RouteConditionAccepted: ConditionTrue` and a matching gateway. Consider adding:
   - A case where the ParentRef matches but the Accepted condition is `false` → exercises the final `meta.IsStatusConditionTrue` return.
   - A case where no matching parent status exists (e.g. different gateway name or empty Parents) → exercises the `!found` path.

3. **No combined SectionName + Port test** — The two `if` checks are independent, but there's no case testing both set simultaneously. Consider adding:
   - `SectionName: "http"` + `Port: 80` → `true` (both match)
   - `SectionName: "http"` + `Port: 443` → `false` (port mismatch with section match)

vibhor-5:
@thomasmaas addresed review PTAL.

maksymvavilov:

vibhor-5:
> Re-opening to pick up new workflows

sorry i missed this , I've rebased the PR @maksymvavilov @thomasmaas ptal lmk if i need to change something.

PR #1953, opened 2026-05-08
Use `EnvoyFilter` in lieu of `WasmPlugin` to install our Wasm Filter

alexsnaps:
The whole point of this refactoring, is being able to set `allow_on_headers_stop_iteration` on the [`extensions.wasm.v3.PluginConfig`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#extensions-wasm-v3-pluginconfig), which is currently not feasible using the [Istio](https://github.com/istio/istio/discussions/60126)

## Summary by CodeRabbit

* **Chores**
  * Switched operator gateway integration from Istio WasmPlugin to EnvoyFilter resources.
  * Reduced cluster RBAC by removing WasmPlugin permissions and updated operator metadata timestamp.
  * Narrowed Istio watcher scope to EnvoyFilter and PeerAuthentication kinds.
* **New Features**
  * EnvoyFilter-based WASM injection now supports image pull secrets and embedded plugin config in patches.
* **Tests**
  * Updated unit and integration tests to validate EnvoyFilter generation and config extraction.
* **Bug Fixes**
  * Improved reconciliation to correctly mark gateways as modified after EnvoyFilter changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-operator/pull/1953)


internal/istio/utils.go
```diff
@@ -62,6 +63,100 @@ func BuildEnvoyFilterClusterPatch(host string, port int, mtls bool, clusterPatch
...
+		"name":    "kuadrant-wasm-shim",
+		"root_id": "kuadrant_wasm_shim",
+		"vm_config": map[string]any{
+			"runtime": "envoy.wasm.runtime.v8",
+			"code": map[string]any{
+				"remote": map[string]any{
+					"http_uri": map[string]any{
+						"uri":     imageURL,
+						"timeout": "10s",
+						"cluster": "wasm_remote_cluster",
```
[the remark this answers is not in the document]
alexsnaps: I think that node is a `config.core.v3.RemoteDataSource`, there should be no `cluster` field to begin with there, should there?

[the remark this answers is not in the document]
alexsnaps: Yes, you are right, my bad - misread that! But given the docs:
>A cluster is created in the Envoy “cluster_manager” config section. This field specifies the cluster name.

This cluster will be created, we don't have to create it ourselves. Let's keep that a hardcoded name

internal/controller/extenstion_reconciler_test.go
```diff
@@ -35,57 +36,186 @@ var (
...
+	imagePullSecret, ok := remote["image_pull_secret"].(string)
+	if !ok {
+		return ""
+	}
+
+	return imagePullSecret
+}
+
+// extractImageURLFromEnvoyFilter extracts the image URL from EnvoyFilter's nested structure
+func extractImageURLFromEnvoyFilter(ef *istioclientgonetworkingv1alpha3.EnvoyFilter) string {
```
guicassolato, line 100: Some commonality between this func and `extractImagePullSecretFromEnvoyFilter`. Worth synthesising them into a single a func that extracts both pieces at once and returns structured data?

config/rbac/role.yaml
```diff
@@ -112,18 +112,6 @@ rules:
...
-  resources:
-  - wasmplugins
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
adam-cattermole, line 126: I see there's still reference to watch the `WasmPlugin` resource in `state_of_the_world.go` but the RBAC is gone so the operator fails to start:

https://github.com/Kuadrant/kuadrant-operator/blob/131810cefdbf596c7228ef1cd8b197e8a6909011/internal/controller/state_of_the_world.go#L390-L395

philbrookes: ah, this didn't happen to me, I think because I was using `make run`. I just tried it with a deployment and see the same thing :+1:

alexsnaps:
@philbrookes should be fine now™

alexsnaps:
@crstrn13 Can you tell me how I'm supposed to debug that smoke test? I wanted to run this locally, but looking at the action, I don't think I can :thinking:

crstrn13:
@alexsnaps you should be able to clone the testsuite repo and do `make local-setup`, followed by `make smoke`. However, from the looks of it, the last failure is most likely because of the enforced status mismatch thingy. There is a slight chance on the next run it won't reproduce, or reproduce :shrug:

alexsnaps:
> @alexsnaps you should be able to clone the testsuite repo and do `make local-setup`, followed by `make smoke`. However, from the looks of it, the last failure is most likely because of the enforced status mismatch thingy. There is a slight chance on the next run it won't reproduce, or reproduce 🤷

All my test locally fail with:
```
E   Failed: Unable to locate Kuadrant installation: Cluster https://127.0.0.1:46471 does not have Kuadrant resource in project kuadrant-system
```
But I also can't run anything that uses all [those (AWS?) secrets](https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/smoke-test.yaml#L69-L106)... Not obviously related afaict, but :shrug:

crstrn13:
use the testsuite's make local-setup target. That's the easiest way to set it up. Those aws-secrets are solely for the dns tests, which are part of the smoke target, but I don't think they are affected by this change. That error is probably because you are missing the Kuadrant resource IIRC.

alexsnaps:
Ok, I'll try to testsuite's target, I used the workfile as a template, so I did [that](https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/smoke-test.yaml#L53-L63)

internal/controller/data_plane_policies_workflow.go
```diff
@@ -34,6 +34,7 @@ const (
 
 var (
 	WASMFilterImageURL = env.GetString("RELATED_IMAGE_WASMSHIM", "quay.io/kuadrant/wasm-shim:latest")
+	WASMFilterImageSHA = env.GetString("RELATED_IMAGE_WASMSHIM_SHA", "b937dee6dd5e496b5b6fc28aba9b23254dfd5ec304d7565125b3e14d69511d7a")
```
alexsnaps, line 37: Actually this can't work as a source for fetching the WASM. This isn't a wasm module, nor probably even an endpoint where the actual "binary" is. It probably should be `oci://`, that envoy wouldn't understand. Then, that'd be an OCI image with the `.wasm` _within it_. But we _only_ care about the `.wasm` here. 

So I don't know how we best tackle this actually. I'm thinking package the `.wasm` within the operator and get it from there directly... Ideas anyone?

/cc @adam-cattermole others?

alexsnaps: This whole process is taken care of by [Istio](https://github.com/istio/istio/blob/ac188f625375b34dde381003d786db985b2de995/pkg/wasm/imagefetcher.go#L381-L428) when using the `WasmPlugin` API

adam-cattermole: Right, that makes sense, it's the binary itself not the image with the single layer. We currently include the binary on [releases](https://github.com/Kuadrant/wasm-shim/releases/tag/v0.12.3). There's some investigation here from a previous attempt at this #686.

I see a few options:
- Package the binary in with the operator like in #686 
- Pull from github releases 
- Implement logic for OCI fetching and unpacking in the operator - this is actually exactly what we want for extensions right so we could re-use this? there's a commit here (specific to extensions) that did this https://github.com/Kuadrant/kuadrant-operator/compare/main...philbrookes:kuadrant-operator:feat/oci-extension-pull

I guess the caveat with #686 was having to host the binary in some way for the gateways to be able to pull it so that's an addition that would be required in all 3 cases.

I guess I tend to think that the 3rd option (having this done by the operator) would make sense but also realise that commit isn't the whole story - needs support for pulling from protected registries etc.

alexsnaps: Option 3 is the "most" transparent, but also the most work. Transparent in the sense that the `RELATED_IMAGE_WASMSHIM` keeps on pointing to an OCI image, the thing that changes would be that the kuadrant operator, instead of Istio would be fetching, unpacking and serving the wasm binary to Envoy. But all those 3 things have to happen, and support credentials et al indeed. 

But I think I agree that's probably the best way forward.

alexsnaps:
Alright, this is 🟢 
This now fixes #1894 too

Next steps _after this got merged_:

 - [ ] Add the `allow_on_headers_stop_iteration` on the `PluginConfig`
 - [ ] Re-enable the `WASMFilterImage` -`URL` & -`SHA` env vars to fetch an alternative OCI image for the wasm-shim at runtime, extract the `/plugin.wasm` and store it to some `/tmp` fs to serve instead of the wasm packaged with the build
 - [ ] Make sure fetching the OCI image works from a `ProtectedRegistry`, with the `RegistryPullSecretName`

internal/controller/data_plane_policies_workflow.go
guicassolato: Let's also take [air gapped installs](https://github.com/Kuadrant/kuadrant-operator/pull/1948) into consideration.

Apparently on OpenShift, the Operator may need to perform a couple tricks to resolve a mirrored `RELATED_IMAGE_WASMSHIM` endpoint.

internal/istio/utils.go
```diff
@@ -62,6 +64,102 @@ func BuildEnvoyFilterClusterPatch(host string, port int, mtls bool, clusterPatch
...
+			"code": map[string]any{
+				"remote": map[string]any{
+					"http_uri": map[string]any{
+						"uri":     wasmURL,
+						"timeout": "10s",
+						"cluster": clusterName,
+					},
+					"sha256": imageSHA,
+				},
+			},
```
guicassolato, line 131: Not specifying `failure_policy` here. `fail_open` (default: `false`) was deprecated in favour of it, with default to `UNSPECIFIED` → `FAIL_CLOSED`.

I reckon this is the behaviour we want. Yet, do we want to be explicit about it or we're OK with working with the defaults (hoping they won't change)?

```suggestion
			},
			"failure_policy": "FAIL_CLOSED",
```

internal/controller/data_plane_policies_workflow.go
alexsnaps: Wait... what?! There is no need to concern ourselves with air gapped installs here, is there?!
Other than you talking about the `ENV_VARS` to overwrite at run/launch time, which wasm-shim to use. But, while I was planning to address this in a future PR, I don't think this should be a _feature_.

guicassolato: Never mind. Now I understand it better. We're actually pulling the image at build time, not at runtime anymore. Nice!

alexsnaps: Implemented in this PR, for clarity, is the wasm-shim is bundled into the Docker image, using the same property _at build time_. From there on, that'll be the wasm module that Envoy will load _from the Kuadrant operator_. There is no OCI registry involved.

internal/istio/utils.go
alexsnaps: This should be whatever ended up before. But I can add this as part of the next PR then, along side the `allow_on_headers_stop_iteration: true`

guicassolato (commented):
What's the story for Envoy Gateway?

The data plane continues to pull the image from a registry at runtime. The change to patch resource after this is marginal though: https://github.com/Kuadrant/kuadrant-operator/blob/4b765a83d1aed8a080f111202496da75903e96eb/internal/controller/envoy_gateway_extension_reconciler.go#L563-L568

Planned for the follow-up?

alexsnaps:
> What's the story for Envoy Gateway?
> Planned for the follow-up?

Yeah. First want to confirm this all works and doesn't create further headaches (mostly the adding of the `allow_on_headers_stop_iteration`) flag actually. But yeah, if this is all fine it should be ported across totally!

internal/controller/state_of_the_world.go
```diff
@@ -387,22 +386,15 @@ func (b *BootOptionsBuilder) getIstioOptions() ([]controller.ControllerOption, e
...
 			istio.EnvoyFilterGroupKind,
-			istio.WasmPluginGroupKind,
 			istio.PeerAuthenticationGroupKind,
 		),
 		controller.WithObjectLinks(
 			istio.LinkGatewayToEnvoyFilter,
-			istio.LinkGatewayToWasmPlugin,
 			istio.LinkKuadrantToPeerAuthentication,
 		),
+		controller.WithRunnable("wasm server", wasmServerRunnable(b.logger)),
```
mikenairn, line 397: This is only being added when istio is detected. Probably fine for now, but wondering if other implementations could also eventually benefit from this eventually, envoygateway for instance.

internal/controller/wasm_server.go
```diff
@@ -0,0 +1,108 @@
...
+	server *http.Server
+	logger logr.Logger
+}
+
+func NewWasmServer(logger logr.Logger) *WasmServer {
+	return &WasmServer{logger: logger.WithName("WasmServer")}
+}
+
+func (s *WasmServer) Run(stopCh <-chan struct{}) {
+	port, err := env.GetInt("WASM_SERVER_PORT", defaultWasmServerPort)
```
mikenairn, line 52: Confusingly we seem to have a mix of approaches already in this code base for how port values are being set. Health and Metrics use CLI flags and this and EXTENSIONS_DESCRIPTOR_SERVICE_PORT use env vars.

That should probably be made consistent, but out of scope here i guess. If you are using an env var though you will need to add it to the deployment https://github.com/Kuadrant/kuadrant-operator/blob/wasm-envoy-filter/config/manager/manager.yaml#L36-L37

internal/istio/utils.go
```diff
@@ -86,14 +184,41 @@ func EqualEnvoyFilters(a, b *istioclientgonetworkingv1alpha3.EnvoyFilter) bool {
...
+				bListener := bConfigPatch.Match.GetListener()
+				if (aListener == nil) != (bListener == nil) {
+					return false
+				}
+				// For HTTP_FILTER patches, we compare the listener match structure if present
+				// Since the structure is complex, we'll compare the JSON representation
+				if aListener != nil && bListener != nil {
+					aMatchJSON, _ := json.Marshal(aConfigPatch.Match)
+					bMatchJSON, _ := json.Marshal(bConfigPatch.Match)
+					if string(aMatchJSON) != string(bMatchJSON) {
```
mikenairn: Could ignoring errors here potentially result in both appearing equal if both fail to marshal for some reason?

internal/istio/utils.go
```diff
@@ -86,14 +184,41 @@ func EqualEnvoyFilters(a, b *istioclientgonetworkingv1alpha3.EnvoyFilter) bool {
...
+					return false
+				}
+				if aCluster.Service != bCluster.Service || aCluster.PortNumber != bCluster.PortNumber || aCluster.Subset != bCluster.Subset {
+					return false
+				}
+			default:
+				// For other patch types, compare the match structure via JSON
+				aMatchJSON, _ := json.Marshal(aConfigPatch.Match)
+				bMatchJSON, _ := json.Marshal(bConfigPatch.Match)
+				if string(aMatchJSON) != string(bMatchJSON) {
```
mikenairn: Same here

internal/controller/istio_extension_reconciler.go
```diff
@@ -58,15 +55,23 @@ func (r *IstioExtensionReconciler) Subscription() controller.Subscription {
...
-	logger.V(1).Info("building istio extension ", "image url", WASMFilterImageURL)
+	operatorNamespace := env.GetString("OPERATOR_NAMESPACE", "kuadrant-system")
+	wasmServerHost := fmt.Sprintf("kuadrant-operator-wasm.%s.svc.cluster.local", operatorNamespace)
+	wasmServerPort, portErr := env.GetInt("WASM_SERVER_PORT", defaultWasmServerPort)
+	if portErr != nil {
+		wasmServerPort = defaultWasmServerPort
+	}
+	wasmURL := fmt.Sprintf("http://%s:%d/plugin.wasm", wasmServerHost, wasmServerPort)
+
+	logger.V(1).Info("building istio extension", "wasm url", wasmURL)
```
mikenairn, line 74: What happens in istio when this url is wrong? How can we diagnose an issue where istio can't connect to this for whatever reason?

mikenairn (commented):
Left a few comments. I think this is the right direction to go. Worth noting however that local development is kind of awkward after this. Using `make run` will no longer work as it does today since the http server won't be running anywhere to serve the wasm plugin.

guicassolato (dismissed):
I've tested this on a local cluster and it works. Well done, @alexsnaps!

internal/controller/state_of_the_world.go
mikenairn: @guicassolato Does it make sense that a runnable like this is added to the policy-machinery controller? The fact it has to implement this interface seems odd. Maybe it would make more sense added directly to the controller manager like health and metrics https://github.com/Kuadrant/kuadrant-operator/blob/wasm-envoy-filter/cmd/main.go#L258

It should also fail if it can't start for some reason which it currently doesn't. Any errors in that http server initialisation will be ignored.

internal/controller/wasm_server.go
alexsnaps: We probably don't want to encourage that tho, I only copied the approach "because we already did it". I slowly wonder how much of that should actually be tweakable tbh

internal/controller/istio_extension_reconciler.go
alexsnaps: Istio doesn't care about this URL there. That URL is used to populate the HTTP2 request by Envoy. The connection itself is established _prior_ that even being hit, as per the `cluster` definition. We "just" use istio to patch the Envoy config with that filter. 

Now if you are asking what happens if that happens to be "broken", then Envoy will log an error when trying to access that `.wasm`. Honestly, the error is really not helpful tho (e.g. when I had an HTTP1 server, while Envoy expecting a HTTP2 one) - it just tells you "this failed  ¯\_(ツ)_/¯"

alexsnaps:
>Using make run will no longer work as it does today since the http server won't be running anywhere to serve the wasm plugin.

I didn't even know about that "run" target. We have so many of them 🫣 
I can have a look at what it does, if it runs a kuadrant operator, it should be able to load it.

internal/istio/utils.go
alexsnaps: I guess those can't fail given how they are created, but... _for now_ indeed. This should be harden probably.

internal/controller/istio_extension_reconciler.go
mikenairn: Yeah it was really how do you know something is wrong if the http server isn't working or not connectible for whatever reason. Everything just appears to be ok from the POV of the policies and envoyfilter, but in fact nothing is working, rate limiting or auth that is. I managed to do this by using `make local-env-setup` and `make run`.

mikenairn: It's not an issue, i was more just wondering how you would know. I guess monitoring of istio might catch issues like this.

mikenairn:
> I didn't even know about that "run" target. We have so many of them 🫣
> I can have a look at what it does, if it runs a kuadrant operator, it should be able to load it.

fyi https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/overviews/development.md#run-as-a-local-process

For iterative development locally it's far easier than having to build and re-deploy an image. This is also what the tests use, `make local-env-setup` which i guess means they don't test the wasm config actually works :thinking:  Probably just that resulting resources are as expected.

mikenairn:
I should also mention that i did test this and it all worked as expected, when using `make local-setup`. Nothing i have mentioned need block merging, but some things we would probably want to come back to sooner rather than later.

internal/controller/state_of_the_world.go
guicassolato: With pros and cons, both ways will work.

If we want to completely separate this listener from the state of the world controller, then yes, we could start the wasm server in a goroutine we manage ourselves, like we do in Authorino for [example](https://github.com/Kuadrant/authorino/blob/fc38f5d4d77e6db871a39e4d62e2c99a7e21d5a7/main.go#L516).

Policy Machinery's `Runnable` basically wraps that initialisation with an additional step for the processes to sync before reconciliation actually starts. In this case, synchronisation is not being used, but eventually it could, to signal _"Please do not start creating EnvoyFilter resources until I'm ready to serve the wasm binary"_ – like if any of those errors where now `*WasmServer.Run(…)` just logs and returns would instead first flip an internal field so `HasSync()` would return `false`.

Though right now I wouldn't do that. First, because of the other kinds of policies (DNS and TLS) that do not depend on the wasm binary being served, and second because of Envoy Gateway, which also does not depend on the wasm server yet. But if we imagine a world where these feature are behind a switch and disabled for a particular deployment, it could make sense.

guicassolato:
> > Using make run will no longer work as it does today since the http server won't be running anywhere to serve the wasm plugin.
> 
> I didn't even know about that "run" target. We have so many of them 🫣 I can have a look at what it does, if it runs a kuadrant operator, it should be able to load it.

I think there's no problem with `make run` itself and even the wasm server would start, but locally. The EnvoyFilters created pointing to a service that does not exist running inside the cluster are the ones to fail, well to fail when the proxy tries to fetch the binary from those endpoints.

```
$ make local-env-setup

$ ls -ls /tmp/plugin.wasm
11112 -rw-r--r--@ 1 mcassola  wheel  5686051 15 May 17:43 /tmp/plugin.wasm

$ export WASM_SERVER_FILE_PATH=/tmp/plugin.wasm

$ make run
[…]
2026-05-15T19:28:30.199+0200	INFO	kuadrant-operator.WasmServer	starting wasm server (h2c)	{"port": 8082, "file": "/tmp/plugin.wasm", "sha256": "220d760ca4d350c637222804279fba54425a4aaf7612c366148bcd823ad40f61"}
[…]
```

internal/controller/istio_extension_reconciler.go
alexsnaps: Sadly, actually, that hasn't changed from before. We always had this issue where if anything goes wrong installing (or bootstrapping) the wasm-shim within envoy, we were always blind to that - still so today. In the sense that whatever these CR's status say is... well, a lie. 

Moving forward, I think we should probably have the wasm-shim callback to the operator and "tell it" how it's "doing", e.g. (thinking out loud) the SHA of the config it's applying. There is plenty of opportunity to harden our stack in that respect.

alexsnaps:
> > > Using make run will no longer work as it does today since the http server won't be running anywhere to serve the wasm plugin.
> > 
> > 
> > I didn't even know about that "run" target. We have so many of them 🫣 I can have a look at what it does, if it runs a kuadrant operator, it should be able to load it.
> 
> I think there's no problem with `make run` itself and even the wasm server would start, but locally. The EnvoyFilters created pointing to a service that does not exist running inside the cluster are the ones to fail, well to fail when the proxy tries to fetch the binary from those endpoints.

Probably should look it all up myself, but I imagine that if that doesn't build the image, but "runs locally", as in "just spins up the process"? Then, `/wasm/plugin.wasm` will not be a location where the plugin can be found for sure!

internal/controller/wasm_server.go
adam-cattermole: As far as I understand this is pretty 'common practice' to have the controller-runtime manager command line flags separate from operator specific configuration as env vars - perhaps that's just a symptom of the scaffolding tools though.

guicassolato:
> > > > Using make run will no longer work as it does today since the http server won't be running anywhere to serve the wasm plugin.
> > > 
> > > 
> > > I didn't even know about that "run" target. We have so many of them 🫣 I can have a look at what it does, if it runs a kuadrant operator, it should be able to load it.
> > 
> > 
> > I think there's no problem with `make run` itself and even the wasm server would start, but locally. The EnvoyFilters created pointing to a service that does not exist running inside the cluster are the ones to fail, well to fail when the proxy tries to fetch the binary from those endpoints.
> 
> Probably should look it all up myself, but I imagine that if that doesn't build the image, but "runs locally", as in "just spins up the process"? Then, `/wasm/plugin.wasm` will not be a location where the plugin can be found for sure!

Maybe we can get the more important part of this rolling and then loop back on the dev workflow at another PR?

alexsnaps:
Fixed `make run` to fetch, unpack and override the location of the `plugin.wasm` to load. 
Comments on JSON parsing failing addressed too.

crstrn13 (approved):

Makefile
```diff
@@ -401,12 +401,21 @@ build: DIRTY=$(shell $(PROJECT_PATH)/utils/check-git-dirty.sh || echo "unknown")
...
+
 run: export LOG_LEVEL = debug
 run: export LOG_MODE = development
 run: export OPERATOR_NAMESPACE := $(OPERATOR_NAMESPACE)
+run: export WASM_SERVER_FILE_PATH := $(WASM_BIN)
 run: GIT_SHA=$(shell git rev-parse HEAD || echo "unknown")
 run: DIRTY=$(shell $(PROJECT_PATH)/utils/check-git-dirty.sh || echo "unknown")
-run: generate fmt vet ## Run a controller from your host.
+run: generate fmt vet $(WASM_BIN) ## Run a controller from your host.
 	go run -ldflags "-X main.version=v$(VERSION) -X main.gitSHA=${GIT_SHA} -X main.dirty=${DIRTY}" -race ./cmd/main.go
```
mikenairn, line 419: I see this got merged. This change doesn't actually work for me, throws this error:

```
$ make run
/home/mnairn/go/src/github.com/kuadrant/kuadrant-operator/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
go fmt ./...
go vet ./...
docker pull quay.io/kuadrant/wasm-shim:latest
latest: Pulling from kuadrant/wasm-shim
Digest: sha256:49222456322255975d72bbf34874bc4a20a0c0016dcf71a06fdd7b16c069eafa
Status: Image is up to date for quay.io/kuadrant/wasm-shim:latest
quay.io/kuadrant/wasm-shim:latest
docker save quay.io/kuadrant/wasm-shim:latest | tar xf - --to-stdout $(docker save quay.io/kuadrant/wasm-shim:latest | tar tf - | grep -E '^[a-f0-9]+\.tar$' | while IFS= read -r layer; do docker save quay.io/kuadrant/wasm-shim:latest | tar xf - --to-stdout "$layer" | tar tf - 2>/dev/null | grep -q plugin.wasm && echo "$layer"; done | head -1) | tar xf - -C /home/mnairn/go/src/github.com/kuadrant/kuadrant-operator/bin/wasm
tar: This does not look like a tar archive
tar: Skipping to next header
tar: Exiting with failure status due to previous errors
make: *** [Makefile:410: /home/mnairn/go/src/github.com/kuadrant/kuadrant-operator/bin/wasm/plugin.wasm] Error 2
```

Even if this did work, it only half fixes the issue anyway as the internal URL is hard coded and will never point back to the server running on the host. I think it also needs to know when it needs to update the local binary, currently it will just create it once and never change it.

tbh, `make run` not working e2e was not a huge issue, but the current changes are not fixing anything, and are actually broken at least for some, so are probably better reverted and a proper follow on issue created.

alexsnaps: >This change doesn't actually work for me, throws this error:

Are you using docker or podman?

>internal URL is hard coded

Ah! You mean how envoy loads it, is that right? 🤦 I checked it served the wasm, but yeah... envoy would need to load it. @adam-cattermole did you/we support `make run` for extensions and loading the protos? I guess this might be equally broken right now, no?

mikenairn: > Are you using docker or podman?

docker

This worked, but i didn't spend any time seeing if there was a better way, or if it worked on both docker and podman.

```
$(WASM_BIN): ## Fetch and extract the wasm-shim binary from the OCI image.
	@mkdir -p $(WASM_BIN_DIR)
	$(CONTAINER_ENGINE) pull $(RELATED_IMAGE_WASMSHIM)
	@echo "Extracting plugin.wasm from image..."
	@CID=$$($(CONTAINER_ENGINE) create --entrypoint="" $(RELATED_IMAGE_WASMSHIM) /bin/true 2>/dev/null || $(CONTAINER_ENGINE) create $(RELATED_IMAGE_WASMSHIM) 2>/dev/null); \
	if [ -n "$$CID" ]; then \
		$(CONTAINER_ENGINE) cp $$CID:/plugin.wasm $(WASM_BIN) && \
		$(CONTAINER_ENGINE) rm $$CID >/dev/null 2>&1; \
	else \
		echo "Failed to create container from image"; \
		exit 1; \
	fi
```


> > internal URL is hard coded
> 
> Ah! You mean how envoy loads it, is that right? 

Yeah, [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/controller/istio_extension_reconciler.go#L67) either needs to be configured (Service) to point back to the host IP or make it override-able via an env var or something.

alexsnaps: > > Ah! You mean how envoy loads it, is that right?
> 
> Yeah, [this](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/controller/istio_extension_reconciler.go#L67) either needs to be configured (Service) to point back to the host IP or make it override-able via an env var or something.

Can we make that resolve instead to the local ip (as I guess `make run` binds `127.0.0.1` "at least", yes)?
Cause ideally, these shouldn't be overrides I'm thinking, but "just" point to the proper thing, wdyt?

azgabur:
Is there some place in the resulting image which can be used to deduce which wasm-shim version/image was used at build? A log line, image artifact, or image manifest?
Building from `latest` tag will create ambiguity. 

What is the current use of `RELATED_IMAGE_WASMSHIM` env var if it does not change the image used at runtime? If its not used anymore, will it be removed? (and `RELATED_IMAGE_WASMSHIM_SHA` ?)

cc: @silvi-t

alexsnaps:
@azgabur See [here](https://github.com/Kuadrant/kuadrant-operator/pull/1953#issuecomment-4453890260)

So, `RELATED_IMAGE_WASMSHIM` is only honored at build time indeed. But there is plan to re-introduce this _at some point_... could be sooner than later if there is a need - where sooner is like now, and later is possibly after the release

Issue #1934, opened 2026-04-27
Non-Deterministic WasmPlugin Reconciliation Bug


Boomatang:
**Describe the bug**

The WasmPlugin comparison function in kuadrant-operator performs strict order-sensitive equality checks on collections (slices) that have no inherent ordering requirements. This causes non-deterministic reconciliation behavior when multiple policies are applied to a cluster.

The `EqualTo` implementation for WASM configuration types was comparing slices element-by-element using positional equality (`a[i] == b[i]`), which led to:
- **Non-deterministic ordering**: When 20+ RateLimitPolicies and AuthPolicies were applied, the order of ActionSets, Actions, and other collections could vary between reconciliation cycles
- **Reconciliation instability**: Two functionally identical WASM configs were considered different purely due to ordering, causing the operator to continuously update WasmPlugin resources without ever reaching a stable state
- **"Lottery effect"**: With a large number of policies, reconciliation would sometimes stabilize but often would not, depending on the random ordering of collections

This issue is directly related to the continuous reconciliation loops and CPU saturation observed in customer environments and documented in issue [#1919](https://github.com/Kuadrant/kuadrant-operator/issues/1919).

**To Reproduce**

1. Deploy kuadrant-operator in a Kubernetes cluster
2. Create multiple HTTPRoutes (20+ recommended to observe the issue)
3. Attach both an AuthPolicy and a RateLimitPolicy to each HTTPRoute
4. Monitor the kuadrant-operator logs: `kubectl logs -f deployment/kuadrant-operator-controller-manager -n kuadrant-system`
5. Observe continuous reconciliation loops with WasmPlugin resources being updated repeatedly
6. Monitor resource usage: `kubectl top pods -n kuadrant-system`
7. See CPU saturation and continuous reconciliation events without reaching a stable state

The issue becomes more pronounced as the number of policies increases:
- At 20-30 policies: Intermittent reconciliation stability ("lottery effect")
- At 100+ policies: Consistent reconciliation instability
- At 300+ policies: Severe CPU saturation and slow reconciliation

**Expected behavior**

The WasmPlugin comparison should use semantic (order-insensitive) equality for collections where order doesn't matter. Specifically:

- Two functionally identical WASM configurations should be considered equal regardless of the ordering of their ActionSets, Actions, Hostnames, Predicates, and ConditionalData items
- Reconciliation should reach a stable state after applying policies
[...]

PR #1957, opened 2026-05-13
Cache Kuadrant CR lookup in reconciliation state to fix scaling performance

### LINKED ISSUE #1919: Status Enforced Scaling Issue

Boomatang:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/1919

# Summary

`GetKuadrantFromTopology` is called by nearly every reconciler in the operator. Each call performs a `lo.FilterMap` over all topology roots, type-asserts, filters by GroupKind and deletion timestamp, then sorts by creation timestamp. In large topologies (hundreds of HTTPRoutes with policies), this becomes a significant bottleneck because the function is invoked many times per reconciliation cycle.

This change introduces a simple cache-on-first-access pattern: `GetKuadrantFromTopology` now accepts the shared `*sync.Map` reconciliation state and stores the result under a `"kuadrant"` key after the first lookup. Subsequent calls within the same reconciliation cycle return the cached value directly, avoiding repeated topology traversal.

All callers of `GetKuadrantFromTopology`, `GetAuthorinoFromTopology`, and `GetLimitadorFromTopology` across 28 files are updated to pass the `state` parameter through. The Authorino and Limitador helper functions benefit transitively since they delegate to `GetKuadrantFromTopology` internally.

# Dependency

This change depends on https://github.com/Kuadrant/kuadrant-operator/pull/1923, which fixes WasmPlugin comparison to use order-insensitive equality checks. Without that fix, non-deterministic map iteration causes phantom spec changes that trigger unnecessary reconciliation cycles, amplifying the very performance problem this PR addresses.

# Acceptance criteria

The issue does not define explicit acceptance criteria, but describes the scaling problem with `enforcedCondition` and related status update functions where topology traversal time grows with resource count. This PR directly addresses the root cause by ensuring the expensive `lo.FilterMap` + sort over topology roots in `GetKuadrantFromTopology` executes at most once per reconciliation cycle, reducing the per-call cost from O(n) to O(1) for all subsequent invocations.

# Findings

Using the test set up in [perftest:boomatang-perf-0](https://github.com/Kuadrant/perftest/tree/boomatang-perf-0) the following table of results were created based on the attached log files. In the naming the "patch" refers to the #1923 code rebase on current main. The "fix" is this change set on the same patched main version. CPU usage is eye balled in k9s, and is a very around that value. The deploy of kuadrant-operator was modified to allow 10 vCPU and  2 GB of RAM. RAM usage was not monitored. 

The test run used 600 resource for the test. That is 600 HTTProutes, each with an AuthPolicy and RateLimitPolicy attached. In total 600 HTTPRoutes, 600 AuthPolicy and 600 RateLimitPolicy.

|run|start log line|start time|end log line |end time | CPU | Total time | 
| --- | --- | --- | ---| --- | --- | --- |
| patch only | 107 | 10:22:48 | 1895 | 11:01:30 | ~3.5 | 00:38:42 |
|patch + fix | 118 | 11:39:51 | 3904 | 11:43:19 | ~2.0 | 00:03:28 |

[kuadrant-operator+patch.log](https://github.com/user-attachments/files/27710057/kuadrant-operator%2Bpatch.log)
[kuadrant-operator+patch+fix.log](https://github.com/user-attachments/files/27710059/kuadrant-operator%2Bpatch%2Bfix.log)

## Note on log lines
The `patch + fix` logs have a far higher number of logs, with a lot being errors similar to `the object has been modified; please apply your changes to the latest version and try again`. This is down to the speed in which the reconcile loop can now run. We are not getting the events fast enough from the API server. 

There is possible later improvements to be made here.

## Summary by CodeRabbit

* **Performance Improvements**
  * Topology lookups now use a shared reconciliation state cache for faster component discovery and fewer redundant operations.

* **Reliability / Behaviour**
  * Reconciliation flows and integrations (auth, rate-limit, tracing, observability) now derive component presence from the shared state, improving consistency of apply/remove decisions.

* **Monitoring**
  * Component readiness and status metrics are computed using the reconciliation state for more accurate status reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-operator/pull/1957)

### DIFF
--- a/internal/controller/auth_policy_status_updater.go
+++ b/internal/controller/auth_policy_status_updater.go
@@ -155,7 +155,7 @@ func (r *AuthPolicyStatusUpdater) UpdateStatus(ctx context.Context, _ []controll
 }
 
 func (r *AuthPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.AuthPolicy, topology *machinery.Topology, state *sync.Map, logger logr.Logger) *metav1.Condition {
-	kObj := GetKuadrantFromTopology(topology)
+	kObj := GetKuadrantFromTopology(topology, state)
 	if kObj == nil {
 		return kuadrant.EnforcedCondition(policy, kuadrant.NewErrSystemResource("kuadrant"), false)
 	}
@@ -291,7 +291,7 @@ func (r *AuthPolicyStatusUpdater) enforcedCondition(policy *kuadrantv1.AuthPolic
 	var componentsToSync []string
 
 	// check the status of Authorino
-	authorino := GetAuthorinoFromTopology(topology)
+	authorino := GetAuthorinoFromTopology(topology, state)
 	if authorino == nil {
 		return kuadrant.EnforcedCondition(policy, kuadrant.NewErrSystemResource("authornio"), false)
 	}
--- a/internal/controller/auth_workflow_helpers.go
+++ b/internal/controller/auth_workflow_helpers.go
@@ -35,8 +35,8 @@ var (
 	ErrMissingStateEffectiveAuthPolicies = fmt.Errorf("missing auth effective policies stored in the reconciliation state")
 )
 
-func GetAuthorinoFromTopology(topology *machinery.Topology) *authorinooperatorv1beta1.Authorino {
-	kuadrant := GetKuadrantFromTopology(topology)
+func GetAuthorinoFromTopology(topology *machinery.Topology, state *sync.Map) *authorinooperatorv1beta1.Authorino {
+	kuadrant := GetKuadrantFromTopology(topology, state)
 	if kuadrant == nil {
 		return nil
 	}
--- a/internal/controller/authconfigs_reconciler.go
+++ b/internal/controller/authconfigs_reconciler.go
@@ -53,7 +53,7 @@ func (r *AuthConfigsReconciler) Subscription() controller.Subscription {
 func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
 	logger := controller.LoggerFromContext(ctx).WithName("AuthConfigsReconciler").WithValues("context", ctx)
 
-	authorino := GetAuthorinoFromTopology(topology)
+	authorino := GetAuthorinoFromTopology(topology, state)
 	if authorino == nil {
 		logger.V(1).Info("authorino resource not found in the topology")
 		return nil
--- a/internal/controller/authorino_istio_integration_reconciler.go
+++ b/internal/controller/authorino_istio_integration_reconciler.go
@@ -62,7 +62,7 @@ func (a *AuthorinoIstioIntegrationReconciler) Run(baseCtx context.Context, _ []c
 	logger.V(1).Info("reconciling authorino integration in istio", "status", "started")
 	defer logger.V(1).Info("reconciling authorino integration in istio", "status", "completed")
 
-	kObj := GetKuadrantFromTopology(topology)
+	kObj := GetKuadrantFromTopology(topology, state)
 
 	if kObj == nil {
 		// Nothing to be done. It is expected that the authorino resource managed by kuadrant
--- a/internal/controller/authorino_reconciler.go
+++ b/internal/controller/authorino_reconciler.go
@@ -40,13 +40,13 @@ func (r *AuthorinoReconciler) Subscription() *controller.Subscription {
 	}
 }
 
-func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, _ *sync.Map) error {
+func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, state *sync.Map) error {
 	span := trace.SpanFromContext(ctx)
 	logger := controller.LoggerFromContext(ctx).WithName("AuthorinoReconciler").WithValues("context", ctx)
 	logger.V(1).Info("reconciling authorino resource", "status", "started")
 	defer logger.V(1).Info("reconciling authorino resource", "status", "completed")
 
-	kobj := GetKuadrantFromTopology(topology)
+	kobj := GetKuadrantFromTopology(topology, state)
 	if kobj == nil {
 		span.AddEvent("no kuadrant object found")
 		span.SetStatus(codes.Ok, "")
@@ -59,7 +59,7 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 		attribute.String("kuadrant.namespace", kobj.Namespace),
 	)
 
-	clusterAuthorino := GetAuthorinoFromTopology(topology)
+	clusterAuthorino := GetAuthorinoFromTopology(topology, state)
 
 	if clusterAuthorino != nil {
 		span.AddEvent("Authorino resource already exists")
--- a/internal/controller/developerportal_reconciler_test.go
+++ b/internal/controller/developerportal_reconciler_test.go
@@ -86,7 +86,7 @@ func TestDeveloperPortalReconciler(t *testing.T) {
 
 	t.Run("Topology with Kuadrant CR", func(subT *testing.T) {
 		topology := buildTopologyWithKuadrant(subT, true)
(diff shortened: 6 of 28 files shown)


Boomatang:
At the time I didn't have time to dig into the `lo.FilterMap` to see if there is better improvements that could be made there. I do think it might be worth while reviewing all usage of `lo` package.

guicassolato (approved):
Spot on!

Boomatang:
@guicassolato as much as I want to merge this now. We can't with out first merging https://github.com/Kuadrant/kuadrant-operator/pull/1923. This PR does not "depend" on that PR, but without that patch when the number of resource is large the number of reconciles will explode. Before the reconcile was slow, but now it is not such. 

Without #1923 I would worry about what side affect we cause to the cluster, operator resource usage and such.

guicassolato:
> @guicassolato as much as I want to merge this now. We can't with out first merging #1923. This PR does not "depend" on that PR […]

Shame, because this seems to be a nice improvement regardless.

Using the state to find the Kuadrant object in the topology just once and then reuse in the next steps seems like a good idea. Unless accessing the `sync.Map` under concurrency turns out to be too much overhead that makes it perform even worse than `lo.FilterMap(topology.Objects().Roots(), …)`, but I don't think so.


> without that patch when the number of resource is large the number of reconciles will explode. Before the reconcile was slow, but now it is not such.
>
> Without #1923 I would worry about what side affect we cause to the cluster, operator resource usage and such.

I share the same worry but about #1923. I don't think the approach adopted in that PR is the correct one yet. It may reduce the number of calls to the reconciliation function it takes until it stabilises for large number of resources, but it does that possibly without fixing the bug it claims and yet introducing a new one.

Either way, whatever it happens with that other PR, it does not make this one wrong, right? Without the "patch", perhaps we would not see the same benchmarks as in the description, but what you did here seems to make things better anyway, no?

What's the harm in merging this PR? How do you worry about resource utilisation getting worse with this fix while #1923 is not merged? If anything, we seem to be reducing CPU utilisation (with less calls to `lo.FilterMap`) in exchange for an almost negligible increment in memory (exactly the number of bytes one copy of the Kuadrant CR occupies in the heap). Am I missing something?

mikenairn (approved):
I have tested this with [this](https://github.com/Kuadrant/kuadrant-operator/pull/1923) change and without and it doesn't depend on it. This is a significant improvement and i think we should merge it.

I don't see any issue here tbh, the other issue might cause more reconciles, but this makes each faster and uses less memory at scale(slight increase of cpu usage at lower usage due to `sync.Map` but this is probably an ok trade-off? ) so it's an improvement on an improvement.

Here is a report of what i tested:

## Benchmark: FullReconciliationCycle (unit benchmark)

This benchmark simulates a single reconciliation cycle by constructing a realistic Kuadrant topology in-memory (1 GatewayClass, configurable Gateways and Listeners, N HTTPRoutes each with an AuthPolicy, RateLimitPolicy, and TokenRateLimitPolicy attached) and then performing the operations that happen in a real cycle:

1. Calls `GetKuadrantFromTopology` 28 times (once per reconciler call site in the operator)
2. Calls `CalculateEffectiveAuthPolicies` to compute effective policies across the full topology

No resources are created on a cluster — the topology is constructed in-memory using the same types the operator uses (`machinery.NewGatewayAPITopology`). This measures the cost of a single reconciliation cycle in isolation.

### CPU (sec/op)

| Routes | Main | PR #1957 | Change | Significant? |
|---|---|---|---|---|
| 10 | 1.89ms | 1.35ms | **-28.5%** | p=0.002 |
| 100 | 69.5ms | 71.8ms | +3.2% | ~ |
| 300 | 565ms | 605ms | **+7.0%** | p=0.002 |
| 1000 | 6.61s | 7.23s | **+9.3%** | p=0.002 |
| 32 listen × 100 | 1.71s | 1.82s | +6.6% | ~ |
| 3gw × 16listen × 300 | 8.61s | 8.07s | **-6.2%** | p=0.009 |

### Memory (B/op)

| Routes | Main | PR #1957 | Change |
|---|---|---|---|
| 10 | 721 KB | 518 KB | **-28.2%** |
| 100 | 23.6 MB | 21.7 MB | **-7.9%** |
| 300 | 188 MB | 183 MB | **-3.0%** |
| 1000 | 1.98 GB | 1.96 GB | **-0.9%** |

### Allocations (allocs/op)

| Routes | Main | PR #1957 | Change |
|---|---|---|---|
| 10 | 19,584 | 10,917 | **-44.3%** |
| 100 | 637,800 | 556,227 | **-12.8%** |
| 300 | 4,913,362 | 4,669,777 | **-5.0%** |
| 1000 | 51,397,739 | 50,587,040 | **-1.6%** |

### Unit Benchmark Analysis

Per-cycle CPU time **increases slightly** (~7-9% at larger scales) due to `sync.Map` overhead from threading the `state` parameter through 28 call sites. However, memory allocations **decrease significantly** (up to 44% at small scale) because the cached lookup avoids repeated `lo.FilterMap` + sort operations.

---

## Performance Test: Reconciliation Scale (real cluster)

This test runs the full operator reconciliation loop in-process against a real Kubernetes cluster with Istio. For each scale level it:

1. Creates a Gateway with a single HTTP listener
2. Creates N HTTPRoutes, each with an AuthPolicy attached
3. Measures wall-clock time until all AuthPolicies reach `Enforced: True` status
4. Cleans up the namespace

Unlike the unit benchmark, this captures the full operator lifecycle: event handling, topology rebuilding, effective policy calculation, AuthConfig creation, Authorino reconciliation, status updates, and API server I/O across **hundreds of reconciliation cycles** triggered by bulk resource creation.

### Reconciliation Time

| Routes | Main | PR #1957 | Change |
|---|---|---|---|
| 10 | 5.0s | 5.0s | ~ |
| 50 | 1m 0s | 5.0s | **-93%** |
| 100 | 15.1s | 10.1s | **-33%** |
| 300 | 3m 47s | 20.3s | **-91%** |

### CPU Usage (pprof, cumulative across all scale levels)

| | Main | PR #1957 | Change |
|---|---|---|---|
| Total CPU samples | 751.6s | 55.0s | **-93%** |
| Test duration | 432s | 141s | -67% |

The operator consumed **93% less CPU** across the full test run. Fewer reconciliation cycles means less time spent in topology traversal, status updates, and API server I/O.

### Memory Allocations (pprof, cumulative across all scale levels)

| | Main | PR #1957 | Change |
|---|---|---|---|
| Total allocated | 127.7 GB | 6.9 GB | **-95%** |
| `LocatorFromObject` | 125.0 GB | 4.5 GB | -96% |
| `fmt.Sprintf` | 50.1 GB | 2.2 GB | -96% |
| `enforcedCondition` | 116.9 GB | — | eliminated from hot path |

The main branch allocated **127 GB** over the test duration across hundreds of reconciliation cycles. With PR #1957, total allocations dropped to **6.9 GB** — a 95% reduction. The `LocatorFromObject` function (which computes string locators via `fmt.Sprintf` on every DFS node visit) dropped from 125 GB to 4.5 GB because far fewer reconciliation cycles execute, and therefore far fewer DFS traversals occur.

### Real Cluster Analysis

The dramatic improvement (3m47s → 20s at 300 routes) comes from the caching benefit **compounding across hundreds of reconciliation cycles** during bulk resource creation — each cycle saves 28 redundant topology lookups. This effect is not visible in unit benchmarks which measure a single cycle.

The 50-route case on main (1m 0s) shows the non-deterministic WasmPlugin comparison bug (Issue #1934) causing reconciliation loops that eventually resolve by luck. PR #1957 makes each cycle cheap enough that the operator converges quickly even with phantom changes.

PR #1986, opened 2026-05-20
feat: Extension SDK action pipeline improvements

### LINKED ISSUE #1958: Extension SDK Action Pipeline Improvements
# Feature: Extension SDK Action Pipeline Improvements

## Summary

Redesign the Phase 3 action pipeline API to unify the action model with a single `Action` interface, inter-action data flow via GRPCMethodAction variables, and phase-agnostic actions. This replaces the separate `RequestAction`/`ResponseAction` type hierarchy from the [original design](doc/proposals/extensions-SDK-action-pipeline-design.md) with a single `Action` interface where all actions can be used in either the HTTP request or response phase.

Reference: [Original Phase 3 Design](doc/proposals/extensions-SDK-action-pipeline-design.md), [Kuadrant/kuadrant-operator#1889](https://github.com/Kuadrant/kuadrant-operator/issues/1889)

## Goals

Demo: a ThreatPolicy extension that uses the redesigned pipeline API to demonstrate:

1. **Register a threat-scoring gRPC service** using `RegisterActionMethod` (unchanged from original design)
2. **Block known-bad paths** with `DenyAction` in the request phase — deny requests to `/blocked` before they reach the backend
3. **Call the threat service** with `GRPCMethodAction` — invoke the gRPC method and store the response in a `threatResponse` variable
4. **Signal failures** with `FailAction` — when the gRPC response indicates an error (e.g. `threatResponse.error_code != 0`), log a message and terminate the action chain
5. **Deny based on gRPC response** with `DenyAction` in the response phase — reference the `threatResponse` variable from step 3 to deny when the threat level is high, demonstrating inter-action data flow and the difference between denying a request (send code to origin) vs denying a response (send code to destination)
6. **Add response headers** with `AddHeadersAction` in the response phase — confirm the threat check ran by adding `x-threat-checked: true`, demonstrating the same action type working in a different phase than step 2's request-phase usage
7. **Observe ordering validation** — attempt to reference a variable before it is populated and observe the immediate error, then attempt to add a request-phase action after a response-phase action and observe the error

## Non-Goals

- Branching or conditional control flow beyond per-action predicates
- Multi-cluster pipeline orchestration
- Variable scoping rules within the wasm-shim (owned by the wasm-shim, not the operator)
- CEL evaluation semantics for variables at request time (owned by the wasm-shim)

### Changes from Original Design

| Area | Original | Redesign |
|------|----------|----------|
| Action interfaces | Separate `RequestAction` / `ResponseAction` | Single `Action` interface |
| Phase methods | `OnRequest` / `OnResponse` | `OnHTTPRequest` / `OnHTTPResponse` |
| Method signature | Variadic `...RequestAction`, takes `ctx` | Variadic `...Action`, no `ctx` (local accumulation) |
| AllowAction | Allows when intention is true | Removed — replaced by `DenyAction` |
| DenyAction | N/A | Denies when predicate is true, with optional status code, headers, and body |
| FailAction | N/A | Logs error message and terminates the action chain |
| AddHeadersAction | Response phase only | Either phase |
| WithResponseCodeAction | Modifies response status code | Removed |
| GRPCMethodAction.Intention | CEL expression evaluated against response | Removed — response stored in `Var` |
| GRPCMethodAction.Var | N/A | Variable name to store gRPC response |
| Intention field | Separate field on GRPCMethodAction and AllowAction | Removed — `Predicate` is the unified condition field |
| Inter-action data flow | Non-Goal | Supported via `Var` on `GRPCMethodAction` |
| Predicate field | `[]string` (multiple, AND semantics) | `string` (single CEL expression) |

### Backwards Compatibility

**Breaking change**: The `Pipeline` interface, action types, and phase method names all change. Extensions using the Phase 3 API from the original design must update. Since the extension SDK is pre-1.0, this is acceptable.

`RegisterActionMethod` and `ActionMethodConfig` are unchanged. `NewPipeline(policy)` signature is unchanged.

### Architecture Changes

```text
Extension Controller                    Operator (gRPC server)
   │                                       │
   │── RegisterActionMethod ──────────────►│── (unchanged from original design)
   │◄── nil / error ───────────────────────│
   │                                       │
   │── NewPipeline(policy) ─── (local) ───►│  (no I/O — creates PipelineImpl)
   │                                       │
   │── pipeline.OnHTTPRequest(actions...) ►│  (local: validate ordering, accumulate)
   │── pipeline.OnHTTPRequest(actions...) ►│  (local: validate ordering, accumulate)
   │── pipeline.OnHTTPResponse(actions...)►│  (local: validate ordering, accumulate)
   │                                       │
   │── pipeline.Commit(ctx) ──────────────►│── Validate CEL expressions
   │                                       │── Cross-action Var→proto schema validation
   │                                       │── Atomically replace pipeline actions
   │                                       │── Trigger reconciliation
   │◄── nil / error ───────────────────────│
```

#### Action Interface (replaces RequestAction / ResponseAction)

```go
// pkg/extension/types/actions.go

// Action is the interface implemented by all pipeline action types.
// Actions can be used in either the request or response phase.
type Action interface {
    actionType() ActionType
}

// ActionType discriminates how the wasm-shim dispatches an action.
type ActionType string

const (
    ActionTypeGRPCMethod ActionType = "grpc_method"
    ActionTypeDeny       ActionType = "deny"
    ActionTypeFail    ActionType = "fail"
    ActionTypeAddHeaders ActionType = "add_headers"
)
```

#### Action Types

All action types share a unified `Predicate` field — a CEL expression that determines whether the action executes. When `Predicate` evaluates to true (or is empty), the action runs. When false, the action is skipped.

```go
// pkg/extension/types/actions.go

// GRPCMethodAction invokes a registered gRPC action method and optionally
// stores the response in a named variable for use by subsequent actions.
type GRPCMethodAction struct {
    Predicate string // CEL — if true, call the gRPC method
    Method    string // Name of a registered ActionMethod
    Var       string // Variable name to store gRPC response (optional)
}

// DenyAction denies the request or response when the predicate evaluates
// to true. All response fields are optional.
//
// Phase semantics:
//   - Request phase: deny sends the response to the origin
//     (request never reaches backend)
//   - Response phase: deny sends the response to the destination
//     (backend response replaced before reaching client)
//
// The operator builds a wasm-shim DenyResponse CEL expression from the
// provided fields: DenyResponse{status: 403u, headers: {...}, body: '...'}
type DenyAction struct {
    Predicate   string // CEL — if true, deny
    WithStatus  int    // HTTP status code (e.g. 403); optional
    WithHeaders string // CEL expression — array of [name, value] pairs; optional
    WithBody    string // Response body string; optional
}

// FailAction logs an error message and terminates the action chain when
// the predicate evaluates to true. Maps to the wasm-shim's "fail" type.
type FailAction struct {
    Predicate  string // CEL — if true, fail with log message
    LogMessage string // Error message to log
}

// AddHeadersAction adds headers to the request or response depending on
// the phase in which it is used, when the predicate evaluates to true.
//
// Phase semantics:
//   - Request phase: headers added to the request before it reaches the backend
//   - Response phase: headers added to the response before it reaches the client
type AddHeadersAction struct {
    Predicate    string // CEL — if true, add the headers
    HeadersToAdd string // CEL expression evaluating to a map of headers
}
```

#### Pipeline Interface

```go
// pkg/extension/types/types.go

// Pipeline provides a builder for composing ordered actions on HTTP request
// and response phases. Actions accumulate locally with immediate ordering
// validation. Commit sends all actions atomically to the operator.
//
// Multiple actions can be passed to a single OnHTTPRequest/OnHTTPResponse
// call. Actions within a call are processed in order — if action 1
// references a variable that action 2 populates, the call returns an error
// because action 1 precedes action 2.
type Pipeline interface {
    OnHTTPRequest(actions ...Action) error
    OnHTTPResponse(actions ...Action) error
    Commit(ctx context.Context) error
}
```

**Why no `ctx` on OnHTTPRequest/OnHTTPResponse?** These methods accumulate actions locally and validate ordering — no I/O occurs. Only `Commit` performs I/O (the gRPC call to the operator) and therefore requires `ctx`.

#### KuadrantCtx Interface

No changes — `RegisterActionMethod` and `NewPipeline` signatures are unchanged.

#### gRPC Proto Changes

- **PipelineCommit RPC** — request message updated: single `repeated ActionEntry` list (replacing separate request/response lists), each entry has a `phase` field (`request` or `response`)
- **ActionType enum** — values changed: `GRPC_METHOD`, `DENY`, `FAIL`, `ADD_HEADERS` (removed `ALLOW`, `WITH_RESPONSE_CODE`)
- **ActionEntry** — gains `var`, `with_status`, `with_headers`, `with_body`, `log_message` fields; drops `intention`, `new_response_code`, `deny_with`

### Pipeline Ordering Validation

`PipelineImpl` tracks state across calls to enforce ordering constraints. Validation applies both within a single call (between actions in the variadic list) and across calls:

1. **Phase ordering**: Once `OnHTTPResponse` has been called, subsequent `OnHTTPRequest` calls return an error. Request actions must be added before response actions.

2. **Variable availability**: When an action's `Predicate` (or `HeadersToAdd`) references a variable name, the pipeline checks that a prior action (either earlier in the same call or in a previous call) populates that variable via a `GRPCMethodAction`'s `Var` field. If not, the call returns an error immediately.

3. **Duplicate variable names**: If two `GRPCMethodAction` entries use the same `Var` name, the call returns an error.

Checks 1–3 happen locally in the client — no gRPC call is needed. Server-side validation (CEL syntax, proto schema validation of Var fields) happens on `Commit`.

### Cross-Action CEL Validation

When `Commit` is called, the operator performs cross-action validation for CEL expressions that reference variables:

1. Build a map of `Var → ActionMethod` from all `GRPCMethodAction` entries
2. For each action with a CEL expression (`Predicate`, `HeadersToAdd`), extract variable references
3. For each referenced variable, look up the originating `ActionMethod` in the `ActionMethodStore`
4. Retrieve the method's response proto schema from the ProtoCache
5. Validate that the CEL expression's field accesses are valid on the response message type
6. Reject the entire commit if any validation fails

Example: `DenyAction{Predicate: "threatResponse.threat_level >= 5", WithStatus: 403}` → find `threatResponse` originates from `GRPCMethodAction{Var: "threatResponse", Method: "assess-threat"}` → look up `"assess-threat"` → get its response proto → validate `threat_level` is a numeric field on that message.

### DenyAction Phase Semantics

DenyAction behaviour changes based on the phase:

- **Request phase** (`OnHTTPRequest`): When the predicate evaluates to true, the request is denied before reaching the backend. The deny response (status, headers, body) is sent to the client.
- **Response phase** (`OnHTTPResponse`): When the predicate evaluates to true, the backend's response is discarded and replaced with the deny response sent to the client.

All response fields (`WithStatus`, `WithHeaders`, `WithBody`) are optional. The operator builds a wasm-shim `DenyResponse` CEL expression from whichever fields are provided, e.g. `DenyResponse{status: 403u, headers: {'x-reason': 'blocked'}, body: 'Forbidden'}`.

This distinction matters for use cases like threat assessment: the gRPC service runs during the request phase, but the deny decision based on its response may need to happen in the response phase after the backend has responded.

### Wasm Changes

The wasm-shim uses `TypedAction` (`internal/wasm/types.go`) for extension pipeline actions. Any action type can be a root-level entry. `GRPCMethodAction` entries use the `onReply` field to nest actions that depend on the gRPC response variable.

#### TypedAction Struct

The `TypedAction` struct's `DenyWith` field holds the constructed `DenyResponse{...}` CEL expression (built by the operator from the SDK's `WithStatus`/`WithHeaders`/`WithBody` fields). The `LogMessage` field supports `FailAction`:

```go
type TypedAction struct {
    Type                 string        `json:"type"`
    Predicate            string        `json:"predicate"`
    Terminal             bool          `json:"terminal"`
    Var                  string        `json:"var,omitempty"`
    Service              string        `json:"service,omitempty"`
    MessageBuilder       string        `json:"messageBuilder,omitempty"`
    OnReply              []TypedAction `json:"onReply,omitempty"`
    DenyWith             string        `json:"denyWith,omitempty"`
    Target               string        `json:"target,omitempty"`
    Headers              string        `json:"headers,omitempty"`
    LogMessage           string        `json:"logMessage,omitempty"` // NEW
    SourcePolicyLocators []string      `json:"sources,omitempty"`
}
```

#### TypedAction Type Values

| SDK ActionType | Wasm `type` value |
|---|---|
| `ActionTypeGRPCMethod` | `grpc` |
| `ActionTypeDeny` | `deny` |
| `ActionTypeFail` | `fail` |
| `ActionTypeAddHeaders` | `headers` |

#### SDK → Wasm Translation

The reconciler translates the flat SDK pipeline into `TypedAction` entries:

- Actions that do not reference a gRPC response variable are emitted as root-level TypedActions
- A `GRPCMethodAction` becomes a root-level `grpc` TypedAction; actions that reference its `Var` are placed in `onReply`
- `Terminal: true` is set on `deny` and `fail` actions (they stop the chain when their predicate matches)
- `Target: "response"` is set on response-phase `headers` actions
- `MessageBuilder` on the `grpc` TypedAction is populated from the registered `ActionMethodConfig.MessageTemplate`
- If the pipeline contains multiple `GRPCMethodAction` entries, each becomes a root-level `grpc` action; var-dependent actions are placed in the `onReply` of the `grpc` action that populates their referenced variable

#### Example Wasm Config

```json
{
  "services": {
    "ext-a1b2c3d4": {
      "type": "dynamic",
      "endpoint": "ext-threat-service-security-svc-cluster-local-8080",
      "failureMode": "deny",
      "timeout": "100ms",
      "grpcService": "threat.ThreatService",
      "grpcMethod": "AssessThreat"
    }
  },
  "actionSets": [
    {
      "name": "abc123-hash",
      "routeRuleConditions": {
        "hostnames": ["api.example.com"]
      },
      "actions": [
        {
          "type": "deny",
          "predicate": "request.url_path == '/blocked'",
          "terminal": true,
          "denyWith": "DenyResponse{status: 403u, body: 'Forbidden'}",
          "sources": ["ThreatPolicy/default/my-threat-policy"]
        },
        {
          "type": "grpc",
          "predicate": "true",
          "terminal": false,
          "var": "threatResponse",
          "service": "ext-a1b2c3d4",
          "messageBuilder": "AssessThreatRequest { uri: request.path, source_ip: source.address }",
          "sources": ["ThreatPolicy/default/my-threat-policy"],
          "onReply": [
            {
              "type": "fail",
              "predicate": "threatResponse.error_code != 0",
              "terminal": true,
              "logMessage": "Threat service returned unexpected error"
            },
            {
              "type": "deny",
              "predicate": "threatResponse.threat_level >= 5",
              "terminal": true,
              "denyWith": "DenyResponse{status: 403u, headers: {'x-threat-assessed': 'true'}, body: 'Blocked by threat policy'}"
            },
            {
              "type": "headers",
              "predicate": "true",
              "terminal": false,
              "target": "response",
              "headers": "{\"x-threat-checked\": \"true\"}"
            }
          ]
        }
      ]
    }
  ]
}
```

#### 1. ActionMethodStore (internal/extension/registry.go)

Unchanged from original design.

#### 2. PipelineActionStore (internal/extension/registry.go)

Updated to store unified action entries. Keyed by `{PolicyResourceID, Phase, Index}`. Each entry holds:
- `ActionType` — one of `grpc_method`, `deny`, `fail`, `add_headers`
- `Predicate` — CEL predicate string (optional)
- `Phase` — `request` or `response`
- Type-specific fields: `Method`, `Var`, `WithStatus`, `WithHeaders`, `WithBody`, `LogMessage`, `HeadersToAdd`

#### 3. Server-Side PipelineCommit Handler (internal/extension/manager.go)

Updated validation:
- Validates all actions: `ActionType` is a valid type, `Predicate` CEL is syntactically valid
- `grpc_method`: `Method` references a registered action method
- `deny`: `WithStatus` (if set) is a valid HTTP status code (100–599), `WithHeaders` (if set) is valid CEL
- `fail`: `LogMessage` is a non-empty string
- `add_headers`: `HeadersToAdd` CEL is syntactically valid
- **Cross-action validation**: variable references in `Predicate` or `HeadersToAdd` CEL expressions are validated against the proto schema of the originating `GRPCMethodAction`'s response type
- Atomically replaces all pipeline actions for both phases

#### 4. Client-Side PipelineImpl (pkg/extension/controller/controller.go)

- **`OnHTTPRequest(actions ...Action) error`** — validates ordering across all accumulated actions (no request after response, var available), accumulates actions with phase `request`
- **`OnHTTPResponse(actions ...Action) error`** — validates ordering (var available), accumulates actions with phase `response`, sets `responsePhaseStarted` flag
- **`Commit(ctx context.Context) error`** — converts accumulated actions to proto entries, sends `PipelineCommit` RPC
- Tracks `populatedVars` set (Var names from `GRPCMethodAction` entries) and `responsePhaseStarted` bool

#### 5. Extension Reconcilers (internal/controller/)

The reconcilers translate the flat pipeline representation from `PipelineActionStore` into `TypedAction` entries for the wasm config:

- Actions that do not reference a gRPC response variable are emitted as root-level TypedActions
- `GRPCMethodAction` entries become root-level `grpc` TypedActions; var-dependent actions are placed in `onReply`
- `Terminal: true` is set on `deny` and `fail` actions
- `Target: "response"` is set on response-phase `headers` actions
- `MessageBuilder` is populated from the `ActionMethodStore` entry for the referenced method

#### Extension Author Usage

```go
func (r *ThreatPolicyReconciler) reconcileSpec(
    ctx context.Context,
    pol *v1alpha1.ThreatPolicy,
    kCtx types.KuadrantCtx,
) (*v1alpha1.ThreatPolicyStatus, error) {
    err := kCtx.RegisterActionMethod(ctx, pol, types.ActionMethodConfig{
        Name:            "assess-threat",
        URL:             threatServiceURL,
        Service:         "threat.ThreatService",
        Method:          "AssessThreat",
        MessageTemplate: `AssessThreatRequest { uri: request.path, source_ip: source.address }`,
    })
    if errors.Is(err, types.ErrUpstreamUnreachable) {
        return calculateErrorStatus(pol, err), err
    }
    if err != nil {
        return calculateErrorStatus(pol, err), err
    }

    pipeline := kCtx.NewPipeline(pol)

    if err := pipeline.OnHTTPRequest(
        types.DenyAction{
            Predicate:   `request.url_path == "/blocked"`,
            WithStatus:  403,
            WithHeaders: `[["x-blocked-reason", "path-denied"]]`,
            WithBody:    "Forbidden",
        },
        types.GRPCMethodAction{
            Method: "assess-threat",
            Var:    "threatResponse",
        },
    ); err != nil {
        return calculateErrorStatus(pol, err), err
    }

    // Denying a response is different to denying a request:
    //   deny a request: send the code to the origin
    //   deny a response: send the code to the destination
    if err := pipeline.OnHTTPResponse(
        types.FailAction{
            Predicate:  `threatResponse.error_code != 0`,
            LogMessage: "Threat service returned unexpected error",
        },
        types.DenyAction{
            Predicate:   `threatResponse.threat_level >= 5`,
            WithStatus:  403,
            WithHeaders: `[["x-threat-assessed", "true"]]`,
            WithBody:    "Blocked by threat policy",
        },
        types.AddHeadersAction{
            HeadersToAdd: `{"x-threat-checked": "true"}`,
        },
    ); err != nil {
        return calculateErrorStatus(pol, err), err
    }

    if err := pipeline.Commit(ctx); err != nil {
        return calculateErrorStatus(pol, err), err
    }

    return calculateEnforcedStatus(pol, nil), nil
}
```

### Security Considerations

- **Unchanged from original design**: Name validation, policy-scoped lifecycle, no user-controlled names in Envoy/wasm config, ActionType validation
- **Variable name validation**: `Var` names are validated to contain only alphanumeric characters, hyphens, and underscores
- **Cross-action validation**: CEL expressions referencing variables are validated against proto schemas at commit time, preventing malformed field access from reaching the wasm-shim
- **WithStatus validation**: Status codes (when provided) are validated to be in the 100–599 range

## Implementation Plan

1. Update action type definitions: remove `AllowAction`, `WithResponseCodeAction`; add `DenyAction`, `FailAction`; update `GRPCMethodAction` (remove `Intention`, add `Var`); update `AddHeadersAction` for either phase; remove `Intention` field entirely; unify under single `Action` interface
2. Update `Pipeline` interface: rename to `OnHTTPRequest`/`OnHTTPResponse` with variadic `...Action`, add `Commit`
3. Implement client-side ordering validation in `PipelineImpl`
4. Update gRPC proto: `ActionType` enum, `ActionEntry` message, `PipelineCommit` request
5. Update server-side `PipelineCommit` handler: new action type validation, cross-action Var→proto validation
6. Update `PipelineActionStore` for unified action entries
7. Add `logMessage` field to wasm `TypedAction` struct; add `fail` type support
8. Update extension reconcilers: translate flat pipeline into `TypedAction` entries with `onReply` for var-dependent actions
9. Wasm-shim already supports `fail` type — no wasm-shim changes needed

## Testing Strategy

- **Unit tests**: Action ordering validation (var before populated, request after response, duplicate vars — both within a single call and across calls), DenyAction/FailAction/AddHeadersAction field validation, cross-action CEL Var resolution, flat pipeline → TypedAction translation (root-level and onReply placement), PipelineImpl accumulation and Commit serialisation
- **Integration tests**: End-to-end pipeline with inter-action data flow — GRPCMethodAction populates var, DenyAction references var in Predicate; verify wasm config produces correct TypedAction structure; verify ordering errors are surfaced
- **E2E tests**: Deploy extension with full pipeline, verify GRPCMethodAction var is available to subsequent DenyAction, verify DenyAction behaves differently in request vs response phase, verify FailAction logs error and terminates chain

philbrookes:
## Summary

Requires: https://github.com/Kuadrant/threat-assessment-service/pull/4 for local testing of FailAction.

Implements #1958 — Extension SDK Action Pipeline Improvements.

- Unify action types under a single `Action` interface with `GRPCMethodAction`, `DenyAction`, `FailAction`, and `AddHeadersAction`
- Replace per-type proto messages with a unified `ActionEntry` and `ActionType` enum
- Add client-side pipeline ordering validation: phase ordering, variable availability, duplicate variable detection
- Implement server-side `PipelineCommit` handler with full validation and `PipelineActionStore`
- Translate pipeline actions to wasm `TypedAction` with automatic `onReply` nesting for var-dependent actions
- Enhance `DenyAction` with optional `WithStatus`, `WithHeaders`, and `WithBody` fields that build a `DenyResponse{}` CEL expression for the wasm-shim
- Rename `FailureAction` to `FailAction` aligned with wasm-shim's `fail` type (log message only)
- Add `EXTRA_EXTENSIONS` build arg for optionally including additional extensions (e.g. `threat-policy`)
- Update threat-policy demo extension to exercise all action types

## Test plan

- [x] Unit tests pass (`make test-unit`)
- [x] Proto tests verify serialization of unified `ActionEntry`
- [x] Pipeline ordering validation tests cover forward references, phase ordering, duplicate vars
- [x] Registry tests verify `onReply` nesting and `DenyResponse{}` expression construction
- [x] End-to-end verified on local Kind cluster with Istio, wasm-shim (main), and threat-policy extension

Closes #1958

## Summary by CodeRabbit

* **New Features**
  * Configure and build extra extensions via EXTRA_EXTENSIONS and deploy a shim image; operator can fetch wasm via OCI.
  * Unified action model with DENY/FAIL/ADD_HEADERS/GRPC_METHOD, stricter cross-action validation and atomic pipeline updates.
  * Threat‑policy now enforces mandatory assessment, consistent threat headers and explicit 403 deny behaviour.

* **Documentation**
  * Local development guide expanded with step‑by‑step shim build, tag/push and operator configuration.

* **Tests**
  * Expanded tests covering unified actions, validation, translation and typed-action behaviours.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Kuadrant/kuadrant-operator/pull/1986?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -827,17 +827,17 @@ func (s *extensionService) validateActions(policyID ResourceID, actions []*extpb
 			}
 
 		case extpb.ActionType_ACTION_TYPE_DENY:
-			if err := validateHTTPStatusCode(action.DenyWith, fmt.Sprintf("actions[%d].deny_with", i)); err != nil {
-				return nil, err
+			if action.WithStatus != 0 {
+				if err := validateHTTPStatusCode(fmt.Sprintf("%d", action.WithStatus), fmt.Sprintf("actions[%d].with_status", i)); err != nil {
+					return nil, err
+				}
 			}
-			entry.DenyWith = action.DenyWith
+			entry.WithStatus = int(action.WithStatus)
+			entry.WithHeaders = action.WithHeaders
+			entry.WithBody = action.WithBody
 
-		case extpb.ActionType_ACTION_TYPE_FAILURE:
-			if err := validateHTTPStatusCode(action.FailureCode, fmt.Sprintf("actions[%d].failure_code", i)); err != nil {
-				return nil, err
-			}
-			entry.FailureCode = action.FailureCode
-			entry.FailureMessage = action.FailureMessage
+		case extpb.ActionType_ACTION_TYPE_FAIL:
+			entry.LogMessage = action.LogMessage
 
 		case extpb.ActionType_ACTION_TYPE_ADD_HEADERS:
 			if action.HeadersToAdd == "" {
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -32,10 +32,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/google/cel-go/cel"
-	"github.com/google/cel-go/common"
-	"github.com/google/cel-go/common/ast"
 	celtypes "github.com/google/cel-go/common/types"
-	"github.com/google/cel-go/parser"
 	authorinov1beta3 "github.com/kuadrant/authorino/api/v1beta3"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -783,9 +780,77 @@ func (s *extensionService) PipelineCommit(_ context.Context, request *extpb.Pipe
 	return &emptypb.Empty{}, nil
 }
 
+type actionValidationCtx struct {
+	store       *RegisteredDataStore
+	policyID    ResourceID
+	varToMethod map[string]string
+}
+
+type actionEntryValidator func(action *extpb.ActionEntry, index int, entry *PipelineActionEntry, vctx *actionValidationCtx) error
+
+var actionEntryValidators = map[extpb.ActionType]actionEntryValidator{
+	extpb.ActionType_ACTION_TYPE_GRPC_METHOD: validateGRPCMethodEntry,
+	extpb.ActionType_ACTION_TYPE_DENY:        validateDenyEntry,
+	extpb.ActionType_ACTION_TYPE_FAIL:        validateFailEntry,
+	extpb.ActionType_ACTION_TYPE_ADD_HEADERS: validateAddHeadersEntry,
+}
+
+func validateGRPCMethodEntry(action *extpb.ActionEntry, index int, entry *PipelineActionEntry, vctx *actionValidationCtx) error {
+	if action.Method == "" {
+		return fmt.Errorf("actions[%d]: method must be specified for grpc_method actions", index)
+	}
+	if !vctx.store.HasUpstreamName(vctx.policyID, action.Method) {
+		return fmt.Errorf("actions[%d]: method %q is not a registered action method for this policy", index, action.Method)
+	}
+	if action.Var != "" && !varNameRegexp.MatchString(action.Var) {
+		return fmt.Errorf("actions[%d]: var %q must match [a-zA-Z_][a-zA-Z0-9_]*", index, action.Var)
+	}
+	entry.Method = action.Method
+	entry.Var = action.Var
+	if action.Var != "" {
+		vctx.varToMethod[action.Var] = action.Method
+	}
+	return nil
+}
+
+func validateDenyEntry(action *extpb.ActionEntry, index int, entry *PipelineActionEntry, _ *actionValidationCtx) error {
+	if action.WithStatus != 0 {
+		if err := validateHTTPStatusCode(fmt.Sprintf("%d", action.WithStatus), fmt.Sprintf("actions[%d].with_status", index)); err != nil {
+			return err
+		}
+	}
+	entry.WithStatus = int(action.WithStatus)
+	entry.WithHeaders = action.WithHeaders
+	entry.WithBody = action.WithBody
+	return nil
+}
+
+func validateFailEntry(action *extpb.ActionEntry, index int, entry *PipelineActionEntry, _ *actionValidationCtx) error {
+	if strings.TrimSpace(action.LogMessage) == "" {
+		return fmt.Errorf("actions[%d]: log_message must be specified for fail actions", index)
+	}
+	entry.LogMessage = action.LogMessage
+	return nil
+}
+
+func validateAddHeadersEntry(action *extpb.ActionEntry, index int, entry *PipelineActionEntry, _ *actionValidationCtx) error {
+	if action.HeadersToAdd == "" {
+		return fmt.Errorf("actions[%d]: headers_to_add must be specified for add_headers actions", index)
+	}
+	if err := validateCELExpression(action.HeadersToAdd); err != nil {
+		return fmt.Errorf("actions[%d].headers_to_add: %w", index, err)
+	}
+	entry.HeadersToAdd = action.HeadersToAdd
+	return nil
+}
+
 func (s *extensionService) validateActions(policyID ResourceID, actions []*extpb.ActionEntry) ([]PipelineActionEntry, error) {
 	entries := make([]PipelineActionEntry, 0, len(actions))
-	varToMethod := make(map[string]string)
+	vctx := actionValidationCtx{
+		store:       s.registeredData,
+		policyID:    policyID,
+		varToMethod: make(map[string]string),
+	}
 
 	for i, action := range actions {
 		if action == nil {
@@ -809,57 +874,19 @@ func (s *extensionService) validateActions(policyID ResourceID, actions []*extpb
 			Phase:      action.Phase,
 		}
 
-		switch action.ActionType {
-		case extpb.ActionType_ACTION_TYPE_GRPC_METHOD:
-			if action.Method == "" {
-				return nil, fmt.Errorf("actions[%d]: method must be specified for grpc_method actions", i)
-			}
-			if !s.registeredData.HasUpstreamName(policyID, action.Method) {
-				return nil, fmt.Errorf("actions[%d]: method %q is not a registered action method for this policy", i, action.Method)
-			}
-			if action.Var != "" && !varNameRegexp.MatchString(action.Var) {
-				return nil, fmt.Errorf("actions[%d]: var %q must match [a-zA-Z_][a-zA-Z0-9_]*", i, action.Var)
-			}
-			entry.Method = action.Method
-			entry.Var = action.Var
-			if action.Var != "" {
-				varToMethod[action.Var] = action.Method
-			}
-
-		case extpb.ActionType_ACTION_TYPE_DENY:
-			if action.WithStatus != 0 {
-				if err := validateHTTPStatusCode(fmt.Sprintf("%d", action.WithStatus), fmt.Sprintf("actions[%d].with_status", i)); err != nil {
-					return nil, err
-				}
-			}
-			entry.WithStatus = int(action.WithStatus)
-			entry.WithHeaders = action.WithHeaders
-			entry.WithBody = action.WithBody
-
-		case extpb.ActionType_ACTION_TYPE_FAIL:
-			if strings.TrimSpace(action.LogMessage) == "" {
-				return nil, fmt.Errorf("actions[%d]: log_message must be specified for fail actions", i)
-			}
-			entry.LogMessage = action.LogMessage
-
-		case extpb.ActionType_ACTION_TYPE_ADD_HEADERS:
-			if action.HeadersToAdd == "" {
-				return nil, fmt.Errorf("actions[%d]: headers_to_add must be specified for add_headers actions", i)
-			}
-			if err := validateCELExpression(action.HeadersToAdd); err != nil {
-				return nil, fmt.Errorf("actions[%d].headers_to_add: %w", i, err)
-			}
-			entry.HeadersToAdd = action.HeadersToAdd
-
-		default:
+		validator, ok := actionEntryValidators[action.ActionType]
+		if !ok {
 			return nil, fmt.Errorf("actions[%d]: unknown action_type %s", i, action.ActionType)
+		}
+		if err := validator(action, i, &entry, &vctx); err != nil {
+			return nil, err
 		}
 
 		entries = append(entries, entry)
 	}
 
-	if len(varToMethod) > 0 {
-		if err := s.validateCrossActionVars(policyID, actions, varToMethod); err != nil {
+	if len(vctx.varToMethod) > 0 {
+		if err := s.validateCrossActionVars(policyID, actions, vctx.varToMethod); err != nil {
 			return nil, err
 		}
 	}
@@ -898,10 +925,12 @@ func (s *extensionService) validateCrossActionVars(policyID ResourceID, actions 
 				cacheKey := ProtoCacheKey{ClusterName: upstreamEntry.ClusterName, Service: upstreamEntry.Service}
 				fds, ok := s.registeredData.GetProtoDescriptor(cacheKey)
 				if !ok {
+					s.logger.V(1).Info("proto descriptor not cached, skipping field validation", "var", varName, "method", methodName)
 					continue
 				}
 				responseType := findResponseMessageType(fds, upstreamEntry.Service, upstreamEntry.Method)
 				if responseType == "" {
+					s.logger.V(1).Info("response message type not found, skipping field validation", "var", varName, "method", methodName, "service", upstreamEntry.Service)
 					continue
 				}
 				for _, chain := range chains {
@@ -924,147 +953,6 @@ func collectCELExpressions(action *extpb.ActionEntry) []string {
 		exprs = append(exprs, action.HeadersToAdd)
 	}
 	return exprs
-}
-
-func extractVarFieldAccesses(expr string, knownVars map[string]string) (map[string][][]string, error) {
-	prsr, err := parser.NewParser()
-	if err != nil {
-		return nil, fmt.Errorf("failed to create CEL parser: %w", err)
-	}
-	parsed, iss := prsr.Parse(common.NewTextSource(expr))
-	if len(iss.GetErrors()) > 0 {
-		return nil, fmt.Errorf("failed to parse CEL expression: %s", iss.ToDisplayString())
-	}
-
-	result := make(map[string][][]string)
-	collectFieldAccesses(parsed.Expr(), knownVars, result)
-	return result, nil
-}
-
-func collectFieldAccesses(expr ast.Expr, knownVars map[string]string, result map[string][][]string) {
-	if expr == nil {
-		return
-	}
-
-	if expr.Kind() == ast.SelectKind {
-		varName, fields := resolveSelectChain(expr)
-		if varName != "" {
-			if _, ok := knownVars[varName]; ok {
-				result[varName] = append(result[varName], fields)
-				return
-			}
-		}
-	}
-
-	switch expr.Kind() {
-	case ast.SelectKind:
-		collectFieldAccesses(expr.AsSelect().Operand(), knownVars, result)
-	case ast.CallKind:
-		call := expr.AsCall()
-		if t := call.Target(); t != nil {
-			collectFieldAccesses(t, knownVars, result)
-		}
-		for _, arg := range call.Args() {
-			collectFieldAccesses(arg, knownVars, result)
-		}
-	case ast.ListKind:
-		for _, elem := range expr.AsList().Elements() {
-			collectFieldAccesses(elem, knownVars, result)
-		}
-	case ast.MapKind:
-		for _, entry := range expr.AsMap().Entries() {
-			mapEntry := entry.AsMapEntry()
-			collectFieldAccesses(mapEntry.Key(), knownVars, result)
-			collectFieldAccesses(mapEntry.Value(), knownVars, result)
-		}
-	}
-}
-
-func resolveSelectChain(expr ast.Expr) (string, []string) {
-	var fields []string
-	current := expr
-	for current.Kind() == ast.SelectKind {
-		fields = append([]string{current.AsSelect().FieldName()}, fields...)
-		current = current.AsSelect().Operand()
-	}
-	if current.Kind() == ast.IdentKind {
-		return current.AsIdent(), fields
-	}
-	return "", nil
-}
-
-func findResponseMessageType(fds *descriptorpb.FileDescriptorSet, serviceName, methodName string) string {
-	for _, file := range fds.File {
-		for _, svc := range file.Service {
-			fullServiceName := svc.GetName()
-			if file.Package != nil && *file.Package != "" {
-				fullServiceName = *file.Package + "." + svc.GetName()
-			}
-			if fullServiceName != serviceName {
-				continue
-			}
-			for _, method := range svc.Method {
-				if method.GetName() == methodName {
-					return method.GetOutputType()
-				}
-			}
-		}
-	}
-	return ""
-}
-
-func findMessageDescriptor(fds *descriptorpb.FileDescriptorSet, messageType string) *descriptorpb.DescriptorProto {
-	normalizedType := strings.TrimPrefix(messageType, ".")
-	for _, file := range fds.File {
-		pkg := file.GetPackage()
-		for _, msg := range file.MessageType {
-			fqn := msg.GetName()
-			if pkg != "" {
-				fqn = pkg + "." + msg.GetName()
-			}
-			if fqn == normalizedType {
-				return msg
-			}
-		}
-	}
-	return nil
-}
-
-func validateFieldAccess(fds *descriptorpb.FileDescriptorSet, messageType string, fieldPath []string) error {
-	if len(fieldPath) == 0 {
-		return nil
-	}
-
-	msg := findMessageDescriptor(fds, messageType)
-	if msg == nil {
-		return fmt.Errorf("message type %q not found in proto descriptors", messageType)
-	}
-
-	currentMsg := msg
-	for i, fieldName := range fieldPath {
-		var found *descriptorpb.FieldDescriptorProto
-		for _, f := range currentMsg.Field {
-			if f.GetName() == fieldName {
-				found = f
-				break
-			}
-		}
-		if found == nil {
-			return fmt.Errorf("field %q not found on message %q", fieldName, currentMsg.GetName())
-		}
-
-		if i < len(fieldPath)-1 {
-			if found.GetType() != descriptorpb.FieldDescriptorProto_TYPE_MESSAGE {
-				return fmt.Errorf("field %q on message %q is not a message type, cannot access sub-fields", fieldName, currentMsg.GetName())
-			}
-			nextMsg := findMessageDescriptor(fds, found.GetTypeName())
-			if nextMsg == nil {
-				return fmt.Errorf("message type %q for field %q not found", found.GetTypeName(), fieldName)
-			}
-			currentMsg = nextMsg
-		}
-	}
-	return nil
 }
 
 // Creates a locator matching the definition in policy-machinery
--- a/internal/extension/manager.go
+++ b/internal/extension/manager.go
@@ -747,7 +747,7 @@ func validateCELExpression(expr string) error {
 	return nil
 }
 
-var varNameRegexp = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_-]*$`)
+var varNameRegexp = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
 
 func (s *extensionService) PipelineCommit(_ context.Context, request *extpb.PipelineCommitRequest) (*emptypb.Empty, error) {
 	if request == nil {
@@ -818,7 +818,7 @@ func (s *extensionService) validateActions(policyID ResourceID, actions []*extpb
 				return nil, fmt.Errorf("actions[%d]: method %q is not a registered action method for this policy", i, action.Method)
 			}
 			if action.Var != "" && !varNameRegexp.MatchString(action.Var) {
-				return nil, fmt.Errorf("actions[%d]: var %q must match [a-zA-Z_][a-zA-Z0-9_-]*", i, action.Var)
+				return nil, fmt.Errorf("actions[%d]: var %q must match [a-zA-Z_][a-zA-Z0-9_]*", i, action.Var)
 			}
 			entry.Method = action.Method
 			entry.Var = action.Var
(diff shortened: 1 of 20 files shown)


internal/extension/manager.go
```diff
@@ -794,103 +783,285 @@ func (s *extensionService) PipelineCommit(_ context.Context, request *extpb.Pipe
...
+				}
+				cacheKey := ProtoCacheKey{ClusterName: upstreamEntry.ClusterName, Service: upstreamEntry.Service}
+				fds, ok := s.registeredData.GetProtoDescriptor(cacheKey)
+				if !ok {
+					continue
+				}
+				responseType := findResponseMessageType(fds, upstreamEntry.Service, upstreamEntry.Method)
+				if responseType == "" {
+					continue
+				}
```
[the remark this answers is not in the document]
philbrookes: The continue here is intentional. Proto descriptors are fetched asynchronously from the gRPC service after registration, and may not be available at the time the first PipelineCommit arrives. Turning these into hard errors would cause the initial reconciliation to fail every time a new action method is registered, forcing unnecessary error/retry cycles. 

The missing-descriptor case is benign: if the descriptor is not available yet, the cross-action variable field validation is simply skipped for that commit. The server-side PipelineCommit handler already validates the structural correctness of the pipeline (phase ordering, action types, required fields). The field-level var validation here is an additional best-effort check that runs when descriptor metadata is present.
On the next reconciliation (triggered by the descriptor becoming available), the full validation will run.

pkg/extension/controller/controller.go
```diff
@@ -335,87 +336,152 @@ func (ec *ExtensionController) RegisterActionMethod(ctx context.Context, policy
...
+		client:        ec.extensionClient.client,
+		populatedVars: make(map[string]bool),
 	}
 }
 
-// PipelineImpl implements Pipeline by accumulating actions locally and sending
-// them to the operator in a single PipelineCommit gRPC call.
+type pipelineEntry struct {
+	action exttypes.Action
+	phase  string // "request" or "response"
```
crstrn13: nit: maybe extract this into an enum-like

```go
type pipelinePhase string

const (
      phaseRequest  pipelinePhase = "request"
      phaseResponse pipelinePhase = "response"
  )
```

pkg/extension/controller/controller.go
```diff
@@ -335,87 +336,152 @@ func (ec *ExtensionController) RegisterActionMethod(ctx context.Context, policy
...
 		}
+	case exttypes.AddHeadersAction:
+		if a.Predicate != "" {
+			exprs = append(exprs, a.Predicate)
+		}
+		if a.HeadersToAdd != "" {
+			exprs = append(exprs, a.HeadersToAdd)
+		}
+	}
+	return exprs
```
crstrn13: this could benefit from the use of an interface with implementation on each type of action, rather than the switch case.

pkg/extension/controller/controller.go
```diff
@@ -335,87 +336,152 @@ func (ec *ExtensionController) RegisterActionMethod(ctx context.Context, policy
...
+		entry.ActionType = extpb.ActionType_ACTION_TYPE_FAIL
+		entry.Predicate = a.Predicate
+		entry.LogMessage = a.LogMessage
+	case exttypes.AddHeadersAction:
+		entry.ActionType = extpb.ActionType_ACTION_TYPE_ADD_HEADERS
+		entry.Predicate = a.Predicate
+		entry.HeadersToAdd = a.HeadersToAdd
 	default:
-		panic(fmt.Sprintf("unknown response action type: %T", action))
+		panic(fmt.Sprintf("unknown action type: %T", pe.action))
```
crstrn13: hmm, I might have missed this panic the last time. But is panicking here, the best course of action? I don't think we recover from this anywhere and I'd rather surface the error here than taking down the whole extension? what do you reckon?

pkg/extension/controller/controller.go
```diff
@@ -335,87 +336,152 @@ func (ec *ExtensionController) RegisterActionMethod(ctx context.Context, policy
...
+		entry.WithHeaders = a.WithHeaders
+		entry.WithBody = a.WithBody
+	case exttypes.FailAction:
+		entry.ActionType = extpb.ActionType_ACTION_TYPE_FAIL
+		entry.Predicate = a.Predicate
+		entry.LogMessage = a.LogMessage
+	case exttypes.AddHeadersAction:
+		entry.ActionType = extpb.ActionType_ACTION_TYPE_ADD_HEADERS
+		entry.Predicate = a.Predicate
+		entry.HeadersToAdd = a.HeadersToAdd
```
crstrn13: this could also benefit from an interface, just like the other switch statement.

internal/extension/manager.go
```diff
@@ -794,103 +783,288 @@ func (s *extensionService) PipelineCommit(_ context.Context, request *extpb.Pipe
...
+			entry.LogMessage = action.LogMessage
+
+		case extpb.ActionType_ACTION_TYPE_ADD_HEADERS:
+			if action.HeadersToAdd == "" {
+				return nil, fmt.Errorf("actions[%d]: headers_to_add must be specified for add_headers actions", i)
+			}
+			if err := validateCELExpression(action.HeadersToAdd); err != nil {
+				return nil, fmt.Errorf("actions[%d].headers_to_add: %w", i, err)
+			}
+			entry.HeadersToAdd = action.HeadersToAdd
```
crstrn13: nit: this might be a bit more difficult to refactor and draw an interface for it. but might be worth trying? Not sure if the action types will increase in the future, but it will be a bit more extensible with the interface.

internal/extension/manager.go
```diff
@@ -794,103 +783,288 @@ func (s *extensionService) PipelineCommit(_ context.Context, request *extpb.Pipe
...
+				}
+				cacheKey := ProtoCacheKey{ClusterName: upstreamEntry.ClusterName, Service: upstreamEntry.Service}
+				fds, ok := s.registeredData.GetProtoDescriptor(cacheKey)
+				if !ok {
+					continue
+				}
+				responseType := findResponseMessageType(fds, upstreamEntry.Service, upstreamEntry.Method)
+				if responseType == "" {
+					continue
+				}
```
crstrn13, line 935: maybe I am missing something, but should't we return an error here? I think the first continue is fine since we can get an action without an upstream and thus doesn't make sense to do the other checks, but if we don't get the descriptor from the cache or the response type, shouldn't be errors here?

internal/extension/manager.go
crstrn13, line 1: I'd suggest moving `extractVarFieldAccesses`, `collectFieldAccesses`, `resolveSelectChain` maybe to a different file or package as they seem like pure AST utilities. The file is quite large already, I think it'd help with readability since they are not tightly bound to this file.

Same for the proto desciptor helpers: `findResponseMessageType`, `findMessageDescriptor`, `validateFieldAccess`

internal/extension/registry.go
```diff
@@ -717,8 +733,15 @@ func (r *RegisteredDataStore) ClearPipelinePhase(policy ResourceID, phase Pipeli
 }
 
 // ReplacePipelineActions atomically clears all pipeline actions for a policy
-// and replaces them with the provided request and response entries.
-func (r *RegisteredDataStore) ReplacePipelineActions(policy ResourceID, requestEntries, responseEntries []PipelineActionEntry) {
+// and replaces them with the provided entries, grouped by their Phase field.
+// Returns an error if any entry has an invalid phase.
+func (r *RegisteredDataStore) ReplacePipelineActions(policy ResourceID, entries []PipelineActionEntry) error {
```
crstrn13, line 738: nit: the name of this method is a bit __confusing__ (at least to me). I think maybe `ApplyPipelineActions` would convey the intention better?

internal/extension/registry.go
```diff
@@ -1012,135 +1034,175 @@ func (m *RegisteredDataMutator[TResource]) mutateWasmConfig(wasmConfig *wasm.Con
...
+			for varName, methodName := range varToMethod {
+				if actionReferencesVar(e, varName) {
+					grpcOnReply[methodName] = append(grpcOnReply[methodName], ta)
+					break
+				}
 			}
+		}
+	}
+	classifyAndConvert(requestEntries, "request")
+	classifyAndConvert(responseEntries, "response")
```
crstrn13: we could use the constants instead

````go
const (
	PipelinePhaseRequest  PipelinePhase = "request"
	PipelinePhaseResponse PipelinePhase = "response"
)

File: internal/extension/registry.go [hunk 10]
```diff
@@ -1012,135 +1034,175 @@ func (m *RegisteredDataMutator[TResource]) mutateWasmConfig(wasmConfig *wasm.Con
...
-
-			ta.OnReply = append(ta.OnReply, onReplyActions...)
-			typedActions = append(typedActions, ta)
+			if !isVarDependent {
+				result = append(result, entryToTypedAction(e, sources, phase))
+			}
 		}
+	}
+	emit(requestEntries, "request")
+	emit(responseEntries, "response")
```
crstrn13: same thing with the constants.

pkg/extension/controller/controller.go
```diff
@@ -335,87 +336,152 @@ func (ec *ExtensionController) RegisterActionMethod(ctx context.Context, policy
...
-			Method:     a.Method,
-			Var:        a.Var,
+	for _, action := range actions {
+		exprs := celExpressionsFrom(action)
+		for _, expr := range exprs {
+			for varName := range batchVars {
+				if !localPopulated[varName] && celReferencesVar(expr, varName) {
+					return fmt.Errorf("action references variable %q before it is populated", varName)
+				}
+			}
```
crstrn13, line 406: I noticed that regex expression could be compiled outside this nested loop. It has a low performance impact, but make sense to compile the expression before and only executing it here.

```go
  // Build once
  patterns := make(map[string]*regexp.Regexp)
  for varName := range batchVars {
      patterns[varName] = regexp.MustCompile(`\b` + regexp.QuoteMeta(varName) + `\b`)
  }

  // Reuse many times
  for _, action := range actions {
      for _, expr := range celExpressionsFrom(action) {
          for varName := range batchVars {
              if patterns[varName].MatchString(expr) {  // fast — already compiled
                  ...
              }
          }
      }
  }

File: internal/extension/registry.go [hunk 12]
```diff
@@ -1012,135 +1034,175 @@ func (m *RegisteredDataMutator[TResource]) mutateWasmConfig(wasmConfig *wasm.Con
...
-				SourcePolicyLocators: sources,
-			}
-			if upstream, ok := upstreamByMethod[policyID][entry.Method]; ok && upstream.MessageTemplate != "" {
-				ta.MessageBuilder = upstream.MessageTemplate
+			ta := entryToTypedAction(e, sources, phase)
+			for varName, methodName := range varToMethod {
+				if actionReferencesVar(e, varName) {
+					grpcOnReply[methodName] = append(grpcOnReply[methodName], ta)
+					break
+				}
```
crstrn13, line 1101: same here, maybe we can create a map with the patterns after we get the varToMethod and only match against the expression here, similar to what I suggested in the controller. But the impact is small in both cases

internal/extension/manager.go
philbrookes: When a proto descriptor isn't cached or the response type can't be resolved, we skip field-level validation rather than erroring and rejecting the pipeline. The action itself is still valid, we just can't statically verify field accesses. I'll add a warning log here to make this more visible.

internal/extension/registry.go
philbrookes: This method clears and replaces all pipeline actions for a policy, so "replace" is more accurate than "apply" which could be read as merge/append.

internal/extension/registry.go
crstrn13, line 1: let's move this if statement from line 987 out of the for loop. The thing we want to prevent is ever having this map marshall to null in the wasm-shim as it will break the whole config.
```go
if wasmConfig.Services == nil {
		wasmConfig.Services = make(map[string]wasm.Service)
	}

crstrn13 (changes requested):

crstrn13 (approved):
ship it :shipit:

PR #2019, opened 2026-05-29
[Release] Kuadrant Operator v1.5.0-rc1

kuadrant-dev:
The following PR for the release candidate of Kuadrant Operator version 1.5.0-rc1 includes:
- Authorino Operator version 0.25.0
- DNS Operator version 0.17.0
- Limitador Operator version 0.18.1
- Console Plugin version 0.4.0
- WASM Shim version 0.13.0
- Developer Portal Controller version 0.2.0

Auto-generated by [create-pull-request][1]

[1]: https://github.com/peter-evans/create-pull-request


thomasmaas (approved):

.github/workflows/test.yaml
```diff
@@ -361,29 +361,12 @@ jobs:
...
-        run: |
-          make verify-manifests
-
   required-checks:
     name: Test Required Checks
     # This check adds a list of checks to one job to simplify adding settings to the repo.
     # If a new check is added in this file, and it should be retested on entry to the merge queue,
     # it needs to be added to the list below aka needs: [ existing check 1, existing check 2, new check ].
-    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-fmt, test-scripts, verify-generate, verify-go-mod, verify-manifests ]
+    needs: [ unit-tests, controllers-integration-tests, bare-k8s-integration-tests, gatewayapi-integration-tests, gatewayapi-provider-integration-tests, verify-fmt, test-scripts, verify-generate, verify-go-mod ]
```
mikenairn, line 369: Can you explain why this needs to be removed?

I have admittedly not done anything with kuadrant releases in a very long time, but i cannot think of any good reason why we wouldn't require this to work consistently across all branches and tags including, if not especially, releases. This suggest we have no way to re-create a released version of kuadrant locally if it's because of component versions not being persisted to the branch.

Boomatang: It reverted all the version to 0.0.0 and didn't use version set in the release.yaml. The why it was doing that I can not answer right now. So the check is still on main, only reverted on this branch.

PR #2024, opened 2026-06-01
fix: reject top-level fail actions in pipeline validation

### LINKED ISSUE #2021: Reject top-level Fail actions in Extensions
## Problem

The wasm-shim does not support top-level `Fail` operations. When an extension policy includes a `fail` action that doesn't reference a gRPC response variable, it remains top-level. At runtime, the wasm-shim silently skips it:

```
error envoy wasm: Top-level Fail operation is currently unsupported. Action 0: blocked by fail action
```

The request passes through with 200 — the fail action is silently ignored. The policy still shows `Accepted: True` and `Enforced: True`, giving no indication that the action won't work.

## Expected behavior

The extension SDK should reject fail actions that would end up as top-level during pipeline validation (`validateAndAppend`), surfacing `Accepted: False` on the policy status with a clear error message. This protects all extensions using the SDK, not just PipelinePolicy.

crstrn13:
## Summary
- Reject `FailAction`s that don't reference any gRPC response variable in `validateAndAppend()`, since they remain top-level and are silently ignored by the wasm-shim at runtime
- Include previously populated vars (`p.populatedVars`) in `varPatterns` so fail actions referencing gRPC vars from earlier pipeline calls are correctly accepted
- Add comprehensive test cases covering standalone fail, fail with/without gRPC var reference, cross-call references, and no-predicate fail

Closes #2021

## Test plan
- [x] All existing pipeline tests pass
- [x] New `TestPipeline_TopLevelFailAction` subtests cover all edge cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for fail actions: must reference required gRPC response variables and cannot be configured as standalone, top-level, or response-phase actions.
  * Improved error messages clarifying invalid fail-action configurations.
  * Expanded variable-tracking to include variables introduced in the same batch for more accurate validation.

* **Tests**
  * Added and updated tests covering fail-action validation rules and predicate behaviour.

### DIFF
--- a/pkg/extension/controller/controller.go
+++ b/pkg/extension/controller/controller.go
@@ -386,9 +386,14 @@ func (p *PipelineImpl) validateAndAppend(phase string, actions []exttypes.Action
 		}
 	}
 
-	varPatterns := make(map[string]*regexp.Regexp, len(batchVars))
-	for varName := range batchVars {
+	varPatterns := make(map[string]*regexp.Regexp, len(batchVars)+len(p.populatedVars))
+	for varName := range p.populatedVars {
 		varPatterns[varName] = regexp.MustCompile(`\b` + regexp.QuoteMeta(varName) + `\b`)
+	}
+	for varName := range batchVars {
+		if _, exists := varPatterns[varName]; !exists {
+			varPatterns[varName] = regexp.MustCompile(`\b` + regexp.QuoteMeta(varName) + `\b`)
+		}
 	}
 
 	localPopulated := make(map[string]bool, len(p.populatedVars))
--- a/pkg/extension/controller/controller.go
+++ b/pkg/extension/controller/controller.go
@@ -397,6 +397,20 @@ func (p *PipelineImpl) validateAndAppend(phase string, actions []exttypes.Action
 	}
 
 	for _, action := range actions {
+		if _, ok := action.(exttypes.FailAction); ok {
+			refsVar := false
+			for _, expr := range action.CelExpressions() {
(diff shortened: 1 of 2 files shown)


pkg/extension/controller/controller.go
```diff
@@ -386,17 +386,36 @@ func (p *PipelineImpl) validateAndAppend(phase string, actions []exttypes.Action
...
 		localPopulated[k] = true
 	}
 
 	for _, action := range actions {
+		if _, ok := action.(exttypes.FailAction); ok {
+			refsVar := false
+			for _, expr := range action.CelExpressions() {
+				for _, pattern := range varPatterns {
+					if pattern.MatchString(expr) {
+						refsVar = true
```
philbrookes, line 410: Could you break here?

pkg/extension/controller/controller.go
```diff
@@ -386,17 +386,36 @@ func (p *PipelineImpl) validateAndAppend(phase string, actions []exttypes.Action
...
+					}
+				}
```
philbrookes, line 413: add here:
```suggestion
				}
				if refsVar {
					break
				}
```

philbrookes (changes requested):
looks good, just 2 small suggestions.

pkg/extension/controller/controller.go
crstrn13: correct, didn't notice that. Thanks for the suggestion :+1:

philbrookes (approved):

Issue #1085, opened 2024-12-18
Investigate options to speed up reconciliation up for large amount of GWs and Listeners and policies


trepel:
## Overview

Recently a scale test has been implemented:
https://github.com/Kuadrant/testsuite/tree/main/scale_test

I did run a few scale test runs. It took quite some time for Kuadrant operator to reconcile all the policies. Many AuthPolicies and RLPs did not get any status for quite some time. After they got some status it often was:
`'AuthPolicy waiting for the following components to sync: [AuthConfig (0cbc22a687a9ff2a57c54007e8ad9b6bc17de3744144196b9b8286fb1593f495)]'`
and
`'RateLimitPolicy waiting for the following components to sync: [Limitador]'`

Everything got reconciled successfully and policies got enforced eventually, but it took quite some time:
1GW 16 Listeners -> 16s to get status for all policies
1 GW 32 Listeners -> 120s to get status for all policies
1 GW 48 Listeners -> 7 min to get status for all policies
1 GW 63 Listeners -> 30 min to get status for all policies

In the operator log there were a lot of entries complaining about invalid paths. So I made the HTTPRoutes target specific Listener rather than the whole Gateway. This make the results much nicer:

1 GW 32 Listeners -> 18s to get status for all policies
1 GW 48 Listeners -> 60s to get status for all policies
1 GW 63 Listeners -> 120s to get status for all policies

I tried with 2 GWs as well:
2 GW 16 Listeners -> 18s to get status for all policies
2 GW 32 Listeners -> 76s to get status for all policies

However, this was still too much:
2 GW 63 Listeners -> 16 min to get status for all policies

## Initial Investigation

Be aware that certificate generation and DNS record creation might affect the results. It takes some for certificates to get created (the scale test uses self signed cluster issuer) and it also takes time for cloud provider to issue that many DNS records.

It seems reasonable that wasm config (only one per GW) is a contention point that said it should eventually get there. 
There are repeated log entries of “failed to update the object has been modified; please apply your changes to the latest version” in Kuadrant operator pod.

Also entries like "failed to create SOMETHING, SOMETHING already exists" appeared in Kuadrant operator pod log (SOMETHING being Certificate/AuthConfig typically) - not sure if this indicates some issue or not.

## Questions / Investigation required

Does it make sense that having many invalid paths is so expensive?
What can be done to improve on that 16 minutes? 2 Gateways and 63 Listeners on each are not super high numbers.

## Steps to reproduce

[...]

PR #2031, opened 2026-06-03
feat: add benchmark and performance tests for data plane policy reconciliation

mikenairn:
## Summary

Adds two types of performance tests for measuring and preventing regressions in the kuadrant-operator reconciliation paths. This PR is shared for discussion — not intended to merge as-is.

### 1. Unit Benchmarks (`internal/controller/reconciliation_bench_test.go`)

Go benchmark tests that construct realistic topologies in-memory and measure specific reconciliation code paths at varying scale (routes, listeners, gateways). No cluster needed.

Benchmarks cover:
- `CalculateEffectiveAuthPolicies` / `EffectiveRateLimitPolicies` / `EffectiveTokenRateLimitPolicies`
- Listener fanout scaling (1–64 listeners)
- Multi-gateway scaling
- `GetKuadrantFromTopology` lookup cost
- `MultiReconcilerCycle` — all three policy reconcilers together
- `FullReconciliationCycle` — topology build + lookups + effective policies

Also includes `TestEqualToNonDeterminism` in `internal/wasm/types_bench_test.go` which reproduces the WasmPlugin comparison bug (#1934).

Supports pprof profiling via `-cpuprofile`/`-memprofile` and `benchstat` for before/after comparison.

### 2. Cluster Performance Tests (`tests/performance/`)

Ginkgo tests that create real HTTPRoutes and AuthPolicies on a Kubernetes cluster and measure wall-clock reconciliation time. The operator runs in-process, so Go's built-in profiling captures the reconciliation code paths.

- Configurable scale via `PERF_SCALE_LEVELS` env var (default: 10, 50, 100, 300)
- Operator logs silenced by default (`PERF_LOG_ERRORS=true` to enable)
- Supports `-cpuprofile`/`-memprofile` for in-process profiling

## Context

These tests were developed during the investigation of reconciliation performance at scale ([#1919](https://github.com/Kuadrant/kuadrant-operator/issues/1919), [#1085](https://github.com/Kuadrant/kuadrant-operator/issues/1085)). Key findings:

- pprof profiling of the unit benchmarks identified `LocatorFromObject` (47% CPU) and `Children()` (23% CPU) in policy-machinery as the dominant costs in the `Paths()` DFS traversal
- Cluster performance tests showed that single-cycle benchmarks don't capture the compounding effect of repeated reconciliation cycles during bulk resource creation — this is why PR #1957 shows dramatic real-cluster improvement despite minimal unit benchmark impact
- At 3,000 routes, the bottleneck shifts from kuadrant-operator to Authorino throughput (~10 AuthConfigs/second)

See `tests/performance/README.md` for full usage instructions.

## Discussion points

- Should unit benchmarks run in CI? If so, with what thresholds or `benchstat` comparison?
- Should the cluster performance tests be part of the existing integration test infrastructure or remain separate?
- Should we parameterise the cluster tests to also scale listeners and gateways, not just routes?
- Future: capture pprof profiles from all cluster components (Authorino, limitador-operator, etc.) via their `--pprof-bind-address` endpoints during the test

### DIFF
--- a/internal/controller/reconciliation_bench_test.go
+++ b/internal/controller/reconciliation_bench_test.go
@@ -1,0 +1,483 @@
+//go:build unit
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/kuadrant/policy-machinery/controller"
+	"github.com/kuadrant/policy-machinery/machinery"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
+	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
+	gatewayapiv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+
+	kuadrantv1 "github.com/kuadrant/kuadrant-operator/api/v1"
+	kuadrantv1alpha1 "github.com/kuadrant/kuadrant-operator/api/v1alpha1"
+	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
+)
+
+type topologyParams struct {
+	numGateways          int
+	listenersPerGateway  int
+	numRoutes            int
+}
+
+func (p topologyParams) label() string {
+	return fmt.Sprintf("gw=%d_listen=%d_routes=%d", p.numGateways, p.listenersPerGateway, p.numRoutes)
+}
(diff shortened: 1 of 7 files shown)


maksymvavilov:
Sharing what Claude "done" so far on my behalf for pref testing https://gist.github.com/maksymvavilov/d304d224f4d8637da914d7d90525f660 
This is for context more than enything else. 
TL;DR - almost everything alligns with this work

maksymvavilov:
I've combined the unit-test part of this PR into #2035

PR #2032, opened 2026-06-03
Fixing OIDCPolicy for MCP Gateway use case

### LINKED ISSUE #2017: OIDCPolicy: three bugs in redirect URI, cookie parsing, and query string handling
## Summary

I have been experimenting with OIDCPolicy on MCP-Gateway. I hit the following issues that I worked around by constructing my own AuthPolices, but I wanted to report and contribute fixes if they make sense.

**Note:**

I don't know the code around this area particularly well. So please verify these are real solutions as AI helped me a lot with this

Three related bugs in `oidcpolicy_reconciler.go` break the OIDC flow in non-trivial setups.

---

## Bug 1: Port dropped from redirect URI

**Problem:** When `spec.provider.redirectURI` is not set, OIDCPolicy auto-constructs the redirect URL from the Gateway listener's `Hostname` and `Protocol` — but ignores `Port`. The redirect becomes `http://example.com/auth/callback` instead of `http://example.com:8001/auth/callback`.

**Root cause:** `ingressGatewayInfo` has no `Port` field. The CEL expression only extracts `protocol` and `hostname`. `GetURL()` sets `Host: g.Hostname` without port.

Setting `spec.provider.redirectURI` fixes the redirect URL itself but does **not** fix everywhere the reconciler uses `igw.Hostname` — specifically the cookie domain and the OPA base URL still omit the port.

Additionally, the listener port isn't necessarily the external port — a load balancer or NodePort could expose a different one.

**Fix:** Add an optional `spec.provider.externalURL` field that provides the authoritative base URL (scheme + host + port). The reconciler derives cookie domain and encoded redirect URI from this single field. When omitted, fall back to the current auto-construction from the Gateway listener.

**Workaround:** None that fully works. `spec.provider.redirectURI` only fixes the redirect, not cookie domain or OPA base URL.

---

## Bug 2: Rego cookie parser breaks on `=` in values

**Problem:** The OPA cookie parser uses `split("=")` + `count(kv) == 2`. Base64-encoded cookie values (like JWTs) contain `=`, producing >2 elements — the cookie is silently dropped.

**Root cause:** `split("=")` on `jwt=eyJhbG...Q==` yields 4 elements → guard fails.

**Fix:** Use `indexof`/`substring` to split on the first `=` only.

---

## Bug 3: Target cookie drops query string

**Problem:** After OIDC auth, the user is redirected to the original path but without query parameters. `?elicitation_id=...` is lost.

**Root cause:** The `setCookie` expression only stores `request.path`, not `request.query`.

**Fix:** Include `request.query` with a guard for the empty case (CEL ternary).

---

## How they interact

Bug 3 stores a path missing query params → Bug 2 may fail to read the stored value if it contains `=` → Bug 1 means the redirect goes to the wrong host:port regardless.

**File:** `cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go`

didierofrivia:
Closes https://github.com/Kuadrant/kuadrant-operator/issues/2017

### Verification Steps

0. Dev cluster setup with some useful envs
```sh
export GATEWAY_NS=test-namespace      # Namespace for the Gateway
export GATEWAY_NAME=test-gateway # Name for the Gateway
kubectl create namespace $GATEWAY_NS
make local-setup
```

1. Deploy Kuadrant Instance

```sh
kubectl apply -f - <<EOF
kind: Kuadrant
metadata:
  name: kuadrant
  namespace: kuadrant-system
EOF
```

2. Install OIDCPolicy and required Gateway and HTTPRoute for verification

```sh
kubectl apply -f - <<EOF
kind: Gateway
metadata:
  name: ${GATEWAY_NAME}
  namespace: ${GATEWAY_NS}
spec:
  gatewayClassName: istio
  listeners:
  - name: http
    hostname: "example.com"
    port: 8001  # Non-standard port
    protocol: HTTP
---
kind: HTTPRoute
metadata:
  name: test-route
  namespace: ${GATEWAY_NS}
spec:
  parentRefs:
  - name: ${GATEWAY_NAME}
  hostnames:
  - "example.com"
  rules:
  - matches:
    - path:
        type: PathPrefix
        value: /
    backendRefs:
    - name: test-app
      port: 8080
---
kind: OIDCPolicy
metadata:
  name: test-oidc
  namespace: ${GATEWAY_NS}
spec:
  targetRef:
    kind: HTTPRoute
    name: test-route
  provider:
    issuerURL: "https://your-oidc-provider.com"
    clientID: "your-client-id" 
EOF
```

#### Bug 1 Verification:

1. Wait for the OIDCPolicy to be enforced:
  ```sh
  kubectl get -n $GATEWAY_NS oidcpolicy test-oidc -o jsonpath='{.status.conditions[?(@.type=="Enforced")].status}'
```
Should return: True
  
2. Check the generated callback AuthPolicy:
Look for the redirectURI in the token request body CEL expression - it should include the port, In the metadata.token.http.body expression

```sh
  kubectl get -n $GATEWAY_NS authpolicy test-oidc-callback -o yaml | yq '.spec.overrides.rules.metadata.token.http.body'
```
You should see: `"&redirect_uri=http%3A%2F%2Fexample.com%3A8001%2Fauth%2Fcallback"`
Note the %3A8001 (URL-encoded :8001)

4. Check the authorize URL in the main AuthPolicy:
Look in .spec.overrides.authScheme.response.unauthenticated.headers.location.value 

```sh
  kubectl get -n $GATEWAY_NS authpolicy test-oidc -o yaml | yq '.spec.overrides.rules.response.unauthenticated.headers.location'
```
the redirect_uri parameter should include :8001
  
5. Test the cookie domain (should NOT include port):
```sh
  kubectl get -n $GATEWAY_NS authpolicy test-oidc -o yaml | yq '.spec.overrides.rules.response.unauthenticated.headers.set-cookie'
```
Should see: domain=example.com (not example.com:8001)

#### Bug 2 Verification:

1. Check the OPA authorization rule in the callback AuthPolicy:
```sh
kubectl get -n $GATEWAY_NS authpolicy test-oidc-callback -o yaml | yq '.spec.overrides.rules.authorization.location.opa.rego' | grep -o "indexof(trimmed"
```
Should return: indexof(trimmed (confirming the fix is present)
  
2. Verify it does NOT use the old broken pattern:
```sh
kubectl get -n $GATEWAY_NS  authpolicy test-oidc-callback -o yaml | yq '.spec.overrides.rules.authorization.location.opa.rego' | grep "count(kv) == 2"
```
Should return nothing (empty)

3. Check the cookie parsing logic:
```sh
kubectl get -n $GATEWAY_NS authpolicy test-oidc-callback -o yaml | yq '.spec.overrides.rules.authorization.location.opa.rego'
```
Should contain:
```yaml
  eq_idx := indexof(trimmed, "=")
  eq_idx != -1
  name := trim(substring(trimmed, 0, eq_idx), " ")
  value := trim(substring(trimmed, eq_idx + 1, -1), " ")
```

#### Bug 3 Verification:

1. Check the target cookie expression in the main AuthPolicy:
```sh
kubectl get -n $GATEWAY_NS authpolicy test-oidc -o yaml | yq '.spec.overrides.rules.response.unauthenticated.headers.set-cookie.expression'
```
Should contain:
```
request.path + (request.query != "" ? "?" + request.query : "")
```

2. Verify it's NOT using the old broken pattern:
```sh
kubectl get -n $GATEWAY_NS authpolicy test-oidc -o yaml | yq '.spec.overrides.authScheme.response.unauthenticated.headers.set-cookie.expression' | grep 'request.path + "; domain='
```
 Should return nothing (empty)

## Summary by CodeRabbit

* **New Features**
  * Redirect handling now preserves non‑standard gateway ports and derives a clean base URL for post‑auth redirects.
  * Gateway listener resolution now carries an explicit port, improving how gateway URLs are formatted and reused.
  * Cookie target values include the request query when present, with protocol-specific cookie attributes (including Secure for HTTPS).
  * Updated authorisation rule generation to build correct redirect locations using gateway/base/authorise URLs.
* **Tests**
  * Expanded unit coverage for URL/port handling, cookie/CEL expression construction, and generated authorisation rules.


maleck13:
👀

crstrn13 (approved):
LGTM! 🎉 :shipit:

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -282,6 +290,11 @@ func (r *OIDCPolicyReconciler) reconcileHTTPRoute(ctx context.Context, desired *
 	return err
 }
 
+func buildTargetCookieExpression(hostname string, protocol gatewayapiv1.ProtocolType) string {
+	return fmt.Sprintf(`
+"target=" + request.path + (has(request.query) && request.query != "" ? "?" + request.query : "") + "; domain=%s; HttpOnly; %s SameSite=Lax; Path=/; Max-Age=3600"`, hostname, getSecureFlag(protocol))
+}
```
crstrn13, line 311: The issue is that request.path in Authorino's CEL context already includes the query string (it's the full request URI, e.g. /get?foo=bar&baz=qux). Appending "?" + request.query on top of that duplicates the query parameters: e.g target=/get?foo=bar&baz=qux?foo=bar&baz=qux

crstrn13 (changes requested):

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
didierofrivia: mhmm it's weird why they had the issue in the first place, I'll replace it with [url_path](https://github.com/Kuadrant/authorino/blob/main/pkg/service/well_known_attributes.go#L65)

didierofrivia: for some reason, it was dropping the query when used with just `request.path` https://github.com/Kuadrant/kuadrant-operator/issues/2017 Bug 3

crstrn13 (approved):

PR #2035, opened 2026-06-08
feat: add Go benchmarks for reconciliation hot paths

maksymvavilov:
## Summary

Go benchmarks for reconciliation hot paths integrated into the existing `code-quality.yaml` workflow. 8 benchmark functions, 10 sub-benchmarks, ~1 min runtime locally, runs in parallel with lint and unit tests in CI.

## Benchmarks

All benchmarks use the `bench` build tag (`//go:build bench`), fully decoupled from unit tests.

| Benchmark | What it measures |
|-----------|-----------------|
| `CalculateEffectiveAuthPolicies` | O(GC×RouteRules) Paths() loop (300 routes, 3gw×100) |
| `TopologyToDot` | DAG-to-DOT serialization (1000 objects) |
| `BuildDesiredAuthConfig` | AuthConfig generation (300 rules) |
| `BuildLimitadorLimits` | Merging effective policies into Limitador limits (300 policies) |
| `TopologyPaths` | Isolated Paths() DFS traversal (300 routes) |
| `ListenerFanout` | Listener scaling impact on Paths() (32 listeners) |
| `TopologyBuild` | Topology construction cost (300 routes, 300×32L) |
| `FullReconciliationCycle` | Topology build + lookups + all 3 reconcilers (300 routes) |

## Running

```bash
# Full suite (10 repetitions for benchstat)
make test-bench

# Single benchmark
go test ./internal/controller/... -bench=BenchmarkTopologyPaths -benchmem -run='^$' -count=1 -tags bench
```

## CI Integration

Benchmark job added to `code-quality.yaml` (no separate workflow file) — runs in parallel with lint, unit tests, and verify on PRs:
- Compares PR vs base branch (`-count=10`, `-benchtime=200ms`)
- Only posts a PR comment when a regression is detected (silent on success)
- Fails the job when any timing regression >= 5% is detected
- 20-minute timeout
- Only runs on `pull_request` events (skipped for push/schedule)

## Credits

Scaling benchmarks (listener fanout, multi-gateway, RLP, topology build, full cycle) adapted from Mike Nairn's [PR #2031](https://github.com/Kuadrant/kuadrant-operator/pull/2031).


.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,97 @@
...
+              echo "$BENCHSTAT_OUTPUT"
+              echo '```'
+              echo '</details>'
+            } > benchmark-report.md
+            echo 'regression=true' >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Post PR comment
+        if: steps.benchstat.outputs.regression == 'true' && github.event.pull_request.head.repo.full_name == github.repository
+        uses: marocchino/sticky-pull-request-comment@f3bbb8300013c686241004a1c1b5d4045b10bc30 # ratchet:marocchino/sticky-pull-request-comment@v2
```
guicassolato: Any particular reason not to use v3.0.x? I believe [v3.0.5](https://github.com/marocchino/sticky-pull-request-comment/releases/tag/v3.0.5) is the latest as we speak.

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+func buildBenchTopology(b testing.TB, params benchTopologyParams) (*machinery.Topology, *kuadrantv1beta1.Kuadrant, *sync.Map) {
+	b.Helper()
+
+	if params.listenersPerGW == 0 {
+		params.listenersPerGW = 1
+	}
+	if params.policyTarget == "" {
+		params.policyTarget = "gateway"
+	}
+
```
guicassolato, line 59: In case you want to include a validation for `numRoutesPerGW`, it has to be greater or equal to 1, otherwise the policies won't reconcile to `Enforced`.

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+			TypeMeta: metav1.TypeMeta{
+				Kind:       machinery.GatewayClassGroupKind.Kind,
+				APIVersion: gatewayapiv1.GroupVersion.String(),
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: gcName,
+				UID:  types.UID(fmt.Sprintf("uid-gc-%d", gc)),
+			},
+			Spec: gatewayapiv1.GatewayClassSpec{
+				ControllerName: "kuadrant.io/policy-controller",
```
guicassolato: The gateway controller name matters for some reconciliation tasks, such as the status updaters and envoy cluster and envoy extension reconcilers. I believe the benchmark tests are not testing any of those tasks at the moment, but consider using a supported gateway controller name in the GatewayClass or adding a comment explaining why benchmarking some funcs won't reflect reality while this gateway controller name is used.

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+				Kind:       machinery.GatewayGroupKind.Kind,
+				APIVersion: gatewayapiv1.GroupVersion.String(),
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      gwName,
+				Namespace: "default",
+				UID:       types.UID(fmt.Sprintf("uid-gw-%d", gc)),
+			},
+			Spec: gatewayapiv1.GatewaySpec{
+				GatewayClassName: gatewayapiv1.ObjectName(gcName),
```
guicassolato: I was a bit surprise by the 1:1 relationship between GatewayClass and Gateway. In real deployments, a GatewayClass represents a controller implementation (Istio, Envoy Gateway), so you typically have 1-2 GatewayClasses with many Gateways under each. The 1:1 mapping conflates the two dimensions and could mask performance characteristics that emerge from the actual 1:N fan-out (e.g., IDK if `Paths()` traversal cost scales differently when all gateways share a single class vs. each being its own root).

Consider adding a `numGatewaysPerGC` field to `benchTopologyParams` and nesting the gateway loop inside the GatewayClass loop. The existing sub-benchmarks could then use something like `numGatewayClasses: 1`, `numGatewaysPerGC: 3` to model the realistic shape.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - 'go.sum'
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  benchmark:
+    name: Benchmark
+    runs-on: ubuntu-latest
```
guicassolato: Should we add `timeout-minutes` here? The Makefile sets it in 30m and the job runs benchmarks twice (PR + base), so it can exceed an hour with no cap – maybe some default enforced by the GHA.

A `concurrency` group would also help cancel superseded runs when new commits land on the same PR. E.g.:


```suggestion
    runs-on: ubuntu-latest
    timeout-minutes: 45
    concurrency:
      group: bench-${{ github.event.pull_request.number }}
      cancel-in-progress: true
```

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,97 @@
...
+          ref: ${{ github.base_ref }}
+          clean: false
+          path: base
+          persist-credentials: false
+      - name: Run benchmarks (base)
+        run: cd base && make test-bench && mv benchmark-results.txt ../base-bench.txt
+        continue-on-error: true
+
+      # Compare with benchstat
+      - name: Compare results
```
guicassolato: What if the build fails in `main` for some reason? Maybe condition the 'Compare results' steps to the existence of `base-bench.txt`?


```suggestion
        run: |
          cd base && make test-bench && mv benchmark-results.txt ../base-bench.txt
          echo "exists=true" >> "$GITHUB_OUTPUT"
        continue-on-error: true

      # Compare with benchstat
      - name: Compare results
        if: steps.base_bench.outputs.exists == 'true'
```

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+	numRoutesPerGW     int
+	numRulesPerRoute   int
+	listenersPerGW     int
+	attachAuthPolicies bool
+	attachRLPolicies   bool
+	attachTRLPolicies  bool
+	policyTarget       string // "gateway" (default) or "route"
+}
+
+func (p benchTopologyParams) label() string {
```
guicassolato: Is this func being called somewhere?

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+	to := httpRouteRules[len(httpRouteRules)-1]
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		targetables.Paths(from, to)
+	}
+}
+
+// BenchmarkListenerFanout measures the impact of listener count on Paths() DFS traversal.
+// Reproduces the scaling pattern from #1085 where 63 listeners caused 30-minute reconciliation.
```
guicassolato: The issue reports "63 listeners causes 30-min reconciliation". However, the benchmark test creates 1 gateway with 32 listeners. If halving was intentional for CI runtime, maybe mention it in the comment, otherwise readers will wonder whether this actually covers the reported scenario.

internal/controller/bench_test.go
```diff
@@ -0,0 +1,485 @@
...
+	authReconciler := &EffectiveAuthPolicyReconciler{}
+	rlpReconciler := &EffectiveRateLimitPolicyReconciler{}
+	trlpReconciler := &EffectiveTokenRateLimitPolicyReconciler{}
+
+	ctx := context.Background()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		topology, _, _ := buildBenchTopology(b, p)
+		state := &sync.Map{}
+		for j := 0; j < 28; j++ {
```
guicassolato: Another magic number: 28. Maybe use a named constant or short comment to help understand the rationale behind it, even if the rationale ends up being: "some arbitrary number of repetitions"?

In time, I wonder what looping through looking up the Kuadrant CR achieves in the test. Is it a way to model call-site fan-out?

.github/workflows/benchmark.yaml
maksymvavilov: No reason, updated :+1:

internal/controller/bench_test.go
maksymvavilov: The guard makes sense

internal/controller/bench_test.go
maksymvavilov: I'll add a comment explaining it

internal/controller/bench_test.go
maksymvavilov: This is a fair point, but it shouldn't matter- the number of GW classes and routes per class. I'll change it to a more "realistic" pattern - 1 GW class only; why not? 

As a note: I do not expect us to change the number of gateways: it will significantly increase the duration of the test and will make the workflow annoying without adding much.

.github/workflows/benchmark.yaml
maksymvavilov: Never knew about the `concurrency`, nice! 

Also, good point on the timer, but 45 min is too generous. I'll set 20 min (we expect runs of about 8 mins on average)

.github/workflows/benchmark.yaml
maksymvavilov: Good catch, I'll add a check

internal/controller/bench_test.go
maksymvavilov: not anymore. Good catch, gone!

internal/controller/bench_test.go
maksymvavilov: Forgot to update the comment here.

internal/controller/bench_test.go
maksymvavilov: That was not supposed to be here. Good catch, thank you. Wonder how AI missed it so many times

maksymvavilov:

PR #2037, opened 2026-06-10
fix: sanitize CEL string interpolation to prevent quote injection

### LINKED ISSUE #2036: Sanitize user-supplied values in extension CEL predicates
### Problem

Several extension reconcilers build CEL predicates and expressions by interpolating user-supplied values directly into format strings without sanitization. This affects both correctness (special characters break CEL syntax) and security (quote injection can alter CEL logic).

#### oidc-policy (`cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go`)

| Line | Field | Issue |
|------|-------|-------|
| 306 | `claims` key (`k`) | Dot notation breaks on non-identifier chars (e.g. `custom:role`) |
| 306 | `claims` value (`v`) | Unescaped string literal — quote injection |
| 329 | `IssuerURL` | Unescaped string literal — quote injection |
| 644 | `AuthorizationHeader.Prefix` | Unescaped string literal in CEL Expression |
| 656-658 | `cookieName` + `igw.Hostname` | Unescaped string literals in CEL Expression |

#### plan-policy (`cmd/extensions/plan-policy/api/v1alpha1/planpolicy_types.go`)

| Line | Field | Issue |
|------|-------|-------|
| 68 | `plan.Tier` | Unescaped string literal in WhenPredicate CEL |
| 80 | `plan.Tier` | Unescaped string literal in filter expression |

#### Not affected

- **threat-policy**: uses `%d` with integer `Threshold` — safe by type
- **telemetry-policy**: no user-value interpolation into expressions

### Practical risk

Low. An attacker would need RBAC permission to create or edit policy resources, and at that privilege level there are likely simpler attack vectors. Still, this is straightforward to fix and eliminates an unnecessary injection surface, especially relevant in multi-tenant clusters where policy creation may be delegated to less-privileged teams.

### Proposed fix

Create a shared `celStringLiteral` helper that JSON-encodes values for safe interpolation:

```go
func celStringLiteral(value string) string {
    encoded, _ := json.Marshal(value)
    return string(encoded)
}
```

Then at each interpolation site, use JSON-encoded literals and bracket notation for claim keys:

```go
// Claims — use bracket notation for keys, escaped literals for values
Predicate: fmt.Sprintf(`auth.identity[%s] == %s`, celStringLiteral(k), celStringLiteral(v))

// Simple string values — escape the value
Predicate: fmt.Sprintf(`auth.identity.iss == %s`, celStringLiteral(pol.Spec.Provider.IssuerURL))
```

`json.Marshal` on a string produces a properly escaped, double-quoted JSON string literal, which is also a valid CEL string literal.

The helper could live in a shared package (e.g. `pkg/cel/` or `internal/cel/`) so all extensions can use it.

### Verification

Each fix should include a unit test with:
- A value containing a double quote (e.g. `admin"||true`)
- A claim key with special characters (e.g. `custom:role`)
- A normal value (regression check)

### How to review

1. Check each site listed above in the current code
2. Verify the `celStringLiteral` helper is used at every interpolation point
3. Run unit tests for each affected extension

### Context

- Spotted during review of #1914 (which fixes the `in` → `==` bug for scalar claims, #1625)
- CodeRabbit flagged the same concern on that PR
- No CEL input sanitization exists anywhere in the codebase today

There was a work done for this in https://github.com/Kuadrant/kuadrant-operator/pull/2037 - it can be used as a starting point

kavirakesh14:
Fixes #2036

## What's the problem?

While going through the codebase I noticed that user-supplied values were being 
dropped directly into CEL expressions using fmt.Sprintf without any escaping. 

This caused two issues:
- Claim keys with special characters like `custom:role` broke dot-notation and 
  made the CEL expression fail to compile entirely
- Unescaped string values could allow quote injection to alter CEL logic 
  (e.g. a value like `admin"||true` could make an expression always evaluate to true)

## What I changed

Added a small `StringLiteral` helper in `pkg/cel` that uses `json.Marshal` to 
safely escape values before they go into CEL expressions. JSON-encoded strings 
are also valid CEL string literals, so this works cleanly without any extra 
dependencies.

Also switched claim key lookups from dot-notation to bracket notation so any 
valid JWT claim name works correctly.

## Files changed

- `pkg/cel/strings.go` — new helper + tests
- `cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go` — 4 sites fixed
- `cmd/extensions/plan-policy/api/v1alpha1/planpolicy_types.go` — 2 sites fixed

## Testing

Added unit tests in `pkg/cel/strings_test.go` covering:
- Normal values
- Quote injection attempt (`admin"||true`)
- Claim key with special characters (`custom:role`)
- Backslash handling
- Empty string

Happy to make any changes if I've missed something or the approach needs adjusting!

## Release Notes

* **Refactoring**
  * Updated CEL expression generation in OIDC policy and plan policy modules to employ consistent, standardised string literal handling across configurations, enhancing code consistency and maintainability.

* **Tests**
  * Added comprehensive test coverage validating CEL string literal generation across various input scenarios, including edge cases with special characters and escape sequences.

### DIFF
--- a/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
+++ b/cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
@@ -26,6 +26,7 @@ import (
 
 	kuadrantv1 "github.com/kuadrant/kuadrant-operator/api/v1"
 	"github.com/kuadrant/kuadrant-operator/cmd/extensions/oidc-policy/api/v1alpha1"
+	pkgcel "github.com/kuadrant/kuadrant-operator/pkg/cel"
 	extcontroller "github.com/kuadrant/kuadrant-operator/pkg/extension/controller"
 	"github.com/kuadrant/kuadrant-operator/pkg/extension/types"
 )
@@ -303,7 +304,7 @@ func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*ku
 		for k, v := range claims {
 			authPatterns = append(authPatterns, authorinov1beta3.PatternExpressionOrRef{
 				CelPredicate: authorinov1beta3.CelPredicate{
-					Predicate: fmt.Sprintf(`"%s" in auth.identity.%s`, v, k),
+					Predicate: fmt.Sprintf(`auth.identity[%s] == %s`, pkgcel.StringLiteral(k), pkgcel.StringLiteral(v)),
 				},
 			})
 		}
@@ -326,7 +327,7 @@ func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*ku
 			},
 		}
 		if pol.Spec.Provider.IssuerURL != "" {
-			authorization["oidc"].Conditions[0].Predicate = fmt.Sprintf(`auth.identity.iss == "%s"`, pol.Spec.Provider.IssuerURL)
+			authorization["oidc"].Conditions[0].Predicate = fmt.Sprintf(`auth.identity.iss == %s`, pkgcel.StringLiteral(pol.Spec.Provider.IssuerURL))
 		}
 	}
 
@@ -641,7 +642,7 @@ func credentialsHeader(tokenSource *authorinov1beta3.Credentials, igw *ingressGa
 	headers["location"] = authorinov1beta3.ValueOrSelector{Expression: "auth.authorization.location.location"}
 	switch tokenSource.GetType() {
 	case authorinov1beta3.AuthorizationHeaderCredentials:
-		headers["Authorization"] = authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(fmt.Sprintf(`"%s " + auth.metadata.token.id_token`, tokenSource.AuthorizationHeader.Prefix))}
+		headers["Authorization"] = authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(fmt.Sprintf(`%s + " " + auth.metadata.token.id_token`, pkgcel.StringLiteral(tokenSource.AuthorizationHeader.Prefix)))}
 	case authorinov1beta3.CustomHeaderCredentials:
 		headers[tokenSource.CustomHeader.Name] = authorinov1beta3.ValueOrSelector{Expression: "auth.metadata.token.id_token"}
 	case authorinov1beta3.CookieCredentials:
@@ -653,11 +654,11 @@ func credentialsHeader(tokenSource *authorinov1beta3.Credentials, igw *ingressGa
 }
 
 func cookieHeader(cookieName string, igw *ingressGatewayInfo) authorinov1beta3.ValueOrSelector {
-	return authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(fmt.Sprintf(`
-"%s=" + auth.metadata.token.id_token + "; domain=%s; HttpOnly; %s SameSite=Lax; Path=/; Max-Age=3600"
-`, cookieName, igw.Hostname, getSecureFlag(igw.Protocol)))}
+	cookiePrefix := pkgcel.StringLiteral(cookieName + "=")
+	domainSuffix := pkgcel.StringLiteral("; domain=" + igw.Hostname + "; HttpOnly; " + getSecureFlag(igw.Protocol) + " SameSite=Lax; Path=/; Max-Age=3600")
+	expr := fmt.Sprintf("%s + auth.metadata.token.id_token + %s", cookiePrefix, domainSuffix)
+	return authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(expr)}
 }
-
 func getSecureFlag(protocol gatewayapiv1.ProtocolType) string {
 	flag := ""
 	if protocol == gatewayapiv1.HTTPSProtocolType {
--- a/pkg/cel/strings_test.go
+++ b/pkg/cel/strings_test.go
@@ -1,0 +1,26 @@
+package cel
+
+import "testing"
+
+func TestStringLiteral(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{name: "normal value", input: "hello", want: `"hello"`},
+		{name: "double quote injection", input: `admin"||true`, want: `"admin\"||true"`},
+		{name: "claim key with special characters", input: "custom:role", want: `"custom:role"`},
+		{name: "backslash", input: `foo\bar`, want: `"foo\\bar"`},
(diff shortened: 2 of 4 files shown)


cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -303,7 +304,7 @@ func buildMainAuthPolicy(pol *v1alpha1.OIDCPolicy, igw *ingressGatewayInfo) (*ku
 		for k, v := range claims {
 			authPatterns = append(authPatterns, authorinov1beta3.PatternExpressionOrRef{
 				CelPredicate: authorinov1beta3.CelPredicate{
-					Predicate: fmt.Sprintf(`"%s" in auth.identity.%s`, v, k),
```
maksymvavilov: I suspect this was mirrored from #1914 
I did not expect this task to get picked up so soon - I'd wait until we have a resolution on #1914 and build on top of it.

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
maksymvavilov, line 297: This part was not flagged by me initially. The question is about the hostname: this is something that the gateway API [validates](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/shared_types.go#L625-L627) extensively. 
But on the other hand, why wouldn't we sanitise it?  It will not hurt. However, the whole sanitisation is a "blow on the cold" situation, and here it might be too much. I've made ~5 circles on this one from "yes, sanitise" to "leave it be" - I'll leave a final decision with you on it.

pkg/cel/strings_test.go
```diff
@@ -0,0 +1,26 @@
...
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
+		{name: "normal value", input: "hello", want: `"hello"`},
+		{name: "double quote injection", input: `admin"||true`, want: `"admin\"||true"`},
+		{name: "claim key with special characters", input: "custom:role", want: `"custom:role"`},
+		{name: "backslash", input: `foo\bar`, want: `"foo\\bar"`},
+		{name: "empty string", input: "", want: `""`},
```
maksymvavilov, line 15: Can these be expanded a bit to include a few more exotic cases such as 
```
  {name: "newline", input: "line1\nline2", want: `"line1\nline2"`},
  {name: "null byte", input: "abc\x00def", want: `"abc\u0000def"`},
```
Those are what came to my mind - maybe there could be more? 🤔

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler.go
```diff
@@ -653,11 +654,11 @@ func credentialsHeader(tokenSource *authorinov1beta3.Credentials, igw *ingressGa
...
 func cookieHeader(cookieName string, igw *ingressGatewayInfo) authorinov1beta3.ValueOrSelector {
-	return authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(fmt.Sprintf(`
-"%s=" + auth.metadata.token.id_token + "; domain=%s; HttpOnly; %s SameSite=Lax; Path=/; Max-Age=3600"
-`, cookieName, igw.Hostname, getSecureFlag(igw.Protocol)))}
+	cookiePrefix := pkgcel.StringLiteral(cookieName + "=")
+	domainSuffix := pkgcel.StringLiteral("; domain=" + igw.Hostname + "; HttpOnly; " + getSecureFlag(igw.Protocol) + " SameSite=Lax; Path=/; Max-Age=3600")
+	expr := fmt.Sprintf("%s + auth.metadata.token.id_token + %s", cookiePrefix, domainSuffix)
+	return authorinov1beta3.ValueOrSelector{Expression: authorinov1beta3.CelExpression(expr)}
 }
-
```
maksymvavilov, line 697: I suspect this removal is an auto-formatting of sorts. For the sake of some tender eyes (mine included), can you, please, bring an empty line back?

maksymvavilov (commented):
Looks great! A few nit-pick comments/suggestions, tho. 
Was done sooner than I'd expected: I'd wait for the resolution of the #1914 to avoid awkward merge conflicts.

maksymvavilov:
hey @kavirakesh14, the PR #1914 that was blocking this one got merged

thomasmaas (changes requested):
Nice work on the `StringLiteral` helper — the injection concern is real and the `json.Marshal` approach is clean. A few things need addressing before this can merge:

## Blocking

### 1. Behavioral regression: list claim matching dropped

The change on line 306 switches from `"%s" in auth.identity.%s` (list membership) to `auth.identity[%s] == %s` (strict equality). This breaks list-type claims like `groups`:

```
# Old: matches when "admin" is in the groups list
"admin" in auth.identity.groups  →  true for groups: ["admin", "user"]

# New: compares the entire list to a string
auth.identity["groups"] == "admin"  →  false for groups: ["admin", "user"]
```

Since #1914 has merged, the fix should target the new `claimPredicate` function on main (line 305) which already handles both scalars and lists. Rebase onto main, then apply `StringLiteral` to `claimPredicate` without changing its matching logic:

```go
func claimPredicate(k, v string) string {
    key := pkgcel.StringLiteral(k)
    val := pkgcel.StringLiteral(v)
    return fmt.Sprintf(
        `has(auth.identity) && has(auth.identity[%s]) && (auth.identity[%s] == %s || (type(auth.identity[%s]) == list && %s in auth.identity[%s]))`,
        key, key, val, key, val, key,
    )
}
```

This also switches from dot-notation to bracket-notation for claim keys (fixing the `custom:role` issue) while preserving the dual scalar/list matching.

### 2. Rebase needed

The PR conflicts with main because #1914 refactored the claim matching code into `claimPredicate`. After rebasing, the fix sites are:
- `claimPredicate` (line 305 on main) — apply `StringLiteral` to k and v, switch to bracket notation
- `issuerURL` (line 356 on main) — already correct in this PR
- `credentialsHeader` (line 680 on main) — already correct in this PR
- `cookieHeader` (line 692 on main) — already correct in this PR
- `planpolicy_types.go` — already correct in this PR

## Nits

### 3. `pkg/cel/strings.go:10` — document why error is discarded

`json.Marshal` can't fail for string input, but a comment makes that explicit:

```go
encoded, _ := json.Marshal(value) // json.Marshal cannot fail for string values
```

### 4. Blank line removal

The removed blank line between `cookieHeader` and `getSecureFlag` (line 660) is unrelated — please restore it per Maksym's earlier comment.

maksymvavilov:
@kavirakesh14 would you be ok if I unassign you from this issue and include your PR as a starting point for the work on this issue?

maksymvavilov:

maksymvavilov:
I'll close this PR as inactive and will mention it as a starting point on the GH issue

PR #2044, opened 2026-06-15
Set failure_policy to FAIL_RELOAD

adam-cattermole:
https://github.com/envoyproxy/envoy/blob/9a5e83205ec49a63ceabfb18f747c521833e5a3a/api/envoy/extensions/wasm/v3/wasm.proto#L24-L42

## Release Notes

* **Bug Fixes**
  * Enhanced wasm filter configuration handling to improve fault tolerance and resilience in filter deployment scenarios.

### DIFF
--- a/internal/istio/utils.go
+++ b/internal/istio/utils.go
@@ -131,6 +131,7 @@ func buildWasmFilterConfig(wasmURL, imagePullSecret, imageSHA, clusterName strin
 			},
 			"allow_precompiled": true,
 		},
+		"failure_policy":                  "FAIL_RELOAD",
 		"allow_on_headers_stop_iteration": true,
 	}
 


internal/istio/utils.go
```diff
@@ -131,6 +131,7 @@ func buildWasmFilterConfig(wasmURL, imagePullSecret, imageSHA, clusterName strin
 			},
 			"allow_precompiled": true,
 		},
+		"failure_policy":                  1, // FAIL_RELOAD
```
alexsnaps: Can't that "just be `"FAIL_RELOAD"`" as per [this](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/wasm/v3/wasm.proto#envoy-v3-api-enum-extensions-wasm-v3-failurepolicy), I'd expect all the JSON/proto magic to do the conversion ?

adam-cattermole: Ah yep you're 100% right - I was initially looking for the go type to be able to use the correct variant but this isn't the right 'layer'

adam-cattermole: Addressed

alexsnaps (approved):

PR #2045, opened 2026-06-18
Fix non-deterministic EnvoyFilter generation

### LINKED ISSUE #2033: EnvoyFilter equality check fails, causing continuous unnecessary updates
## Description

After the migration from `WasmPlugin` to `EnvoyFilter` in PR #1953, the `EqualEnvoyFilters` comparison function does not correctly handle the new `HTTP_FILTER` patch shape. This causes the reconciler to treat unchanged wasm EnvoyFilters as different on every pass, resulting in continuous unnecessary updates.

The symptom is that `EnvoyFilter` resource generation keeps incrementing even when no configuration has changed. This prevents the operator from reaching a stable state, and the problem worsens as the number of resources increases.

This was observed on commit `b727d75f`.

## Observed behavior

The operator logs show a repeating cycle of `EnvoyFilter` and `ConfigMap` update events followed by wasm configuration rebuilds, even with no user-initiated changes:

```json
{"level":"info","ts":"2026-06-05T14:02:20.681Z","logger":"kuadrant-operator.event logger","msg":"new events","resources":["EnvoyFilter","ConfigMap"],"eventTypes":{"update":2}}
{"level":"info","ts":"2026-06-05T14:02:30.883Z","logger":"kuadrant-operator.IstioExtensionReconciler.buildWasmConfigs","msg":"build Wasm configuration","status":"started"}
{"level":"info","ts":"2026-06-05T14:02:30.883Z","logger":"kuadrant-operator.IstioExtensionReconciler.buildWasmConfigs","msg":"build Wasm configuration","status":"completed"}
```

Inspecting the `EnvoyFilter` resources on the cluster shows the resource generation continuously increasing, confirming the compare function is incorrectly detecting a difference and writing a new configuration each time.

## Expected behavior

Once the operator has reconciled all resources, the `EnvoyFilter` generation should stabilize and the operator should not re-process resources unless an actual change occurs.

## Impact

- The operator will never become stable when resources are present.
- The higher the number of resources, the worse the impact as the reconciler spins continuously.
- Observed with as few as 5 resource sets (HTTPRoute + AuthPolicy + RateLimitPolicy).

### Prerequisites

- A KinD cluster (a large multi-node cluster is not required)
- The `kuadrant-operator` repository checked out at commit `b727d75f` or later

### 1. Deploy kuadrant

From the kuadrant-operator repo:

```sh
make local-setup
```

### 2. Increase operator resources (optional, speeds up reproduction)

```sh
kubectl set resources deployment/kuadrant-operator-controller-manager -n kuadrant-system \
  --limits=cpu=4,memory=2Gi \
  --requests=cpu=4,memory=2Gi

kubectl wait --for=condition=Available deployment/kuadrant-operator-controller-manager \
  -n kuadrant-system --timeout=120s
```

### 3. Create the Kuadrant CR and test namespace

```sh
kubectl apply -f examples/toystore/kuadrant.yaml -n kuadrant-system
kubectl create namespace poc-0
```

### 4. Create resource sets

Apply at least 5 sets of HTTPRoute + AuthPolicy + RateLimitPolicy. Each set consists of:

```sh
NAMESPACE="poc-0"
GATEWAY_NAMESPACE="gateway-system"
GATEWAY_NAME="kuadrant-ingressgateway"
NUMBER=50

for i in $(seq 0 $((NUMBER - 1))); do
  cat <<YAML | kubectl apply -f - > /dev/null
kind: HTTPRoute
metadata:
  name: httpbin-route-$i
  namespace: $NAMESPACE
spec:
  parentRefs:
  - name: $GATEWAY_NAME
    namespace: $GATEWAY_NAMESPACE
  hostnames:
  - "httpbin-$i.local"
  rules:
  - backendRefs:
    - name: httpbin
      port: 80
YAML

  cat <<YAML | kubectl apply -f - > /dev/null
apiVersion: kuadrant.io/v1
kind: AuthPolicy
metadata:
  name: httpbin-auth-$i
  namespace: $NAMESPACE
spec:
  targetRef:
    kind: HTTPRoute
    name: httpbin-route-$i
  rules:
    authentication:
      "api-key":
        apiKey:
          selector:
            matchLabels:
              app: httpbin
YAML

  cat <<YAML | kubectl apply -f - > /dev/null
apiVersion: kuadrant.io/v1
kind: RateLimitPolicy
metadata:
  name: httpbin-ratelimit-$i
  namespace: $NAMESPACE
spec:
  targetRef:
    kind: HTTPRoute
    name: httpbin-route-$i
  limits:
    "general-user":
      rates:
      - limit: 5
        window: 10s
      counters:
      when:
      - predicate: "auth.identity.userid != 'bob'"
    "bob-limit":
      rates:
      - limit: 2
        window: 10s
      when:
      - predicate: "auth.identity.userid == 'bob'"
YAML
done
```

### 5. Observe the logs

```sh
kubectl logs -f deployment/kuadrant-operator-controller-manager -n kuadrant-system
```

Wait for the operator to finish processing the initial resource creation. After it settles, the repeating `EnvoyFilter`/`ConfigMap` update events and wasm configuration rebuilds described above will be visible in the logs.

### 6. Confirm via resource generation

```sh
kubectl get envoyfilters -A -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.generation}{"\n"}{end}'
```

Running this command repeatedly will show the generation number continuously increasing.

## References

- PR #1953: Use `EnvoyFilter` in lieu of `WasmPlugin` to install our Wasm Filter

adam-cattermole:
## Verification

Follow the steps outlined in #2033  before and after this change

## Bug Fixes

* Improved consistency in data-plane policy merging for WASM actions sharing scope and service configuration
* Enhanced deterministic processing order for rate-limit policy rule handling in token-rate and generic rate-limit scenarios
* Refined Envoy filter comparison methodology for more reliable configuration validation

### DIFF
--- a/internal/controller/istio_extension_reconciler_test.go
+++ b/internal/controller/istio_extension_reconciler_test.go
@@ -636,4 +636,86 @@ func TestMergeAndVerifyEdgeCases(t *testing.T) {
 		_, err := mergeAndVerify(context.TODO(), actions)
 		assert.ErrorContains(t, err, "duplicate key '' with different values")
 	})
+
+	t.Run("deterministic conditional data ordering after merge", func(t *testing.T) {
+		// Create two sets of actions in different orders
+		actionsOrder1 := []wasm.Action{
+			{
+				ServiceName: wasm.RateLimitCheckServiceName,
+				Scope:       "route-0",
+				ConditionalData: []wasm.ConditionalData{
+					{
+						Predicates: []string{"auth.identity.bob"},
+						Data: []wasm.DataType{
+							{Value: &wasm.Static{Static: wasm.StaticSpec{Key: "limit.bob", Value: "1"}}},
+						},
+					},
+				},
+			},
+			{
+				ServiceName: wasm.RateLimitCheckServiceName,
+				Scope:       "route-0",
+				ConditionalData: []wasm.ConditionalData{
+					{
+						Predicates: []string{"auth.identity.alice"},
+						Data: []wasm.DataType{
+							{Value: &wasm.Static{Static: wasm.StaticSpec{Key: "limit.alice", Value: "1"}}},
+						},
+					},
+				},
+			},
+		}
+
+		actionsOrder2 := []wasm.Action{
+			{
+				ServiceName: wasm.RateLimitCheckServiceName,
+				Scope:       "route-0",
+				ConditionalData: []wasm.ConditionalData{
+					{
+						Predicates: []string{"auth.identity.alice"},
+						Data: []wasm.DataType{
+							{Value: &wasm.Static{Static: wasm.StaticSpec{Key: "limit.alice", Value: "1"}}},
+						},
+					},
+				},
+			},
+			{
+				ServiceName: wasm.RateLimitCheckServiceName,
+				Scope:       "route-0",
+				ConditionalData: []wasm.ConditionalData{
+					{
+						Predicates: []string{"auth.identity.bob"},
+						Data: []wasm.DataType{
+							{Value: &wasm.Static{Static: wasm.StaticSpec{Key: "limit.bob", Value: "1"}}},
+						},
+					},
+				},
+			},
+		}
+
+		result1, err := mergeAndVerify(context.TODO(), actionsOrder1)
+		assert.NilError(t, err)
+		assert.Equal(t, 1, len(result1))
+
+		result2, err := mergeAndVerify(context.TODO(), actionsOrder2)
+		assert.NilError(t, err)
+		assert.Equal(t, 1, len(result2))
+
+		// Both results should have exactly the same ConditionalData in the same order
+		assert.Equal(t, len(result1[0].ConditionalData), len(result2[0].ConditionalData))
+		for i := range result1[0].ConditionalData {
+			cd1 := result1[0].ConditionalData[i]
+			cd2 := result2[0].ConditionalData[i]
+
+			// Predicates should be in the same order
+			assert.DeepEqual(t, cd1.Predicates, cd2.Predicates)
+
+			// Data should be in the same order
+			assert.Equal(t, len(cd1.Data), len(cd2.Data))
+		}
+
+		// The ConditionalData should be sorted by predicates (alice before bob)
+		assert.Equal(t, "auth.identity.alice", result1[0].ConditionalData[0].Predicates[0])
+		assert.Equal(t, "auth.identity.bob", result1[0].ConditionalData[1].Predicates[0])
+	})
 }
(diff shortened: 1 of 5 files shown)


internal/istio/utils.go
didierofrivia: I reckon here would be needed too: 

```go
 if aListener != nil && bListener != nil {
     if !proto.Equal(aConfigPatch.Match, bConfigPatch.Match) {
         return false
     }
 }
```

adam-cattermole: Yeah nice I think that looks good and potentially the default case also

internal/controller/istio_extension_reconciler_test.go
```diff
@@ -636,4 +636,86 @@ func TestMergeAndVerifyEdgeCases(t *testing.T) {
...
+		assert.Equal(t, 1, len(result2))
+
+		// Both results should have exactly the same ConditionalData in the same order
+		assert.Equal(t, len(result1[0].ConditionalData), len(result2[0].ConditionalData))
+		for i := range result1[0].ConditionalData {
+			cd1 := result1[0].ConditionalData[i]
+			cd2 := result2[0].ConditionalData[i]
+
+			// Predicates should be in the same order
+			assert.DeepEqual(t, cd1.Predicates, cd2.Predicates)
```
didierofrivia, line 711: Nice, this successfully _fails_ without the fix.

internal/istio/utils_test.go
```diff
@@ -209,6 +209,100 @@ func TestEqualEnvoyFilters(t *testing.T) {
 
 		assert.Assert(subT, !EqualEnvoyFilters(a, b))
 	})
+
+	t.Run("equal patch values with maps in different order", func(subT *testing.T) {
```
didierofrivia, line 213: These 2 tests are passing even without the fix, which are similar to the ones I was writing... couldn't make them fail with the buggy code neither...

adam-cattermole: I think it's because it's hard to have the maps be out of order for these static ones.. even though the ordering is non-deterministic. I'm not sure how else to verify though other than introduce something that 'could fail' if we broke the behaviour - i.e. if we remove the determinism we'd get a flakey test?

adam-cattermole: I'm on the fence, do you think it'd be better to just remove these ones?

didierofrivia (approved):
🥇

internal/istio/utils_test.go
didierofrivia: leave them, at least they test what they supposed to do... even tho the old code was buggy, from now on it's good.

PR #2048, opened 2026-06-19
FIX: Prevent operator crash when topology ConfigMap exceeds 1MB limit

Boomatang:
# Problem

The kuadrant-operator crashes when the topology ConfigMap exceeds the Kubernetes 1MB ConfigMap size limit. This happens at scale when approximately 1,500 or more resource sets (HTTPRoute + AuthPolicy + RateLimitPolicy) are present in the cluster. The topology DOT graph grows linearly (~785 bytes per route), and once it exceeds 1MB the API server rejects the write, causing the reconciler to enter a permanent error loop. The operator becomes unable to process any further events, effectively a denial of service.

# Reproduction

1. Deploy kuadrant from the operator repo:
   ```sh
   make local-setup
   ```

2. Increase operator resources (speeds up reproduction, does not affect outcome):
   ```sh
   kubectl set resources deployment/kuadrant-operator-controller-manager -n kuadrant-system \
     --limits=cpu=4,memory=2Gi \
     --requests=cpu=4,memory=2Gi
   kubectl wait --for=condition=Available deployment/kuadrant-operator-controller-manager \
     -n kuadrant-system --timeout=120s
   ```

3. Create the Kuadrant CR and a test namespace:
   ```sh
   kubectl apply -f examples/toystore/kuadrant.yaml -n kuadrant-system
   kubectl create namespace poc-0
   ```

4. Create ~1500 resource sets (HTTPRoute + AuthPolicy + RateLimitPolicy each):
   ```sh
   NAMESPACE="poc-0"
   GATEWAY_NAMESPACE="gateway-system"
   GATEWAY_NAME="kuadrant-ingressgateway"
   NUMBER=1500

   for i in $(seq 0 $((NUMBER - 1))); do
     kubectl apply -f - <<YAML > /dev/null
   kind: HTTPRoute
   metadata:
     name: httpbin-route-$i
     namespace: $NAMESPACE
   spec:
     parentRefs:
     - name: $GATEWAY_NAME
       namespace: $GATEWAY_NAMESPACE
     hostnames:
     - "httpbin-$i.local"
     rules:
     - backendRefs:
       - name: httpbin
         port: 80
   YAML

     kubectl apply -f - <<YAML > /dev/null
   apiVersion: kuadrant.io/v1
   kind: AuthPolicy
   metadata:
     name: httpbin-auth-$i
     namespace: $NAMESPACE
   spec:
     targetRef:
       kind: HTTPRoute
       name: httpbin-route-$i
     rules:
       authentication:
         "api-key":
           apiKey:
             selector:
               matchLabels:
                 app: httpbin
   YAML

     kubectl apply -f - <<YAML > /dev/null
   apiVersion: kuadrant.io/v1
   kind: RateLimitPolicy
   metadata:
     name: httpbin-ratelimit-$i
     namespace: $NAMESPACE
   spec:
     targetRef:
       kind: HTTPRoute
       name: httpbin-route-$i
     limits:
       "general-user":
         rates:
         - limit: 5
           window: 10s
         counters:
         when:
         - predicate: "auth.identity.userid != 'bob'"
       "bob-limit":
         rates:
         - limit: 2
           window: 10s
         when:
         - predicate: "auth.identity.userid == 'bob'"
   YAML
   done
   ```

5. Watch the operator logs:
   ```sh
   kubectl logs -f deployment/kuadrant-operator-controller-manager -n kuadrant-system
   ```

   Without this fix the operator will enter a permanent error loop with:
   ```
   ConfigMap "topology" is invalid: []: Too long: may not be more than 1048576 bytes
   ```

# Solution

This PR adds a size guard to the `TopologyReconciler`. Before writing the topology ConfigMap, the reconciler checks the size of the serialized DOT graph against a 900KB threshold (safely under the 1MB Kubernetes limit). When the topology exceeds this threshold, a small placeholder DOT graph is written instead, preventing the API server rejection and allowing the operator to continue reconciling other resources.

## Changes

- Add a `maxTopologyBytes` constant (900KB) and an `oversizedPlaceholder` DOT string
- Check topology size before writing and substitute the placeholder when oversized
- Add OpenTelemetry tracing spans, attributes, and events for observability
- Change `Client` field type from `*dynamic.DynamicClient` to `dynamic.Interface` to support test fakes
- Add span instrumentation for all error and success paths
- Add unit tests covering create, update, no-op, oversized placeholder validation, and panic on empty namespace

## Limitations

This is a temporary workaround. The placeholder means the console-plugin will not display the full topology graph at scale. A proper solution will require a different storage or serialization strategy, coordinated with the console-plugin.

## Release Notes

* **Bug Fixes**
  * Automatically caps oversized topology DOT payloads and replaces them with a valid placeholder to prevent reconciliation failures.

* **New Features**
  * Improved observability for topology reconciliation using OpenTelemetry tracing, including span attributes and events covering the ConfigMap create/update lifecycle.

* **Tests**
  * Added unit tests for placeholder sizing, topology ConfigMap creation, updates, and handling of unchanged content.

### DIFF
--- a/internal/controller/topology_reconciler.go
+++ b/internal/controller/topology_reconciler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kuadrant/policy-machinery/machinery"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -41,9 +42,7 @@ func NewTopologyReconciler(client dynamic.Interface, namespace string) *Topology
 
 func (r *TopologyReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, _ *sync.Map) error {
 	logger := controller.LoggerFromContext(ctx).WithName("topology file").WithValues("context", ctx)
-	tracer := controller.TracerFromContext(ctx)
-	ctx, span := tracer.Start(ctx, "TopologyReconciler.Reconcile")
-	defer span.End()
+	span := trace.SpanFromContext(ctx)
 
 	span.SetAttributes(
 		attribute.String("configmap.name", TopologyConfigMapName),
--- a/internal/controller/topology_reconciler.go
+++ b/internal/controller/topology_reconciler.go
@@ -2,11 +2,14 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 
 	"github.com/kuadrant/policy-machinery/controller"
 	"github.com/kuadrant/policy-machinery/machinery"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,14 +20,19 @@ import (
 
 const (
 	TopologyConfigMapName = "topology"
+	// TODO: This size cap is a temporary workaround. The topology can outgrow a single ConfigMap,
+	// and other resources that consume it (e.g. the console-plugin) will need coordinated changes
+	// to support a different storage or serialization strategy.
+	maxTopologyBytes     = 900 * 1024 // ~900KB, safely under the 1MB ConfigMap limit
+	oversizedPlaceholder = `digraph { "error" [label="Topology exceeds ConfigMap 1MB limit"] }`
 )
 
 type TopologyReconciler struct {
-	Client    *dynamic.DynamicClient
+	Client    dynamic.Interface
 	Namespace string
 }
 
-func NewTopologyReconciler(client *dynamic.DynamicClient, namespace string) *TopologyReconciler {
+func NewTopologyReconciler(client dynamic.Interface, namespace string) *TopologyReconciler {
 	if namespace == "" {
 		panic("namespace must be specified and can not be a blank string")
 	}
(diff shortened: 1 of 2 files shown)


Boomatang:
@R-Lawton I want to  give you a head up on this. It will have an effect on the console-plugin if the topology becomes to large. I am not sure how the console-plugin would handle the placeholder topology that is added.

Boomatang:
I have check what happens in the console plugin if the topology config map holds the error state diagram. The UI gives an `Error parsing topology: Failed to parse topology data.` error. This can be seen in the right hand side of the screen shot below. The left hand side is the configmap that was being rendered.

image

internal/controller/topology_reconciler.go
```diff
@@ -33,6 +41,27 @@ func NewTopologyReconciler(client *dynamic.DynamicClient, namespace string) *Top
 
 func (r *TopologyReconciler) Reconcile(ctx context.Context, _ []controller.ResourceEvent, topology *machinery.Topology, _ error, _ *sync.Map) error {
 	logger := controller.LoggerFromContext(ctx).WithName("topology file").WithValues("context", ctx)
+	tracer := controller.TracerFromContext(ctx)
+	ctx, span := tracer.Start(ctx, "TopologyReconciler.Reconcile")
```
eguzki: I believe there is already a span for the topology reconciler. Consider reusing it.

```
span := trace.SpanFromContext(ctx)
```

internal/controller/topology_reconciler_test.go
```diff
@@ -0,0 +1,180 @@
+//go:build unit
+
```
eguzki, line 2: I do not see a test case exercising the new behavior. Worth doing it?

internal/controller/topology_reconciler_test.go
```diff
@@ -0,0 +1,180 @@
...
+	fakeClient := dfake.NewSimpleDynamicClient(scheme)
+	unstructuredCM, err := controller.Destruct(existingCM.Object.(*corev1.ConfigMap))
+	assert.NilError(t, err)
+	_, err = fakeClient.Resource(controller.ConfigMapsResource).Namespace(namespace).Create(context.Background(), unstructuredCM, metav1.CreateOptions{})
+	assert.NilError(t, err)
+
+	reconciler := NewTopologyReconciler(fakeClient, namespace)
+	err = reconciler.Reconcile(context.Background(), nil, topology, nil, &sync.Map{})
+	assert.NilError(t, err)
```
eguzki, line 149: doesn't actually assert that Update was skipped. Maybe compare the `ResourceVersion`.

internal/controller/topology_reconciler_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+func TestNewTopologyReconciler_PanicsOnEmptyNamespace(t *testing.T) {
+	defer func() {
+		r := recover()
+		assert.Assert(t, r != nil, "expected panic for empty namespace")
+	}()
+	NewTopologyReconciler(nil, "")
+}
+
+func unstructuredNestedString(obj map[string]any, fields ...string) (string, bool, error) {
```
eguzki: I believe https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#NestedString is doing the exact same thing.

eguzki (commented):
Overall looks good.

Left some comments

internal/controller/topology_reconciler.go
Boomatang: The tracing stuff is new to me, I will change that.

internal/controller/topology_reconciler_test.go
Boomatang: Not sure how we can do this. There is no `ResourceVersion` updated on the configmap, and even at that I don't think we would have access to the real configmap in the unit tests.

internal/controller/topology_reconciler_test.go
Boomatang: Though about it but was sure the best way to construct the test. The topology would need to be about 1400 sets of the resources (HTTPRoute, Authpolicy, RateLimitPolicy), and as we don't have access to the posted configmap it is hard to test that it worked as expected.

I am not sure what is best to do.

Boomatang: Is it possible to get access to the tracing span data during a unit test?

internal/controller/topology_reconciler_test.go
eguzki: If it's unittest, you can provide a fake client (already done) and then inspect the fake client for actions being done (or not done) on it.

internal/controller/topology_reconciler_test.go
eguzki: You can build a big topology in memory that meets `topologySize > maxTopologyBytes`, then assert the "stored" topology is the oversized placeholder. 

> Is it possible to get access to the tracing span data during a unit test?

if the tracer is set in the context, find a mock tracer. Quick search led me to go.opentelemetry.io/otel/sdk/trace/tracetest. After the Reconcile call, you can inspect the mock tracer for spans and attributes and events stored in the span.

eguzki (approved):
LGTM

Good guardrail mechanism. 

Anyway, I believe that the topology rendering in a configmap should be done by the consumer (the console plugin) at the frontend side. In the meantime, this PR should protect the operator with clusters at scale

eguzki:
BTW, it's working. The verification steps led to the oversized topology which turned into a topology with a warning message
Screenshot From 2026-07-14 12-21-03

PR #2053, opened 2026-06-22
Cleanup orphaned WasmPlugin objects

adam-cattermole:
We recently migrated away from `WasmPlugin` to `EnvoyFilter` for our wasm configuration to configure `allow_on_headers_stop_iteration`, to move toward supporting request body processing. 

As part of this work we missed the cleanup of existing `WasmPlugin` objects. This would result in the envoy configuration having two merged copies of the wasm config. The protection policy configuration for one of these would be snapshotted at the state at the time of upgrade, and the other would reflect current state. This would result in overly restrictive policies where double auth/double rate limiting is happening on each request.

A workaround to mitigate, the existing `WasmPlugin` objects in gateway namespaces, with the kuadrant ownership label should be removed, once it’s been confirmed that the new `EnvoyFilter` is in place - guide outlines this here https://gist.github.com/adam-cattermole/5dd1fa53aa36af57254170776ad53601.

### Questions

This PR implements a naive solution to remove the orphaned `WasmPlugin` objects as part of the regular reconcile logic. This comes with a major racey caveat that there is no guarantee the `EnvoyFilter` has been applied correctly before the `WasmPlugin` is removed (there's no status on these and it requires extracting the envoy config from the gateway) - leaving the upstream unprotected.

- What's our stance on upgrades, and uptime - do we have guides on expected behaviour for an istio version upgrade for example?
- Is it acceptable to have this behaviour documented with a warning indicating that for workloads where service protection is critical you cut off access to your cluster/gateways prior to upgrade?

Curious on thoughts and how to proceed here, perhaps @maleck13, @guicassolato

### Alternatives

- I've considered using an initContainer as part of the upgrade process instead of the reconciler. I see one major drawback which is that there's no guarantee the `EnvoyFilter` would ever be ready for a gateway - does the initContainer have timeout conditions, could this make our upgrade process flakey and error prone?
- We have discussed having proper communication between the wasm-shim and the kuadrant-operator. We now have a channel where these communicate, and so there's the option of having the wasm-shim _confirm_ it has loaded the config for a specific identifier - this would mean we can update policy status' much more reliably and say at what time they are `Enforced`, and in this scenario, we would know when it's safe to remove the `WasmPlugin`. This however is not a small amount of work and I'm concerned about rushing this for the fix here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of configuration updates so failed changes are reported properly instead of being silently ignored.
  * Added safer cleanup for related Istio resources during reconciliation, helping prevent stale resources from lingering.
* **Chores**
  * Updated generated deployment metadata and permissions to match the latest resource access requirements.

### DIFF
--- a/internal/controller/istio_extension_reconciler.go
+++ b/internal/controller/istio_extension_reconciler.go
@@ -97,6 +97,7 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 	}
 
 	modifiedGateways := make([]string, 0, len(gateways))
+	var reconcileErr error
 
 	for _, gateway := range gateways {
 		gatewayKey := k8stypes.NamespacedName{Name: gateway.GetName(), Namespace: gateway.GetNamespace()}
@@ -124,11 +125,12 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 			desiredEnvoyFilterUnstructured, err := controller.Destruct(desiredEnvoyFilter)
 			if err != nil {
 				logger.Error(err, "failed to destruct envoyfilter object", "gateway", gatewayKey.String(), "envoyfilter", desiredEnvoyFilter)
+				reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to destruct envoyfilter %s/%s: %w", gateway.GetNamespace(), desiredEnvoyFilter.GetName(), err))
 				continue
 			}
 			if _, err = resource.Create(ctx, desiredEnvoyFilterUnstructured, metav1.CreateOptions{}); err != nil {
 				logger.Error(err, "failed to create envoyfilter object", "gateway", gatewayKey.String(), "envoyfilter", desiredEnvoyFilterUnstructured.Object)
-				// TODO: handle error
+				reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to create envoyfilter %s/%s: %w", gateway.GetNamespace(), desiredEnvoyFilter.GetName(), err))
 			}
 			continue
 		}
@@ -136,7 +138,8 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 		// Clean up old WasmPlugin for this specific gateway - temporary to be removed
 		wasmPluginName := wasm.ExtensionName(gateway.GetName())
 		if err := r.client.Resource(kuadrantistio.WasmPluginsResource).Namespace(gateway.GetNamespace()).Delete(ctx, wasmPluginName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
-			logger.Error(err, "failed to delete old wasmplugin", "gateway", gatewayKey.String(), "wasmplugin", wasmPluginName)
+			logger.Error(err, "failed to delete wasmplugin", "gateway", gatewayKey.String(), "wasmplugin", wasmPluginName)
+			reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to delete wasmplugin %s/%s: %w", gateway.GetNamespace(), wasmPluginName, err))
 		}
 
 		existingEnvoyFilter := existingEnvoyFilterObj.(*controller.RuntimeObject).Object.(*istioclientgonetworkingv1alpha3.EnvoyFilter)
@@ -145,7 +148,7 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 		if utils.IsObjectTaggedToDelete(desiredEnvoyFilter) && !utils.IsObjectTaggedToDelete(existingEnvoyFilter) {
 			if err := resource.Delete(ctx, existingEnvoyFilter.GetName(), metav1.DeleteOptions{}); err != nil {
 				logger.Error(err, "failed to delete envoyfilter object", "gateway", gatewayKey.String(), "envoyfilter", fmt.Sprintf("%s/%s", existingEnvoyFilter.GetNamespace(), existingEnvoyFilter.GetName()))
-				// TODO: handle error
+				reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to delete envoyfilter %s/%s: %w", existingEnvoyFilter.GetNamespace(), existingEnvoyFilter.GetName(), err))
 			}
 			continue
 		}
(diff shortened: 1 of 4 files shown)


thomasmaas (commented):
**Overall:** Reasonable temporary fix for a real upgrade problem. The approach is sound — reuse the known `ExtensionName` convention, delete-with-NotFound-guard, minimal RBAC additions. A few observations below.

**Race condition in the create branch (line 129-138):**
The WasmPlugin delete runs even when `resource.Create()` fails — there's no `continue` after the failed create (the existing `// TODO: handle error` already signals this gap). This means a persistent EnvoyFilter creation failure would leave the gateway unprotected. Easy fix: add `continue` after the create error log, or gate the cleanup on `err == nil`. I realize the broader race (EnvoyFilter created but not yet applied by Envoy) is acknowledged and harder to solve, but this specific case is avoidable.

**Duplication (lines 134-138 and 145-149):**
The cleanup block is identical in both branches. Since it should run regardless of whether an EnvoyFilter exists, it can move to a single location before the `if !found` check — right after `desiredEnvoyFilter` is built, before the branching logic.

**Tracking removal:**
The comments say "temporary to be removed" — worth filing a tracking issue and referencing it in a TODO so this doesn't become permanent.

**RBAC (config/rbac/role.yaml):**
`get` and `list` are included but the code only calls `Delete`. Fine to keep if you plan to add a list-before-delete guard later, otherwise they're unnecessary permissions.

**No tests added:** Given this is temporary and idempotent, not a hard objection, but a simple integration test (create a WasmPlugin, trigger reconcile, assert it's gone) would catch regressions if the `ExtensionName` convention ever changes.

adam-cattermole:
Thanks for the feedback @thomasmaas, addressed those so it's ready once we decide how to proceed (whether we use this or not).

Although I disagree with this section so I've disregarded - we should still best effort only remove once the `EnvoyFilter` is present:

> Since it should run regardless of whether an EnvoyFilter exists, it can move to a single location before the if !found check — right after desiredEnvoyFilter is built, before the branching logic.

maleck13 (commented):
change looks reasonable to me. Only thought was around the logging of the error rather than returning it? Do we want to force a re-try if it fails

adam-cattermole:
> Only thought was around the logging of the error rather than returning it? Do we want to force a re-try if it fails

Yeah good question - I tend to agree that while both the `EnvoyFilter` and `WasmPlugin` are present things could be overly restrictive / fail so we should probably return an error and trigger a retry. The only caveat being that we're within the loop over gateways so I think we should 'try attempt them all' first before we return the error instead of erroring out of the loop - or at least I assume that's why the `EnvoyFilter` creation errors are also ignored

adam-cattermole:
I'm collecting errors and return after we've tried all at the end - addressed the creation/update/delete/destruct errors to also retrigger (the TODO has been there for 1 year 8 months..)

internal/controller/istio_extension_reconciler.go
```diff
@@ -126,11 +129,17 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
...
-				// TODO: handle error
+				reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to create envoyfilter %s/%s: %w", gateway.GetNamespace(), desiredEnvoyFilter.GetName(), err))
 			}
 			continue
 		}
 
+		// Clean up old WasmPlugin for this specific gateway - temporary to be removed
+		wasmPluginName := wasm.ExtensionName(gateway.GetName())
+		if err := r.client.Resource(kuadrantistio.WasmPluginsResource).Namespace(gateway.GetNamespace()).Delete(ctx, wasmPluginName, metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
+			reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to delete wasmplugin %s/%s: %w", gateway.GetNamespace(), wasmPluginName, err))
```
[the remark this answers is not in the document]
adam-cattermole: Thoughts on this @maleck13? I'm in two minds - we do have the ownership label but we'd need to add get perms, retrieve first, double check the labels, then delete. On the flip-side the name is prefixed `kuadrant-` so perhaps that's sufficient for us to assume ownership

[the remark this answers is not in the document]
maleck13: kuadrant- is sufficient for us to assume ownership IMO. The risk of another WASMPlugin resource created with a kuadrant- is tiny

didierofrivia:
This should be OK _alongside clear documentation on the upgrade path_, besides, it would be required that this approach is understood/accepted by Product / team leads / architects, since there are a couple of questions in the description that require some answers. It would be ideal to have a test in our testsuite asserting the failure to upgrade before this fix and the correct enforcement of the policies afterwards.

I'm particularly drawn to the alternative of having communication between the shim and operator, making sure the correct configuration is there and as a corollary the actual enforcement state in the policies

thomasmaas (approved):
All feedback from prior review addressed. The code is in good shape:

- **Cleanup placement is correct** — only runs when EnvoyFilter already exists in topology, avoiding the race where WasmPlugin is removed before the replacement is in place.
- **Error handling** — the second commit properly accumulates all write errors into `reconcileErr` and returns them for retry. This also fixes the pre-existing `// TODO: handle error` gaps for EnvoyFilter create/update/delete.
- **RBAC minimal** — only `delete` on `wasmplugins`, nothing more.
- **Ownership** — `kuadrant-` prefix convention is sufficient per maleck13's comment; no need for get+label-check.

Minor: the `controller.Destruct` error at line 167-169 (update branch) still only logs+continues without adding to `reconcileErr`, but that's a pre-existing pattern for a non-API-server error and not introduced by this PR.

Agree with Didier that this needs product/architect sign-off on the upgrade path and documentation before shipping, but the code itself is ready.

internal/controller/istio_extension_reconciler.go
```diff
@@ -122,22 +125,30 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
...
+		}
+
 		existingEnvoyFilter := existingEnvoyFilterObj.(*controller.RuntimeObject).Object.(*istioclientgonetworkingv1alpha3.EnvoyFilter)
 
 		// delete
 		if utils.IsObjectTaggedToDelete(desiredEnvoyFilter) && !utils.IsObjectTaggedToDelete(existingEnvoyFilter) {
 			if err := resource.Delete(ctx, existingEnvoyFilter.GetName(), metav1.DeleteOptions{}); err != nil {
 				logger.Error(err, "failed to delete envoyfilter object", "gateway", gatewayKey.String(), "envoyfilter", fmt.Sprintf("%s/%s", existingEnvoyFilter.GetNamespace(), existingEnvoyFilter.GetName()))
-				// TODO: handle error
+				reconcileErr = errors.Join(reconcileErr, fmt.Errorf("failed to delete envoyfilter %s/%s: %w", existingEnvoyFilter.GetNamespace(), existingEnvoyFilter.GetName(), err))
```
guicassolato, line 151: Slightly worried about the proliferation of this pattern in the STOW reconciliation workflow. Failing the entire workflow because one task returned an error won't automatically trigger a retry as one coming from a controller-runtime background would expect.

I sincerely believe we should give #2043 some attention.

adam-cattermole: I recalled seeing that PR but that's my mistake for not investigating what it was doing before returning the error here. What would you advise to proceed here - a follow up removing the errors or leaving as is for the imminent patch release?

guicassolato: I guess the error handling is not required for the patch which is more about deleting the WasmPlugin leftovers, right? I would rollback https://github.com/Kuadrant/kuadrant-operator/pull/2053/changes/ed89ee94c75c45c7a8c7be9a9aae4a885b88d677.

adam-cattermole: That's right, luckily I kept the changes separate here so here's a follow up to revert that commit only https://github.com/Kuadrant/kuadrant-operator/pull/2067

PR #2061, opened 2026-06-23
fix: use non-blocking owner refs for Authorino and Limitador

### LINKED ISSUE #1578: Cannot set blockOwnerDeletion linked to Kuadrant CR
**Describe the bug**

Sometime, when reconciling a `Kuadrant` CR, the Operator fails to create the `Authorino` resource with the following error:

```
{"level":"error","ts":"2025-09-18T16:06:05Z","logger":"kuadrant-operator.AuthorinoReconciler","msg":"failed to create authorino resource","status":"error","error":"authorinos.operator.authorino.kuadrant.io \"authorino\" is forbidden: cannot set blockOwnerDeletion in this case because cannot find RESTMapping for APIVersion kuadrant.io/v1beta1 Kind Kuadrant: no matches for kind \"Kuadrant\" in version \"kuadrant.io/v1beta1\"","stacktrace":"github.com/kuadrant/kuadrant-operator/internal/controller.(*AuthorinoReconciler).Reconcile\n\t/remote-source/app/internal/controller/authorino_reconciler.go:106\ngithub.com/kuadrant/policy-machinery/controller.Subscription.Reconcile\n\t/remote-source/deps/gomod/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/subscriber.go:34\ngithub.com/kuadrant/policy-machinery/controller.(*Workflow).Run.func1\n\t/remote-source/deps/gomod/pkg/mod/github.com/kuadrant/policy-machinery@v0.6.4/controller/workflow.go:42\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/remote-so...
```

Restarting the Operator pod fixes the issue.

**To Reproduce**

N/A

The issue was observed in an automated OpenShift deployment that installs the Kuadrant Operator using OLM and creates a `Kuadrant` CR right away.

**Expected behavior**

The `Authorino` resource should be created without errors.

**Additional context**

- It is not clear if other resources attempted to create by the Operator out of reconciling a `Kuadrant` CR also fail, other than `Authorino`.

- According to the [documentation](https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/apis/meta/v1#OwnerReference.BlockOwnerDeletion), blockOwnerDeletion requires `delete` on the owner resource granted for the user. Currently, this is **not** the case for the Kuadrant Operator Service Account, which does not have permission to delete `Kuadrant` resources. However, it is not clear whether this is indeed the root cause of the bug, given that:
  1. the Operator sets `blockOwnerDeletion` on other kinds of resources linked to the `Kuadrant` CR (e.g.: `Limitador`) as well as other resources linked to kinds for which the Operator does not hold delete permission (e.g.: many different kinds owned by `Gateway`), and yet there's no report of same issue related to those kinds;
  2. Restarting the Operator fixes the issue even though nothing has changed in the permissions

Priyanshu-u07:
### Changes
- Set `BlockOwnerDeletion: false` on the owner references pointing at the Kuadrant CR in the Authorino and Limitador reconcilers. The GC-admission RESTMapping check only runs for blocking refs, so this avoids the startup race entirely.
- The owner reference is retained (`Controller: true`), so garbage collection of the children is unchanged, only foreground-deletion ordering (unused by the operator) is dropped.
- Added a regression test asserting both resources carry a non-blocking owner ref.

### Note
`PeerAuthentication` also sets a blocking owner ref to the Kuadrant CR (via `SetControllerReference`, default `true`), so it carries the same latent risk. Left unchanged here as it's created after startup (outside the race window).

Fixes #1578 

## Bug Fixes

* Fixed resource creation failures that could occur during Operator Lifecycle Manager (OLM) startup by adjusting resource ownership configurations.

## Tests

* Added integration test validating resource ownership behaviour in Gateway API scenarios.

### DIFF
--- a/internal/controller/authorino_reconciler.go
+++ b/internal/controller/authorino_reconciler.go
@@ -149,11 +149,13 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 			Namespace: kobj.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				{
-					APIVersion:         kobj.GroupVersionKind().GroupVersion().String(),
-					Kind:               kobj.GroupVersionKind().Kind,
-					Name:               kobj.Name,
-					UID:                kobj.UID,
-					BlockOwnerDeletion: ptr.To(true),
+					APIVersion: kobj.GroupVersionKind().GroupVersion().String(),
+					Kind:       kobj.GroupVersionKind().Kind,
+					Name:       kobj.Name,
+					UID:        kobj.UID,
+					// Intentionally false: a blocking owner ref to the Kuadrant CR
+					// races OLM startup and fails the Create.
+					BlockOwnerDeletion: ptr.To(false),
 					Controller:         ptr.To(true),
 				},
 			},
--- a/internal/controller/limitador_reconciler.go
+++ b/internal/controller/limitador_reconciler.go
@@ -71,11 +71,13 @@ func (r *LimitadorReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 			Namespace: kobj.Namespace,
 			OwnerReferences: []metav1.OwnerReference{
 				{
-					APIVersion:         kobj.GroupVersionKind().GroupVersion().String(),
-					Kind:               kobj.GroupVersionKind().Kind,
-					Name:               kobj.Name,
-					UID:                kobj.UID,
-					BlockOwnerDeletion: ptr.To(true),
+					APIVersion: kobj.GroupVersionKind().GroupVersion().String(),
+					Kind:       kobj.GroupVersionKind().Kind,
+					Name:       kobj.Name,
+					UID:        kobj.UID,
+					// Intentionally false: a blocking owner ref to the Kuadrant CR
+					// races OLM startup and fails the Create.
+					BlockOwnerDeletion: ptr.To(false),
 					Controller:         ptr.To(true),
 				},
 			},
--- a/tests/bare_k8s/kuadrant_controller_test.go
+++ b/tests/bare_k8s/kuadrant_controller_test.go
@@ -93,6 +93,27 @@ var _ = Describe("Kuadrant controller when Gateway API is missing", func() {
 			}).WithContext(ctx).Should(Succeed())
 		}, testTimeOut)
 
+		It("Kuadrant-owned resources use non-blocking owner references", func(ctx SpecContext) {
+			assertNonBlockingKuadrantOwnerRef := func(g Gomega, owner []metav1.OwnerReference) {
+				g.Expect(owner).To(ContainElement(SatisfyAll(
+					HaveField("Kind", "Kuadrant"),
+					HaveField("Name", kuadrantCR.Name),
+					HaveField("Controller", ptr.To(true)),
+					HaveField("BlockOwnerDeletion", ptr.To(false)),
+				)))
+			}
+
+			Eventually(func(g Gomega) {
+				limitador := &limitadorv1alpha1.Limitador{}
+				g.Expect(testClient().Get(ctx, client.ObjectKey{Name: kuadrant.LimitadorName, Namespace: testNamespace}, limitador)).To(Succeed())
+				assertNonBlockingKuadrantOwnerRef(g, limitador.GetOwnerReferences())
+
+				authorino := &authorinoopapi.Authorino{}
+				g.Expect(testClient().Get(ctx, client.ObjectKey{Name: "authorino", Namespace: testNamespace}, authorino)).To(Succeed())
+				assertNonBlockingKuadrantOwnerRef(g, authorino.GetOwnerReferences())
+			}).WithContext(ctx).Should(Succeed())
+		}, testTimeOut)
+
 		It("Limitador CR should retain user fields and restore default", func(ctx SpecContext) {
 			By("Patching Limitador CR with user fields")
 			limitador := &limitadorv1alpha1.Limitador{


eguzki:
I think you are fixing the surface symptom and only preventing the effect from happening. There is an underlying issue which I do not fully understand. Why is the Kubernetes API server refusing to accept the owner reference? What does it mean that "cannot find RESTMapping for APIVersion kuadrant.io/v1beta1 Kind Kuadrant"? At the time Limitador/Authorino are created, the Kuadrant object has already been read from the Kubernetes API server, so kuadrant.io/v1beta1 Kind Kuadrant should already be registered at that point. What is failing? This is the primary question I want answered, as it would lead us to the proper fix. In the meantime, a quick fix would be to retry. I do not think that losing foreground cascading deletion is a worthwhile trade-off.

guicassolato:
> I think you are fixing the surface symptom and only preventing the effect from happening. There is an underlying issue which I do not fully understand. Why is the Kubernetes API server refusing to accept the owner reference? What does it mean that "cannot find RESTMapping for APIVersion kuadrant.io/v1beta1 Kind Kuadrant"? At the time Limitador/Authorino are created, the Kuadrant object has already been read from the Kubernetes API server, so kuadrant.io/v1beta1 Kind Kuadrant should already be registered at that point. What is failing? This is the primary question I want answered, as it would lead us to the proper fix. In the meantime, a quick fix would be to retry. I do not think that losing foreground cascading deletion is a worthwhile trade-off.

+1

For retrying, it would be useful to have #2043 in.

Priyanshu-u07:
Thank you @eguzki and @guicassolato  for the feedback! Here's my analysis:

>Why is the Kubernetes API server refusing to accept the owner reference?

The Kubernetes API server refuses the owner reference because on a fresh install, the gc admission plugin can't resolve the owner kind. The create carries an owner reference with `blockOwnerDeletion: true` so it goes through the apiserver's `OwnerReferencesPermissionEnforcement` (gc) admission plugin. To authorize that flag, the plugin must first resolve the owner kind (`Kuadrant`) to its resource via a discovery-backed RESTMapper internal to the plugin, before it can authorize the flag against the owner. On a fresh install that mapper hasn't caught up to the just-installed Kuadrant CRD, so `RESTMappings(Kuadrant)` returns "no matches for kind" and the create is rejected at the mapping step. Because it fails at the mapping step before the owner-authorization check, granting delete/finalizer RBAC on kuadrants wouldn't help either.

> What does it mean that "cannot find RESTMapping for APIVersion kuadrant.io/v1beta1 Kind Kuadrant"?

 RESTMapping is the apiserver's translation from a kind (`kuadrant.io/v1beta1`, `Kind=Kuadrant`) to the actual REST resource it's served at (`kuadrants.kuadrant.io`) along with its scope. This table is built from discovery. "cannot find RESTMapping ... no matches for kind" means that table has no entry for the Kuadrant kind at that moment i.e., the gc-admission plugin's RESTMapper can't translate `kuadrant.io/v1beta1 Kind=Kuadrant `into its resource and scope, so it can't build the authorization check for the `blockOwnerDeletion` flag. The entry is missing not because the CRD is absent (the resource is already servable as our reads succeed) but because this mapper hasn't refreshed from discovery to include the freshly installed Kuadrant CRD yet.

> At the time Limitador/Authorino are created, the Kuadrant object has already been read from the Kubernetes API server, so kuadrant.io/v1beta1 Kind Kuadrant should already be registered at that point. What is failing? 

The question assumes there's a single RESTMapper, but two different subsystems inside the apiserver are involved. Our read is served by the apiserver's request-serving path (the apiextensions CRD handler) which goes live the moment the Kuadrant CRD is Established that's why the read succeeds. The failure happens in a separate, discovery-backed RESTMapper cache used only by the `OwnerReferencesPermissionEnforcement` admission plugin, which refreshes independently and lags a freshly installed CRD. So "the apiserver knows the kind" is true for its serving path but not yet for its admission mapper. The apiserver returning the Kuadrant CR tells us nothing about whether that admission cache has caught up and that admission cache is what fails. This also explains why it's intermittent (the create only fails in the up-to-30s window before that mapper's next reset; in an HA control plane it's further amplified, since each apiserver replica runs its own mapper and warms independently) and why a restart "fixes" it (it just delays the retry until the cache refreshes).

This looks like a good fit for a retry-based approach that keeps `blockOwnerDeletion: true`, recording the transient create error and letting it requeue with backoff on top of #2043, by extending the `ErrorRegistry` pattern that #2043 introduces (and applies to `AuthConfigsReconciler`/`LimitadorLimitsReconciler`) to the `Authorino`/`Limitador` reconcilers.

Would love to hear if you have a different approach in mind.

eguzki:
The two subsystems might explain what we see, but does not explain why it's happening _now_. Something else might have changed, maybe the kubernetes API Server itself, and it's happening since we upgraded testing kubernetes cluster version. who knows :shrug:. Currently running v1.35.0 since January 14, 2025.

> This looks like a good fit for a retry-based approach that keeps blockOwnerDeletion: true, recording the transient create error and letting it requeue with backoff on top of https://github.com/Kuadrant/kuadrant-operator/pull/2043, by extending the ErrorRegistry pattern that https://github.com/Kuadrant/kuadrant-operator/pull/2043 introduces (and applies to AuthConfigsReconciler/LimitadorLimitsReconciler) to the Authorino/Limitador reconcilers.

Looks good to me. Let's close this PR then.

Priyanshu-u07:
Closing this in favor of the retry-based fix. Now that #2043 is merged, I'll open a PR that keeps `blockOwnerDeletion: true` and records the transient create error by the `ErrorRegistry` so it requeues with backoff.

On the "why now" question, I'll look into that separately, but it doesn't block the fix since the retry approach is robust regardless of the root timing cause.

PR #2062, opened 2026-06-24
feat: add per-subscriber counters to PlanPolicy

### LINKED ISSUE #2047: PlanPolicy: support per-subscriber rate limit counters
## Summary

`PlanPolicy` currently treats all subscribers on the same plan tier as a single group, meaning they share one rate limit counter. If the **gold** plan allows 5 requests per day, that quota is consumed collectively by everyone holding a gold-annotated Secret, not individually per subscriber.

This behaviour is counterintuitive: the word _plan_ and _subscription_ imply that each subscriber gets their own independent quota.

## Background

`PlanPolicy` is a meta policy — a higher-level abstraction that composes a `RateLimitPolicy` and requires an `AuthPolicy` to enforce plan-based rate limiting. Unlike a plain `RateLimitPolicy`, it integrates with the control plane, which patches the `AuthConfig` to map requests to plan IDs.

When `PlanPolicy` was first implemented, the design chose a single bucket/tier counter: all subscribers on the same plan share one counter. There was no requirement to identify individual subscribers at the rate-limiting layer. The plan tier is resolved from a Kubernetes Secret annotation (`auth.kuadrant.plan`) and used as the `When` predicate in the generated `RateLimitPolicy`.

`ToRateLimits()` in `cmd/extensions/plan-policy/api/v1alpha1/planpolicy_types.go` generates a `Limit` for each tier with only a `When` predicate and `Rates`, but **no `Counters`**:

```go
// planpolicy_types.go:65
func (p *PlanPolicy) ToRateLimits() map[string]kuadrantv1.Limit {
    return utils.Associate(p.Spec.Plans, func(plan Plan) (string, kuadrantv1.Limit) {
        return plan.Tier, kuadrantv1.Limit{
            When:  kuadrantv1.NewWhenPredicates(fmt.Sprintf(`auth.kuadrant.plan == "%s"`, plan.Tier)),
            Rates: plan.Limits.ToRates(),
            // No Counters → single shared counter per tier
        }
    })
}
```

Because `RateLimitPolicy` `Limit.Counters` (`api/v1/ratelimitpolicy_types.go:228`) is empty, Limitador tracks one counter per tier, shared by all subscribers on that tier.

## Problem

The shared-counter model conflicts with user expectations. When a user subscribes to a plan, they expect their quota to be independent of other subscribers on the same tier. Shared counters mean one heavy user can exhaust the quota for all other gold-plan subscribers.

The current behaviour is also not clearly documented, which causes confusion when users or operators review the generated `RateLimitPolicy`.

zyzzmohit:
## Motivation
This PR addresses issue #2047 where `PlanPolicy` currently generates `RateLimitPolicy` limits with no `Counters`. This causes all subscribers on a specific plan tier to share a single rate-limit bucket, which conflicts with user expectations for per-subscriber rate limiting.

## Changes
- **API Extension**: Added a `Counters` field to the `Plan` struct in `PlanPolicySpec` using the standard `kuadrantv1.Counter` type.
- **Generator Update**: Updated `ToRateLimits()` in `planpolicy_types.go` to correctly map `plan.Counters` into the generated `kuadrantv1.Limit`.
- **CRD Generation**: Ran `make generate` and `make manifests` to update the OpenAPI schema and CRDs to reflect the new `Counters` property.

Fixes #2047


eguzki:
I am not entirely sure the best option is to add `counter` field to the `Plan` object. What do you set in that field? 

The planpolicy is an abstraction layer. The user needs to map plan tiers from auth based data. I was thinking on doing the same for the user id. So instead of exposing raw counters, expose mapping function to user id. What other use case has the counters in the context of the plan policy? 

This requires some thiking and design first. Feel free to make your proposal in the source issue https://github.com/Kuadrant/kuadrant-operator/issues/2047 or RFC in https://github.com/Kuadrant/architecture

maksymvavilov:

eguzki:
I think we can close this. This requires RFC first

PR #2064, opened 2026-06-24
fix: return created object instead of nil in ReconcileResource

### LINKED ISSUE #2046: AuthPolicy/RateLimitPolicy changes cause EnvoyFilter update storm, dropping TLS on all gateway routes
**Describe the bug**
Applying an AuthPolicy or RateLimitPolicy (via PlanPolicy) to an HTTPRoute causes 
all routes on the shared gateway to become unreachable with SSL_ERROR_SYSCALL for 
several seconds. The outage affects ALL routes on the gateway, not just the one 
being configured.

Additionally, a nil pointer panic occurs in the plan-policy-extension-controller 
when a PlanPolicy CREATE races with RateLimitPolicy creation.

**To Reproduce**
1. Have a stable gateway with TLSPolicy enforced and multiple HTTPRoutes attached
2. Have at least one existing AuthPolicy + RateLimitPolicy (PlanPolicy) already enforced
3. Deploy a new HTTPRoute with no auth (verify it works)
4. Apply a new AuthPolicy + PlanPolicy (RateLimitPolicy) to that HTTPRoute via GitOps/ArgoCD
5. Observe all existing routes (including unrelated ones) drop TLS connections during reconciliation

**Expected behavior**
Applying an AuthPolicy or RateLimitPolicy to one HTTPRoute should not affect 
traffic on other routes. The `kuadrant-prod-gateway` EnvoyFilter should be 
updated atomically or with a mechanism that prevents Istio from reloading 
mid-storm. Existing routes should continue serving TLS throughout.

**Observed behavior**
- The `kuadrant-prod-gateway` EnvoyFilter generation jumped from ~135 to ~1426 
  within a few seconds as multiple reconcilers (IstioExtensionReconciler, 
  plan-policy-extension-controller) raced to update it concurrently
- Each update triggers an Istio WASM filter reload, dropping TLS connections
- Hundreds of optimistic locking conflicts logged:
  `"failed to update envoyfilter object" error="Operation cannot be fulfilled on 
  envoyfilters.networking.istio.io \"kuadrant-prod-gateway\": the object has been 
  modified; please apply your changes to the latest version and try again"`
- curl to any route returns: `SSL_ERROR_SYSCALL` during the storm window

**Bug 2: Nil pointer panic in plan-policy-extension-controller**
On PlanPolicy CREATE, the controller panics if the underlying RateLimitPolicy 
hasn't been fully populated in the cache yet:
panic: interface conversion: client.Object is nil, not *v1.RateLimitPolicy

at planpolicy_reconciler.go:97
The controller recovers and retries, but contributes to the reconciliation storm.

**Environment**
- Kuadrant operator (Red Hat Connectivity Link)
- Gateway: Istio via gateway.networking.k8s.io/v1
- TLSPolicy: Let's Encrypt via cert-manager
- Deployment: OpenShift + ArgoCD GitOps

zyzzmohit:
## Motivation
This PR addresses **Bug 2** described in issue #2046: *Nil pointer panic in plan-policy-extension-controller*.

During a `PlanPolicy` creation (or an `OIDCPolicy` creation), the controller delegates to `BaseReconciler.ReconcileResource` to apply the newly generated underlying policies to the cluster. Previously, upon successfully creating a new object via the Kubernetes client, `ReconcileResource` returned a `nil` object instead of returning the newly created object.

This resulted in a nil pointer panic in downstream reconcilers when they attempted to inspect the `Status.Conditions` of the returned `nil` interface to determine if the object was enforced.

## Changes
- **Core Update**: Modified `internal/reconcilers/base_reconciler.go`'s `ReconcileResource` to return the `desired` object upon successful creation instead of `nil`.
- **Impact**: Instead of panicking, the calling reconcilers now safely receive the newly created object. They correctly identify that it lacks an `Enforced` status, return an error, and gracefully requeue the reconciliation until the policy is enforced.

Fixes #2046

### DIFF
--- a/internal/reconcilers/base_reconciler.go
+++ b/internal/reconcilers/base_reconciler.go
@@ -125,7 +125,10 @@ func (b *BaseReconciler) ReconcileResource(ctx context.Context, obj, desired cli
 
 		// Not found
 		if !utils.IsObjectTaggedToDelete(desired) {
-			return nil, b.CreateResource(ctx, desired)
+			if err := b.CreateResource(ctx, desired); err != nil {
+				return nil, err
+			}
+			return desired, nil
 		}
 
 		// Marked for deletion and not found. Nothing to do.


thomasmaas (commented):
Good fix — the change is correct and `client.Create` populates the passed object with server-set fields, so returning `desired` gives callers a usable object.

Two small requests:

1. **Update the docstring** (line 117): it currently says "It returns the object applied to the cluster in the case of updates or an error." — should be "in the case of creates or updates" since the behavior has changed.

2. **Add a regression test**: the existing tests in `base_reconciler_test.go` all discard the returned object with `_, err`. A test that asserts the create path returns a non-nil object would prevent this from regressing. Something like:

```go
obj, err := baseReconciler.ReconcileResource(ctx, &v1.ConfigMap{}, desiredConfigmap, CreateOnlyMutator)
if err != nil {
    t.Fatal(err)
}
if obj == nil {
    t.Fatal("expected non-nil object on create, got nil")
}
```

maksymvavilov:

PR #2068, opened 2026-06-26
feat: add Prometheus metrics for reconciliation performance

maksymvavilov:
## Summary

Prometheus histograms and gauges for reconciliation performance observability. These metrics are the interface for QE scale tests — QE scrapes them via kube-burner and displays in Grafana dashboards.

### Histograms

| Metric | Labels | What it measures |
|--------|--------|-----------------|
| `kuadrant_reconciliation_duration_seconds` | `workflow` (init, dns, tls, data_plane, observability, finalize) | End-to-end duration of each workflow phase |
| `kuadrant_effective_policy_duration_seconds` | `policy_type` (auth, ratelimit, tokenratelimit) | Time to calculate effective policies |
| `kuadrant_topology_rebuild_duration_seconds` | — | Time for topology reconciliation (ToDot + ConfigMap) |
| `kuadrant_authconfig_generation_duration_seconds` | — | Time for AuthConfig reconciliation |
| `kuadrant_limitador_limits_generation_duration_seconds` | — | Time for Limitador limits reconciliation |

### Gauges

| Metric | Labels | What it measures |
|--------|--------|-----------------|
| `kuadrant_topology_objects` | `kind` | Number of objects in the topology DAG by kind |
| `kuadrant_authconfigs_generated` | — | Number of AuthConfigs generated per cycle |

## Verification

```bash
make local-env-setup && make run
curl localhost:8080/metrics | grep kuadrant_reconciliation
curl localhost:8080/metrics | grep kuadrant_effective_policy
curl localhost:8080/metrics | grep kuadrant_topology
```

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics for reconciliation phase durations and effective policy processing.
  * Added visibility into topology rebuilds, generated AuthConfigs, topology object counts, and limit generation times.
  * Added workflow-level timing for key reconciliation stages.

* **Tests**
  * Added coverage validating metric observations, gauges, resets, and duration buckets.

### DIFF
--- a/internal/controller/state_of_the_world.go
+++ b/internal/controller/state_of_the_world.go
@@ -726,15 +726,15 @@ func additionalMainTraceAttributes(resourceEvents []controller.ResourceEvent, _ 
 
 func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
 	mainWorkflow := &controller.Workflow{
-		Precondition: metricsReconcileFunc("init", traceReconcileFunc("workflow.init", initWorkflow(b.client).Run)),
+		Precondition: metricsReconcileFunc("workflow.init", traceReconcileFunc("workflow.init", initWorkflow(b.client).Run)),
 		Tasks: []controller.ReconcileFunc{
-			metricsReconcileFunc("dns", traceReconcileFunc("workflow.dns", NewDNSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isDNSOperatorInstalled).Run)),
-			metricsReconcileFunc("tls", traceReconcileFunc("workflow.tls", NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run)),
-			metricsReconcileFunc("data_plane", traceReconcileFunc("workflow.data_plane_policies", NewDataPlanePoliciesWorkflow(b.manager, b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run)),
-			metricsReconcileFunc("observability", traceReconcileFunc("workflow.observability", NewObservabilityReconciler(b.client, b.manager, operatorNamespace).Subscription().Reconcile)),
-			traceReconcileFunc("workflow.developer_portal", NewDeveloperPortalReconciler(b.manager).Subscription().Reconcile),
+			metricsReconcileFunc("workflow.dns", traceReconcileFunc("workflow.dns", NewDNSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isDNSOperatorInstalled).Run)),
+			metricsReconcileFunc("workflow.tls", traceReconcileFunc("workflow.tls", NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run)),
+			metricsReconcileFunc("workflow.data_plane_policies", traceReconcileFunc("workflow.data_plane_policies", NewDataPlanePoliciesWorkflow(b.manager, b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run)),
+			metricsReconcileFunc("workflow.observability", traceReconcileFunc("workflow.observability", NewObservabilityReconciler(b.client, b.manager, operatorNamespace).Subscription().Reconcile)),
+			metricsReconcileFunc("workflow.developer_portal", traceReconcileFunc("workflow.developer_portal", NewDeveloperPortalReconciler(b.manager).Subscription().Reconcile)),
 		},
-		Postcondition: metricsReconcileFunc("finalize", traceReconcileFunc("workflow.finalize", b.finalStepsWorkflow().Run)),
+		Postcondition: metricsReconcileFunc("workflow.finalize", traceReconcileFunc("workflow.finalize", b.finalStepsWorkflow().Run)),
 	}
 
 	if b.isConsolePluginInstalled && b.isClusterVersionInstalled {
--- a/internal/controller/state_of_the_world.go
+++ b/internal/controller/state_of_the_world.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"sort"
 	"sync"
+	"time"
 
 	istiosecurity "istio.io/client-go/pkg/apis/security/v1"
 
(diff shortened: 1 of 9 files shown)


internal/metrics/reconciliation.go
```diff
@@ -0,0 +1,100 @@
...
+	authconfigGenerationDuration.Observe(time.Since(start).Seconds())
+}
+
+func ObserveLimitadorLimitsGenerationDuration(start time.Time) {
+	limitadorLimitsGenerationDuration.Observe(time.Since(start).Seconds())
+}
+
+func SetTopologyObjectsTotal(kind string, count int) {
+	topologyObjectsTotal.WithLabelValues(kind).Set(float64(count))
+}
```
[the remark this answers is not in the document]
maksymvavilov: Fixed in 8c1580d — added `ResetTopologyObjects()` in `internal/metrics/reconciliation.go` and calling it from the topology reconciler before the `objectsByKind` loop to clear stale kinds.

internal/controller/state_of_the_world.go
```diff
@@ -717,15 +726,15 @@ func additionalMainTraceAttributes(resourceEvents []controller.ResourceEvent, _
...
-			traceReconcileFunc("workflow.observability", NewObservabilityReconciler(b.client, b.manager, operatorNamespace).Subscription().Reconcile),
+			metricsReconcileFunc("dns", traceReconcileFunc("workflow.dns", NewDNSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isDNSOperatorInstalled).Run)),
+			metricsReconcileFunc("tls", traceReconcileFunc("workflow.tls", NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run)),
+			metricsReconcileFunc("data_plane", traceReconcileFunc("workflow.data_plane_policies", NewDataPlanePoliciesWorkflow(b.manager, b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run)),
+			metricsReconcileFunc("observability", traceReconcileFunc("workflow.observability", NewObservabilityReconciler(b.client, b.manager, operatorNamespace).Subscription().Reconcile)),
 			traceReconcileFunc("workflow.developer_portal", NewDeveloperPortalReconciler(b.manager).Subscription().Reconcile),
```
guicassolato: Dev Portal reconciler not wrapped in `metricsReconcileFunc`. Intentional?

internal/metrics/reconciliation.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	reconciliationDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "kuadrant_reconciliation_duration_seconds",
+			Help:    "Duration of each top-level workflow phase in seconds",
+			Buckets: prometheus.DefBuckets,
```
guicassolato: I think the default histogram buckets are `{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10}` (in seconds). However, we've had reports of reconciliation workflows taking up to several minutes to finish. Do we risk all observations landing in the `+Inf` bucket and lose distribution visibility in exactly the scenario these metrics exist to diagnose?

internal/metrics/reconciliation.go
```diff
@@ -0,0 +1,104 @@
...
+	topologyObjects.Reset()
+}
+
+func SetTopologyObjects(kind string, count int) {
+	topologyObjects.WithLabelValues(kind).Set(float64(count))
+}
+
+func SetAuthconfigsGenerated(count int) {
+	authconfigsGenerated.Set(float64(count))
+}
```
guicassolato, line 106: Maybe add unit tests for these new metrics? (Similar to [`internal/metrics/operator_health_test.go`](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/metrics/operator_health_test.go))

internal/controller/state_of_the_world.go
```diff
@@ -717,15 +726,15 @@ func additionalMainTraceAttributes(resourceEvents []controller.ResourceEvent, _
...
 func (b *BootOptionsBuilder) Reconciler() controller.ReconcileFunc {
 	mainWorkflow := &controller.Workflow{
-		Precondition: traceReconcileFunc("workflow.init", initWorkflow(b.client).Run),
+		Precondition: metricsReconcileFunc("init", traceReconcileFunc("workflow.init", initWorkflow(b.client).Run)),
 		Tasks: []controller.ReconcileFunc{
-			traceReconcileFunc("workflow.dns", NewDNSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isDNSOperatorInstalled).Run),
-			traceReconcileFunc("workflow.tls", NewTLSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isCertManagerInstalled).Run),
-			traceReconcileFunc("workflow.data_plane_policies", NewDataPlanePoliciesWorkflow(b.manager, b.client, b.isGatewayAPIInstalled, b.isIstioInstalled, b.isEnvoyGatewayInstalled, b.isLimitadorOperatorInstalled, b.isAuthorinoOperatorInstalled).Run),
-			traceReconcileFunc("workflow.observability", NewObservabilityReconciler(b.client, b.manager, operatorNamespace).Subscription().Reconcile),
+			metricsReconcileFunc("dns", traceReconcileFunc("workflow.dns", NewDNSWorkflow(b.client, b.manager.GetScheme(), b.isGatewayAPIInstalled, b.isDNSOperatorInstalled).Run)),
```
guicassolato: nit: while the trace spans are named `workflow.<sth>`, the metrics are labeled just `<sth>`. Not a problem, but I wonder if we don't want to use the same value at both. Keeping them consistent could help avoid to confuse one trying to correlate metrics with traces.

internal/controller/state_of_the_world.go
maksymvavilov: It was intentional, but at the second tohught consistency worth more. I'll wrap it

internal/metrics/reconciliation.go
maksymvavilov: The slowest I observed was 3.8 sec with 3 gateways and 16 listeners on each. The approximation test that run about what we expect from the scale test had reconcile loops of ~7 sec. 
The bottleneck, afaik, is in Authorino. 
My concern is an increase in cardinality if we add more, but we could have two more: 30 and 60 sec. 
And then we observe and adjust buckets accordingly, wdyt?

internal/metrics/reconciliation.go
maksymvavilov: good point, added

internal/controller/state_of_the_world.go
maksymvavilov: alligned them

maksymvavilov:

PR #2089, opened 2026-07-07
Migrate WASM config to new types

### LINKED ISSUE #1995: Convert rate limiting policy reconciler to use new WASM plugin format
Now that we have a new WASM plugin format, we should rewrite the Rate Limiting Reconciler to output the WASM plugin in that format.

### LINKED ISSUE #1996: Convert Auth policy reconciler to use new WASM plugin format
just as #1995 .

adam-cattermole:
Migration from the original legacy wasm config to the newer declarative style. The one major caveat in this approach is the more verbose nature of the config and much larger `EnvoyFilter` as a result. I propose that we see how this performs, and if it's causing issues we could look into alternative methods of sourcing the config for the wasm-shim

## Summary by CodeRabbit

* **New Features**
  * Introduced a specification-driven WASM action pipeline for auth, rate limiting, token rate limiting, headers, storage, denial, and failures.
  * Added deterministic policy merging with improved CEL rendering, including support for nested reply actions.
  * Added request-binding collection to improve how generated configs wire policy inputs.

* **Bug Fixes**
  * CEL validation now evaluates configured predicates end-to-end and rejects internal predicates.
  * Improved detection and messaging for conditional-data key conflicts during merging.

* **Tests**
  * Updated unit and integration tests to align with the new spec-based action model.


internal/cel/kuadrant_validator.go
```diff
@@ -103,14 +105,19 @@ func NewRootValidatorBuilder() *ValidatorBuilder {
...
 
-func ValidateWasmAction(action wasm.Action, validator *Validator) error {
-	pol := policyKindFromWasmServiceName(action.ServiceName)
-	for _, predicate := range action.Predicates {
+const internalPredicatePrefix = "kuadrant.internal."
+
+func ValidateWasmActionSpec(spec wasm.ActionSpec, validator *Validator) error {
+	pol := policyKindFromWasmServiceName(spec.ServiceName)
+	for _, predicate := range spec.Predicates {
+		if strings.HasPrefix(predicate, internalPredicatePrefix) {
```
eguzki: Then, users using CEL expressions starting with `kuadrant.internal.` would not be validated. Is that a real use case? Why skipping validation when CEL starts with `kuadrant.internal.` ?

eguzki (dismissed):
LGTM

Tested locally with some authenticated rate limiting and token rate limiting.

internal/cel/kuadrant_validator.go
adam-cattermole: Yeah that's a good question. I think you're right - ideally we'd actually prevent users from authoring expressions that reference `kuadrant.internal.` whereas this ignores them in the validation step if they're in predicates. I'm thinking about two options to fix this:

- We validate first then inject our internal predicates afterward
- Store the internal separately and validate only user defined

Leaning toward the first, will add a commit to remove this check so that `kuadrant.` is an invalid root binding again

eguzki (approved):

maksymvavilov:

PR #2091, opened 2026-07-07
Propagate OpenShift TLS security profile to Authorino

philbrookes:
## Summary
- Reads the cluster's `APIServer` CR (`apiserver.config.openshift.io/cluster`) and resolves its TLS security profile
- Propagates `tlsMinVersion` and `tlsCipherSuites` to the Authorino CR
- On non-OpenShift clusters (no APIServer CRD), falls back to Intermediate profile defaults
- Translates OpenSSL cipher names to IANA names; DHE ciphers silently skipped (unsupported by Go)
- Includes RBAC for `apiservers` (get/list/watch) and test script covering 9 scenarios

## Context
Part of OCP5 TLS profile compliance. Authorino is the only Kuadrant component serving TLS. This change ensures its TLS configuration matches the cluster-wide TLS security profile.

The operator looks up the `APIServer` CR named `cluster` — this follows the OpenShift convention where all `config.openshift.io` resources are singletons with a canonical instance name of "cluster". See:
- [OpenShift API Conventions](https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md) — configuration APIs are "typically cluster-scoped, a singleton within the cluster"
- [APIServer config.openshift.io/v1 Reference](https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html/config_apis/apiserver-config-openshift-io-v1) — "The canonical name of an instance is 'cluster'"

Depends on: https://github.com/Kuadrant/authorino-operator/pull/335

## Test plan
- [ ] Unit tests for `ResolveTLSProfile()` (all profile types, nil input, OpenSSL→IANA mapping)
- [ ] No CRD → Intermediate defaults
- [ ] CRD exists, no CR → Intermediate defaults
- [ ] Intermediate, Modern, Old, Custom profiles
- [ ] Profile change triggers reconciliation
- [ ] CR deletion falls back to Intermediate
- [ ] Multiple CRs — only "cluster" is used

## Summary by CodeRabbit

- **New Features**
  - Added detection and monitoring of the OpenShift `APIServer` resource when available.
  - Authorino TLS settings now use the configured OpenShift TLS security profile, including support for custom profiles and cipher-suite translation.

- **Bug Fixes**
  - Improved TLS updates to preserve existing certificate secret and enablement settings while applying the new minimum TLS version and cipher suites.
  - Refined OIDC cookie query handling to avoid adding empty query strings.

- **Compatibility**
  - Updated RBAC permissions and installation manifests to allow access to OpenShift `APIServer` resources.


internal/controller/authorino_reconciler.go
```diff
@@ -213,3 +227,20 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
 	return nil
 }
+
+func (r *AuthorinoReconciler) resolveTLSProfileFromTopology(topology *machinery.Topology) (string, []string) {
+	if !r.IsAPIServerInstalled {
+		return openshift.ResolveTLSProfile(nil)
+	}
+
+	apiServers := topology.Objects().Items(func(object machinery.Object) bool {
+		return object.GetName() == "cluster" && object.GroupVersionKind().GroupKind() == openshift.APIServerGVK.GroupKind()
```
guicassolato: Should we make the object name (`cluster`) configurable via env var?

internal/controller/authorino_reconciler.go
```diff
@@ -213,3 +227,20 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 
 	return nil
 }
+
+func (r *AuthorinoReconciler) resolveTLSProfileFromTopology(topology *machinery.Topology) (string, []string) {
+	if !r.IsAPIServerInstalled {
+		return openshift.ResolveTLSProfile(nil)
+	}
+
+	apiServers := topology.Objects().Items(func(object machinery.Object) bool {
```
guicassolato: This is fine, but if you want to reduce computation you can define a linking function for the `APIServer` objects and the `Kuadrant` CR ([example](https://github.com/Kuadrant/kuadrant-operator/blob/f601936bb8a45f35517a5ba788aac81d55b9f633/api/v1beta1/topology.go#L30)), and then navigate the CR's relationships. For large topology graphs, the performance improvement can be considerable.

internal/controller/state_of_the_world.go
```diff
@@ -228,6 +228,7 @@ type BootOptionsBuilder struct {
 	isLimitadorOperatorInstalled  bool
 	isAuthorinoOperatorInstalled  bool
 	isPrometheusOperatorInstalled bool
+	isAPIServerInstalled          bool
```
guicassolato: nit:

I reckon "API Server" here is in reference to OpenShift's [`APIServer`](https://github.com/openshift/api/blob/8bc26b0fcc2d981007c4c725a73df8ed05578b33/config/v1/types_apiserver.go#L11-L35) kind. I worry that first-time readers of the code may be slightly thrown off by our name choice though, think this is some kind of flag about presence of _the_ API server itself – I know, absurd as it sounds.

How would you see choosing a different name for the field? E.g.: `isOpenShiftServerConfigInstalled`.

Same for other all places where the name is used.

philbrookes: that's fair. I'll update it.

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler_test.go
```diff
@@ -333,9 +333,9 @@ func TestBuildTargetCookieExpression_Examples(t *testing.T) {
 
 	for _, ex := range examples {
 		t.Run(ex.scenario, func(t *testing.T) {
-			// The CEL expression uses a ternary: request.url_path + (has(request.query) ? "?" + request.query : "")
+			// The CEL expression uses a ternary: request.url_path + (has(request.query) && request.query != "" ? "?" + request.query : "")
 			// This should construct the full path with query when query is present
-			if !strings.Contains(expression, `request.url_path + (has(request.query) ? "?" + request.query : "")`) {
+			if !strings.Contains(expression, `request.url_path + (has(request.query) && request.query != "" ? "?" + request.query : "")`) {
```
maksymvavilov: This is a "pre-existing" part, but does this test check for nothing? And the comment seems useless.

internal/controller/authorino_reconciler.go
```diff
@@ -224,3 +249,34 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
+	for _, child := range children {
+		if child.GroupVersionKind().GroupKind() == openshift.APIServerGroupKind && child.GetName() == targetName {
+			rObj, ok := child.(*controller.RuntimeObject)
+			if !ok {
+				continue
+			}
+			apiServer, ok := rObj.Object.(*configv1.APIServer)
+			if !ok {
+				continue
+			}
```
maksymvavilov: Why do we have two casts/checks?

internal/openshift/tls_profile.go
```diff
@@ -0,0 +1,120 @@
...
+		spec = configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
+	} else if profile.Type == configv1.TLSProfileCustomType && profile.Custom != nil {
+		spec = &profile.Custom.TLSProfileSpec
+	} else {
+		spec = configv1.TLSProfiles[profile.Type]
+	}
+
+	if spec == nil {
+		spec = configv1.TLSProfiles[configv1.TLSProfileIntermediateType]
+	}
```
maksymvavilov: This defines what the default profile is in two places: the first if block and this `nil` check. It should work the same if we remove the first `if` and would leave us with the single "default" source of truth 🤔

internal/openshift/tls_profile_test.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+
+	t.Run("Intermediate profile excludes DHE ciphers", func(t *testing.T) {
+		_, ciphers := ResolveTLSProfile(nil)
+		for _, c := range ciphers {
+			if c == "DHE-RSA-AES128-GCM-SHA256" || c == "DHE-RSA-AES256-GCM-SHA384" {
+				t.Errorf("DHE cipher should not appear in translated output: %s", c)
+			}
+		}
+	})
```
maksymvavilov, line 135: If someone were to add DHE ciphers to the `opensslToIANA` map this will not catch it. We should be looking for `"TLS_DHE_*"` or matching the count to what we expect

internal/controller/authorino_reconciler.go
```diff
@@ -224,3 +249,34 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 
 	return nil
 }
+
+func apiServerCRName() string {
+	if name := os.Getenv("APISERVER_CR_NAME"); name != "" {
```
maksymvavilov: How will people know that this envar exists? 
If there is any documentation that we expect to produce, we would benefit from including this envar in there.

internal/controller/authorino_reconciler.go
```diff
@@ -224,3 +249,34 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
+		return name
+	}
+	return openshift.DefaultAPIServerCRName
+}
```
guicassolato: Wondering if this func belongs in the `openshift` package too 🤔

internal/openshift/tls_profile.go
```diff
@@ -0,0 +1,120 @@
...
+
+	minVersion = openshiftVersionToShort(spec.MinTLSVersion)
+
+	for _, opensslName := range spec.Ciphers {
+		if ianaName, ok := opensslToIANA[opensslName]; ok {
+			cipherSuites = append(cipherSuites, ianaName)
+		}
+	}
+
+	return minVersion, cipherSuites
```
maksymvavilov, line 162: There is a mention of a test script in the description and a few test scenarios. Are those manual testing steps?
We could have an integration test case that would register a new CRD/CR and verify this logic. Should we have it?

internal/openshift/tls_profile.go
```diff
@@ -0,0 +1,120 @@
...
+	APIServerGroupKind = APIServerGVK.GroupKind()
+	APIServersResource = configv1.SchemeGroupVersion.WithResource("apiservers")
+)
+
+func IsOpenShiftServerConfigInstalled(restMapper meta.RESTMapper) (bool, error) {
+	return utils.IsCRDInstalled(restMapper, APIServerGVK.Group, APIServerGVK.Kind, APIServerGVK.Version)
+}
+
+func LinkKuadrantToAPIServer(objs controller.Store) machinery.LinkFunc {
+	kuadrants := lo.Map(objs.FilterByGroupKind(kuadrantv1beta1.KuadrantGroupKind), controller.ObjectAs[machinery.Object])
```
guicassolato, line 44: Since the `cluster` APIServer CR is a cluster-scoped singleton, I wonder if the link function couldn't be filtering it by name already. I.e., linking this one and only, well-known CR to the Kuadrant CR. Then, later, the reconcilers could always navigate to the first child of the Kuadrant CR of this type, with no need to filter by name again. WDYT?

internal/controller/state_of_the_world.go
```diff
@@ -480,6 +488,32 @@ func (b *BootOptionsBuilder) getConsolePluginOptions() ([]controller.ControllerO
 	return opts, nil
 }
 
+func (b *BootOptionsBuilder) getOpenShiftServerConfigOptions() ([]controller.ControllerOption, error) {
+	opts := make([]controller.ControllerOption, 0, 2)
```
guicassolato: Line 504 appends 3 elements to the slice, but the capacity is preset to 2. Won't this require a realloc?

internal/controller/authorino_reconciler.go
```diff
@@ -224,3 +249,34 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
+
+func (r *AuthorinoReconciler) resolveTLSProfileFromTopology(topology *machinery.Topology, kobj *v1beta1.Kuadrant) (string, []string) {
```
guicassolato: Maybe add a unit test for this func?

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler_test.go
```diff
@@ -333,9 +333,9 @@ func TestBuildTargetCookieExpression_Examples(t *testing.T) {
 
 	for _, ex := range examples {
 		t.Run(ex.scenario, func(t *testing.T) {
-			// The CEL expression uses a ternary: request.url_path + (has(request.query) ? "?" + request.query : "")
+			// The CEL expression uses a ternary: request.url_path + (has(request.query) && request.query != "" ? "?" + request.query : "")
```
guicassolato: Brought from an unrelated and uncommitted local change perhaps, or was it @coderabbit complaining about it?

internal/controller/authorino_reconciler.go
```diff
@@ -79,10 +90,20 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
+						CipherSuites: tlsCipherSuites,
+					},
 				},
 				Listener: authorinoopapi.Listener{
-					Tls: clusterAuthorino.Spec.Listener.Tls,
+					Tls: authorinoopapi.Tls{
+						Enabled:      clusterAuthorino.Spec.Listener.Tls.Enabled,
+						CertSecret:   clusterAuthorino.Spec.Listener.Tls.CertSecret,
+						MinVersion:   tlsMinVersion,
+						CipherSuites: tlsCipherSuites,
```
guicassolato: Should we avoid setting these options when `tls.enabled` is `false`? Thinking about reducing the noise in the CR spec (and command-line args). WDYT?

cmd/extensions/oidc-policy/internal/controller/oidcpolicy_reconciler_test.go
guicassolato: It's unrelated to the main issue addressed by this PR though. For the sake of cohesion, I personally think it would be better addressed in a separate PR.

guicassolato: Really up to @philbrookes to decide.

philbrookes: It was coderabbit. Same with the comment, I'll undo these changes.

internal/openshift/tls_profile.go
philbrookes: I did originally  create a test script to go through some local testing, that I commited originally as well. Eventually I decided that it was probably not useful enough to commit and removed it. I guess some left over references made it through!

internal/controller/authorino_reconciler.go
philbrookes: I'm not seeing anywhere to put documentation on envvars (or any existing docs on envvars).

internal/controller/state_of_the_world.go
philbrookes: Thanks! I don't know how this got past my testing...

internal/controller/authorino_reconciler.go
philbrookes: the first cast is required to access the `.Object` property, then that object can be cast.

Trying:
```go
apiServer, ok := child.Object(*configv1.APIServer)

```
Generates the error: `child.Object undefined`.

internal/openshift/tls_profile.go
philbrookes: I agree, let's make what was in that script into integration tests. I'll get on that tomorrow.

internal/controller/authorino_reconciler.go
maksymvavilov: Will there be a doc describing TLS propagation? Something with "hey, if there is this OS CR present, the TLS config comes from it, otherwise it is this default. Also, you can override the name of the CR"

philbrookes: I don't think it makes sense for us to have that, as this is an openshift API that we are choosing to honour when present. I would expect Openshift to document the use of the config CR (although the best I've been able to find is [here](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md)). 

Overriding the name of the CR, is present to allow for troubleshooting weird situations (e.g. a customer must rename the CR from the default for some reason).

tests/bare_k8s/suite_test.go
```diff
@@ -99,11 +107,75 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	err := testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
 })
 
+func installOpenShiftAPIServerCRD(cfg *rest.Config) {
+	By("installing OpenShift APIServer CRD")
+	extClient, err := apiextensionsclient.NewForConfig(cfg)
+	Expect(err).NotTo(HaveOccurred())
+
+	crd := &apiextensionsv1.CustomResourceDefinition{
```
guicassolato: Is this a common practice: installing a shallow representation of the CRD for the tests? I reckon other integration tests in the Kuadrant Operator run against the actual distributed versions of the CRDs they depend upon (e.g., Gateway API, Istio, etc - see `local-env-setup` make target). Or is this maybe only for testing the check for presence of the `APIServer` CRD while the rest is all unit test?

tests/bare_k8s/suite_test.go
```diff
@@ -67,6 +73,8 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
+	installOpenShiftAPIServerCRD(cfg)
```
guicassolato: Wondering about why not having this in `tests/common` and run it for all test suites? Or perhaps we don't expect any difference and therefore testing on bare K8s suffices, running for all suites would be redundant?

internal/openshift/tls_profile.go
```diff
@@ -0,0 +1,161 @@
...
+func ResolveTLSProfile(profile *configv1.TLSSecurityProfile) (minVersion string, cipherSuites []string) {
+	var spec *configv1.TLSProfileSpec
+
+	if profile != nil && profile.Type == configv1.TLSProfileCustomType && profile.Custom != nil {
+		spec = &profile.Custom.TLSProfileSpec
+	} else if profile != nil {
+		spec = configv1.TLSProfiles[profile.Type]
+	}
+
+	if spec == nil {
```
guicassolato, line 150: nit:

I imagine this case was thought for when `ResolveTLSProfile` is called passing a `nil` profile. It could also trigger if `profile.Type` is set to a value not in `configv1.TLSProfiles`. Worth maybe logging an exception in this case (before the correct falling back to `configv1.TLSProfileIntermediateType`)?

Also, what happens if `profile.Type == configv1.TLSProfileCustomType` and `profile.Custom` is `nil`? Does OpenShift's API server operator allow that? The [CRD](https://github.com/openshift/api/blob/master/payload-manifests/crds/0000_10_config-operator_01_apiservers-Default.crd.yaml) doesn't seem to have any validation. Maybe at the controller level?

guicassolato (dismissed):
PR looks good.

I left a couple of comments on:
- small nit about logging edge cases, and
- clarifying questions about the test design.

But more importantly, I've tested the happy path locally and I was indeed happy with the outcome 🙂

**Setup the environment:**

```sh
make local-setup
```

**Install the `APIServer` CRD:**

```sh
kubectl apply -f https://raw.githubusercontent.com/openshift/api/refs/heads/master/payload-manifests/crds/0000_10_config-operator_01_apiservers-Default.crd.yaml
```

**Request an instance of Kuadrant:**

```sh
kubectl apply -f - <<EOF
kind: Kuadrant
metadata:
  name: kuadrant
  namespace: kuadrant-system
spec: {}
EOF
```

**Enable TLS in Authorino:**

```sh
kubectl apply -f - <<EOF
kind: Issuer
metadata:
  name: server-certs-issuer
  namespace: kuadrant-system
spec:
  selfSigned: {}
---
kind: Certificate
metadata:
  name: authorino-server-cert
  namespace: kuadrant-system
spec:
  dnsNames:
  - authorino-authorino-authorization.kuadrant-system.svc
  - authorino-authorino-authorization.kuadrant-system.svc.cluster.local
  issuerRef:
    kind: Issuer
    name: server-certs-issuer
  secretName: authorino-server-cert
EOF
```

```sh
kubectl patch authorino/authorino -n kuadrant-system --type=merge -p '{"spec":{"listener":{"tls":{"enabled":true,"certSecretRef":{"name":"authorino-server-cert"}}}}}'
```

**Simulate enabling a custom TLS profile cluster-wide:**

```sh
kubectl apply -f -<<EOF
apiVersion: config.openshift.io/v1
kind: APIServer
metadata:
  name: cluster
spec:
  tlsSecurityProfile:
    type: Custom # tip: omit this line to verify what I was talking about at one of my nitpick comments.
    custom:
      minTLSVersion: VersionTLS11
      ciphers:
      - ECDHE-ECDSA-CHACHA20-POLY1305
      - ECDHE-RSA-CHACHA20-POLY1305
      - ECDHE-RSA-AES128-GCM-SHA256
      - ECDHE-ECDSA-AES128-GCM-SHA256
EOF
```

**Bounce the Operator pod:** (In case in doesn't pick up the `APIServer` CR right away, since the CRD may have been installed _after_ the Operator initialised.)

```sh
kubectl rollout restart deployment/kuadrant-operator-controller-manager -n kuadrant-system
```

**Check the state of the affected resources:**

```sh
kubectl get authorino/authorino -n kuadrant-system -o yaml | yq '.spec.listener.tls'
# certSecretRef:
#   name: authorino-server-cert
# cipherSuites:
#   - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
#   - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
#   - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
#   - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
# enabled: true
# minVersion: "1.1"
```

```sh
kubectl get deployment/authorino -n kuadrant-system -o yaml | yq '.spec.template.spec.containers[0]'.args
# - --allow-superseding-host-subsets
# - --tls-cert=/etc/ssl/certs/tls.crt
# - --tls-cert-key=/etc/ssl/private/tls.key
# - --tls-min-version=1.1
# - --tls-cipher-suites=TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
```

tests/bare_k8s/suite_test.go
philbrookes: I think I left this over from an attempt to have a test for when the CRD was absent that ultimately didn't work. I'll revert this to work the usual way.

internal/openshift/tls_profile.go
philbrookes: yeah, good point. I'll update this too.

tests/bare_k8s/suite_test.go
philbrookes: related to your other comment, I'll move the installation of this CRD to kustomize.

internal/openshift/tls_profile.go
```diff
@@ -139,10 +140,17 @@ func ResolveTLSProfileFromTopology(topology *machinery.Topology, kobj *kuadrantv
 func ResolveTLSProfile(profile *configv1.TLSSecurityProfile) (minVersion string, cipherSuites []string) {
 	var spec *configv1.TLSProfileSpec
 
-	if profile != nil && profile.Type == configv1.TLSProfileCustomType && profile.Custom != nil {
-		spec = &profile.Custom.TLSProfileSpec
+	if profile != nil && profile.Type == configv1.TLSProfileCustomType {
+		if profile.Custom != nil {
+			spec = &profile.Custom.TLSProfileSpec
+		} else {
+			slog.Warn("TLS profile type is Custom but Custom spec is nil, falling back to Intermediate")
```
guicassolato: I don't recall using `log/slog` in a Kubernetes controller before. Will it respect the logging guidelines described in https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/overviews/logging.md (structured, 3-levels, etc)?

philbrookes: no it doesn't. I'll update it!

maksymvavilov (dismissed):

maksymvavilov (approved):

guicassolato (approved):

PR #2093, opened 2026-07-07
Change docs to use the OPA Rego v1 syntax

averevki:
The addition to the authorino docs change https://github.com/Kuadrant/authorino/pull/648, updating the docs residing in the kuadrant-operator to also use the new OPA Rego v1 syntax. 

I verified the new syntax with the OPA CLI, and also simplified some of the conditions where it was possible.

### DIFF
--- a/doc/overviews/token-rate-limiting.md
+++ b/doc/overviews/token-rate-limiting.md
@@ -217,8 +217,9 @@ spec:
         opa:
           rego: |
             groups := split(object.get(input.auth.identity.metadata.annotations, "kuadrant.io/groups", ""), ",")
-            allow { groups[_] == "free" }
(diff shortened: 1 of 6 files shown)


guicassolato (changes requested):
The changes to the examples look good. Blocking the PR for now because first we need to bump the version of the AuthConfig-inherited types used in the definition of the AuthPolicy CRD. Until then, fields like `opa.version` will not be available to the users.

guicassolato (commented):
A new version of Authorino has been released containing the new version of OPA available and therefore also the new field: https://github.com/Kuadrant/authorino/releases/tag/v0.27.0. This PR should now be unblocked.

maksymvavilov:

PR #2094, opened 2026-07-08
feat: implement ErrorRegistry retry handling for all reconcilers

guicassolato:
## Summary
Implements comprehensive ErrorRegistry retry handling for all reconciliation errors across the operator. This covers errors that were previously silently absorbed (25 original TODO comments) plus additional error sites discovered through systematic audit, including upstream cluster reconciliation operations.

## What Changed
**Total: 43 error handling sites across 14 reconciler files**

### Reconcilers Enhanced
**Extension Reconcilers (2 files, 12 error sites):**
- EnvoyGatewayExtensionReconciler (Reconcile + reconcileUpstreamClusters)
- IstioExtensionReconciler (Reconcile + reconcileUpstreamClusters)

**Cluster Reconcilers (6 files, 18 error sites):**
- EnvoyGatewayAuthClusterReconciler
- EnvoyGatewayRateLimitClusterReconciler
- EnvoyGatewayTracingClusterReconciler
- IstioAuthClusterReconciler
- IstioRateLimitClusterReconciler
- IstioTracingClusterReconciler

**Policy Reconcilers (2 files, 6 error sites):**
- EffectiveDNSPoliciesReconciler
- EffectiveTLSPoliciesReconciler

**Status Updaters (2 files, 2 error sites):**
- AuthPolicyStatusUpdater
- RateLimitPolicyStatusUpdater

**Pre-existing Reconcilers (2 files, 5 error sites):**
- AuthConfigsReconciler
- LimitadorLimitsReconciler

### Error Handling Pattern
All reconcilers now follow the established pattern from AuthConfigsReconciler, LimitadorLimitsReconciler, AuthorinoReconciler, and LimitadorReconciler:

```go
const ReconcilerName = "..."

errorRegistry := GetOrCreateErrorRegistry(state)

// For each Create/Update/Delete operation
if err != nil {
    errorRegistry.Record(
        ReconcilerName,
        OperationCreate/Update/Delete,
        k8stypes.NamespacedName{Name: ..., Namespace: ...},
        ResourceGroupKind,
        err,
    )
}
```

### Idempotent Operation Handling
**Delete Operations (15 sites):**
- ✅ Skip retry for NotFound errors with `!apierrors.IsNotFound(err)`
- Rationale: Resource already deleted, retry unnecessary

**Create Operations (11 sites):**  
- ✅ Record AlreadyExists errors for retry
- Rationale: Allows manual intervention to clean up conflicting state before next reconciliation
- Future work: Implement automatic fallback to Update operations

**Update Operations (14 sites):**
- ✅ Record all errors for retry
- Rationale: Updates should succeed on properly constructed resources

## ErrorRegistry Features
- ✅ Exponential backoff retry (2s → 4s → 8s → 16s → 32s → 60s max)
- ✅ Maximum 5 retry attempts before giving up
- ✅ Structured error logging with resource identification
- ✅ Automatic error resolution tracking
- ✅ Non-blocking errors (other reconciliation work continues)

### Function Signature Changes
Modified signatures to accept errorRegistry parameter:
- `EffectiveDNSPoliciesReconciler.deleteRecord()`
- `EffectiveDNSPoliciesReconciler.deleteOrphanDNSRecords()`
- `EffectiveTLSPoliciesReconciler.reconcileCertificates()`
- `EnvoyGatewayExtensionReconciler.reconcileUpstreamClusters()`
- `IstioExtensionReconciler.reconcileUpstreamClusters()`

### New Constants Added
Each reconciler now has a name constant for error tracking:
- `AuthPolicyStatusUpdaterName`
- `RateLimitPolicyStatusUpdaterName`
- `EnvoyGatewayExtensionReconcilerName`
- `IstioExtensionReconcilerName`
- `EnvoyGatewayAuthClusterReconcilerName`
- `EnvoyGatewayRateLimitClusterReconcilerName`
- `EnvoyGatewayTracingClusterReconcilerName`
- `IstioAuthClusterReconcilerName`
- `IstioRateLimitClusterReconcilerName`
- `IstioTracingClusterReconcilerName`
- `EffectiveDNSPoliciesReconcilerName`
- `EffectiveTLSPoliciesReconcilerName`

## Test Plan
- [x] Code formatted with `make fmt`
- [x] Code verified with `make vet`
- [x] Project builds successfully with `make build`
- [ ] Integration tests pass with `make test-integration`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when applying authentication, rate-limit, tracing, extension, and gateway policies.
  * Failed resource creations, updates, deletions, and status updates are now recorded for deferred retry instead of being lost.
  * Cleanup operations now correctly ignore resources that have already been removed.
  * DNS record and TLS certificate reconciliation failures are also tracked for retry.
  * Added clearer handling for missing gateway addresses and routes.

### DIFF
--- a/internal/controller/effective_dnspolicies_reconciler.go
+++ b/internal/controller/effective_dnspolicies_reconciler.go
@@ -267,14 +267,12 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 
 	state.Store(StateDNSPolicyErrorsKey, policyErrors)
 
-	return r.deleteOrphanDNSRecords(controller.LoggerIntoContext(ctx, logger), topology, state)
+	return r.deleteOrphanDNSRecords(controller.LoggerIntoContext(ctx, logger), topology, errorRegistry)
 }
 
 // deleteOrphanDNSRecords deletes any DNSRecord resources that exist in the topology but have no parent targettable, policy or path back to the policy.
-func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSRecords(ctx context.Context, topology *machinery.Topology, state *sync.Map) error {
+func (r *EffectiveDNSPoliciesReconciler) deleteOrphanDNSRecords(ctx context.Context, topology *machinery.Topology, errorRegistry *ErrorRegistry) error {
 	logger := controller.LoggerFromContext(ctx).WithName("deleteOrphanDNSRecords").WithValues("context", ctx)
-
-	errorRegistry := GetOrCreateErrorRegistry(state)
 
 	orphanRecords := lo.Filter(topology.Objects().Items(), func(item machinery.Object, _ int) bool {
 		if item.GroupVersionKind().GroupKind() == DNSRecordGroupKind {
--- a/internal/controller/effective_dnspolicies_reconciler.go
+++ b/internal/controller/effective_dnspolicies_reconciler.go
@@ -13,6 +13,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
@@ -35,6 +36,8 @@ func NewEffectiveDNSPoliciesReconciler(client *dynamic.DynamicClient, scheme *ru
 		scheme: scheme,
 	}
 }
+
+const EffectiveDNSPoliciesReconcilerName = "EffectiveDNSPoliciesReconciler"
 
 type EffectiveDNSPoliciesReconciler struct {
 	client *dynamic.DynamicClient
@@ -64,6 +67,8 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 	policyErrors := map[string]error{}
 
 	logger.V(1).Info("updating dns policies", "policies", len(policies))
+
+	errorRegistry := GetOrCreateErrorRegistry(state)
 
 	clusterID, err := utils.GetClusterUID(ctx, r.client)
 	if err != nil {
@@ -176,13 +181,13 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 					} else {
 						rLogger.V(1).Info("no endpoint addresses for DNSRecord, deleting record for listener")
 					}
-					r.deleteRecord(ctx, existingRecordObj)
+					r.deleteRecord(ctx, existingRecordObj, errorRegistry)
 					continue
 				}
 
 				if !canUpdateDNSRecord(ctx, existingRecord, desiredRecord) {
 					rLogger.V(1).Info("unable to update record, deleting record for listener and re-creating")
-					r.deleteRecord(ctx, existingRecordObj)
+					r.deleteRecord(ctx, existingRecordObj, errorRegistry)
 					break
 				}
 
@@ -201,6 +206,15 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 				rLogger.V(1).Info("updating record for listener")
 				if _, uErr := resource.Update(ctx, un, metav1.UpdateOptions{}); uErr != nil {
 					rLogger.Error(uErr, "unable to update dns record")
+
+					// Record error for deferred retry
+					errorRegistry.Record(
+						EffectiveDNSPoliciesReconcilerName,
+						OperationUpdate,
+						k8stypes.NamespacedName{Name: existingRecord.GetName(), Namespace: existingRecord.GetNamespace()},
+						DNSRecordGroupKind,
+						uErr,
+					)
 				}
 				continue
 			}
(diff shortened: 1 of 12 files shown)


internal/controller/effective_dnspolicies_reconciler.go
```diff
@@ -244,13 +267,15 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 
 	state.Store(StateDNSPolicyErrorsKey, policyErrors)
 
-	return r.deleteOrphanDNSRecords(controller.LoggerIntoContext(ctx, logger), topology)
+	return r.deleteOrphanDNSRecords(controller.LoggerIntoContext(ctx, logger), topology, state)
```
maksymvavilov: Why are we passing in a state to recreate the registry instead of passing down the error registry itself?

internal/controller/effective_dnspolicies_reconciler.go
```diff
@@ -223,8 +237,17 @@ func (r *EffectiveDNSPoliciesReconciler) reconcile(ctx context.Context, _ []cont
 
 			//Create
 			lLogger.V(1).Info("creating DNS record for listener")
-			if _, cErr := resource.Create(ctx, un, metav1.CreateOptions{}); cErr != nil && !apierrors.IsAlreadyExists(cErr) {
```
maksymvavilov, line 226: This is being called out in the PR description. I just want to clarify that I understand it well. 
If we get `alreadyExists`, we go for a retry; on a retry, we get a refreshed topology with the object already there, which leads to an update call instead?

internal/controller/envoy_gateway_extension_reconciler.go
```diff
@@ -251,7 +280,7 @@ func (r *EnvoyGatewayExtensionReconciler) reconcileUpstreamClusters(ctx context.
 			!desired
 	})
 	for _, pp := range stale {
-		if err := r.client.Resource(kuadrantenvoygateway.EnvoyPatchPoliciesResource).Namespace(pp.GetNamespace()).Delete(ctx, pp.GetName(), metav1.DeleteOptions{}); err != nil {
+		if err := r.client.Resource(kuadrantenvoygateway.EnvoyPatchPoliciesResource).Namespace(pp.GetNamespace()).Delete(ctx, pp.GetName(), metav1.DeleteOptions{}); err != nil && !apierrors.IsNotFound(err) {
```
maksymvavilov, line 299: Why are we not recording an error here?

internal/controller/effective_dnspolicies_reconciler.go
guicassolato: The explanation why this check is being removed is two-fold:
1. Silently ignoring an already-exists error on CREATE is not a great idea overall. A resource existing in the cluster does't mean it exits in the correct state, and if it is in the cluster then it should be represented in the topology and therefore an UPDATE should have been performed in the first place, and not a CREATE. It smells like a situation where the resource was manually created by the user and missing its linking properties, or a concurrency problem. Either way, it's better not to hide the error, but schedule a retry. If the error is consistent, it will be flagged enough times so the user can act on it.
2. The retry itself does not get a refreshed topology snapshot. All retries run with the same snapshot from when it failed the first time. If the state of the ~world~ cluster changes while there are retries waiting in the queue, the mechanism will abort the retries (i.e., clear the queue) and fallback to a normal reconciliation with the updated topology. So, it's not a retry per-se, it's a fresh new reconciliation call due to new events that changed the state of the cluster. If the event was 'an object that was previously missing in the topology has been created', the new snapshot will include de object, which, in the right state, will trigger an UPDATE this time.

internal/controller/effective_dnspolicies_reconciler.go
guicassolato: No particular reason really. I suppose we could pass `errorRegistry` instead and save the extra call to `GetOrCreateErrorRegistry(state)`. Will do. Thanks!

internal/controller/envoy_gateway_extension_reconciler.go
guicassolato: This isn't much of a special case in this file, which lacks error recording for retries in the entire `reconcileUpstreamClusters` function.

It probably slipped due to the order of changes: first I looked up for all `// TODO: handle error` placeholders (and similar) to add the retry, then I standardised `&& !apierrors.IsNotFound(err)` across DELETE operations.

I'll add error recording to all API server operations related to Envoy clusters of the wasm extension.

internal/controller/effective_dnspolicies_reconciler.go
maksymvavilov: Does it mean that now on `AlreadyExists` we will potentially "waste" 5(?) reconcile loops on retries? 
I agree that ignoring this error is not a good path, but if we only log the error and nothing else happens, should we record the error in the registry at all?

guicassolato: I agree the benefit is only marginal. It's essentially to cover the case where the resource exists but it doesn't show in the topology. The classical scenario is: the user tried to recreate the resource manually after messing with the state of the cluster and got it wrong. The retries give the user a chance to simply delete the resource, a remediation event that might not be caught by the reconciler depending on the reconciliation filters (e.g. label selectors).

maksymvavilov: I see. Thank you for the explanation. This is preference-land now, but I'd show an error log in this case and would not requeue.
But even if the cluster is a mess with thousands of policies/routes/listeners, this would still be relatively cheap "repeat".

maksymvavilov (approved):

PR #2098, opened 2026-07-08
fix: separate error handling in IsCertManagerInstalled

### LINKED ISSUE #2097: Bug: `IsCertManagerInstalled` logs `Error` with nil error when CRD is simply not installed
## Describe the bug
When checking if cert-manager CRDs are installed, the `IsCertManagerInstalled` function in `internal/gatewayapi/utils.go` combines checking the error and checking the boolean result into a single condition (`!ok || err != nil`). 

If the CRD is simply not installed, the function receives `ok=false` and `err=nil`. Because of the combined condition, it proceeds to log an error with a `nil` error object, like so:
`logger.V(1).Error(err, "CertManager CRD was not installed", ...)`

This produces confusing log output. A missing dependency (which is expected in some cluster setups if the user hasn't installed cert-manager) should not log as a generic "error", and logging a nil error is an anti-pattern. 

*Note: This is related to issue #1264, where connection failures were misidentified as missing CRDs. This issue focuses specifically on the logging semantics when the CRD is genuinely absent (i.e. `err` is nil).*

## To Reproduce
1. Ensure cert-manager is NOT installed in the cluster.
2. Start the kuadrant operator.
3. Observe the logs. It will log an error message for CertManager CRD not installed, despite `err` being nil.

## Expected behavior
The function should distinguish between "not installed" and "failed to check". 
If the CRD is not installed (`!ok` and `err == nil`), it should log an informational message (e.g. `logger.V(1).Info(...)`) and return `false, nil`. 
If there was an actual error checking the CRD (`err != nil`), it should log a true error and return `false, err`.

## Proposed Solution
Split the condition in `internal/gatewayapi/utils.go`:

```go
ok, err := utils.IsCRDInstalled(...)
if err != nil {
    logger.Error(err, "failed to check if CertManager CRD is installed", ...)
    return false, err
}
if !ok {
    logger.V(1).Info("CertManager CRD was not installed", ...)
    return false, nil
}
```

shellyco-code:
## Description
This PR fixes a bug in `internal/gatewayapi/utils.go` where `IsCertManagerInstalled` was logging an error with a `nil` error object if the CertManager CRD was simply not installed in the cluster.

Previously, the function combined the CRD presence check (`!ok`) and the error check (`err != nil`) into a single condition, logging a generic error in both scenarios. This led to confusing log output for users who intentionally did not install cert-manager (a valid configuration for some policies).

**Changes made:**
- Split the combined condition into two explicit checks.
- If an actual error occurs when communicating with the API server (`err != nil`), it logs an `Error` and returns `false, err`.
- If the CRD is just missing (`!ok`), it gracefully logs an `Info` message at verbosity level 1 and returns `false, nil`.

**Related Issue:**
Fixes #2097

## Verification
- Run the operator locally (`make run`) against a cluster without cert-manager installed.
- Ensure that the operator logs an informational message (`CertManager CRD was not installed`) instead of a nil error trace.

## Checklist
- [x] Commits are signed off (`git commit -s`)
- [x] Code builds locally (`make build`)
- [x] Relevant tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cert-Manager detection so the app more reliably recognises whether required components are available.
  * Added clearer handling when a required component is missing or cannot be checked, helping avoid incorrect installation status results.

### DIFF
--- a/internal/gatewayapi/utils.go
+++ b/internal/gatewayapi/utils.go
@@ -209,19 +209,34 @@ func IsGatewayAPIInstalled(restMapper meta.RESTMapper) (bool, error) {
 }
 
 func IsCertManagerInstalled(restMapper meta.RESTMapper, logger logr.Logger) (bool, error) {
-	if ok, err := utils.IsCRDInstalled(restMapper, certmanager.GroupName, certmanv1.CertificateKind, certmanv1.SchemeGroupVersion.Version); !ok || err != nil {
-		logger.V(1).Error(err, "CertManager CRD was not installed", "group", certmanager.GroupName, "kind", certmanv1.CertificateKind, "version", certmanv1.SchemeGroupVersion.Version)
+	ok, err := utils.IsCRDInstalled(restMapper, certmanager.GroupName, certmanv1.CertificateKind, certmanv1.SchemeGroupVersion.Version)
+	if err != nil {
+		logger.Error(err, "failed to check if CertManager CRD is installed", "group", certmanager.GroupName, "kind", certmanv1.CertificateKind, "version", certmanv1.SchemeGroupVersion.Version)
 		return false, err
 	}
-
-	if ok, err := utils.IsCRDInstalled(restMapper, certmanager.GroupName, certmanv1.IssuerKind, certmanv1.SchemeGroupVersion.Version); !ok || err != nil {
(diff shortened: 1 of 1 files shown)


maksymvavilov:

PR #2101, opened 2026-07-09
Revert "feat(wasm): `allow_on_headers_stop_iteration` on our `PluginConfig`"

### LINKED ISSUE #388: Request body data loss (16KiB chunk delivered empty) when combined with ext_proc FULL_DUPLEX on Envoy 1.35+
## Summary

When the wasm-shim (auth-only AuthPolicy) shares a filter chain with an ext_proc filter running `request_body_mode: FULL_DUPLEX_STREAMED`, large request bodies (>1MB) intermittently lose **exactly one 16,384-byte chunk**: the ext_proc server receives the *second* body message with **zero bytes** where a full 16KiB chunk should be. Failure rate 40–80% at ~1.6MB bodies. The corrupted body is then forwarded upstream (broken base64 in image blocks, or JSON parse failures when the hole lands on structure).

## Environment

- wasm-shim served by kuadrant-operator v1.4.1 (`plugin.wasm` sha256 `7ac45a63…`), phase-configured via generated EnvoyFilters, gRPC auth to Authorino
- Istio gateway (tested on upstream Istio 1.27.9 / Envoy 1.35.13, 1.28.8 / 1.36.9, and 1.29.5 / 1.37.5 — reproduces on all three)
- **Not reproducible on Envoy 1.34 (Istio 1.26.2)** — same shim generation ran the identical topology cleanly under weeks of production-like traffic
- ext_proc server: llm-d-inference-payload-processor (Go, go-control-plane)

## Evidence

Instrumented the ext_proc server to record per-chunk `(offset:length:hash)` and a whole-body integrity record, comparing against the client-side constant payload (1,640,098 bytes):

- Failing runs receive exactly **1,623,714 bytes = 16,384 short**, same head/tail hashes, same message count (102)
- Chunk map of every failing run: 97×16384-byte chunks, fragments, and **message index 1 with `length=0`** — deterministic, always the second body message, always exactly one 16KiB unit lost

A/B attribution on the same gateway, same minutes, byte-identical payload:

| Configuration | Result |
|---|---|
| wasm + ext_proc (Envoy 1.37.5) | 5/8 and 10/12 corrupted |
| **AuthPolicy removed (no wasm), same ext_proc** | **12/12 clean** |
| wasm restored | corruption returns immediately |
| wasm + ext_proc on Envoy 1.34 | clean (long-running production-like traffic) |

We also observed rarer variants under the same combination (duplicate slice mid-stream; duplicate final chunk after EndOfStream) that look like the same buffer mismanagement.

## Reproduction

1. Gateway with a wasm-shim AuthPolicy (auth-only; no body-inspecting rules needed) + one ext_proc filter with `request_body_mode: FULL_DUPLEX_STREAMED`
2. POST a >1MB request with **dense content** (e.g. a base64 image block — whitespace-padded test bodies mask the corruption since a hole in padding is invisible)
3. Compare bytes received by the ext_proc server against Content-Length — failures show a deficit of exactly 16384 with the second body message empty

## Suspicion

Envoy's proxy-wasm body-buffer ownership/semantics changed around 1.35; the shim's request-body passthrough (pause for the gRPC auth callout, then resume) appears to consume or fail to re-inject the first buffered chunk after resume, so the message is delivered to the next filter (ext_proc) with an empty body. Happy to provide the full chunk maps, instrumented-server logs, or run patched shim builds against the live reproduction environment.

adam-cattermole:
This reverts commit 65068c9ac902b509d86fbd3929453a3f6047f805.

## Summary by CodeRabbit

~~* **Bug Fixes**~~
  ~~* Adjusted the generated wasm filter runtime configuration to remove header-based early termination behaviour.~~
  ~~* Kept the existing reload failure handling in place while simplifying the filter’s runtime settings.~~

### DIFF
--- a/internal/istio/utils.go
+++ b/internal/istio/utils.go
@@ -132,8 +132,7 @@ func buildWasmFilterConfig(wasmURL, imagePullSecret, imageSHA, clusterName strin
 			},
 			"allow_precompiled": true,
 		},
-		"failure_policy":                  "FAIL_RELOAD",
-		"allow_on_headers_stop_iteration": true,
+		"failure_policy": "FAIL_RELOAD",
 	}
 
 	if pluginConfig != nil {


alexsnaps (approved):
😭

hexfusion:
@adam-cattermole could maintainers provide a bit more context on why this revert was needed? I see that it closed https://github.com/Kuadrant/wasm-shim/issues/388 but is there any more context?

adam-cattermole:
As far as we could determine, there's a bug in Envoy when a wasm filter shares a filter chain with an ext_proc filter using `FULL_DUPLEX_STREAMED` in conjunction with the `allow_on_headers_stop_iteration` flag, where large request bodies lose one chunk (the second one) streamed to subsequent filters.

My vague understanding is that while the request is paused at the wasm filter subsequent chunks are buffered. Upon resuming envoy passes the first chunk down the filter chain to subsequent filters (ext_proc), but due to a stale flag, continues decoding data from the 3rd chunk (skipping the buffered one), resulting in one envoy buffer frame being lost. This is the result of differing behaviour due to `allow_on_headers_stop_iteration` and the way that `StopSingleIteration` works vs `StopAllIterationAndWatermark`

PR #2107, opened 2026-07-13
feat: nightly benchmark workflow (alternative to per-PR)

maksymvavilov:
## Summary

Alternative to [PR #2035](https://github.com/Kuadrant/kuadrant-operator/pull/2035) — runs benchmarks nightly instead of per-PR.

Same benchmarks, same `make test-bench` target, different trigger and comparison strategy.

## How it works

1. Runs nightly at 3:17am UTC on `main`
2. Downloads previous night's results from artifacts
3. Compares with `benchstat` (same 5% threshold, same p<0.05 filter)
4. If regression detected: creates a GitHub issue

## Trade-offs vs per-PR (#2035)

| | Per-PR (#2035) | Nightly (this PR) |
|---|---|---|
| **When regression found** | Before merge | ~24h after merge |
| **Which PR caused it** | Known immediately | Requires bisection |
| **Baseline** | Always available (main branch) | Depends on previous run's artifact |
| **First run / artifact expiry** | Works immediately | No comparison possible |
| **Storage** | None needed | Artifacts (90-day expiry) |
| **CI overhead** | ~8 min per PR (parallel with integration tests) | ~5 min nightly |
| **Workflow complexity** | 95 lines | 105 lines (artifact download, issue creation) |
| **Action on regression** | Author fixes before merge | Triage needed to find which PR |

## Recommendation

This PR exists to demonstrate the trade-offs. I recommend [PR #2035](https://github.com/Kuadrant/kuadrant-operator/pull/2035) (per-PR) because:
- It catches regressions **before** they merge, not after
- It identifies **which PR** caused the regression with zero ambiguity
- It runs in parallel with integration tests that take longer, so it adds **no wall-clock time** to CI
- It requires **no artifact storage** or API calls for baseline retrieval

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,78 @@
+name: Benchmark
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'internal/**'
+      - 'pkg/**'
+      - 'api/**'
+      - 'go.mod'
+      - 'go.sum'
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  benchmark:
+    name: Benchmark
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Check out PR branch
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Install benchstat
+        run: go install golang.org/x/perf/cmd/benchstat@latest
+
+      # Run benchmarks on PR branch
+      - name: Run benchmarks (PR)
+        run: make test-bench
+      - name: Save PR results
+        run: mv benchmark-results.txt pr-bench.txt
+
+      # Run benchmarks on base branch
+      - name: Check out base branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.base_ref }}
+          clean: false
+          path: base
+      - name: Run benchmarks (base)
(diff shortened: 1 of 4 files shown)


maksymvavilov:
## Demo result: nightly approach fails due to CI runner inconsistency

Attempted to demo this workflow on [my fork](https://github.com/maksymvavilov/kuadrant-operator):

1. Triggered first run to create baseline — completed successfully on **AMD EPYC 9V74 80-Core**
2. Merged an intentional regression (50μs sleep per route rule in the Paths() inner loop)
3. Triggered second run — completed successfully on **AMD EPYC 7763 64-Core**

**Result:** benchstat could not compare the two runs because Go benchmark output includes a `cpu:` header line that must match for comparison. Different CI runners = different CPUs = incomparable results.

```
# Run 1 baseline
cpu: AMD EPYC 9V74 80-Core Processor

# Run 2 (with regression)
cpu: AMD EPYC 7763 64-Core Processor
```

benchstat treated these as separate benchmark sets and showed only one column — the regression was completely invisible.

### Why per-PR (#2035) doesn't have this problem

The per-PR workflow runs both PR branch and base branch benchmarks **in the same job, on the same runner**. The CPU is always identical because both `make test-bench` calls happen sequentially on the same machine.

### Workarounds for nightly (all add complexity)

1. **Pin to a specific runner label** — not available on public GitHub-hosted runners
2. **Strip the `cpu:` line** from benchmark output before comparison — works but masks real hardware differences
3. **Self-hosted runners** — consistent hardware but operational overhead

This is a fundamental limitation of comparing benchmark results across separate workflow runs.

PR #2119, opened 2026-07-20
fix(tracing): propagate observability config to extension-only gateways

### LINKED ISSUE #2095: Observability config not propagated to Envoy filter when only extension policies are used
### Description

  When using only extension policies (e.g., `telemetry-policy`, `threat-policy`) without any core data plane policy (`AuthPolicy`, `RateLimitPolicy`, `TokenRateLimitPolicy`), the observability/tracing configuration from the `Kuadrant` CR is not propagated into the wasm plugin (Istio `EnvoyFilter` / Envoy Gateway `EnvoyExtensionPolicy`). This means no traces are emitted even though `spec.observability.tracing` is correctly configured.

  ### Root Cause

  There are two gaps in the current implementation:

  1. **Observability config not attached to extension-only wasm configs.** The observability config is built inside `buildWasmConfigs()` in both `istio_extension_reconciler.go` and `envoy_gateway_extension_reconciler.go`, but it is only embedded into configs that have core-policy-driven action sets (via `BuildConfigForActionSet()`). When no core policies exist, `buildWasmConfigs()` returns no entry for the gateway. In `Reconcile()`, the code retrieves a zero-value `wasm.Config{}` from the map, and `ApplyWasmConfigMutators` injects extension `TypedActions` into it — but the `Observability` field and `Services` (including the tracing service) remain nil.

  2. **Tracing cluster not created for extension-only gateways.** `IstioTracingClusterReconciler` and `EnvoyGatewayTracingClusterReconciler` use `targetIsInEffectivePolicyPath()` (`tracing_workflow_helpers.go:77-117`)
  to decide which gateways need a tracing cluster. This function only checks effective auth/ratelimit/tokenratelimit policy paths — gateways with only extension policies are skipped, so the Envoy cluster definition for the tracing backend is never created.

  ### Affected files

  - `internal/controller/istio_extension_reconciler.go` — observability built at L346-355 but not connected to zero-value config at L548
  - `internal/controller/envoy_gateway_extension_reconciler.go` — same pattern at L331-339 / L516
  - `internal/controller/istio_tracing_cluster_reconciler.go` — L74, `targetIsInEffectivePolicyPath` excludes extension-only gateways
  - `internal/controller/envoy_gateway_tracing_cluster_reconciler.go` — L74, same check
  - `internal/controller/tracing_workflow_helpers.go` — L77-117, only checks core policies

  ### Steps to reproduce

  1. Configure `Kuadrant` CR with `spec.observability.tracing` pointing to a collector endpoint
  2. Deploy only an extension policy (e.g., `TelemetryPolicy`) targeting a Gateway/HTTPRoute — no `AuthPolicy` or `RateLimitPolicy`
  3. Send traffic through the gateway
  4. Observe that the wasm plugin is created (extension `TypedActions` are present) but the `observability` and tracing `services` sections are missing from the wasm config
  5. No traces are reported to the collector

  ### Expected behavior

  Observability/tracing configuration should be propagated to the wasm plugin whenever *any* policy (core or extension) produces action sets for a gateway, not only when core data plane policies are present.

silvi-t:
Closes #2095 

  ### Problem

  When a gateway had only extension policies:
  1. The tracing cluster EnvoyFilter was not created because the tracing reconciler only checked for core effective policies
  2. The wasm config lacked `observability` and `services` (including `tracing-service`) because these were only injected when core policies produced action sets

  ### Changes

  **Extension reconcilers** (`istio_extension_reconciler.go`, `envoy_gateway_extension_reconciler.go`):
  - `buildWasmConfigs` now returns `observability` and `serviceBuilder` alongside wasm configs
  - After `ApplyWasmConfigMutators`, if extensions added action sets to a gateway that had none from core policies, observability config and services are injected
  - Existing extension-added services are preserved (no overwrite)

  **Tracing cluster reconcilers** (`istio_tracing_cluster_reconciler.go`, `envoy_gateway_tracing_cluster_reconciler.go`):
  - Removed `targetIsInEffectivePolicyPath` check — tracing clusters are now created for all gateways when tracing is configured
  - This is necessary because extension policies can't be detected without importing `internal/extension`, and workflow tasks run concurrently so state can't be shared between them
  - The tracing cluster is a lightweight Envoy cluster definition with zero overhead when unused

  **Cleanup** (`tracing_workflow_helpers.go`):
  - Removed `targetIsInEffectivePolicyPath` function and unused imports

  **Integration tests** (Istio + Envoy Gateway):
  - Removed `createAuthPolicy` helper and all calls — tracing clusters no longer require core policies
  - Updated "no policy exists" tests to assert tracing cluster IS created
  - Renamed tests to reflect new behavior
  

## Summary by CodeRabbit

* **New Features**
  * Extension policies can now target gateways when an associated relevant policy pipeline is present.
  * Added support for deriving request bindings from registered extension configuration.
  * Improved observability and service configuration when new extension actions are introduced.
* **Bug Fixes**
  * Enhanced gateway selection for tracing reconciliation to include gateways targeted by extension policies.
  * Improved validation and handling of extension actions, including pre-authorisation rate-limiting signal behaviour.

### DIFF
--- a/internal/controller/istio_extension_reconciler.go
+++ b/internal/controller/istio_extension_reconciler.go
@@ -115,6 +115,9 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 		if len(wasmConfig.ActionSets) > 0 && !hadActionSets {
 			if observability != nil {
 				wasmConfig.Observability = observability
+				if wasmConfig.Services == nil {
+					wasmConfig.Services = make(map[string]wasm.Service)
+				}
 				for k, v := range serviceBuilder.Build() {
 					if _, exists := wasmConfig.Services[k]; !exists {
 						wasmConfig.Services[k] = v
@@ -422,7 +425,7 @@ func (r *IstioExtensionReconciler) buildWasmConfigs(ctx context.Context, topolog
 
 	effectiveAuthPolicies, ok := state.Load(StateEffectiveAuthPolicies)
 	if !ok {
-		return nil, nil, nil, ErrMissingStateEffectiveAuthPolicies
+		return nil, observability, serviceBuilder, ErrMissingStateEffectiveAuthPolicies
 	}
 	effectiveAuthPoliciesMap := effectiveAuthPolicies.(EffectiveAuthPolicies)
 
--- a/internal/controller/istio_extension_reconciler.go
+++ b/internal/controller/istio_extension_reconciler.go
@@ -91,7 +91,7 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 	r.reconcileUpstreamClusters(ctx, topology, gateways, errorRegistry)
 
 	// build wasm plugin configs for each gateway
-	wasmConfigs, err := r.buildWasmConfigs(ctx, topology, state)
+	wasmConfigs, observability, serviceBuilder, err := r.buildWasmConfigs(ctx, topology, state)
 	if err != nil {
 		if errors.Is(err, ErrMissingStateEffectiveAuthPolicies) || errors.Is(err, ErrMissingStateEffectiveRateLimitPolicies) {
 			logger.V(1).Info(err.Error())
@@ -107,8 +107,20 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
 
 		// Get the wasm config for this gateway and apply mutators
 		wasmConfig := wasmConfigs[gateway.GetLocator()]
+		hadActionSets := len(wasmConfig.ActionSets) > 0
 		if err := extension.ApplyWasmConfigMutators(&wasmConfig, gateway, topology); err != nil {
 			logger.Error(err, "failed to apply wasm config mutators", "gateway", gatewayKey.String())
+		}
+
+		if len(wasmConfig.ActionSets) > 0 && !hadActionSets {
+			if observability != nil {
+				wasmConfig.Observability = observability
+				for k, v := range serviceBuilder.Build() {
+					if _, exists := wasmConfig.Services[k]; !exists {
+						wasmConfig.Services[k] = v
+					}
+				}
+			}
 		}
 
 		desiredEnvoyFilter := buildIstioEnvoyFilterForGateway(gateway, wasmConfig, wasmURL, wasmServerHost, wasmServerPort, WasmFileSHA256)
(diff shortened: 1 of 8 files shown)


internal/controller/istio_tracing_cluster_reconciler.go
```diff
@@ -71,7 +71,7 @@ func (r *IstioTracingClusterReconciler) Reconcile(ctx context.Context, _ []contr
 					return nil, false
 				}
 				isIstioGateway := lo.Contains(istioGatewayControllerNames, gatewayClass.(*machinery.GatewayClass).Spec.ControllerName)
-				return gateway, isIstioGateway && targetIsInEffectivePolicyPath(gateway, state)
+				return gateway, isIstioGateway
```
adam-cattermole: I'm a little unsure on deleting this and creating for every cluster. This introduces an `EnvoyFilter` for every gateway when we might only affect a small number - whilst the overhead of the filter is small, reconciling up to `N` of them could be somewhat costly.

The `istio_extension_reconciler.go` and envoy variant currently read information from the extensions so it's worth having a look at what they're doing (the concurrency/import issue is solved). I think the function we need is `extension.hasRelevantPipelineActions` (I'd maybe rename this and export it something like `HasPipelineForTarget`) and you'd end up doing:

```
return gateway, isIstioGateway && (targetIsInEffectivePolicyPath(gateway, state) || extension.HasPipelineForTarget(...))
```

internal/controller/istio_extension_reconciler.go
```diff
@@ -103,10 +103,25 @@ func (r *IstioExtensionReconciler) Reconcile(ctx context.Context, _ []controller
...
+				if wasmConfig.Services == nil {
+					wasmConfig.Services = make(map[string]wasm.Service)
+				}
+				for k, v := range serviceBuilder.Build() {
+					if _, exists := wasmConfig.Services[k]; !exists {
+						wasmConfig.Services[k] = v
+					}
+				}
+			}
+		}
```
adam-cattermole: I think this bit here can be simplified a little using samber/lo
```suggestion
        if len(wasmConfig.ActionSets) > 0 && !hadActionSets && observability != nil {
            wasmConfig.Observability = observability
            wasmConfig.Services = lo.Assign(wasmConfig.Services, serviceBuilder.Build())
        }
```

bundle/manifests/devportal.kuadrant.io_apikeys.yaml
```diff
@@ -67,6 +67,11 @@ spec:
                 required:
                 - name
                 type: object
+              expiresAt:
+                description: ExpiresAt is the time after which the API key should
+                  be revoked
+                format: date-time
+                type: string
               planTier:
```
adam-cattermole, line 75: Looks like maybe these changes to the manifests got pulled into your changes here, maybe should be updated separately - just created a PR for that https://github.com/Kuadrant/kuadrant-operator/pull/2146

internal/extension/registry.go
```diff
@@ -180,11 +180,11 @@ func ApplyWasmConfigMutators(wasmConfig *wasm.Config, gateway *machinery.Gateway
 	return GlobalMutatorRegistry.ApplyWasmConfigMutators(wasmConfig, gateway, topology)
 }
 
-// hasRelevantPipelineActions checks whether any wasm config mutator has pipeline actions
+// HasRelevantPipelineActions checks whether any wasm config mutator has pipeline actions
 // for policies targeting the given gateway. This checks both policies with registered
 // upstream methods and policies that only have pipeline actions (e.g., pure allow or
 // response-only extensions).
-func (r *MutatorRegistry) hasRelevantPipelineActions(targetRefs []machinery.PolicyTargetReference) bool {
+func (r *MutatorRegistry) HasRelevantPipelineActions(targetRefs []machinery.PolicyTargetReference) bool {
```
adam-cattermole: Ah I see you introduced a new `HasPipelineForTarget` convenience function - I think this means this `HasRelevantPipelineActions` could stay private (`hasRelevantPipelineActions`) and the new `HasPipelineForTarget` could be the public API as you have now - although this is just a minor nitpick

maksymvavilov:

adam-cattermole (approved):
Looks good to me, thanks!

PR #2133, opened 2026-07-24
fix: read OLM dependency versions from release.yaml instead of Quay

### LINKED ISSUE #2131: prepare-release: OLM dependency versions silently stale when bundle images not yet on Quay
## Problem

`bundle-post-generate` updates `bundle/metadata/dependencies.yaml` by calling `update-operator-dependencies.sh` for each dependency operator. This script calls `parse-bundle-version.sh` which runs `opm render $BUNDLE_IMAGE` to extract the version from the actual bundle image on Quay.

If the bundle image doesn't exist yet on Quay (common during a release — component images may not be built yet when the orchestrator runs `make prepare-release`), `opm render` fails and the old version stays in `dependencies.yaml`. The failure is silent because the error is swallowed by the pipeline.

## Impact

During the v1.4.6-rc1 release, `dependencies.yaml` had stale versions:
- `authorino-operator: "0.23.3"` (should be `0.23.4` — bundle image not yet on Quay)
- `dns-operator: "0.16.2"` (should be `0.16.3` — bundle image not yet on Quay)
- `limitador-operator: "0.17.3"` (correct — bundle image already existed)

Re-running `make prepare-release` for rc2 after all bundle images were on Quay produced the correct versions.

OLM dependency versions are minimum versions so this doesn't break installations, but it's incorrect metadata.

## Suggested Fix

Read versions from `release.yaml` directly instead of inspecting Quay images. `release.yaml` is the source of truth and is always available locally. The `opm render` approach is fragile because it creates a circular dependency: you need to build component images before you can generate the orchestrator's manifests, but the orchestrator's manifests are supposed to declare what versions it depends on.

thomasmaas:
\`update-operator-dependencies.sh\` used \`opm render\` to pull bundle images from Quay to extract the OLM package version. If the images weren't pushed yet (common during releases), it failed silently and \`dependencies.yaml\` kept stale versions.

Replace the Quay image lookup with direct version strings from the \`*_OPERATOR_VERSION\` Makefile variables (set from \`release.yaml\` by the release scripts). Delete \`parse-bundle-version.sh\` (no remaining callers), remove \`opm\` from \`bundle-post-generate\` prerequisites.

Fixes #2131

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -523,17 +523,17 @@ bundle: opm yq manifests dependencies-manifests kustomize operator-sdk ## Genera
 bundle-post-generate: OPENSHIFT_VERSIONS_ANNOTATION_KEY="com.redhat.openshift.versions"
 # Supports Openshift v4.12+ (https://redhat-connect.gitbook.io/certified-operator-guide/ocp-deployment/operator-metadata/bundle-directory/managing-openshift-versions)
 bundle-post-generate: OPENSHIFT_SUPPORTED_VERSIONS="v4.14"
-bundle-post-generate: yq opm
+bundle-post-generate: yq
 	# Set Openshift version in bundle annotations
 	$(YQ) -i '.annotations[$(OPENSHIFT_VERSIONS_ANNOTATION_KEY)] = $(OPENSHIFT_SUPPORTED_VERSIONS)' bundle/metadata/annotations.yaml
 	$(YQ) -i '(.annotations[$(OPENSHIFT_VERSIONS_ANNOTATION_KEY)] | key) headComment = "Custom annotations"' bundle/metadata/annotations.yaml
-	# Update operator dependencies
+	# Update operator dependencies from version strings (not Quay images)
 	PATH=$(PROJECT_PATH)/bin:$$PATH; \
-			 $(PROJECT_PATH)/utils/update-operator-dependencies.sh limitador-operator $(LIMITADOR_OPERATOR_BUNDLE_IMG)
+			 $(PROJECT_PATH)/utils/update-operator-dependencies.sh limitador-operator $(LIMITADOR_OPERATOR_VERSION)
 	PATH=$(PROJECT_PATH)/bin:$$PATH; \
-			 $(PROJECT_PATH)/utils/update-operator-dependencies.sh authorino-operator $(AUTHORINO_OPERATOR_BUNDLE_IMG)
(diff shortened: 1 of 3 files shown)


eguzki:
This fixes the version lookup in `bundle-post-generate` (`dependencies.yaml`), but the same Quay image dependency this PR describes as a "circular dependency" still exists in the catalog generation path and isn't addressed here:

`utils/generate-catalog.sh:21-24` still does `opm render` against `LIMITADOR_OPERATOR_BUNDLE_IMG` / `AUTHORINO_OPERATOR_BUNDLE_IMG` / `DNS_OPERATOR_BUNDLE_IMG` to pull the dependency bundles from Quay for the file-based catalog. so `make catalog` / `catalog-build` will still fail if those images aren't pushed yet, same timing issue as before, just one step later in the pipeline.

thomasmaas:
Good catch — you're right that `generate-catalog.sh` still uses `opm render` against the bundle images. That's a different problem though: the catalog genuinely needs the full rendered bundle content to build the FBC index, so it can't be replaced with a version string like `dependencies.yaml` can.

The timing constraint for catalog generation is inherent — bundle images must exist on Quay before the catalog can be built. That's already the natural release order (operator → bundle → catalog are built sequentially within each component's tag workflow). The issue we hit during 1.4.6-rc1 was specifically that `dependencies.yaml` was generated before the images existed, which this PR fixes.

If the catalog timing is also a concern we could track it separately, but I think it's already handled by the release workflow ordering.

eguzki (approved):
Worth noting: this only closes the loop because #2134 is coming right after. On its own, #2133 doesn't change *when* a missing dependency image is detected. `generate-catalog.sh` is untouched here, so `build-catalog` in `build-images-for-tag-release.yaml` will still fail the same way it always has if a dependency's bundle image isn't on Quay yet by the time the RC PR is merged and tagged. 

What #2133 *does* fix on its own, regardless of #2134, is a real, separate bug: `dependencies.yaml` no longer silently keeps a stale version when the Quay image lookup fails. That's worth merging by itself.

But the goal is to catch missing dependency images before merge (as candidate-PR CI), that only happens once #2134's pre-validation check lands. I'd have preferred landing both together, but approving this now on the understanding that #2134 follows shortly is fine.

thomasmaas:
Right, you would have preferred https://github.com/Kuadrant/kuadrant-operator/pull/2132

PR #2134, opened 2026-07-24
feat: verify dependency images exist on Quay during prepare-release

### LINKED ISSUE #2129: Validate dependency images exist on Quay before release PR merges
## Problem

The `Validate release data` CI check verifies that `make prepare-release` output is self-consistent, but does not verify that the component images referenced in `release.yaml` actually exist on Quay. This means a release PR can pass CI while upstream component images are missing, broken, or only partially built (e.g. operator image exists but bundle/catalog images don't).

During the v1.4.6-rc1 release, authorino-operator v0.23.4 had its operator image on Quay but the bundle and catalog images were missing due to a `workflow_dispatch` parameter issue. This was only caught by manual `curl` checks against the Quay API.

## Proposal

Add a CI step (or extend `verify-prepare-release`) that checks all dependency images are pullable before the release PR can merge:

```bash
# For each dependency in release.yaml, verify operator + bundle + catalog images exist
for comp in authorino-operator limitador-operator dns-operator; do
  version=$(yq ".dependencies.\"$comp\"" release.yaml)
  for suffix in "" "-bundle" "-catalog"; do
    curl -sf "https://quay.io/api/v1/repository/kuadrant/${comp}${suffix}/tag/?specificTag=v${version}" | jq -e '.tags | length > 0' || echo "MISSING: ${comp}${suffix}:v${version}"
  done
done

# Also check operand images
for comp in authorino limitador wasm-shim console-plugin developer-portal-controller; do
  # resolve version from the operator's pinned RELATED_IMAGE or release.yaml
done
```

## Impact

Without this, broken releases can ship with missing OLM bundle/catalog images, which breaks OLM-based installations. The check is cheap (HTTP calls to Quay API) and could run as part of the existing `Validate release data` job.

## Context

Discovered during the 1.4.6 patch release cycle. The authorino-operator build workflow was triggered via `workflow_dispatch` without the `authorinoVersion` parameter, causing `verify-manifests` to fail on the bundle build. The operator image was pushed but bundle and catalog were not.

thomasmaas:
Add \`verify_images.sh\` to the pre-validation phase of \`prepare-release\`. It checks all dependency images (operator + bundle + catalog for each operator dependency, plus supporting component images) exist on Quay before generating manifests. Skipped in dev mode (version 0.0.0).

Previously, missing images were only discovered after tagging and attempting OLM installation.

Fixes #2129

## Summary by CodeRabbit

- **Bug Fixes**
  - Release validation now checks that all required dependency container images are available before proceeding.
  - Missing images are reported clearly, and validation fails early to prevent incomplete releases.
  - Development configurations can skip image checks when using the default unpublished version.

### DIFF
--- a/utils/release/pre-validation/verify_images.sh
+++ b/utils/release/pre-validation/verify_images.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+
+echo "Verifying dependency images exist on Quay"
+file=$env/release.yaml
+FAILED=0
+
+check_image() {
+  local repo=$1 tag=$2
+  if curl -sf "https://quay.io/api/v1/repository/kuadrant/$repo/tag/?specificTag=$tag" | grep -q '"name"'; then
+    echo "  OK: quay.io/kuadrant/$repo:$tag"
+  else
+    echo "  MISSING: quay.io/kuadrant/$repo:$tag"
+    FAILED=1
+  fi
+}
+
+mod_version() {
+  local v=$1
+  if [[ "$v" == "0.0.0" ]]; then
+    echo "latest"
+  else
+    echo "v$v"
+  fi
+}
+
+# Skip for development (any dependency at 0.0.0 means dev mode)
+kuadrant_version=$(yq '.kuadrant-operator.version' "$file")
+if [[ "$kuadrant_version" == "0.0.0" ]]; then
+  echo "Skipping image verification for development version"
+  exit 0
+fi
+
+# Operator dependencies: check operator + bundle + catalog images
+operators=("authorino-operator" "limitador-operator" "dns-operator")
+for op in "${operators[@]}"; do
+  version=$(yq "(.dependencies.\"$op\")" "$file")
+  tag=$(mod_version "$version")
+  echo "Checking $op $tag images..."
+  check_image "$op" "$tag"
+  check_image "$op-bundle" "$tag"
+  check_image "$op-catalog" "$tag"
+done
+
+# Supporting components: check operator image only
+components=("console-plugin" "wasm-shim" "developer-portal-controller")
+for comp in "${components[@]}"; do
+  version=$(yq "(.dependencies.\"$comp\")" "$file")
+  tag=$(mod_version "$version")
+  echo "Checking $comp $tag image..."
+  check_image "$comp" "$tag"
+done
+
+if [[ $FAILED -ne 0 ]]; then
+  echo "Some dependency images are missing on Quay. Release cannot proceed."
+  exit 1
+fi
+
+echo "All dependency images verified on Quay"
--- a/utils/release/pre-validation/verify_images.sh
+++ b/utils/release/pre-validation/verify_images.sh
@@ -4,6 +4,12 @@ echo "Verifying dependency images exist on Quay"
 file=$env/release.yaml
 FAILED=0
 
+# Check Quay API is reachable before proceeding
+if ! curl -sf "https://quay.io/api/v1/discovery" > /dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


eguzki (commented):
`check_image` can't tell *why* a check failed. Quay's API doesn't map to normal HTTP semantics: a missing tag returns HTTP 200 with an empty `tags` array, while a nonexistent/misspelled repo returns 401 "Requires authentication". Both end up printed identically as `MISSING: quay.io/kuadrant/<repo>:<tag>`, so a genuinely unpublished dependency, a typo'd repo name, and a transient auth/rate-limit/network error are indistinguishable in the log.

Suggestion: replace the `curl` call against Quay's app API with `crane digest quay.io/kuadrant/$repo:$tag` (already used elsewhere in this repo's CI, e.g. `build-images-nightly.yaml`). It hits the standard OCI Distribution API, where a missing tag is a real, unambiguous error, separating "not published" from other failure modes instead of folding them all into one bucket.

thomasmaas:
Thanks @eguzki — good catch. Replaced `curl` with `crane digest` which uses the OCI Distribution API and gives unambiguous errors (missing tag vs nonexistent repo vs network failure). Also prints the digest on success for traceability.

Added `imjasonh/setup-crane@v0.1` to the `verify-release` CI job (same action + pin already used in `build-images-nightly.yaml`).

maksymvavilov:

eguzki:
Conflicts with .github/workflows/test.yaml that must be resolved.

utils/release/pre-validation/verify_images.sh
```diff
@@ -0,0 +1,90 @@
...
+set -euo pipefail
+
+if [[ -z "${env:-}" ]]; then
+  echo "ERROR: \$env is not set"
+  exit 1
+fi
+
+if ! command -v crane &>/dev/null; then
+  echo "WARNING: crane not installed, skipping image verification"
+  exit 0
```
eguzki, line 12: for a release blocking gate, it arguably should fail loudly

utils/release/pre-validation/verify_images.sh
```diff
@@ -0,0 +1,90 @@
...
+  fi
+  tag=$(mod_version "$version")
+  echo "Checking $op $tag images..."
+  check_image "quay.io/kuadrant/$op:$tag"
+  check_image "quay.io/kuadrant/$op-bundle:$tag"
+  check_image "quay.io/kuadrant/$op-catalog:$tag"
+done
+
+# Supporting components: check image only (Quay)
+components=("authorino" "limitador" "console-plugin" "wasm-shim" "developer-portal-controller")
```
eguzki, line 62: nit (feel free to ignore):  Hardcoded operator/component name lists can quickly go stale. why not test whatever found in the release.yaml file?

eguzki (commented):
Looks good. 

Some comments left

.github/workflows/code-quality.yaml
```diff
@@ -107,6 +107,9 @@ jobs:
       - name: Verify manifests
         if: ${{ !cancelled() }}
         run: make verify-manifests
+      - name: Install crane
```
eguzki, line 110: automated-release.yaml workflow should also get crane installed

eguzki (changes requested):

utils/release/pre-validation/verify_images.sh
thomasmaas: Yeah but  [release.yaml ](https://github.com/Kuadrant/kuadrant-operator/blob/feat/verify-quay-images/release.yaml) mixes components with different image conventions so then we would need a script to encode the mapping?

eguzki (approved):

PR #2147, opened 2026-07-28
fix(observability): use annotated metrics port in istio-pod-monitor

dwivedipranav-dev:
Raw-string regexes in istioPodMonitorBuild double-escaped \d/\., so __address__ relabel never matched prometheus.io/port and Prometheus fell back to scraping status port 15021 (/stats/prometheus → 404). Fixes RHAIENG-5557.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed observability relabelling patterns for Istio pod monitoring so IPv4 and IPv6 addresses are matched correctly.
  * Prevented double-escaped regular expressions from affecting pod monitor address processing.
  * Preserved existing monitoring behaviour while improving reliability for metric port rewrites.

### DIFF
--- a/internal/controller/observability_reconciler.go
+++ b/internal/controller/observability_reconciler.go
@@ -216,7 +216,10 @@ func istioPodMonitorBuild(ns string) *monitoringv1.PodMonitor {
 								"__meta_kubernetes_pod_annotation_prometheus_io_port",
 								"__meta_kubernetes_pod_ip",
 							},
-							Regex:       `(\\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
+							// Raw-string regex must use \d/\. (not \\d/\\.) so Prometheus
+							// rewrites __address__ to the annotated metrics port (15020)
+							// instead of falling back to status port 15021.
+							Regex:       `(\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
 							Replacement: &istioPodMonitorPortReplacement1,
 							TargetLabel: "__address__",
 						},
@@ -226,7 +229,7 @@ func istioPodMonitorBuild(ns string) *monitoringv1.PodMonitor {
 								"__meta_kubernetes_pod_annotation_prometheus_io_port",
 								"__meta_kubernetes_pod_ip",
 							},
-							Regex:       `(\\d+);((([0-9]+?)(\\.|$)){4})`,
+							Regex:       `(\d+);((([0-9]+?)(\.|$)){4})`,
 							Replacement: &istioPodMonitorPortReplacement2,
 							TargetLabel: "__address__",
 						},
--- a/internal/controller/observability_reconciler.go
+++ b/internal/controller/observability_reconciler.go
@@ -216,9 +216,6 @@ func istioPodMonitorBuild(ns string) *monitoringv1.PodMonitor {
 								"__meta_kubernetes_pod_annotation_prometheus_io_port",
 								"__meta_kubernetes_pod_ip",
 							},
-							// Raw-string regex must use \d/\. (not \\d/\\.) so Prometheus
-							// rewrites __address__ to the annotated metrics port (15020)
-							// instead of falling back to status port 15021.
 							Regex:       `(\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
 							Replacement: &istioPodMonitorPortReplacement1,
 							TargetLabel: "__address__",
--- a/internal/controller/observability_reconciler_test.go
+++ b/internal/controller/observability_reconciler_test.go
@@ -1,0 +1,27 @@
+package controllers
+
+import (
+	"strings"
+	"testing"
+)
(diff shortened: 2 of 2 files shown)


maksymvavilov:

internal/controller/observability_reconciler.go
```diff
@@ -216,7 +216,10 @@ func istioPodMonitorBuild(ns string) *monitoringv1.PodMonitor {
 								"__meta_kubernetes_pod_annotation_prometheus_io_port",
 								"__meta_kubernetes_pod_ip",
 							},
-							Regex:       `(\\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
+							// Raw-string regex must use \d/\. (not \\d/\\.) so Prometheus
+							// rewrites __address__ to the annotated metrics port (15020)
+							// instead of falling back to status port 15021.
+							Regex:       `(\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
```
maksymvavilov, line 219: **blocking — missing regression test**

This is a bug fix but there's no unit test asserting the correct regex values. The existing integration test (`tests/istio/observability_reconciler_test.go`) only checks that the PodMonitor is created with the right label — it never inspects the `RelabelConfig.Regex` fields. This exact class of bug (raw-string double-escaping) is easy to reintroduce.

Suggestion — add a unit test that calls `istioPodMonitorBuild("test-ns")` and asserts the two `__address__` relabel regexes contain `\d` (not `\\d`):

```go
func TestIstioPodMonitorBuild_RegexEscaping(t *testing.T) {
    pm := istioPodMonitorBuild("test-ns")
    for _, rc := range pm.Spec.PodMetricsEndpoints[0].RelabelConfigs {
        if rc.TargetLabel == "__address__" {
            if strings.Contains(rc.Regex, `\\d`) {
                t.Errorf("regex %q contains double-escaped \\d in raw string", rc.Regex)
            }
        }
    }
}
```

internal/controller/observability_reconciler.go
```diff
@@ -216,7 +216,10 @@ func istioPodMonitorBuild(ns string) *monitoringv1.PodMonitor {
 								"__meta_kubernetes_pod_annotation_prometheus_io_port",
 								"__meta_kubernetes_pod_ip",
 							},
-							Regex:       `(\\d+);(([A-Fa-f0-9]{1,4}::?){1,7}[A-Fa-f0-9]{1,4})`,
+							// Raw-string regex must use \d/\. (not \\d/\\.) so Prometheus
+							// rewrites __address__ to the annotated metrics port (15020)
+							// instead of falling back to status port 15021.
```
maksymvavilov: **nit — comment consistency**

The escaping rationale comment is added here on the IPv6 regex but not on the second (IPv4) regex at line 232, even though both suffered the same bug.

Suggestion: either add the same comment to the second regex, or (preferred) remove it from both and keep the explanation in the commit message / PR description only — the fix is now the obvious correct form and the comment will age poorly.

maksymvavilov (changes requested):
Thanks for the fix — the analysis is correct and the regex change itself is right. In Go raw strings (backticks), `\\d` produces the literal two-character sequence `\\` + `d`, which Prometheus interprets as "literal backslash then d" — never matching a port number. The fix to single `\d` / `\.` is the correct form.

Two things to address before merge:

1. **Regression test (blocking):** A unit test on the regex values would prevent this class of bug from being reintroduced. See inline comment for a concrete suggestion.

2. **Upgrade path (nit):** The reconciler uses `CreateOnlyMutator` — if the PodMonitor already exists with the broken regex, upgrading the operator alone won't fix it. The stale monitor persists. Please add an upgrade note to the PR description: users on existing clusters need to either delete the `istio-pod-monitor` PodMonitor or toggle `observability.enable` off then on to pick up the fix. Consider whether a follow-up to use a mutating reconciler for monitors is warranted.

internal/controller/observability_reconciler.go
dwivedipranav-dev: Hello @maksymvavilov 
I have made the changes that you have asked for and below is the test result:
```
go test ./internal/controller/ -run TestIstioPodMonitorBuild_RegexEscaping -v -count=1
=== RUN   TestIstioPodMonitorBuild_RegexEscaping
--- PASS: TestIstioPodMonitorBuild_RegexEscaping (0.00s)
PASS
ok  	github.com/kuadrant/kuadrant-operator/internal/controller	0.023s
```

maksymvavilov (approved):
Looks good and fixes the [RHAIENG-5557](https://redhat.atlassian.net/browse/RHAIENG-5557)

One nit:
The regression test catches double-escaped `\d` but not double-escaped `\.`. The IPv4 regex also had a `\\.` → `\.` fix. If someone reintroduces `\\.` in the future, this test won't catch it. Consider adding:

```go
if strings.Contains(rc.Regex, `\\.`) {
    t.Errorf("regex %q contains double-escaped \\. in raw string", rc.Regex)
}
```

LGTM otherwise

maksymvavilov:
I'll re-run the integration test: it seems to be a fluke

Issue #2151, opened 2026-07-29
Port existing egress docs to downstream (RHCL) format


maksymvavilov:
## Context

The RHCL 1.4 developer preview produced three upstream egress user guides that have no downstream counterparts yet:

- `doc/user-guides/egress/egress-gateway.md` — egress gateway setup
- `doc/user-guides/egress/dns-routing.md` — DNS routing for egress
- `doc/user-guides/egress/credential-injection.md` — credential injection via Vault

The TP feature issues (#2083, #2084, #2085, #2086) each include "upstream + downstream" docs for **new** capabilities, but there is no task covering the migration of these **existing** guides to the downstream RHCL format in openshift-docs.

## Acceptance Criteria

- Each of the three existing upstream egress guides is ported to downstream AsciiDoc format
- Downstream docs follow RHCL conventions (product names, prerequisites, admonitions, etc.)
- PRs are submitted to the downstream docs repo
- Content is reviewed for accuracy against the current upstream guides

PR #1896, opened 2026-04-16
docs: egress gateway setup and rate limiting guide

maksymvavilov:
## Summary

- Adds `doc/user-guides/egress/egress-gateway.md` — the entry point for egress documentation covering setup, workload identity, and rate limiting
- Part of egress documentation restructure (Approach B): setup+policies in this doc, routing in dns-routing.md (#1856), credential injection in credential-injection.md (#1893)

### What the guide covers

- Egress gateway infrastructure setup (Gateway, ServiceEntry, DestinationRule, HTTPRoute)
- Workload identity via `kubernetesTokenReview` — shared infrastructure for both RLP and AuthPolicy
- Rate limiting: basic global limits and per-workload limits using SA identity counters
- Reference table linking to existing ingress rate limiting guides
- Auth overview with link to credential injection guide

### Key findings from validation

- `source.address` cannot be used as a rate limit counter (includes ephemeral port, making each connection unique; pod IPs are also ephemeral)
- Per-workload rate limiting requires AuthPolicy (for identity) + RLP (with `auth.identity.username` counter)
- AuthPolicy selectors and RLP expressions resolve identity paths differently by design (documented)

### Merge order

This PR should merge first, then #1856 (dns-routing.md), then #1893 (credential-injection.md).

Closes #1801

## Test plan

- [x] Basic egress RLP: 3 requests pass, 4th returns 429
- [x] Per-workload RLP with SA counter: two different SAs get independent rate limit buckets
- [x] `source.address` counter validated as non-functional (documented as gap)
- [x] Workload identity fields verified via kubernetesTokenReview

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive user guide for configuring Istio egress gateways with Kuadrant, including setup instructions, rate limiting policies for outbound traffic, workload identity and authorisation controls, per-workload rate limiting, and operational considerations for managing egress traffic.

PR #2152, opened 2026-07-30
fix(docs): add identity filter to egress workload identity AuthPolicy

maksymvavilov:
## Summary

- Adds `response.success.filters` section to the workload-identity `AuthPolicy` in the egress gateway guide
- Without this, the wasm-shim cannot resolve `auth.identity.username` in RateLimitPolicy counter expressions, causing per-workload rate limiting to silently fail (`failureMode: allow` lets all requests through)
- Matches the pattern used in the ingress authenticated rate limiting guide (`authenticated-rl-with-jwt-and-k8s-authnz.md`)

## Context

The gap was introduced in PR #1896 (April 2026) which added the egress gateway doc. The workload-identity AuthPolicy was missing the `response.success.filters` section needed to expose identity as dynamic metadata for RLP counter expressions. The ingress guide already had this section, but it was missed for egress.

Discovered during downstream doc porting (#2151). The `NoSuchKey("identity")` CEL error in wasm-shim logs confirmed the AuthConfig had `numResponseItems: 0` — Authorino authenticated requests correctly but did not pass identity data back through dynamic metadata.

## Test plan

- [x] Verified on Kind cluster: without the filter, all 10 requests return 200 (counter fails silently); with the filter, requests 1-5 return 200, requests 6-10 return 429
- [x] Confirmed AuthConfig shows `numResponseItems: 1` after the fix

Fixes part of #2151

### DIFF
--- a/doc/user-guides/egress/egress-gateway.md
+++ b/doc/user-guides/egress/egress-gateway.md
@@ -209,11 +209,20 @@ spec:
         patternMatching:
           patterns:
             - predicate: auth.identity.user.username.startsWith('system:serviceaccount:egress-test:')
+    response:
+      success:
+        filters:
+          "identity":
+            json:
+              properties:
+                username:
+                  selector: auth.identity.user.username
 EOF
 ```
 
 - `authentication.workload-sa` - validates the SA token via TokenReview. Unauthenticated requests are rejected (401).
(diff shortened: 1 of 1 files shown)


averevki (approved):
Makes sense, this corresponds with configuration from the ingress doc that you mentioned:
https://github.com/Kuadrant/kuadrant-operator/blob/77006cd7f009e05ab807745aba3e5e2818670021/doc/user-guides/ratelimiting/authenticated-rl-with-jwt-and-k8s-authnz.md?plain=1#L199-L206

PR #2164, opened 2026-08-04
fix(docs): rename wasm-shim service to kuadrant-filter

silvi-t:
## Summary
  - Rename `wasm-shim` to `kuadrant-filter` in tracing documentation to reflect the service rename in [wasm-shim#399](https://github.com/Kuadrant/wasm-shim/pull/399)

## Summary by CodeRabbit

* **Documentation**
  * Updated tracing documentation to consistently reference `kuadrant-filter`.
  * Revised configuration, request-flow, trace attributes, Jaeger queries, correlation, and troubleshooting guidance.
  * Updated documented service names, resource descriptions, and log-level guidance.

### DIFF
--- a/doc/observability/tracing.md
+++ b/doc/observability/tracing.md
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-This guide outlines the steps to enable tracing in Istio and Kuadrant components (Authorino, Limitador, and wasm-shim), directing traces to a central collector for improved observability and troubleshooting. We'll also explore a typical troubleshooting flow using traces and logs.
+This guide outlines the steps to enable tracing in Istio and Kuadrant components (Authorino, Limitador, and Wasm-shim), directing traces to a central collector for improved observability and troubleshooting. We'll also explore a typical troubleshooting flow using traces and logs.
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


doc/observability/tracing.md
```diff
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-This guide outlines the steps to enable tracing in Istio and Kuadrant components (Authorino, Limitador, and wasm-shim), directing traces to a central collector for improved observability and troubleshooting. We'll also explore a typical troubleshooting flow using traces and logs.
+This guide outlines the steps to enable tracing in Istio and Kuadrant components (Authorino, Limitador, and kuadrant-filter), directing traces to a central collector for improved observability and troubleshooting. We'll also explore a typical troubleshooting flow using traces and logs.
```
adam-cattermole: I'm not sure the best way to write this - I think there's the chance that using `kuadrant-filter` when referencing the component name is a bit confusing here as the component is still packaged and referenced as the `wasm-shim` but where the service in the traces changed. I wonder if it's best to leave it as the `wasm-shim` when talking about the component and then quickly mention `wasm-shim` exports as service `kuadrant-filter` in the bit before talking about the services/spans specifically? wdyt?

doc/observability/tracing.md
```diff
@@ -482,7 +482,7 @@ Request Flow (29.4ms total):
 
 Data plane traces include rich context from multiple services:
 
-**Wasm-shim Spans** (service: `wasm-shim`):
+**kuadrant-filter Spans** (service: `kuadrant-filter`):
```
adam-cattermole: Maybe this switches to:
```suggestion
**wasm-shim Spans** (service: `kuadrant-filter`):
```

doc/observability/tracing.md
silvi-t: Thanks for thinking this through. You're right, using `kuadrant-filter` when referring to the component is confusing. I've updated the docs to use `wasm-shim` when talking about the component and only `kuadrant-filter` where it refers to the service name in traces. I also added a note on clarifying that `wasm-shim` exports traces as service `kuadrant-filter`.

adam-cattermole (approved):
Thanks, lgtm!

PR #2166, opened 2026-08-05
Extensions handshake and session management

### LINKED ISSUE #2137: Design/implement extensions handshake
A means to allow 'permitted' extensions to connect to the operator/announce their presence. We should decided whether this also behaves as the liveness check, or expand on the existing. Requires updating the protos

We need to think about authN for the extensions - initial thoughts are a secret in the kuadrant-system namespace with pub/private key exchange. We need a solution that also takes into account the local in process extensions communicating over localhost - one thing to decide is if these require the deployment of the secret by the operator or 'circumvent auth' and are assumed safe.

adam-cattermole:
Resolves #2137

## Summary by CodeRabbit

- **New Features**
  - Added an extension handshake for exchanging identity, credentials, version information and policy details.
  - Added session-based authentication for extension requests, including streaming connections.
  - Added secure session tokens with validation, revocation and connection-management safeguards.
  - Added clear responses for rejected handshakes, including invalid credentials, missing information and conflicting policy types.

- **Tests**
  - Added comprehensive coverage for successful and rejected handshakes, session lifecycle behaviour and concurrent access.


silvi-t (approved):
LGTM

PR #2171, opened 2026-08-07
Replace EnvoyFilter targetRef with workloadSelector

adam-cattermole:
I've seen that on Istio v1.26.x we unintentionally affect all gateways in the target namespace rather than the targeted gateway only - due to this istio issue https://github.com/istio/istio/issues/56417

While we still support istio v1.26.x we should use `workloadSelector` instead of `targetRef`.

To reproduce, I created two `Gateway`s and `HTTPRoute`s with the same hostname and backend, and created an `AuthPolicy` targeting just one. By inspecting envoy config on the two gateways I could see both were affected by our `EnvoyFilter`s, and requests to either `Gateway` are affected by the `AuthPolicy` 

## Enhancements

- EnvoyFilters now target gateways using gateway-name workload selectors.
- Improved consistency across authentication, extension, rate-limiting and tracing integrations.
- Existing configurations with outdated target references are migrated automatically.

## Bug Fixes

- Resolved reconciliation issues where EnvoyFilters could retain stale gateway associations.
- Improved gateway-to-EnvoyFilter matching by considering gateway name and namespace.

### DIFF
--- a/internal/controller/istio_auth_cluster_reconciler.go
+++ b/internal/controller/istio_auth_cluster_reconciler.go
@@ -10,7 +10,6 @@ import (
 	"github.com/kuadrant/policy-machinery/machinery"
 	"github.com/samber/lo"
 	istioapinetworkingv1alpha3 "istio.io/api/networking/v1alpha3"
-	istiov1beta1 "istio.io/api/type/v1beta1"
 	istioclientgonetworkingv1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


maleck13 (commented):
changes look well contained. One thought is whether it would be valuable to add a test that proves that when the reconciler compares a pre-upgrade EnvoyFilter against the desired new-style one, EqualEnvoyFilters returns false and the update path
  fires?

adam-cattermole:
> changes look well contained. One thought is whether it would be valuable to add a test that proves that when the reconciler compares a pre-upgrade EnvoyFilter against the desired new-style one, EqualEnvoyFilters returns false and the update path fires?

Yep good idea, I think I tried to cover this with the integration test but I could also add a unit test for that function that confirms the change works

maleck13 (approved):

PR #2173, opened 2026-08-10
feat: kuadrant finalizer removes AuthConfig

DaliborD45:
## Summary

  - Add a `kuadrant.io/authconfigs` finalizer to the Kuadrant CR so that managed AuthConfig resources are cleaned up before the CR is deleted
  - When the Kuadrant CR is marked for deletion, the `AuthConfigsReconciler` now deletes all managed AuthConfigs (matched by label selector) and removes the finalizer
  - When the Kuadrant CR is not being deleted, the reconciler ensures the finalizer is present before proceeding with normal reconciliation

  ## Changes
  - **`authconfigs_reconciler.go`**: Added finalizer lifecycle management (add on reconcile, remove on deletion after cleaning up AuthConfigs). Changed `client` field type from `*dynamic.DynamicClient` to `dynamic.Interface` to support testing with fake clients. Embedded `BaseReconciler` for finalizer helper methods.
  - **`data_plane_policies_workflow.go`**: Updated `AuthConfigsReconciler` instantiation to pass a `BaseReconciler` with the manager's client, scheme, and API reader.
  - **`authconfigs_reconciler_test.go`**: Added comprehensive tests covering deletion with and without Authorino present, no-op when Kuadrant CR is absent from topology, and no-op when Kuadrant CR is not being deleted.

## Summary by CodeRabbit

- **Bug Fixes**
  - AuthConfig resources are now automatically cleaned up when the associated Kuadrant resource is deleted.
  - Cleanup completes safely before deletion finishes, preventing leftover managed resources.
  - Normal AuthConfig reconciliation is paused while deletion is in progress.
  - Finalisers are added and removed appropriately, allowing deletion to complete correctly.
  - Missing resources are handled safely during cleanup.
  - Cleanup and deletion errors are recorded for follow-up without blocking reconciliation.
  - Developer Portal resource updates and cleanup now handle errors more reliably.


internal/controller/authconfigs_reconciler_test.go
```diff
@@ -4,14 +4,224 @@ package controllers
...
+				&controller.RuntimeObject{Object: authConfig2},
+			),
+		)
+		assert.NilError(subT, err)
+
+		err = reconciler.Reconcile(context.TODO(), nil, topology, nil, &sync.Map{})
+		assert.NilError(subT, err)
+
+		// AuthConfigs should still exist
+		_, err = fakeDynClient.Resource(kuadrantauthorino.AuthConfigsResource).Namespace("kuadrant-system").Get(context.TODO(), "authconfig1", metav1.GetOptions{})
```
maksymvavilov: nit: The Kuadrant CR is added to the topology but not registered with the `fakeCtrlClient`. When `Reconcile` runs, `AddFinalizer` calls `UpdateResource` which returns `NotFound` — silently ignored by `client.IgnoreNotFound`. So this test passes without actually verifying the finalizer is added.

Consider registering the CR with the fake client and asserting the finalizer is present after reconciliation:

```go
fakeCtrlClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(kuadrantCR).Build()
// ... after reconcile ...
updatedKuadrant := &kuadrantv1beta1.Kuadrant{}
err = fakeCtrlClient.Get(context.TODO(), client.ObjectKeyFromObject(kuadrantCR), updatedKuadrant)
assert.NilError(subT, err)
assert.Assert(subT, controllerutil.ContainsFinalizer(updatedKuadrant, authConfigFinalizer),
    "finalizer should be added to Kuadrant CR")
```

internal/controller/authconfigs_reconciler_test.go
```diff
@@ -4,14 +4,224 @@ package controllers
...
+		assert.NilError(subT, err, "authconfig1 should exist before reconciliation")
+		_, err = fakeDynClient.Resource(kuadrantauthorino.AuthConfigsResource).Namespace("kuadrant-system").Get(context.TODO(), "authconfig2", metav1.GetOptions{})
+		assert.NilError(subT, err, "authconfig2 should exist before reconciliation")
+
+		topology := buildTopologyWithKuadrantForDeletion(subT, authConfig1, authConfig2)
+		err = reconciler.Reconcile(context.TODO(), nil, topology, nil, &sync.Map{})
+		assert.NilError(subT, err)
+
+		// verify AuthConfigs were deleted
+		_, err = fakeDynClient.Resource(kuadrantauthorino.AuthConfigsResource).Namespace("kuadrant-system").Get(context.TODO(), "authconfig1", metav1.GetOptions{})
```
maksymvavilov, line 160: nit: Same issue here — the Kuadrant CR from the topology isn't in the `fakeCtrlClient`, so `RemoveFinalizer` → `UpdateResource` returns `NotFound` which is silently ignored. The finalizer removal path isn't truly exercised.

Consider adding the Kuadrant CR to the fake client (`.WithObjects(kuadrantCR)` where `kuadrantCR` is the one with the deletion timestamp and finalizer) and asserting the finalizer is gone after reconciliation, to get true end-to-end coverage of the finalizer lifecycle.

Note: you'd need to build the `kuadrantCR` used in `buildTopologyWithKuadrantForDeletion` separately so it can be passed to both the topology builder and the fake client.

internal/controller/authconfigs_reconciler.go
```diff
@@ -55,6 +59,48 @@ func (r *AuthConfigsReconciler) Subscription() controller.Subscription {
...
+				return err
+			}
+		}
+
+		logger.Info("auth configs deleted, removing finalizer from Kuadrant CR")
+		kObjCopy := kObj.DeepCopy()
+		if err := r.RemoveFinalizer(ctx, kObjCopy, authConfigFinalizer); err != nil {
+			logger.Error(err, "failed to remove finalizer")
+			return err
+		}
```
maksymvavilov, line 109: So here we remove finalizer once we send a "delete" request to all the CRs. We are not checking that CRs are gone. As far as I know, we don't put any finalizers on `AuthConfig`, but we might, eventually. 

The [example](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/controller/developerportal_reconciler.go#L80-L92) in the dev portal suffers from the same problem - controller deletes `Deployment` and does not check that it is gone. This is a problem with the pattern

maksymvavilov (approved):
This is a good implementation of existing pattern. 
The pattern is questinable, but this is still an improvement. 
There are a few nits that are "good to have"s and the tests are not happy here. 
The failures seem to be caused by naughty GH workflows.

internal/controller/authconfigs_reconciler.go
```diff
@@ -55,6 +59,48 @@ func (r *AuthConfigsReconciler) Subscription() controller.Subscription {
...
+	}
+
+	// if deletion of kuadrant CR initiated, return
+	if kObj.GetDeletionTimestamp() != nil {
+		return nil
+	}
+
+	if !controllerutil.ContainsFinalizer(kObj, authConfigFinalizer) {
+		kObjCopy := kObj.DeepCopy()
+		if err := r.AddFinalizer(ctx, kObjCopy, authConfigFinalizer); err != nil {
```
maksymvavilov: Never mind - the failing test is this PR's work (kind of). 
We should either `Patch()` or fetch a fresh copy just before updating. 

When deleting, we now have 2 controllers [updating](https://github.com/Kuadrant/kuadrant-operator/blob/c7bad9eed7c31cf1e45d22150752648f40717eb1/internal/reconcilers/base_reconciler.go#L200-L203) CR and one of them will loose the race.

maksymvavilov (changes requested):

internal/controller/authconfigs_reconciler.go
DaliborD45: I do not really understand why we should check if AuthConfigs are deleted again. On line 76 we have a check that if it does not delete and raises a err, we log it. Can you maybe elaborate more please ?

internal/controller/authconfigs_reconciler.go
```diff
@@ -55,6 +59,48 @@ func (r *AuthConfigsReconciler) Subscription() controller.Subscription {
...
+				logger.Error(err, "failed to delete auth config", "namespace", ac.GetNamespace(), "name", ac.GetName())
+				return err
+			}
+		}
+
+		logger.Info("auth configs deleted, removing finalizer from Kuadrant CR")
+		kObjCopy := kObj.DeepCopy()
+		if err := r.RemoveFinalizer(ctx, kObjCopy, authConfigFinalizer); err != nil {
+			logger.Error(err, "failed to remove finalizer")
+			return err
```
maksymvavilov: Only now I noticed that we are not using the `ErrorRegistry` here and that this copy does not have it. This technically does not change functionality, but the `ErrorRegistry` was added to centralise the way we deal with errors and prevent premature cancellation of the workflows. Meaning, if we return here, all workflows/controllers/reconcilers/whatewer that are next in line after this one will be skipped until the retry. 

I'd encourage a rebase on the fresh main and using an error registry in the PR

internal/reconcilers/base_reconciler.go
```diff
@@ -233,3 +233,23 @@ func (b *BaseReconciler) RemoveFinalizer(ctx context.Context, obj client.Object,
...
 }
+
+// GetAndRemoveFinalizer fetches a fresh copy of the object from the API server,
+// which sends the stale `desired` object in a dry-run before the mutateFn runs.
+func (b *BaseReconciler) GetAndRemoveFinalizer(ctx context.Context, obj client.Object, finalizer string) error {
+	key := client.ObjectKeyFromObject(obj)
+	fresh := obj.DeepCopyObject().(client.Object)
+	if err := b.Client().Get(ctx, key, fresh); err != nil {
+		return client.IgnoreNotFound(err)
+	}
```
Boomatang: Getting resources directly from the cluster is an anti pattern of using the topology and the policy-machinery. Is there any alternative approach that doesn't require direct access to the cluster.

DaliborD45: In this case I see only two options: Either get resources directly from cluster ( because if we do not, we run into race condition when removing finalizer), or the second option is to use Patch, which I dont know if it is a best option either ( as discussed here https://github.com/Kuadrant/kuadrant-operator/pull/2173#discussion_r3750775176). However I am not really experienced with this so maybe you can also share your opinion, thanks.

DaliborD45: Okay so I have tried something like: 
```
patch := client.RawPatch(
		types.StrategicMergePatchType,
		[]byte(fmt.Sprintf(
			`{"metadata":{"$deleteFromPrimitiveList/finalizers":[%q]}}`,
			finalizer,
		)),
	)
	err := b.Client().Patch(ctx, obj, patch)
	
```
Unfortunately, this seems to trigger the race condition, which I do not know really why. The only working option I got is with Get(), which as you mention is not ideal. Can you maybe help me with this ?

DaliborD45: So I have tried couple of other approaches, like fetching new state in the kuadrant_controller file, when operator is trying to retry finalizers, using Patch instead of Get. Unfortunately, none of it works and this anti-pattern approach is the only approach that is actually working.

DaliborD45:
Closing this PR, because I eventually chose completely different approach, continuing here: https://github.com/Kuadrant/kuadrant-operator/pull/2185

PR #2177, opened 2026-08-12
fix: stop EnvoyFilter reconcile storm via merge patch

eformat:
● Problem

  IstioExtensionReconciler in the kuadrant-operator enters a self-perpetuating write loop, continuously updating the kuadrant-<gateway-name> EnvoyFilter resource at ~0.5–1 writes/second indefinitely.

  The loop:
  1. Reconciler computes desired EnvoyFilter spec and calls resource.Update()
  2. Update() bumps resourceVersion in etcd — even when the spec bytes are identical to what's already stored
  3. The reconciler is subscribed to EnvoyFilterGroupKind watch events, so the write fires a watch event that immediately re-enters Reconcile
  4. If the informer cache hasn't yet synced the post-update state, the reconciler reads the stale pre-update object, compares it against desired, concludes they differ, and issues another Update()
  5. Go to 2

  The EqualEnvoyFilters guard is intended to short-circuit this, but is unreliable when the informer is behind the API server — which is exactly the situation created by the high-frequency writes.

  Impact on single-node clusters (SNO):

  - etcd sustains continuous high write load, increasing 99th-percentile write latency
  - All operators sharing the API server begin missing leader lease renewals (context deadline exceeded) and crash-loop — including the kuadrant-operator itself, kube-controller-manager, and the
  etcd-operator
  - Istiod receives continuous xDS push updates from EnvoyFilter churn, causing gateway Envoy pods to repeatedly reload their Wasm extension
  - Each Wasm reload fetches the binary remotely; under rapid reload cycles memory accumulates and gateway pods OOMKill
  - The HPA responds to the CPU/memory spike by scaling gateway pods to maximum replicas, amplifying the OOM cascade
  - The RHCL ClusterServiceVersion cycles between Succeeded and Installing because the kuadrant-operator deployment repeatedly loses availability

  The cluster remains in this degraded state until the kuadrant-operator is scaled to zero.

---

●  Fix 1 — resource.Update() → resource.Patch() (MergePatchType)

  File: internal/controller/istio_extension_reconciler.go
  Two call sites: the wasm EnvoyFilter update and the upstream cluster EnvoyFilter update.

  Why: resource.Update() always writes to etcd and bumps resourceVersion, even when the object spec hasn't changed. The reconciler watches EnvoyFilterGroupKind events (it subscribes to its own writes),
  so every Update fires a watch event → re-enters Reconcile → compares existing vs desired → issues another Update. Infinite loop.

  A merge patch (resource.Patch() with MergePatchType) sends only the spec payload. The API server diffs it against the stored object — if nothing changed, no write, no resourceVersion bump, no watch 
  event. Loop cannot start.

  Change:
  // Before — both call sites look like this:
  if _, err = resource.Update(ctx, existingEnvoyFilterUnstructured, metav1.UpdateOptions{}); err != nil {
      logger.Error(err, "failed to update envoyfilter object", ...)
      
  // After:
  patchBytes, err := json.Marshal(map[string]interface{}{"spec": existingEnvoyFilterUnstructured.Object["spec"]})
  if err != nil {
      logger.Error(err, "failed to marshal envoyfilter patch", ...)
      continue
  }   
  if _, err = resource.Patch(ctx, existingEnvoyFilter.GetName(), k8stypes.MergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
      logger.Error(err, "failed to patch envoyfilter object", ...)

  Add "encoding/json" to the import block (k8stypes alias for k8s.io/apimachinery/pkg/types is already imported).

  ---
 ● Fix 2 — EqualEnvoyFilters CLUSTER nil guard

  File: internal/istio/utils.go
  Function: EqualEnvoyFilters, CLUSTER case.

  Why: The nil guard uses || (OR) instead of !=. When both aCluster and bCluster are nil — two CLUSTER patches with no explicit match selector, which are equal — the function returns false (not equal).
  This is a false negative that causes the reconciler to issue an Update when it shouldn't, undermining the equality check that's supposed to prevent redundant writes.

  The HTTP_FILTER case directly above it uses the correct != pattern.

  Change:
  // Before:
  if aCluster == nil || bCluster == nil {
      return false
  }   

  // After — consistent with HTTP_FILTER nil guard at line 188:
  if (aCluster == nil) != (bCluster == nil) {
      return false
  }

---

● Fix 3: 
  cmd/extensions/telemetry-policy/internal/controller/telemetrypolicy_reconciler.go

  When the RHCL developer portal is not configured on the cluster, kuadrantCtx.AddDataTo returns
  rpc error: code = Unauthenticated desc = missing session token for every TelemetryPolicy
  reconcile. The current code treats this as a hard error:

  return calculateErrorStatus(pol, err), err   // returns the error

  Returning the error causes the controller to requeue the reconcile. On the next run the same
  Unauthenticated error occurs, triggering another requeue — an infinite retry loop. Each retry
  calls reconcileStatus which updates the TelemetryPolicy status conditions, and each status
  update emits a watch event. Those watch events cascade into the IstioExtensionReconciler (via
  ConfigMap updates in the topology), which rebuilds and re-patches the
  kuadrant-maas-default-gateway EnvoyFilter — causing the same write storm on SNO clusters as the
  resource.Update() bug in the main reconciler.

  Fix

  Detect the Unauthenticated gRPC status code and treat it as "developer portal not configured"
  rather than an error. Record an informational "not enforced" condition on the TelemetryPolicy
  (so the status reflects the feature is unavailable) but return nil — no requeue, no status
  churn, no downstream storm:

// Use string check because the gRPC error may be wrapped by the extension framework.
if strings.Contains(err.Error(), "code = Unauthenticated") {
    r.Logger.Info("developer portal not configured, telemetry binding skipped", "binding", binding)
    return calculateEnforcedStatus(pol, err), nil
}

  Effect

  - TelemetryPolicy reconciler stops looping on clusters without a developer portal
  - Status conditions correctly reflect "enforced with caveat" rather than perpetual error
  - ConfigMap update churn stops → IstioExtensionReconciler stops being triggered by telemetry
  policy failures → EnvoyFilter write rate drops to zero when content is unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved updates to gateway and upstream networking configuration for more reliable application.
  * Corrected comparison handling when both cluster match values are absent.
  * Preserved existing error reporting when configuration updates fail.


eformat:
Test after applying new image with these fixes - Generation frozen at 3017 across 40 seconds — storm stopped

```
oc get envoyfilter kuadrant-maas-default-gateway -n openshift-ingress -o jsonpath='gen={.metadata.generation}{"\n"}'
sleep 10
oc get envoyfilter kuadrant-maas-default-gateway -n openshift-ingress -o jsonpath='gen={.metadata.generation}{"\n"}'
gen=3017
gen=3017
```

eformat:
will change this to DRAFT on new PR

PR #2170, opened 2026-08-06
feat: add component chart sync tool and automation

mikenairn:
## Summary

Implements chart sync tooling and GitHub Actions automation for syncing upstream component Helm charts into kuadrant-operator, per [RFC 0019](https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md).

- Go CLI tool (`hack/sync-components/`) with `sync`, `find-targets`, and `query` subcommands
- Config driven via `component-charts/sync.yaml` declaring repo, chart path, tracked branch, and auto-merge per component
- `sync-component-chart.yaml` workflow receives `repository_dispatch` from component repos, syncs charts, and creates PRs
- Ref pinning: sync resolves tracked branch to a commit SHA and updates `sync.yaml`
- PR body includes upstream compare link and actor attribution

Starting with dns-operator only. Other components to be enabled incrementally.

### Follow-ups
- Scheduled daily reconciliation as safety net for missed events
- Release branch sync support (deferred until branch mapping strategy is defined)
- Image digest pinning ([release process simplification](https://gist.github.com/mikenairn/4c729ddd60a6129f84c6919b007430e6))

closes #2154
closes #2155

## Summary by CodeRabbit

* **New Features**
  * Added the DNS Operator Helm chart, including custom resources, monitoring, health checks and access controls.
  * Added automated synchronisation for component Helm charts, with branch tracking, change detection, pull request creation and optional automatic merging.
  * Added commands to synchronise all charts or a selected chart, query configuration and identify synchronisation targets.

* **Documentation**
  * Added installation guidance for the DNS Operator chart and documentation for chart synchronisation, configuration and generated files.

Issue #2154, opened 2026-07-30
Implement chart sync tool for child operator Helm charts


mikenairn:
## Summary

Implement a sync tool (`hack/sync-components/`) that downloads complete upstream Helm charts as-is into `config/sync-components/charts/`. Add `make sync-component-charts` target. Must be runnable manually at any time to pull the latest from tracked branches. Sync dns-operator as the first component, extend to all four child operators.

## Acceptance Criteria

- [x] Sync tool downloads complete upstream charts (Chart.yaml, values.yaml, templates/, crds/) without modification
- [x] `make sync-component-charts` target runs the tool for all child operators
- [x] Can be run manually at any time to update charts from tracked branches
- [x] dns-operator synced as first component
- [x] Synced charts committed to the repo

## References

- [RFC 0019: Kuadrant Operator Consolidation](https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md)
- Parent issue: Kuadrant/architecture#164

PR #2178, opened 2026-08-12
feat: consolidate dns-operator into kuadrant-operator

### LINKED ISSUE #2156: Consolidate dns-operator into kuadrant-operator
## Summary

Deploy dns-operator as a child controller managed by kuadrant-operator at runtime via Helm chart rendering. This is the first child operator and establishes the pattern for all others.

## Scope

- Startup deployer: render chart, apply CRDs following Helm install ordering, deploy controller in operator namespace
- RBAC: scoped CRD create/update permissions, `escalate` for ClusterRoles, `bind` for ClusterRoleBindings
- `RELATED_IMAGE_DNS_OPERATOR` env var in `config/manager/manager.yaml`
- Remove dns-operator from OLM `dependencies.yaml` and bundle
- OLM migration: programmatic cleanup of orphaned dns-operator Subscription/CSV
- Helm chart build updates
- OLM upgrade test

## Acceptance Criteria

- [x] dns-operator controller deployed at kuadrant-operator startup from embedded Helm chart
- [x] CRDs applied first and established before controller manager starts
- [x] Controller runs in operator namespace (e.g. `kuadrant-system`)
- [x] dns-operator CRDs not in OLM bundle (no GVK conflicts during upgrade)
- [x] OLM upgrade from current multi-operator installation works without manual intervention
- [x] Orphaned dns-operator Subscription/CSV cleaned up programmatically
- [x] `make local-setup` deploys dns-operator without separate dependency step

## References

- [RFC 0019: Kuadrant Operator Consolidation](https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md)
- Depends on: Kuadrant/kuadrant-operator#2154
- Parent issue: Kuadrant/architecture#164

mikenairn:
> **Blocked by #2170** (chart sync tool and automation)

## Summary

Deploy dns-operator as a child controller managed by the kuadrant-operator at runtime via Helm chart rendering. This is the first child operator consolidated under [RFC 0019](https://github.com/Kuadrant/architecture/blob/main/rfcs/0019-olmv1-operator-consolidation.md), establishing the pattern for authorino-operator, limitador-operator, and mcp-gateway.

Introduces a `KuadrantControlPlane` CRD (cluster-scoped singleton) that drives component deployment and reports control plane health, separate from the data plane `Kuadrant` CR.

### Changes

- **KuadrantControlPlane CRD** (`api/v1alpha1/`) — cluster-scoped singleton with CEL validation, self-healing on deletion, `Ready` condition, per-component status with CRD establishment tracking, deployed image reporting
- **Helm renderer** (`internal/helm/`) — chart loading, Helm SDK rendering (`ClientOnly`, `DryRun`), CRD extraction from both `templates/` and `crds/` directories
- **Control plane controller** (`internal/controlplane/`) — standard controller-runtime reconciler with Deployment and CRD watches for event-driven status convergence, periodic 5-minute drift reconciliation, independent component deployment (fail-continue not fail-fast)
- **Kubernetes Events** — lifecycle audit trail for component install/upgrade/failure, dependency detection, control plane bootstrap, and OLM migration cleanup (visible via `kubectl events` / `kubectl describe`)
- **Two-phase startup** — pre-manager CRD bootstrap (required for PolicyMachineryController boot detection), then controller-driven deployment via KuadrantControlPlane CR
- **OLM migration** — removes dns-operator from `dependencies.yaml`, dynamically discovers and cleans up all OLM-owned resources (labels, ownerReferences) via label selectors. Hardened with metadata-first cleanup, per-component event reporting, and Forbidden error handling. One-off code, removable after ~3 releases
- **Resource labelling** — all rendered resources stamped with `app.kubernetes.io/managed-by: kuadrant-operator`
- **RBAC** — scoped CRD/ClusterRole/ClusterRoleBinding permissions with `escalate`/`bind` and `resourceNames` restrictions, `list`/`watch` on unscoped CRD rule for informer cache, dual API group grants for events (`""` and `events.k8s.io`)
- **Image override** — `RELATED_IMAGE_DNS_OPERATOR` env var with post-render patching (fallback) and chart values path (for charts that support it). Status reports actual deployed images extracted from rendered manifests
- **Integration tests** (`tests/common/controlplane/`) — Serial execution, included in `test-integration` via `tests/common/...`
- **Build changes** — Dockerfile embeds `component-charts/`, `CHARTS_PATH` env var for local dev, dns-operator dev ConfigMap for `inmemory` provider

### What's NOT changed

- Wrapper CRs (Authorino CR, Limitador CR) preserved
- Authorino-operator and limitador-operator still deployed via OLM (consolidation in later phases)
- No changes to data plane behaviour or policy reconciliation
- Existing CRDs, Deployments, and resources all remain

Depends on: #2154 (chart sync tool — merged)  
Relates to: Kuadrant/architecture#164

### Local (Kind)

```bash
# Set up cluster and deploy
make local-setup

# Verify dns-operator deployed by kuadrant-operator
kubectl get kcp
kubectl get deployments -n kuadrant-system
kubectl get crds | grep dns

# Verify status
kubectl get kcp default -o yaml

# Verify events
kubectl events --for kuadrantcontrolplane.kuadrant.io/default

# Test drift reconciliation
kubectl delete deployment dns-operator-controller-manager -n kuadrant-system
# Wait ~5 minutes or watch for immediate recreation (Deployment watch)
kubectl get deployments -n kuadrant-system -w

# Test self-healing
kubectl delete kcp default
kubectl get kcp  # should be re-created immediately
```

### Integration tests

```bash
# Controlplane tests only
make test-integration INTEGRATION_TEST_PACKAGES=tests/common/controlplane/...

# All common tests
make test-integration
```

## Summary by CodeRabbit

* **New Features**
  * Added the `KuadrantControlPlane` resource (v1alpha1) for managing component deployments and reporting readiness, health, image and CRD status
  * Components are now deployed and updated automatically, including required CRDs and configurable images
  * Added Kubernetes Events for control plane lifecycle visibility (install, upgrade, failure, dependency detection, OLM cleanup)
  * Components deploy independently — one failure doesn't block others
  * Added configurable DNS operator providers and image support
* **Bug Fixes**
  * Added self-healing for missing or drifted control-plane resources
  * Improved cleanup of legacy operator metadata during migration with ownership cascade prevention
  * Fixed RBAC to grant both core and events.k8s.io API groups for event creation
  * Status now reports actual deployed images from rendered manifests
* **Compatibility**
  * Updated the minimum supported Kubernetes version to 1.25
  * Removed the DNS operator bundle dependency

## Summary by CodeRabbit

* **New Features**
  * Added the Kuadrant control plane resource for managing component deployment and health.
  * Added automatic installation, readiness tracking, version reporting, and image reporting for managed components.
  * Added support for deploying the DNS Operator as a related runtime image.
  * Added migration cleanup for legacy OLM-installed DNS Operator resources.
  * Added warning events when required dependencies are unavailable.
  * Added component charts to the container image.

* **Compatibility**
  * Updated the minimum supported Kubernetes version to 1.25.
  * Improved local development and integration-test configuration.


mikenairn:
@coderabbitai review

charts/kuadrant-operator/templates/manifests.yaml
```diff
@@ -14570,6 +14774,52 @@ rules:
...
+  - operators.coreos.com
+  resources:
+  - clusterserviceversions
+  - subscriptions
+  verbs:
+  - delete
+  - get
+  - list
+  - patch
+  - update
```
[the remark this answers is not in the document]
mikenairn: OLM Subscription and CSV names are not predictable at compile time (versions and catalog source names vary), so we can't scope delete by resourceNames. We can however move these permissions from the ClusterRole to a namespace-scoped Role limiting access to the operator namespace only. This is a temporary set of permissions only to deal with migration and would be better to be scoped to as few perms as possible.

cmd/main.go
```diff
@@ -246,11 +248,26 @@ func main() {
...
+	// finds them when it checks for CRD availability at boot.
+	componentDeployer, err := controlplane.NewDeployer(restConfig, operatorNamespace, setupLog)
+	if err != nil {
+		setupLog.Error(err, "unable to create component deployer")
+		os.Exit(1)
+	}
+	if err := componentDeployer.ApplyCRDsForComponents(context.Background(), componentDeployer.EnabledComponents()); err != nil {
+		setupLog.Error(err, "failed to bootstrap child operator CRDs")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
mikenairn: Re-using `ctx := ctrl.SetupSignalHandler()` and using throughout seems more appropriate.

internal/controlplane/olm_cleanup.go
```diff
@@ -0,0 +1,339 @@
...
+						LabelSelector: labelSelector,
+					})
+				} else {
+					list, err = c.client.Resource(gvr).List(ctx, metav1.ListOptions{
+						LabelSelector: labelSelector,
+					})
+				}
+				if err != nil {
+					continue
+				}
```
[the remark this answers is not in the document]
mikenairn: This should be resolved

cmd/main.go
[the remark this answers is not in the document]
mikenairn: changes have been applied

charts/kuadrant-operator/templates/manifests.yaml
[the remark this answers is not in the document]
mikenairn: changes have been applied

config/rbac/olm_migration_role.yaml
```diff
@@ -0,0 +1,19 @@
...
+rules:
+  - apiGroups:
+      - operators.coreos.com
+    resources:
+      - subscriptions
+      - clusterserviceversions
+    verbs:
+      - get
+      - list
+      - delete
```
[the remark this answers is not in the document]
mikenairn: The Update calls are on core/RBAC/apiextensions resources, not on operators.coreos.com resources. The existing get, list, delete verbs on the OLM migration role are correct and sufficient.

charts/kuadrant-operator/templates/manifests.yaml
```diff
@@ -14570,6 +14815,41 @@ rules:
...
+  resources:
+  - clusterroles
+  verbs:
+  - bind
+  - escalate
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
mikenairn: The escalate and bind verbs are scoped to specific resourceNames (dns-operator-manager-role, dns-operator-remote-cluster-role), not cluster-wide

bundle/manifests/kuadrant.io_kuadrantcontrolplanes.yaml
```diff
@@ -0,0 +1,208 @@
...
+    - jsonPath: .status.conditions[?(@.type=="Ready")].status
+      name: Ready
+      type: string
+    - jsonPath: .status.conditions[?(@.type=="Ready")].reason
+      name: Reason
+      type: string
+    - jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
+    name: v1beta1
```
maleck13: This jumped out at me. I learned that v1beta is not really used anymore in kube. Instead using v1alpha(n) until v1

internal/controlplane/olm_cleanup.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil
+	}
+
+	if !c.isOLMInstalled() {
+		c.logger.V(1).Info("OLM not detected, skipping cleanup")
+		c.cleaned = true
+		return nil
+	}
+
+	c.removeOrphanedOLMMetadata(ctx)
```
maleck13: If this fails to remove metadata (conflict for example) what happens? would the subscription get deleted anyway and then the related components also removed? For DNS that might be an ok thing, but less so for something in the data plane.

config/rbac/role.yaml
```diff
@@ -327,6 +350,41 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
```
maleck13, line 371: should we consider adding resourceNames here also similar to what is below under cluster roles?

internal/controlplane/kuadrant_control_plane_controller.go
```diff
@@ -0,0 +1,286 @@
...
+		r.logger.Info("ignoring KuadrantControlPlane with non-default name", "name", cp.Name)
+		return ctrl.Result{}, nil
+	}
+
+	var deployErr error
+	for _, component := range r.deployer.EnabledComponents() {
+		if err := r.deployer.DeployComponent(ctx, component); err != nil {
+			r.logger.Error(err, "failed to deploy component", "component", component.Name)
+			deployErr = err
+			break
```
maleck13: do we want to break out here? or allow it continue ?

maleck13: not major deal breaker with only dns currently but as we add more it is worth thinking about whether we bail or continue

internal/controlplane/kuadrant_control_plane_controller.go
```diff
@@ -0,0 +1,286 @@
...
+		Namespace: r.deployer.Namespace(),
+		Name:      component.DeploymentName,
+	}, deploy); err != nil {
+		if !apierrors.IsNotFound(err) {
+			r.logger.V(1).Info("unexpected error checking deployment readiness",
+				"component", component.Name, "error", err)
+		}
+		return false
+	}
+	return deploy.Status.ReadyReplicas > 0
```
maleck13: nit: could check against the spec replicas to see are all instances running. Minor though as most of these will be single instance

internal/controlplane/olm_cleanup.go
```diff
@@ -0,0 +1,335 @@
...
+			if rt.namespaced {
+				list, err = c.client.Resource(rt.gvr).Namespace(c.namespace).List(ctx, metav1.ListOptions{
+					LabelSelector: labelSelector,
+				})
+			} else {
+				list, err = c.client.Resource(rt.gvr).List(ctx, metav1.ListOptions{
+					LabelSelector: labelSelector,
+				})
+			}
+			if err != nil {
```
maleck13: if there is an error here prob should at lease log be hard to understand what was happening if it was silently failing to clean up.
Wondering if some debug/info logs would also be useful in here to audit what it was doing? I see the one below but this is more on the continues

internal/helm/renderer.go
```diff
@@ -0,0 +1,119 @@
...
+
+	return result, nil
+}
+
+func setManagedByLabel(obj *unstructured.Unstructured) {
+	labels := obj.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels["app.kubernetes.io/managed-by"] = "kuadrant-operator"
```
maleck13, line 95: just an FYI https://github.com/Kuadrant/architecture/pull/196/changes this is recommending a label that we might want to include here also

maleck13 (commented):
Some comments as I went through  this, main items were:

- Seems like OLM cleanup can cascade-delete resources — if stripping ownerReferences
     fails, the CSV still gets deleted, triggering GC. Recoverable for
     dns-operator today, destructive for authorino/limitador later. 

- ClusterRoleBindings unscoped — the operator can modify/delete any
     ClusterRoleBinding cluster-wide. Should consider split into unscoped create +
     scoped mutation on named bindings. 

- API version should be v1alpha1 — new CRD with no Spec field.  Can't find the reference but I know v1beta(n) is no longer something recommended. v1alpha(n) then v1 

- Deploy loop breaks on first component failure. Not sure if we want this or not. Something spotted not a blocker right now

internal/helm/renderer.go
mikenairn: The `app.kubernetes.io/managed-by` is a [standard recommended label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels)used to indicate:

> The tool being used to manage the operation of an application

All of our component manifests set this to "Helm" https://github.com/Kuadrant/dns-operator/blob/main/config/helm/kustomization.yaml#L8 during chart generation. It could be left as is, but considering its not really managed by "Helm" it felt wrong.

re. https://github.com/Kuadrant/architecture/pull/196/files, presumably you are talking about `kuadrant.io/managed: "true"`. I have seen that kind of thing before, i think olm uses a similar label for crds.

bundle/manifests/kuadrant.io_kuadrantcontrolplanes.yaml
mikenairn: I used v1beta1 since that's what the Kuadrant CR was using, but v1alpha1 would be more correct considering its a new API. I'll change it.

config/rbac/role.yaml
mikenairn: I _think_ it might be worth doing it for ClusterRoleBindings since its another cluster scoped resource same as ClusterRole. Idea is you can create "any" ClusterRole|ClusterRoleBinding (we can't restrict verb:create by resourceName afaik), but we can only modify/delete ones that we explicitly name.

Roles/RoleBindings are namespaced so we are limited to what we can do with these anyway.

Worth noting, the more explicit naming we add, the more we have to modify here if they ever change. Since changes to RBAC is fairly rare it's probably not too much of a concern.

internal/controlplane/kuadrant_control_plane_controller.go
mikenairn: Good question. With one this is fine, but with more we need to decide should it try all of them and report potentially multiple deploy errors in the KCP status. This does seem more like what it should do.

internal/controlplane/kuadrant_control_plane_controller.go
mikenairn: I think it might be better to check the Available condition instead of ReadyReplicas > 0. This should take into account multiple replicas.

internal/controlplane/olm_cleanup.go
mikenairn: I'm going to update this to break out if it the metadata update fails.

mikenairn:
@maleck13 Addressed most of your comments in https://github.com/Kuadrant/kuadrant-operator/pull/2178/commits/14b48b329843ce3f75bda1ead9d6efe95a6f1ace

Main changes since you looked last are:


**API & RBAC fixes**
- Moved KuadrantControlPlane from v1beta1 → v1alpha1
- Scoped ClusterRoleBinding RBAC to specific resourceNames instead of a broad grant

**OLM migration cleanup hardening**
- Gated Subscription/CSV deletion on successful metadata stripping first, to avoid an ownership cascade if stripping fails partway through
- Skip Forbidden errors during metadata stripping instead of failing the whole cleanup
- Cleanup now reports removed Subscription/CSV names per component (there's at most one of each per component), not just a total count

**Status accuracy**
- Component readiness now checks the Deployment's Available condition instead of using replicas > 0
- Deployed image versions are now extracted from the actually-rendered Deployment manifests (post-patch), rather than read back from env vars — so status reflects what's really running

**New: Kubernetes Events for control plane lifecycle**

Added `kubectl events`/`kubectl describe`-visible events across the control plane, as a lightweight audit trail alongside `status.conditions`:
- `KuadrantControlPlaneCreated` — first install or recreation after deletion
- `ComponentInstalled` / `ComponentVersionChanged` / `ComponentDeployFailed` — per child component (dns-operator, mcp-gateway, etc.)
- `CertManagerNotFound` / `DNSOperatorNotFound` / `LimitadorOperatorNotFound` / `AuthorinoOperatorNotFound` — dependency detection at startup
- `OLMMigrationComplete` / `OLMMigrationIncomplete` plus a new per-component `OLMComponentCleaned` breakdown

Each event correctly aggregates repeated occurrences (via a stable/synthetic object reference) rather than spamming duplicates, and per-component events use distinct references so multiple components don't collapse into a single event, hiding all but the first message.

**Key technical details:**
- Event dedup keys on `(eventType, action, reason, reportingController, regarding, related)` but **not** message text
- Used synthetic `componentReference(name)` for `related` param to keep per-component events distinct (otherwise they collapse into one event with counter)
- Both `apiGroups: [""]` and `apiGroups: ["events.k8s.io"]` RBAC grants required for `GetEventRecorder()` API

Example Events:

```
$ kubectl events --for kuadrantcontrolplane.kuadrant.io/default --watch
LAST SEEN   TYPE      REASON                OBJECT                         MESSAGE
62m         Warning   CertManagerNotFound   KuadrantControlPlane/default   cert-manager is not installed; TLSPolicy will not be reconciled
61m         Normal    KuadrantControlPlaneCreated   KuadrantControlPlane/default   created default KuadrantControlPlane
61m         Normal    OLMMigrationComplete          KuadrantControlPlane/default   stripped OLM metadata from 6 resources, deleted 1 Subscriptions and 1 CSVs
61m         Normal    OLMComponentCleaned           KuadrantControlPlane/default   removed Subscription dns-operator-preview-kuadrant-operator-catalog-kuadrant-system and CSV dns-operator.v0.0.0 (stripped OLM metadata from 6 resources)
61m         Normal    ComponentInstalled            KuadrantControlPlane/default   component dns-operator installed at version 0.0.0
61m         Normal    ComponentInstalled            KuadrantControlPlane/default   component mcp-gateway installed at version 0.9.0
```

**Component deployment independence**

Changed component deployment from fail-fast (break on first error) to attempt all components independently. If dns-operator fails but mcp-gateway succeeds, mcp-gateway will still be deployed. All errors are collected and returned joined, so status and logs show which specific components failed rather than stopping at the first failure and leaving subsequent components undeployed.

bundle/manifests/kuadrant.io_kuadrantcontrolplanes.yaml
mikenairn: Changed to v1alpha1

internal/controlplane/olm_cleanup.go
mikenairn: Updated to break out of cleanup if metadata cleanup fails.

config/rbac/role.yaml
mikenairn: I added additional name scoped RBAC for ClusterRoleBindings. If you think there is more to do here, re-open this conversation.

internal/controlplane/kuadrant_control_plane_controller.go
mikenairn: Updated to continue, and aggregate errors into the status for multiple failures. Components are independent so doesn't make much sense to just break out and not continue installing all components.

internal/controlplane/kuadrant_control_plane_controller.go
mikenairn: Updated to use Available condition, should be more accurate in terms of a deployment actually being ready based on it's spec.

internal/controlplane/olm_cleanup.go
mikenairn: updated

internal/controlplane/resources.go
```diff
@@ -0,0 +1,236 @@
...
+
+func CRDNames(crds []*unstructured.Unstructured) []string {
+	names := make([]string, 0, len(crds))
+	for _, crd := range crds {
+		names = append(names, crd.GetName())
+	}
+	return names
+}
+
+// PatchDeploymentImage overrides the first container's image on all Deployment
```
maleck13: looks like a stale comment?

maleck13 (dismissed):
reviewed the changes. Looks good. Went through the manual verification

internal/controlplane/resources.go
mikenairn: Yeah, it was in the wrong place. Updated.

maleck13 (dismissed):

internal/controlplane/resources.go
mikenairn: updated. Was in the wrong place.

internal/helm/renderer.go
mikenairn: Leaving this for now. If other labels are desirable across all resources managed by the kaudrant-operator they can be added later.

maleck13 (dismissed):

internal/controlplane/bootstrap.go
```diff
@@ -0,0 +1,109 @@
...
+func NewBootstrapRunnable(restConfig *rest.Config, scheme *runtime.Scheme, deployer *Deployer, recorder events.EventRecorder, namespace string, logger logr.Logger) *BootstrapRunnable {
+	return &BootstrapRunnable{
+		restConfig: restConfig,
+		scheme:     scheme,
+		deployer:   deployer,
+		recorder:   recorder,
+		namespace:  namespace,
+		logger:     logger.WithName("bootstrap"),
+	}
+}
```
maksymvavilov: This is a lot for a cleanup. Why this approach instead of a throwaway function in main with a `delete me in a few releases` comment?
Or have this as a manual-ish task via `kuadrantctl` or a script?

maksymvavilov (dismissed):
Went over the verification steps and was trying to be as picky as I can to the code: all looks good! 
One question for my curiosity

internal/controlplane/bootstrap.go
mikenairn: Well, it is really a single execute once "throwaway" function in main, only managed by the controller runtime manager instead of us having to do it ourselves. This allows us to automatically have it run only on a single replica (leader elected replica), use client/record/schema already configured and deal with shutdown signals gracefully. So ... why do we want to deal with all that ourselves when we can just have a Runnable registered via mgr.Add.

A manual option was discussed during the rfc process, but dismissed pretty quickly as we wanted a better user experience. It's also not clear how the timing of that would work, cleanup needs to happen before an upgrade starts, and that could be automatic on some clusters. It was decided a fully testable solution built into the operator itself was a better approach.

config/rbac/role.yaml
```diff
@@ -327,6 +351,55 @@ rules:
...
+  - rbac.authorization.k8s.io
+  resourceNames:
+  - dns-operator-manager-rolebinding
+  - dns-operator-remote-cluster-rolebinding
+  resources:
+  - clusterrolebindings
+  verbs:
+  - delete
+  - get
+  - list
```
eguzki: _resourceNames_ on list/watch verbs doesn't actually authorize anything, since list requests carry no name for the RBAC authorizer to match against

internal/controlplane/olm_cleanup.go
```diff
@@ -0,0 +1,441 @@
...
+	namespaced bool
+}{
+	{schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}, true},
+	{schema.GroupVersionResource{Group: "", Version: "v1", Resource: "serviceaccounts"}, true},
+	{schema.GroupVersionResource{Group: "", Version: "v1", Resource: "services"}, true},
+	{schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}, true},
+	{schema.GroupVersionResource{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "roles"}, true},
+	{schema.GroupVersionResource{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "rolebindings"}, true},
+	{schema.GroupVersionResource{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "clusterroles"}, false},
+	{schema.GroupVersionResource{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "clusterrolebindings"}, false},
```
eguzki: _cleanResourcesOwnedByCSV_ lists clusterroles/clusterrolebindings cluster-wide by label selector. Every single run of this cleanup will get Forbidden listing these two types

internal/controlplane/olm_cleanup.go
```diff
@@ -0,0 +1,441 @@
...
+			errs = append(errs, err)
+		}
+	}
+	return countByCSV, errors.Join(errs...)
+}
+
+func (c *OLMCleaner) findOrphanedCSVNames(ctx context.Context) []string {
+	csvs, err := c.client.Resource(csvGVR).Namespace(c.namespace).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil
```
eguzki: it this error swallowing intended?

config/rbac/role.yaml
```diff
@@ -22,17 +22,39 @@ rules:
...
   - patch
 - apiGroups:
-  - ""
+  - apiextensions.k8s.io
   resources:
-  - namespaces
+  - customresourcedefinitions
+  verbs:
+  - create
+  - list
```
eguzki, line 44: resourceNames on list/watch verbs doesn't actually authorize anything, since list requests carry no name for the RBAC authorizer to match against

eguzki (commented):
Really solid piece of engineering. The two-phase startup, drift reconciliation, and OLM migration path all show a lot of careful thought about edge cases (self-healing, fail-continue vs fail-fast, event-based lifecycle visibility).

That said, it's worth naming honestly: this adds a meaningful amount of new complexity. A new controller, a new CRD, a Helm-rendering engine, and a one-off cluster-mutating migration path on top of an already complex operator. I hope the packaging simplification this unlocks (one operator instead of several, no separate OLM bundle/dependency graph for dns-operator) ends up paying for that complexity over time.

Is the migration path tested on a cluster (OCP)  where kuadrant has been deployed using OLM?

Left a few comments along the way. None of them blocking, just things worth a look either now or as follow-ups.

mikenairn:
@coderabbitai review

.github/workflows/build-images-base.yaml
```diff
@@ -192,7 +192,7 @@ jobs:
             VERSION=${{ inputs.kuadrantOperatorVersion }} \
             AUTHORINO_OPERATOR_BUNDLE_IMG=${{ inputs.authorinoBundleImg }} \
             LIMITADOR_OPERATOR_BUNDLE_IMG=${{ inputs.limitadorBundleImg }} \
-            DNS_OPERATOR_BUNDLE_IMG=${{ inputs.dnsOperatorBundleImg }} \
+            RELATED_IMAGE_DNS_OPERATOR=${{ inputs.relatedImageDnsOperator }} \
```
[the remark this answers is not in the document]
mikenairn: leaving this now. It's consistent with the rest currently, will follow up with a general fix for all.

config/rbac/role.yaml
mikenairn: no longer adding list/watch due to changes made to OLMCleanup

internal/controlplane/olm_cleanup.go
mikenairn: no longer adding list/watch due to changes made to OLMCleanup

internal/controlplane/olm_cleanup.go
mikenairn: no longer relevant

config/rbac/role.yaml
mikenairn: no longer adding list/watch due to changes made to OLMCleanup

mikenairn:
> That said, it's worth naming honestly: this adds a meaningful amount of new complexity. A new controller, a new CRD, a Helm-rendering engine, and a one-off cluster-mutating migration path on top of an already complex operator. 

Agreed, but i think we were going to be dealing with additional complexity one way or the other re. OLMv1 support, any kube (Helm) install equivalence and likely modular component installation coming down the line.  

> I hope the packaging simplification this unlocks (one operator instead of several, no separate OLM bundle/dependency graph for dns-operator) ends up paying for that complexity over time.

Me too

> Is the migration path tested on a cluster (OCP) where kuadrant has been deployed using OLM?

Yes, i have been constantly testing upgrades on a local CRC cluster. I have a script that i created to test going from our last released version to whatever is on this branch. My hope is to make that into a GH workflow for continued upgrade testing going forward. Only OLM currently, but should also include Helm.

mikenairn:
@maleck13 @maksymvavilov @eguzki based on the feedback, and the fact that the OLMCleanup had some issues (highlighted by Eguzki), i went back and basically started again with the OLMCLeanup task. All the changes i made since anyone last looked are in [this](https://github.com/Kuadrant/kuadrant-operator/pull/2178/commits/0feaefb50a27e6a5290e1edff4394b0b53461283) commit and are _mostly_ only about the OLMCleanup, the thing that rightly got the most attention in reviews. 

Main changes are:

- Scoped OLM metadata stripping down to only the specific resources whose cascade-delete matters (Deployment, CRDs, and the dns-operator-controller-env ConfigMap), targeted by exact known name instead of a broad cluster-wide discovery scan.
- Replaced the generic, config-driven migration engine with an explicit `migrateDNSOperator` function (one per component, going forward) documenting exactly which resources it touches and why. This is more explicit in what it's doing for each componet and is (hopefully) clearer to understand when looking at it.
- Fully self-contained, it no longer depends on the Deployer's component registry or its dynamic/discovery clients.

eguzki:
> Yes, i have been constantly testing upgrades on a local CRC cluster. I have a script that i created to test going from our last released version to whatever is on this branch. My hope is to make that into a GH workflow for continued upgrade testing going forward. Only OLM currently, but should also include Helm.

I did try that in the past:  https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/test-upgrade.yaml

It broke after dependencies were included. Once dependencies are dropped, we can bring it back and make it a first-class check for releasing.

eguzki (approved):
LGTM

I am not able to spot significant issues. Just one nit, but unsure about it. `ImageReporter` and `ChartValueOverride` mechanisms are unused, but they could be left there for future use.

maksymvavilov (approved):
Changes make sence and work well.

mikenairn:
> I am not able to spot significant issues. Just one nit, but unsure about it. `ImageReporter` and `ChartValueOverride` mechanisms are unused, but they could be left there for future use.

 `ImageReporter` I'll double check and remove in a follow up if it's unused, might have been left over from a previous iteration.  `ChartValueOverride`  will be used by the mcp-gateway chart, the one I'll be adding next [see](https://github.com/Kuadrant/kuadrant-operator/pull/2202).

PR #2179, opened 2026-08-12
feat(telemetry): extend TelemetryPolicy with logging stanza

andresllh:
## Summary

Adds optional `logging.default.fields` to TelemetryPolicy so users can enrich authorization decision logs with CEL-derived fields. Policies can configure metrics, logging, or both. Logging bindings are delivered through `io.kuadrant.logging.fields` filter metadata.

Logging enrichment requires an administrator to explicitly set `spec.enableLoggingFields: true` on the managed Authorino CR. Kuadrant Operator does not set this field, and creating a TelemetryPolicy does not enable it. No operator environment variable is introduced.

### Configuration

With an Authorino Operator version that includes https://github.com/Kuadrant/authorino-operator/pull/360, the administrator enables logging on the Authorino CR:

```yaml
spec:
  enableLoggingFields: true
```

Then configure the desired fields:

```yaml
kind: TelemetryPolicy
metadata:
  name: audit-logging
spec:
  targetRef:
    kind: Gateway
    name: my-gateway
  logging:
    default:
      fields:
        client_identity: auth.identity.sub
        request_path: request.path
```

Requires the Authorino implementation in https://github.com/Kuadrant/authorino/pull/671. Authorino Operator dependency and component-chart revision updates are handled separately.

## Validation

- `make pre-commit` passed (verification, lint, vet, and race-enabled unit suite).
- `go test -mod=readonly -tags unit ./cmd/extensions/telemetry-policy/...` passed.
- Integration tests were not run.
- Binding and metadata domain routing tests.
- TelemetryPolicy reconciler tests for logging-only, mixed metrics/logging, and error handling.
- Generated CRD, OLM bundle, and Helm manifests updated.

## Summary by CodeRabbit

- **New Features**
  - Telemetry policies can enrich authorisation decision logs with custom fields derived from CEL expressions.
  - Metric configuration is optional, allowing policies to configure logging independently.
  - Logging field names are validated, and at least one field is required.
  - Custom logging fields can be enabled through operator environment configuration; unset or false values preserve existing settings.

- **Documentation**
  - Updated TelemetryPolicy reference documentation to explain logging fields, configuration requirements, environment-based enablement, and restart requirements.


internal/controller/authorino_reconciler.go
```diff
@@ -232,6 +238,12 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 	return nil
 }
 
+// enableAuthorinoLoggingFields uses the unstructured representation until the
+// independently managed authorino-operator dependency revision is updated.
+func enableAuthorinoLoggingFields(authorino *unstructured.Unstructured) error {
+	return unstructured.SetNestedField(authorino.Object, true, "spec", "enableLoggingFields")
```
guicassolato: Not sure if we want to have this feature (a) enabled by default, or (b) somehow conditioned by the presence of TelemetryPolicies that specify logging fields, or yet (c) as an explicit opt-in action of the user (by editing the Authorino CR).

(b) probably implies non trivial relationship to maintain, given the phases of the resources involved. Having this option flipped based on the state of multiple TelemetryPolicy CRs doesn't sound very feasible to me. Moreover, the TelemetryPolicy is still alpha, so there's also the question of API coupling across different maturity levels. In practice, this path would force us to promote the TelemetryPolicy, an effect we don't usually expect from introducing a new field.

(c) would be my preferred choice, but because I know some deployments are usually adverse to bootstrap configuration, especially when it involves manual steps, maybe open a possibility for (a), but still on the user's volition? Concretely, I'm thinking an env var of the Operator that controls the default about setting this field in the Authorino CR, disabled by default.

andresllh: My mistake, this should have been opt-in from the beginning. Implementing option C now.

internal/controller/authorino_reconciler.go
```diff
@@ -117,6 +118,10 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
 			}
 		}
 
+		if err := configureAuthorinoLoggingFields(unstructuredAuthorino); err != nil {
+			return err
+		}
+
 		// Only force ownership if tracing field is available for ownership
 		force := clusterAuthorino.Spec.Tracing.Endpoint == ""
```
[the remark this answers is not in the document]
andresllh: Addressed in 32834eaa: the operator now only sets enableLoggingFields when AUTHORINO_ENABLE_LOGGING_FIELDS parses as true, preserving administrator-managed values otherwise.

internal/controller/authorino_reconciler.go
```diff
@@ -232,6 +241,16 @@ func (r *AuthorinoReconciler) Reconcile(ctx context.Context, _ []controller.Reso
...
 }
 
+// configureAuthorinoLoggingFields uses the unstructured representation until the
+// independently managed authorino-operator dependency revision is updated.
+func configureAuthorinoLoggingFields(authorino *unstructured.Unstructured) error {
+	enabled, _ := env.GetBool("AUTHORINO_ENABLE_LOGGING_FIELDS", false)
+	if !enabled {
+		return nil
+	}
+	return unstructured.SetNestedField(authorino.Object, true, "spec", "enableLoggingFields")
```
guicassolato, line 251: Sorry, I should have pointed this out before. On setting `enableLoggingFields` via unstructured, I think we should consider cutting a release of Authorino Operator first, bump the dependency, and use the proper type to set the field on the CR.

The field currently lives only on Authorino Operator `main` — it isn't in any tagged release, and this repo pins v0.26.0 — so the typed `Spec.EnableLoggingFields` doesn't exist in the imported package yet. That's what makes unstructured the only option today, and it's also what the "cut a release first" ask is about.

The parts that worry me are two-fold:

1. **Runtime behaviour:** With Authorino Operator v0.26.0's CRD installed, `enableLoggingFields` is an unknown field. CRD structural schemas prune unknown fields by default, so there's no error and no event on create/update — the API server silently drops it. An administrator who sets `AUTHORINO_ENABLE_LOGGING_FIELDS=true` against the current stack gets a feature that quietly does nothing, with no signal to debug against.
2. **Loss of type safety and maintainability:** Using the strongly typed struct from the imported package keeps the components in sync code-wise and avoids pieces like this going stale. If Authorino Operator ever drops `enableLoggingFields`, this code in Kuadrant Operator would be easy to overlook — nothing would flag it.

The typed route gives us a compile-time guarantee the field exists, a pinned CRD that won't prune it at runtime, and refactor-safety — no stringly-typed `"spec", "enableLoggingFields"` path to silently break on a rename, and no dead code left behind if the field ever goes away.

I know this probably just followed the approach used for `spec.tracing.insecure`. That case uses unstructured to work around `Destruct` omitting a field that _does_ exist in the typed API — a serialisation workaround for a present field, not a way to reference a field that's absent from the dependency.

andresllh: Got it. I'll keep an eye out for when they tag the new Authorino Operator Release that contains the required code.

PR #2180, opened 2026-08-12
Extensions open socket

### LINKED ISSUE #2139: Expose extensions port
Once we have the handshake verifying authN we can look to open up the extensions port. We should align on usage of `NetworkPolicies`

adam-cattermole:
## Summary by CodeRabbit

* **New Features**
  * Added a dedicated network service for connecting to operator extensions over TCP.
  * Extension processes now connect using a configurable service address and port.
  * Added authenticated gRPC communication with graceful startup and shutdown handling.

* **Bug Fixes**
  * Improved extension connection lifecycle management and cleanup when startup fails.
  * Updated validation to provide clearer feedback when the extension address is unavailable.

* **Chores**
  * Removed obsolete local socket configuration and related deployment settings.

### DIFF
--- a/bundle/manifests/kuadrant-operator-extensions_v1_service.yaml
+++ b/bundle/manifests/kuadrant-operator-extensions_v1_service.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  creationTimestamp: null
+  labels:
+    app: kuadrant
+    control-plane: controller-manager
+  name: kuadrant-operator-extensions
+spec:
+  ports:
+  - name: extensions
+    port: 50052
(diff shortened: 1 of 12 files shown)


silvi-t (approved):
LGTM - clean migration from Unix sockets to TCP. Lifecycle management is correct, auth interceptor properly applied, tests improved.

One non-blocking nit (pre-existing, not introduced by this PR): `internal/extension/oop.go:126-130 — timer.Stop()` is called immediately after `time.AfterFunc` so the 2s force-kill fallback never fires. Should be `defer timer.Stop()` so the timer is only cancelled after the process has actually exited.

PR #2185, opened 2026-08-14
feat: kuadrant CR has now owner reference to Authconfigs

### LINKED ISSUE #2143: AuthConfig resources are stranded if the operator is removed before AuthPolicies are deleted
## Summary

If the Kuadrant operator is removed before `AuthPolicy` resources are deleted, the `AuthConfig` objects created by those policies remain in the cluster indefinitely.

## Root cause

`AuthConfig` objects are created and deleted by `AuthConfigsReconciler` in a reconciler-driven pattern — there are no `ownerReferences` on `AuthConfig` objects. Cleanup only happens while the operator is running and reconciling effective auth policies.

`AuthConfig` objects live in the Authorino namespace (`authorino.GetNamespace()`, typically `kuadrant-system`), as confirmed in `authconfigs_reconciler.go:64`.

If the operator is removed first and AuthPolicies are deleted afterwards (or never deleted), the `AuthConfigs` are never cleaned up.

## Expected behavior

`AuthConfig` resources should be cleaned up as part of `Kuadrant` CR deletion, regardless of whether `AuthPolicy` resources were deleted first.

## Proposed fix

Add a `Kuadrant` CR finalizer (following the precedent in `developerportal_reconciler.go`) that walks all `AuthConfig` objects labeled `kuadrant.io/auth=true` in the Authorino namespace and deletes them before the `Kuadrant` CR is fully removed. This makes cleanup order-independent.

The infrastructure already exists: `AddFinalizer`/`RemoveFinalizer` are implemented in `base_reconciler.go`, and `kuadrants/finalizers` RBAC is already granted (`state_of_the_world.go:77`).

## Relevant files

- `internal/controller/authconfigs_reconciler.go` — AuthConfig lifecycle management (no owner refs)
- `internal/controller/developerportal_reconciler.go` — existing finalizer pattern to follow
- `internal/reconcilers/base_reconciler.go:221` — `AddFinalizer`/`RemoveFinalizer` helpers

## Impact

Users who remove the operator before deleting their `AuthPolicy` resources are left with stale `AuthConfig` objects in `kuadrant-system` that must be cleaned up manually.

DaliborD45:
### What

Adds a controller `ownerReference` pointing from every operator-managed Authorino `AuthConfig` back to the `Kuadrant` CR. As a result, deleting the `Kuadrant` CR now cascades — Kubernetes' garbage collector automatically deletes all managed `AuthConfig`s — without any explicit teardown code.

### Changes

`internal/controller/authconfigs_reconciler.go`:
- New helper `authConfigOwnerReferences(kuadrantCR)` builds the owner reference (`Controller: true`, `BlockOwnerDeletion: true`); returns `nil` when the CR is unavailable.

- `Reconcile` fetches the `Kuadrant` CR from the topology and threads it through `reconcileAuthConfigForPath` → `buildDesiredAuthConfig`, which now stamps the owner reference onto every desired `AuthConfig`.

- `equalAuthConfigs` now compares owner references, so **pre-existing** `AuthConfig`s (created before this change) are detected as out-of-date and get the reference back-filled on the next reconcile.

- `applyAuthConfigUpdate` writes the owner reference onto existing objects during updates.

Both `equalAuthConfigs` and `applyAuthConfigUpdate` guard on `len(desiredOwners) > 0`, so a momentarily-unavailable `Kuadrant` CR (e.g. mid-deletion) never strips an existing reference or causes an infinite update loop.

`internal/controller/authconfigs_reconciler_test.go`:
- `TestAuthConfigOwnerReferences` — verifies the reference is built correctly and is `nil` for a `nil` CR.
- `TestEqualAuthConfigs` — added cases for missing/matching owner refs and the no-desired-ref guard.
- `TestApplyAuthConfigUpdate` — added cases confirming the ref is applied, and never stripped when none is desired.

## Summary by CodeRabbit

* **New Features**
  * Managed AuthConfig resources now receive an owner reference to the associated Kuadrant resource.
  * Added support for retrieving a Kuadrant resource’s owner reference.

* **Bug Fixes**
  * Improved AuthConfig comparison and reconciliation to ensure updates converge correctly.
  * Unrelated owner references are retained, while outdated Kuadrant references are replaced.
  * Owner references are applied consistently during AuthConfig updates.

### DIFF
--- a/api/v1beta1/kuadrant_types.go
+++ b/api/v1beta1/kuadrant_types.go
@@ -79,7 +79,7 @@ func (k *Kuadrant) IsDeveloperPortalEnabled() bool {
 }
 
 // GetOwnerReference returns the owner reference pointing to this Kuadrant CR,
-func (k *Kuadrant) GetOwnerReference() []metav1.OwnerReference {
+func (k *Kuadrant) BuildOwnerReference() []metav1.OwnerReference {
 	if k == nil {
 		return nil
 	}
--- a/api/v1beta1/kuadrant_types.go
+++ b/api/v1beta1/kuadrant_types.go
@@ -78,6 +78,23 @@ func (k *Kuadrant) IsDeveloperPortalEnabled() bool {
 	return k.Spec.Components.DeveloperPortal.Enabled
 }
 
+// GetOwnerReference returns the owner reference pointing to this Kuadrant CR,
+func (k *Kuadrant) GetOwnerReference() []metav1.OwnerReference {
+	if k == nil {
+		return nil
+	}
+	return []metav1.OwnerReference{
+		{
+			APIVersion:         GroupVersion.String(),
+			Kind:               KuadrantGroupKind.Kind,
+			Name:               k.GetName(),
+			UID:                k.GetUID(),
+			BlockOwnerDeletion: ptr.To(true),
+			Controller:         ptr.To(true),
+		},
+	}
+}
+
 // KuadrantSpec defines the desired state of Kuadrant
 type KuadrantSpec struct {
 	Observability Observability `json:"observability,omitempty"`
--- a/internal/controller/authconfigs_reconciler.go
+++ b/internal/controller/authconfigs_reconciler.go
@@ -85,7 +85,7 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	desiredAuthConfigs := make(map[k8stypes.NamespacedName]struct{})
 	modifiedAuthConfigs := []string{}
 
-	kuadrantCROwnerReference := kuadrantCR.GetOwnerReference()
+	kuadrantCROwnerReference := kuadrantCR.BuildOwnerReference()
 
 	for pathID, effectivePolicy := range effectivePoliciesMap {
 		authConfigName, modified := r.reconcileAuthConfigForPath(ctx, pathID, effectivePolicy, authConfigsNamespace, kuadrantCROwnerReference, topology, errorRegistry)
--- a/internal/controller/authconfigs_reconciler.go
+++ b/internal/controller/authconfigs_reconciler.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/utils/ptr"
 
 	kuadrantv1 "github.com/kuadrant/kuadrant-operator/api/v1"
 	kuadrantv1beta1 "github.com/kuadrant/kuadrant-operator/api/v1beta1"
@@ -65,6 +64,10 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	authConfigsNamespace := authorino.GetNamespace()
 
 	kuadrantCR := GetKuadrantFromTopology(topology, state)
+	if kuadrantCR == nil {
+		logger.V(1).Info("kuadrant resource not found in the topology")
+		return nil
+	}
 
 	effectivePolicies, ok := state.Load(StateEffectiveAuthPolicies)
 	if !ok {
@@ -81,8 +84,10 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	desiredAuthConfigs := make(map[k8stypes.NamespacedName]struct{})
 	modifiedAuthConfigs := []string{}
 
+	kuadrantCROwnerReference := kuadrantCR.GetOwnerReference()
+
 	for pathID, effectivePolicy := range effectivePoliciesMap {
-		authConfigName, modified := r.reconcileAuthConfigForPath(ctx, pathID, effectivePolicy, authConfigsNamespace, kuadrantCR, topology, errorRegistry)
+		authConfigName, modified := r.reconcileAuthConfigForPath(ctx, pathID, effectivePolicy, authConfigsNamespace, kuadrantCROwnerReference, topology, errorRegistry)
 		if authConfigName == "" {
 			continue
 		}
@@ -127,7 +132,7 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 
 // reconcileAuthConfigForPath reconciles the AuthConfig for a single effective policy path.
 // It returns the authConfigName (empty if the path is invalid) and whether the object was modified.
-func (r *AuthConfigsReconciler) reconcileAuthConfigForPath(ctx context.Context, pathID string, effectivePolicy EffectiveAuthPolicy, authConfigsNamespace string, kuadrantCR *kuadrantv1beta1.Kuadrant, topology *machinery.Topology, errorRegistry *ErrorRegistry) (string, bool) {
+func (r *AuthConfigsReconciler) reconcileAuthConfigForPath(ctx context.Context, pathID string, effectivePolicy EffectiveAuthPolicy, authConfigsNamespace string, kuadrantCROwnerReference []metav1.OwnerReference, topology *machinery.Topology, errorRegistry *ErrorRegistry) (string, bool) {
 	logger := controller.LoggerFromContext(ctx).WithName("AuthConfigsReconciler")
 
 	parsed, err := kuadrantpolicymachinery.ParseTopologyPath(effectivePolicy.Path)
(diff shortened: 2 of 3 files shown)


internal/controller/authconfigs_reconciler.go
```diff
@@ -79,7 +82,7 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	modifiedAuthConfigs := []string{}
 
 	for pathID, effectivePolicy := range effectivePoliciesMap {
-		authConfigName, modified := r.reconcileAuthConfigForPath(ctx, pathID, effectivePolicy, authConfigsNamespace, topology, errorRegistry)
+		authConfigName, modified := r.reconcileAuthConfigForPath(ctx, pathID, effectivePolicy, authConfigsNamespace, kuadrantCR, topology, errorRegistry)
```
Boomatang: Can we calculate the owner reference and pass it into the `reconcileAuthConfigForPath`. That way if there is a 1000 AuthConfigs being generated we don't have to recalculate the owner reference for each.

internal/controller/authconfigs_reconciler.go
```diff
@@ -63,6 +64,8 @@ func (r *AuthConfigsReconciler) Reconcile(ctx context.Context, _ []controller.Re
 	}
 	authConfigsNamespace := authorino.GetNamespace()
 
+	kuadrantCR := GetKuadrantFromTopology(topology, state)
```
Boomatang, line 67: There is no nil check for the kuadrant CR here. If there is no kuadrant CR we should not create any resources. As we would not be able to create them successfully.

internal/controller/authconfigs_reconciler.go
```diff
@@ -248,7 +251,24 @@ func (r *AuthConfigsReconciler) reconcileAuthConfigForPath(ctx context.Context,
...
+		{
+			APIVersion:         kuadrantv1beta1.GroupVersion.String(),
+			Kind:               kuadrantv1beta1.KuadrantGroupKind.Kind,
+			Name:               kuadrantCR.GetName(),
+			UID:                kuadrantCR.GetUID(),
+			BlockOwnerDeletion: ptr.To(true),
+			Controller:         ptr.To(true),
+		},
+	}
+}
```
Boomatang: I don't think this should take in a pointer, or if it does I think it needs to return an error of some kind. 

There is a different question I would like to ask. How could this be more generic? How could others use this pattern easily? 

Could we add a function on to the kuadrant type that returns owner reference? Then we could do some thing like.
```go
kuadrantCR := GetKuadrantFromTopology(topology, state)
if kuadrantCR == nil {
    // some logging
    return nil // or return an error state
}

kuadrantOwnerReference := kuadrantCR.GenerateOwnerReference()

// Start looping over the authconfigs

```

internal/controller/authconfigs_reconciler.go
```diff
@@ -260,7 +280,8 @@ func (r *AuthConfigsReconciler) buildDesiredAuthConfig(ctx context.Context, effe
 			Annotations: map[string]string{
 				annotationKey: routeRuleLocator,
 			},
-			Labels: AuthObjectLabels(),
+			Labels:          AuthObjectLabels(),
+			OwnerReferences: authConfigOwnerReferences(kuadrantCR),
```
Boomatang: As I mentioned above, we really don't want this to be nil. Currently it is possible for it to be nil.

internal/controller/authconfigs_reconciler.go
```diff
@@ -383,6 +404,13 @@ func equalAuthConfigs(existing, desired *authorinov1beta3.AuthConfig) bool {
 		return false
 	}
 
+	// owner references (only enforced when a desired owner is set, so we don't
+	// churn or strip references when the Kuadrant CR is momentarily unavailable)
+	if desiredOwners := desired.GetOwnerReferences(); len(desiredOwners) > 0 &&
+		!reflect.DeepEqual(existing.GetOwnerReferences(), desiredOwners) {
+		return false
+	}
```
Boomatang: Two things. 
1. Can you refactor this how the same code style as the other checks in the function.
2. I think this could create churn if the list of ownerReferences are generated out of order between runs. In this code [A,B] != [B,A], while in truth for what we are doing they are equal.

internal/controller/authconfigs_reconciler.go
```diff
@@ -408,6 +436,10 @@ func applyAuthConfigUpdate(ctx context.Context, existing, desired *authorinov1be
 	existing.Spec = desired.Spec
 	existing.Labels = desired.Labels
 
+	if desiredOwners := desired.GetOwnerReferences(); len(desiredOwners) > 0 {
+		existing.SetOwnerReferences(desiredOwners)
+	}
```
Boomatang: I have ran the code to play test this, but are we sure this wont remove existing references that maybe a third party added.

Boomatang (commented):
This looking good, there is a couple of improvements that would be nice to, and a few error cases that needs to be handled. But is looking much cleaner.

internal/controller/authconfigs_reconciler.go
DaliborD45: well there will be only ownerReference, kuadrantCR, so it will be just [A] == [A], can it somehow get into state, where others can put there authConfigs, with more than one ownerReference ?

internal/controller/authconfigs_reconciler.go
DaliborD45: I dont think I really understand the problem here, can you maybe explain it more please ?

internal/controller/authconfigs_reconciler.go
Boomatang: The happy path assumption is that there will ever only be one ownerReference, KuadrantCR, but users can do odd things.  For the happy path the code is fine, but if there is more than one ownerReference then that code can introduce churn.

internal/controller/authconfigs_reconciler.go
Boomatang: If the AuthConfig has an existing ownerReference set by a third party. When we add the kuadrantCr ownerReference will the third party reference be maintained?

DaliborD45: But how can AuthConfig be created with ownerReference set by a third party in the first place ?

DaliborD45: I have created fix for this

internal/controller/authconfigs_reconciler.go
```diff
@@ -383,10 +393,60 @@ func equalAuthConfigs(existing, desired *authorinov1beta3.AuthConfig) bool {
...
+// Kuadrant-managed reference replaced by the desired one. Third-party references
+// are preserved, and stale Kuadrant references
+func upsertKuadrantOwnerReference(existing, desired []metav1.OwnerReference) []metav1.OwnerReference {
+	preserved := lo.Filter(existing, func(ref metav1.OwnerReference, _ int) bool {
+		return !isKuadrantOwnerReference(ref)
+	})
+	if ref, found := lo.Find(desired, isKuadrantOwnerReference); found {
+		preserved = append(preserved, ref)
+	}
+	return preserved
```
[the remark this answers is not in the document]
DaliborD45: But here can be only one Kuadrant CR

Boomatang (dismissed):
Tested locally, the authconfigs are removed when the kuadrant cr is removed, even if the operator is scaled down provide the dev portal finalizers has being removed.

api/v1beta1/kuadrant_types.go
```diff
@@ -78,6 +78,23 @@ func (k *Kuadrant) IsDeveloperPortalEnabled() bool {
...
+		return nil
+	}
+	return []metav1.OwnerReference{
+		{
+			APIVersion:         GroupVersion.String(),
+			Kind:               KuadrantGroupKind.Kind,
+			Name:               k.GetName(),
+			UID:                k.GetUID(),
+			BlockOwnerDeletion: ptr.To(true),
+			Controller:         ptr.To(true),
```
maksymvavilov, line 93: Do we want these configurable? 
Also, when I saw `kuadrantCR.GetOwnerReference()`, the first thing that came to me was "get reference to the owner of kuadrant CR". Maybe `buildOwnerReference()` would be better? WDYT?

maksymvavilov (dismissed):
Two questions; both non-clocking. 
I've checked and we are guaranteed to match namespaces with `Kuadrant` CR. I'm not sure why I was convinced that they live in separate namespaces but here we are 🤷  looks like I did need that break after all. 
Correct minimal changes that are vell-tested - awesome work! 

Can we do the same for the dev-portal?

DaliborD45:
> Can we do the same for the dev-portal?

Yes we can, but that would be good to implement it in separate PR

api/v1beta1/kuadrant_types.go
DaliborD45: I think this is quite subjective, but yeah I see your point, and I kinda agree with you so I will change the naming

maksymvavilov (approved):

PR #2188, opened 2026-08-18
Extensions secret reconciler

adam-cattermole:
Add the extensions secret reconciler and fix two outstanding issues:

- The credential was assumed hex which didn't match the intended opaque string
- Built-in extensions can leverage the warmup gate to ensure they claim their policy kinds before standalone extensions

## Summary by CodeRabbit

- **New Features**
  - Extension authentication credentials can now be synchronised automatically from a Kubernetes Secret.
  - Added a configurable warm-up period before standalone extensions can connect.
  - Built-in extensions remain available during warm-up, with access opening once readiness conditions are met.
  - Credential additions, rotations and removals are handled automatically, including session revocation when required.

- **Bug Fixes**
  - Extension credentials are now handled consistently as raw values rather than encoded strings.
  - Added secure, namespace-scoped permissions for reading authentication Secrets.

### DIFF
--- a/config/rbac/extension_auth_role.yaml
+++ b/config/rbac/extension_auth_role.yaml
@@ -1,0 +1,15 @@
+# Permissions to read the Secret supplying standalone extension credentials
+# Secret read permissions for the operator namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: extension-auth-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - list
+      - watch
--- a/internal/controller/extension_auth_secret_reconciler.go
+++ b/internal/controller/extension_auth_secret_reconciler.go
@@ -47,7 +47,12 @@ func (r *ExtensionAuthSecretReconciler) Reconcile(eventCtx context.Context, _ []
 		credentials = secret.Data
 	}
 
-	logger.V(1).Info("syncing extension auth secret credentials", "count", len(credentials))
-	r.store.SyncSecretCredentials(credentials)
+	result := r.store.SyncSecretCredentials(credentials)
+	if result.Changed() {
+		logger.Info("extension auth secret credentials changed",
+			"added", result.Added, "rotated", result.Rotated, "removed", result.Removed, "total", len(credentials))
+	} else {
+		logger.V(1).Info("extension auth secret credentials unchanged", "count", len(credentials))
+	}
 	return nil
 }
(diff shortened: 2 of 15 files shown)


internal/controller/extension_auth_secret_reconciler.go
```diff
@@ -0,0 +1,53 @@
...
+		secret, ok := runtimeObj.Object.(*corev1.Secret)
+		if !ok {
+			return fmt.Errorf("auth secret %s/%s has unexpected object type %T", r.namespace, r.secretName, runtimeObj.Object)
+		}
+		credentials = secret.Data
+	}
+
+	logger.V(1).Info("syncing extension auth secret credentials", "count", len(credentials))
+	r.store.SyncSecretCredentials(credentials)
+	return nil
```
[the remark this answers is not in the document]
adam-cattermole: I think there's some confusion here, the pre, tasks, and post conditions allow you to compose multiple reconcilers into a single workflow - in this case we have one simple reconciler that only has one 'task' there are no pre/post conditions here.

adam-cattermole:
@coderabbitai review

config/rbac/extension_auth_role.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: extension-auth-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - watch
```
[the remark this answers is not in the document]
adam-cattermole: `resourceNames` only applies to the verbs that carry a name. It doesn't work with `list` and `watch`. Our reconciler is informer-based (it does list and watch on secrets) so the `resourceNames` would be ignored here.

config/rbac/extension_auth_role.yaml
```diff
@@ -0,0 +1,14 @@
+# permissions to read the Secret supplying standalone extension credentials.
```
silvi-t: I would suggest updating this comment to reflect that the operator can actually read all Secrets in its own namespace, not just this one.

adam-cattermole: It was more just indicating why we need the role itself, rather than describing the permissions in the list below the comment - do you think add another line below stating that it can read all secrets in its namespace?

silvi-t: That sounds good. Adding a note below with the actual scope makes sense.

silvi-t (approved):
LGTM! I tested the standalone extension deployment on both Kind and OCP, and the Secret-based credential mechanism worked correctly for me on both — including the full flow of credential sync, handshake, and real policy reconciliation.

PR #2195, opened 2026-08-20
Add NetworkPolicy reconciler for Authorino and Limitador operands

Boomatang:
# Summary

Introduces a new `NetworkPolicyReconciler` that automatically creates and manages Kubernetes `NetworkPolicy` resources for the Authorino and Limitador operand deployments. These policies restrict ingress traffic to only the ports each operand requires and only from namespaces that contain Gateway resources managed by the topology.

- Authorino NetworkPolicy allows ingress on gRPC (50051), HTTP (5051), and OIDC discovery (8083) ports by default, with port values sourced from the Authorino CR spec when available. OIDC discovery is left open (empty `from`) while gRPC and HTTP are scoped to gateway namespaces.
- Limitador NetworkPolicy allows ingress on gRPC (8081) and HTTP (8080) ports by default, with port values sourced from the Limitador CR spec when available. Both rules are scoped to gateway namespaces.
- Pod selectors are derived from the linked operand Deployment's pod template labels via the topology DAG, falling back to `kuadrant.io/managed: true` when the deployment is not yet available.
- NetworkPolicies are owned by their respective operand CR (Authorino/Limitador) via `ownerReferences`, enabling garbage collection on operand deletion.
- Merge logic preserves user-added labels and only triggers updates when spec or managed labels actually change.
- New topology links (`LinkAuthorinoToDeployment`, `LinkAuthorinoToNetworkPolicy`, `LinkLimitadorToNetworkPolicy`) and watchers are added to wire the reconciler into the existing policy machinery controller.
- RBAC rules for `networking.k8s.io/networkpolicies` are added to the operator role, bundle CSV, and Helm chart.

## Summary by CodeRabbit

* **New Features**
  * Added automatic Kubernetes NetworkPolicy management for Authorino and Limitador.
  * Network policies restrict ingress to relevant gateway namespaces and services.
  * Policies adapt to configured or default service ports while preserving existing ownership details.
  * Added monitoring and reconciliation for policy and Authorino deployment changes.
  * Added automatic bootstrapping of enabled component resources.
  * Added the required permissions for the operator to manage NetworkPolicies.

* **Bug Fixes**
  * Improved handling of missing resources, duplicate gateway peers and policy updates.
  * Added clearer warnings when required dependencies are unavailable.

### DIFF
--- a/api/v1beta1/topology.go
+++ b/api/v1beta1/topology.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kuadrant/policy-machinery/machinery"
 	"github.com/samber/lo"
 	appsv1 "k8s.io/api/apps/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	gatewayapiv1 "sigs.k8s.io/gateway-api/apis/v1"
@@ -25,6 +26,9 @@ var (
 
 	DeploymentGroupKind = appsv1.SchemeGroupVersion.WithKind("Deployment").GroupKind()
 	DeploymentsResource = appsv1.SchemeGroupVersion.WithResource("deployments")
+
+	NetworkPolicyGroupKind = schema.GroupKind{Group: networkingv1.SchemeGroupVersion.Group, Kind: "NetworkPolicy"}
+	NetworkPolicyResource  = networkingv1.SchemeGroupVersion.WithResource("networkpolicies")
 )
 
 func LinkKuadrantToGatewayClasses(objs controller.Store) machinery.LinkFunc {
@@ -66,6 +70,21 @@ func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
 		Func: func(child machinery.Object) []machinery.Object {
 			return lo.Filter(kuadrants, func(k machinery.Object, _ int) bool {
 				return k.GetNamespace() == child.GetNamespace() && child.GetName() == "authorino"
+			})
+		},
+	}
+}
+
+func LinkAuthorinoToDeployment(objs controller.Store) machinery.LinkFunc {
+	authorinos := utils.Map(objs.FilterByGroupKind(AuthorinoGroupKind), ControllerObjectToMachineryObject)
+
+	return machinery.LinkFunc{
+		From: AuthorinoGroupKind,
+		To:   DeploymentGroupKind,
+		Func: func(deployment machinery.Object) []machinery.Object {
+			return lo.Filter(authorinos, func(authorino machinery.Object, _ int) bool {
+				// the name of the deployment is hardcoded. This deployment is owned by the authorino operator.
+				return authorino.GetNamespace() == deployment.GetNamespace() && deployment.GetName() == "authorino"
 			})
 		},
 	}
@@ -121,6 +140,34 @@ func LinkKuadrantToPodMonitor(objs controller.Store) machinery.LinkFunc {
 					}
 				}
 				return false
+			})
+		},
+	}
+}
+
+func LinkAuthorinoToNetworkPolicy(objs controller.Store) machinery.LinkFunc {
+	authorinos := utils.Map(objs.FilterByGroupKind(AuthorinoGroupKind), ControllerObjectToMachineryObject)
+
+	return machinery.LinkFunc{
+		From: AuthorinoGroupKind,
+		To:   NetworkPolicyGroupKind,
+		Func: func(networkPolicy machinery.Object) []machinery.Object {
+			return lo.Filter(authorinos, func(authorino machinery.Object, _ int) bool {
+				return authorino.GetNamespace() == networkPolicy.GetNamespace() && networkPolicy.GetName() == "kuadrant-authorino"
+			})
+		},
+	}
+}
+
+func LinkLimitadorToNetworkPolicy(objs controller.Store) machinery.LinkFunc {
+	limitadors := utils.Map(objs.FilterByGroupKind(LimitadorGroupKind), ControllerObjectToMachineryObject)
+
+	return machinery.LinkFunc{
+		From: LimitadorGroupKind,
+		To:   NetworkPolicyGroupKind,
+		Func: func(networkPolicy machinery.Object) []machinery.Object {
+			return lo.Filter(limitadors, func(limitador machinery.Object, _ int) bool {
+				return limitador.GetNamespace() == networkPolicy.GetNamespace() && networkPolicy.GetName() == "kuadrant-limitador"
 			})
 		},
 	}
--- a/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
@@ -225,7 +225,7 @@ metadata:
     categories: Integration & Delivery
     console.openshift.io/plugins: '["kuadrant-console-plugin"]'
     containerImage: quay.io/kuadrant/kuadrant-operator:latest
-    createdAt: "2026-08-25T15:38:41Z"
+    createdAt: "2026-09-01T08:36:51Z"
     description: A Kubernetes Operator to manage the lifecycle of the Kuadrant system
     operators.operatorframework.io/builder: operator-sdk-v1.33.0
     operators.operatorframework.io/project_layout: go.kubebuilder.io/v4
@@ -261,6 +261,9 @@ spec:
       kind: DNSPolicy
       name: dnspolicies.kuadrant.io
       version: v1
+    - kind: KuadrantControlPlane
+      name: kuadrantcontrolplanes.kuadrant.io
+      version: v1alpha1
     - description: Kuadrant configures installations of Kuadrant Service Protection
         components
       displayName: Kuadrant
@@ -412,17 +415,39 @@ spec:
         - apiGroups:
           - ""
           resources:
+          - namespaces
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
+          - events.k8s.io
+          resources:
           - events
           verbs:
           - create
           - patch
         - apiGroups:
-          - ""
+          - apiextensions.k8s.io
           resources:
-          - namespaces
+          - customresourcedefinitions
+          verbs:
+          - create
+          - list
+          - watch
+        - apiGroups:
+          - apiextensions.k8s.io
+          resourceNames:
+          - dnshealthcheckprobes.kuadrant.io
+          - dnsrecords.kuadrant.io
+          resources:
+          - customresourcedefinitions
           verbs:
           - get
           - list
+          - patch
+          - update
           - watch
         - apiGroups:
           - apps
@@ -437,6 +462,12 @@ spec:
           - update
           - watch
         - apiGroups:
+          - authentication.k8s.io
+          resources:
+          - tokenreviews
+          verbs:
+          - create
+        - apiGroups:
           - authorino.kuadrant.io
           resources:
           - authconfigs
@@ -448,6 +479,12 @@ spec:
           - patch
           - update
           - watch
+        - apiGroups:
+          - authorization.k8s.io
+          resources:
+          - subjectaccessreviews
+          verbs:
+          - create
         - apiGroups:
           - cert-manager.io
           resources:
@@ -605,6 +642,7 @@ spec:
           - kuadrant.io
           resources:
           - authpolicies
+          - kuadrantcontrolplanes
           verbs:
           - create
           - get
@@ -628,6 +666,7 @@ spec:
           resources:
           - authpolicies/status
           - dnspolicies/status
+          - kuadrantcontrolplanes/status
           - kuadrants/status
           - ratelimitpolicies/status
           - tlspolicies/status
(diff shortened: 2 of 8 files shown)


guicassolato:
https://github.com/Kuadrant/authorino/pull/673 may be relevant.

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+	labels := linkedDeploymentLabels(&controller.RuntimeObject{Object: aObj}, topology)
+
+	if labels == nil {
+		labels = map[string]string{"kuadrant.io/managed": "true"}
+	}
+
+	// These default port values are hardcode into the authServerCmd in the authorino repo
+	// https://github.com/Kuadrant/authorino/blob/58fecc6cdec38376fa7dba5638f1f7ecb6964cd0/main.go#L178-L218
+	gRPCport := 50051
+	HTTPport := 5051
```
fabikova: On our cluster authorino-authorino-authorization exposes 50051 and 5001 (not 5051). Also here in docs it's 5001 https://github.com/Kuadrant/authorino/pull/673/changes

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+			Labels:    CommonLabels(),
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{
+				MatchLabels: labels,
+			},
+			PolicyTypes: []networkingv1.PolicyType{"Ingress"},
+			Ingress: []networkingv1.NetworkPolicyIngressRule{
+				// gRPC ext-auth from Envoy
+				ingressRule(fromNamespaces, gRPCport),
```
fabikova: if there are no gateways yet, "from" for the gRPC/HTTP rules is empty = “allow all”.  Would it be safer to skip those two rules until a gateway exists, so those ports stay closed until we know which namespaces should be allowed? - same for limitador (OIDC can stay as it is)

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+	return peers
+}
+
+func generateAuthorinoNetworkPolicy(kObj *v1beta1.Kuadrant, aObj *authorinooperatorv1beta1.Authorino, topology *machinery.Topology) *networkingv1.NetworkPolicy {
+	fromNamespaces := gatewayNamespacePeers(topology)
+
+	labels := linkedDeploymentLabels(&controller.RuntimeObject{Object: aObj}, topology)
+
+	if labels == nil {
+		labels = map[string]string{"kuadrant.io/managed": "true"}
```
fabikova, line 292: Instead of falling back to "kuadrant.io/managed: true", could we skip creating/updating the network policy until "linkedDeploymentLabels" returns real Deployment labels? That avoids briefly selecting every managed pod in the namespace.

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+
+	minAuthorinoNetworkPolicy := generateAuthorinoNetworkPolicy(kObj, authorinoObj, topology)
+
+	var existingAuthorinoNetworkPolicy *networkingv1.NetworkPolicy
+	for _, policy := range policies {
+		if policy.GetName() == AuthorinoNetworkPolicy {
+			existingAuthorinoNetworkPolicy = policy
+			break
+		}
+	}
```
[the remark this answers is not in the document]
Boomatang: Policies are filter at the controller-runtime. Non issue

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+	desiredAuthorinoNetworkPolicy, update := mergeNetworkPolicy(*minAuthorinoNetworkPolicy, existingAuthorinoNetworkPolicy)
+
+	if authorinoObj != nil {
+		ownerRef := metav1.OwnerReference{
+			APIVersion:         authorinoObj.GroupVersionKind().GroupVersion().String(),
+			Kind:               authorinoObj.Kind,
+			Name:               authorinoObj.GetName(),
+			UID:                authorinoObj.GetUID(),
+			BlockOwnerDeletion: new(true),
+			Controller:         new(true),
```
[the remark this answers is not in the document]
Boomatang: Did you try compile. This is the gopls recommend method. Update your informant. No issue

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,432 @@
...
+
+	for idx, policy := range policies {
+		p, ok := policy.(*controller.RuntimeObject).Object.(*networkingv1.NetworkPolicy)
+		if ok && p != nil {
+			output[idx] = p
+		}
+	}
+
+	return output
+}
```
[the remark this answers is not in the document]
Boomatang: getting policies already guards against this

internal/controller/networkpolicy_reconciler.go
Boomatang: Good point, my test environment always had gateways present. While check with no gateways

internal/controller/networkpolicy_reconciler.go
Boomatang: Sure that is fair.

jjkiely:
**Output of the claude go review skill from the dev-team-plugin:**


> Verdict
> 
> REQUEST CHANGES
> 
> Issues
> 
> 1. [major — confirm intent] Operand metrics/monitoring ingress is silently blocked
> internal/controller/networkpolicy_reconciler.go (generateAuthorinoNetworkPolicy ~L496, generateLimitadorNetworkPolicy ~L583)
> The policies only open OIDC (Authorino) and gRPC/HTTP (both, gateway-scoped). Since PolicyTypes: [Ingress] with a limited rule set denies everything else, Prometheus scraping of the operands' /metrics endpoints will be blocked once these policies land — unless metrics ride on an already-allowed port. For Limitador with no gateways, the ingress list is empty, i.e. a deny-all-ingress policy. Kuadrant ships observability tooling, so this seems likely to break metrics collection in the field. Please confirm whether an ingress allowance for the monitoring namespace is intended to be omitted (the follow-up operator PR does open metrics, which makes the omission here look accidental).

I will take a look into where exactly this lives in my full review so we can decide if it's valid

> 2. [major — verify at runtime] Owner reference built from possibly-empty TypeMeta

 
 **_I dug into this one and it's a false positive in our case. ```typed objects from a controller-runtime informer cache having TypeMeta stripped```  doesn't apply to how these objects reach the topology._**

> ~networkpolicy_reconciler.go:294-301 and 352-359
> ownerRef := metav1.OwnerReference{
>     APIVersion: authorinoObj.GroupVersionKind().GroupVersion().String(),
>     Kind:       authorinoObj.Kind,
>     ...
> }
> Typed objects pulled from the topology/informer cache frequently have TypeMeta cleared, so Kind/GVK can be empty at runtime → an owner ref with empty apiVersion/kind, which the API server rejects. Combined with the "log errors, never requeue, always return nil" design, that would silently prevent policy creation. Tests set TypeMeta by hand, so they can't catch it. Please verify against a live cluster that authorinoObj/lObj carry populated TypeMeta; if not, source Kind/APIVersion from known constants. (Same pattern recurs in the downstream fork PR.)~

I removed the rest as I didn't think they were worth the characters.

jjkiely:
~Point number one in the above seems most important, it would not be good to lose metrics because of these polcies.~ 

But I guess since it's an ingress only policy atm, maybe the egress alone is enough to keep the metrics working?

jjkiely:
Will be back to do a full review on this shortly

bundle/manifests/kuadrant-operator.clusterserviceversion.yaml
```diff
@@ -744,6 +744,18 @@ spec:
...
+        - apiGroups:
+          - networking.k8s.io
+          resources:
+          - networkpolicies
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
```
jjkiely, line 756: If we go ahead with this I'll need to take this change downstream too, I didn't include patch since it's pretty much equivalent to update, not an issue just flagging it for my own memory

api/v1beta1/topology.go
```diff
@@ -71,6 +75,21 @@ func LinkKuadrantToAuthorino(objs controller.Store) machinery.LinkFunc {
 	}
 }
 
+func LinkAuthorinoToDeployment(objs controller.Store) machinery.LinkFunc {
```
jjkiely, line 78: Biggest question I have is where does the dns-operator tie into this work? 

It's not included here, just authorino and limitador, which is fine because it's not an operand anymore

~But it's also not included in the kuadrant-operator namespace policy since it doesn't live there~ Nevermind, just checked and it does now after the consolidation 😅

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,483 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      AuthorinoNetworkPolicy,
+			Namespace: kObj.GetNamespace(),
+			Labels:    CommonLabels(),
+		},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{
+				MatchLabels: labels,
+			},
+			PolicyTypes: []networkingv1.PolicyType{"Ingress"},
```
jjkiely, line 335: ### Did some exploring and this is what the kuadrant dev team plugin was on about.

> These policies set `PolicyTypes: [Ingress]` with an explicit allow-list, so any ingress not in the rule set is denied. Right now that's OIDC/gRPC/HTTP for Authorino and gRPC/HTTP for Limitador — and for Limitador with no gateways the ingress list is empty (`generateLimitadorNetworkPolicy`), i.e. an effective deny-all-ingress.
> 
> That means Prometheus scraping of the operands' `/metrics` endpoints gets blocked once these land, since metrics are served on separate ports that aren't in the allow-list.

Your other PR for the operator (Boomatang/kuadrant-operator#1) adds a metrics ingress rule for the operator's own manager using this same `ingressRule` helper, so should we add something similar here?

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,483 @@
...
+	// https://github.com/Kuadrant/limitador/blob/f73e5f4b3d9af3756d4d772b35d2798693b961f9/limitador-server/src/config.rs#L101-L103
+	gRPCport := 8081
+	HTTPport := 8080
+
+	if lObj != nil {
+		gRPCport = int(lObj.GRPCPort())
+		HTTPport = int(lObj.HTTPPort())
+	}
+
+	ingress := []networkingv1.NetworkPolicyIngressRule{}
```
jjkiely, line 400: Same deny-all-except-listed issue as the Authorino policy above

jjkiely (commented):
Looks Good to my eyes at least, but worth seeing if Claude's take on the metrics stuff is valid before I approve

edit; All threads resolved, LGTM

internal/controller/networkpolicy_reconciler.go
jjkiely: The documentation for the networkpolicy work says to allow metrics for operators but does not mention operands, since we don't make heavy use of those endpoints, I guess it's fine to keep them excluded

internal/controller/networkpolicy_reconciler.go
jjkiely: Also not really needed.

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,483 @@
...
+
+	if len(fromNamespaces) > 0 {
```
fabikova, line 402: With no Gateways, Limitador gets a NetworkPolicy with empty ingress (deny-all for selected pods), while Authorino still opens OIDC. Could you explain why?

internal/controller/networkpolicy_reconciler.go
fabikova: Authorino still falls back to kuadrant.io/managed: true, but Limitador no longer does, and we only generate when hasLinkedDeployment is true. Is the Authorino fallback still needed?

internal/controller/networkpolicy_reconciler.go
fabikova: no problem with that, just to make sure I understand the intended behaviour

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,483 @@
...
+			HTTPport = int(*aObj.Spec.Listener.Ports.HTTP)
+		}
+		if aObj.Spec.OIDCServer.Port != nil {
+			OIDCdiscoveryPort = int(*aObj.Spec.OIDCServer.Port)
+		}
+	}
+
+	ingress := []networkingv1.NetworkPolicyIngressRule{
+		// OIDC discovery endpoint
+		ingressRule([]networkingv1.NetworkPolicyPeer{}, OIDCdiscoveryPort),
```
fabikova, line 315: Why is OIDC left open to all and gRPC/HTTP require Gateway namespaces?

internal/controller/networkpolicy_reconciler.go
```diff
@@ -0,0 +1,483 @@
...
+	if len(errs) > 0 {
+		span.SetStatus(codes.Error, "reconciliation completed with errors")
+		for _, err := range errs {
+			logger.Error(err, "reconciliation error")
+		}
+	} else {
+		span.SetStatus(codes.Ok, "")
+	}
+	// Don't return errors as it can cancel the context of workflows running in parallel.
+	return nil
```
fabikova, line 211: We always return nil even after create/update failures so parallel workflows aren’t cancelled. If a write fails, how does it get retried only when another watch/topology event runs the workflow again? Is there any requeue if Create keeps failing and nothing else changes in the cluster?

fabikova (approved):
LGTM, thank you for addressing previous comments. Just a few questions now

PR #2199, opened 2026-08-25
docs: TRLP on egress user guide with mock AI API

### LINKED ISSUE #2078: Validate TRLP on egress, write user guide
**JIRA:** [CONNLINK-1248](https://redhat.atlassian.net/browse/CONNLINK-1248)
## What

Validate that TokenRateLimitPolicy works on an Istio egress gateway and produce a user guide with a worked AI inference example. This follows the same validation pattern as AuthPolicy and RateLimitPolicy in 1.4 — those policies worked on egress without code changes, and TRLP uses the same wasm-shim/Limitador mechanism.

### What we expect

TokenRateLimitPolicy uses the same wasm-shim and Limitador infrastructure as RateLimitPolicy. Since RateLimitPolicy was validated on egress in 1.4 without code changes, TRLP should work the same way. The key difference is that TRLP extracts token counts from response bodies (e.g., OpenAI `usage.total_tokens`) rather than counting requests.

### Key things to validate

- TRLP attached to an egress gateway is picked up by the topology builder and reconciled correctly
- Token extraction from AI API responses works through TLS-originating egress (the egress gateway terminates internal mTLS and originates a new TLS connection to the external service — response body must still be accessible for token counting)
- Per-workload and per-tier token limits using workload identity (same SA-based identity as validated for RateLimitPolicy in 1.4)

### Potential risks

- TLS origination at the egress gateway could affect response body access for token extraction
- If the external AI API uses streaming responses (SSE), token extraction may behave differently

## Acceptance Criteria

- [ ] Validate TRLP enforces token-based rate limits on egress traffic to an AI-compatible API (or mock)
- [ ] Validate per-workload and per-tier token limits work with workload identity
- [ ] Document any limitations with streaming responses or specific API formats
- [ ] Write user guide with a worked AI inference example

maksymvavilov:
## Summary

- Add a standalone user guide for TokenRateLimitPolicy on Istio egress gateways (`doc/user-guides/egress/token-rate-limiting.md`)
- Include a mock AI API setup (llm-d-inference-sim) with a setup script and example manifests
- Cover basic token limits, per-workload budgets via workload identity, per-tier quotas, and streaming
- Add a placeholder section for upcoming token limit reservations ([architecture#190](https://github.com/Kuadrant/architecture/pull/190))
- Fix misleading "graceful fallback to request counting" claim in the TRLP overview — TRLP silently does nothing when `usage.total_tokens` is missing from the response

### Key finding during validation

The AuthPolicy for workload identity **must** include a `response.success.filters.identity` block to expose `auth.identity.username` to the wasm-shim. Without it, counter expressions referencing `auth.identity.*` fail silently (`CelError::Resolve { NoSuchKey("identity") }`) and rate limiting is bypassed. This applies to both TRLP and RLP when using `kubernetesTokenReview` with identity-based counters.

## Test plan

- [x] Walked through the entire guide end-to-end on a local Kind cluster with Kuadrant + Istio
- [x] Validated basic token limiting (429 at ~10 requests with 100 token/min limit)
- [x] Validated per-workload limits (independent budgets per SA, exhausting one does not affect the other)
- [x] Validated per-tier limits (default tier 429 at ~11 requests, gold tier unaffected)
- [x] Validated streaming responses (SSE with `include_usage: true`)
- [x] Validated cleanup scripts work correctly
- [x] SME review of TRLP behavior and limitations

Closes #2078

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive guide for applying token rate limiting to Istio egress gateways, covering global limits, workload budgets, tiered quotas, streaming responses and troubleshooting considerations.
  * Clarified behaviour when token usage cannot be extracted, including options to allow or deny requests.
  * Added a link to the new guide from the egress gateway documentation.

* **Examples**
  * Added a mock OpenAI-compatible service and setup script for validating token rate-limiting configurations, connectivity and cleanup.


eguzki (commented):
Happy to see this working for the egress use case. 

Overall the technical content is solid and LGTM

One real accuracy issue: the doc claims the default reservation.amount (when omitted) falls back to the request's max_tokens field. RFC 0021 actually specifies a flat constant default of uint(5000).  Deriving this from the request's own max_tokens would be desirable, unfortunately reading request body fields  isn't feasible today. The `requestBodyJSON` CEL function should not be used in this guide either.

maksymvavilov:
> Happy to see this working for the egress use case.
> 
> Overall the technical content is solid and LGTM
> 
> One real accuracy issue: the doc claims the default reservation.amount (when omitted) falls back to the request's max_tokens field. RFC 0021 actually specifies a flat constant default of uint(5000). Deriving this from the request's own max_tokens would be desirable, unfortunately reading request body fields isn't feasible today. The `requestBodyJSON` CEL function should not be used in this guide either.

Thank you!

eguzki:
We are tracking when the request body will be available https://redhat.atlassian.net/browse/CONNLINK-1608. then your statement would be valid.

hack/setup-egress-ai-mock.sh
```diff
@@ -0,0 +1,221 @@
...
+}
+
+kubectl get pod test-client -n "$EGRESS_TEST_NS" > /dev/null 2>&1 || {
+    error "test-client pod not found in $EGRESS_TEST_NS. Run setup-egress.sh first."
+    exit 1
+}
+
+# ── Ensure Kuadrant CR exists ────────────────────────────────────────
+if kubectl get kuadrant -A --request-timeout=5s > /dev/null 2>&1; then
+    KUADRANT_NS=$(kubectl get kuadrant -A -o jsonpath='{.items[0].metadata.namespace}' 2>/dev/null)
```
philbrookes, line 75:
```suggestion
    KUADRANT_NS=$(kubectl get kuadrant -A -o jsonpath='{.items[0].metadata.namespace}' 2>/dev/null || true) 
```

Otherwise the script crashes here when no Kuadrant CR exists.

philbrookes (commented):
This looks basically fine. I just ran into a small bug with the setup script on a fresh cluster

hack/setup-egress-ai-mock.sh
maksymvavilov: Hm, you are right - that got pushed away for some reason. There is a bigger issue here - we should not succeed if kuadrant CRD is missing. I've changed the logic a bit; it should be working now

hack/setup-egress-ai-mock.sh
```diff
@@ -0,0 +1,237 @@
...
+else
+    KUADRANT_NS="$KUADRANT_SYSTEM_NS"
+    KUADRANT_NAME="kuadrant"
+    info "Kuadrant CR not found. Creating $KUADRANT_NAME in $KUADRANT_NS..."
+    kubectl apply -f - <<EOF
+apiVersion: kuadrant.io/v1beta1
+kind: Kuadrant
+metadata:
+  name: $KUADRANT_NAME
+  namespace: $KUADRANT_NS
```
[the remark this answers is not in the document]
maksymvavilov: This is a default namespace we use in the installations, and it is hardcoded in several places. Making the default ns configurable is not a task or a scope for this doc

philbrookes (approved):

PR #2200, opened 2026-08-25
fix: component chart sync permissions and missing charts

mikenairn:
## Summary

Two fixes to `hack/sync-components`:

- **Missing chart directory:** `syncOne` treated a resolved-SHA match as "up-to-date" and skipped downloading whenever the ref already matched `sync.yaml` — even when the local chart directory was missing (e.g. a component was just uncommented). It now checks the directory actually exists before treating a ref match as up-to-date.
- **Permissions under a restrictive umask:** `os.MkdirAll`/`os.OpenFile`'s mode argument is masked by the process umask, so extracted files/directories could land more restrictive than the requested `0644`/`0755` (e.g. `0600`/`0700` under a `077` umask) — regardless of nesting depth or how many ancestor directories get created in one call. Extraction now zeroes the process umask (`syscall.Umask(0)`) for its duration, so every created path gets exactly the requested mode with no follow-up `os.Chmod` needed at all.
- Separately, a failed `os.Chmod` fixing `os.MkdirTemp`'s hardcoded `0700` on the newly-installed chart directory was only logged as a warning, letting sync report success and rewrite `sync.yaml`'s ref while leaving the chart directory broken. This is now a hard error that restores the backup (or removes the broken directory), matching the existing `updateRef`-failure error path.

Add tests for the missing-directory fix, and for umask-independent extraction (including a deeply nested path with no explicit tar directory entries for any ancestor, proving the fix isn't limited to the innermost directory a tar entry happens to name). Wire `hack/sync-components` into `UNIT_DIRS` so `make test-unit` covers this package.

## Test plan

- [x] `go test ./hack/sync-components/...` passes
- [x] Verified new tests fail against the pre-fix code and pass with the fix
- [x] `make sync-component-charts` manually verified to produce `0755`/`0644` permissions and to re-download charts with a missing local directory
- [x] `golangci-lint run hack/sync-components/...` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected component synchronisation when the configured revision matches but the local chart directory is missing.
  - Ensured downloaded chart files and directories consistently receive the intended permissions.
  - Failed permission updates now correctly report an error and restore the previous chart where possible.

- **Tests**
  - Added coverage for missing and existing chart directories.
  - Added validation of file and directory permissions under restrictive system settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ endif
 IMG ?= $(IMAGE_TAG_BASE):$(IMAGE_TAG)
 
 # Directories containing unit & integration test packages
-UNIT_DIRS := ./pkg/... ./api/... ./internal/...
+UNIT_DIRS := ./pkg/... ./api/... ./internal/... ./hack/sync-components...
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
--- a/hack/sync-components/cmd_sync.go
+++ b/hack/sync-components/cmd_sync.go
@@ -88,7 +88,13 @@ func syncOne(name string, c component, configPath string) syncResult {
 		return syncResult{Repo: c.Repo, OldRef: c.Ref, Status: "error"}
 	}
 
-	if sha == c.Ref {
+	// Check if chart directory exists before declaring up-to-date
+	chartExists := false
+	if stat, err := os.Stat(chartDir); err == nil && stat.IsDir() {
+		chartExists = true
+	}
+
+	if sha == c.Ref && chartExists {
 		fmt.Fprintf(os.Stderr, "Syncing %s from %s@%s\n", name, c.Repo, sha[:12])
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -104,7 +104,7 @@ endif
 IMG ?= $(IMAGE_TAG_BASE):$(IMAGE_TAG)
 
 # Directories containing unit & integration test packages
-UNIT_DIRS := ./pkg/... ./api/... ./internal/...
+UNIT_DIRS := ./pkg/... ./api/... ./internal/... ./hack/sync-components
```
maksymvavilov: Not sure if this matters, but `./hack/sync-components/...` might save someone in the future from headaches of "why my freshly added tests are not running"

maksymvavilov (dismissed):

maksymvavilov:
Looks good. Works locally

Makefile
mikenairn: Good point, will update it

mikenairn: updated

maksymvavilov (approved):

PR #2201, opened 2026-08-25
Extensions authenticate via TokenReview and SAR

### LINKED ISSUE #2198: Replace pre-shared secret auth with tokenreview and SAR
We opted for a pre-shared key as the auth model initially, but as part of the RFC review process we've come up with a better plan (https://github.com/Kuadrant/architecture/pull/197/changes/1b3fc34b510cf28b6c9e2649fc1cfd8c6dda1bbd):

- The extension mounts it's service account token and is provided a rolebinding giving access to register their policykind
- The handshake sends this token
- The operator no longer reads the secret but performs a `TokenReview` and `SubjectAccessReview` to confirm that the service account token is valid and who they are, and that they're allowed to register the extension for this policy kind

This issue covers this migration from the old approach toward the new

adam-cattermole:
Resolves #2198 

## Summary by CodeRabbit

* **New Features**
  * Extensions now authenticate using Kubernetes service-account tokens.
  * Added token validation and policy authorisation through Kubernetes APIs.
  * Tokens can be supplied through environment configuration or a file, with support for token rotation.
  * Extension handshakes now include a token and policy kind.

* **Changes**
  * Removed dependence on shared authentication Secrets.
  * Updated extension identity handling and session management.
  * Reduced permissions for reading Secrets while adding required authentication and authorisation permissions.


silvi-t (dismissed):
LGTM. I tested e2e on an OCP cluster — extensions authenticate and register successfully.

One optional suggestion (non-blocking): `tokenAuthenticator` (`internal/extension/authenticator.go`) is currently tested indirectly through the handshake tests in `manager_test.go`. Dedicated unit tests in an `authenticator_test.go` covering edge cases (API failure, audience mismatch, Status.Error path, convertExtra helper) would strengthen coverage — up to you whether it's worth adding here or not.

silvi-t (approved):

PR #2214, opened 2026-09-01
Extensions session liveness

### LINKED ISSUE #2175: Session lifecycle and liveness
As part of supporting standalone extensions we should ensure we implement session lifecycle logic:

- Handle transient errors
- Retry handshakes
- Reconnection logic
- Liveness detection

adam-cattermole:
Resolves #2175

## Summary by CodeRabbit

* **New Features**
  * Added automatic session expiry and cleanup using configurable time-to-live settings.
  * Added the ability to release an active session so another connection can take over immediately.
  * Added automatic session heartbeats, reconnection retries and stream recovery.
* **Bug Fixes**
  * Improved session activity tracking to prevent active sessions expiring unexpectedly.
  * Prevented stalled subscriptions from blocking updates and ensured subscribers receive the latest changes.
  * Improved handling of disconnected or unavailable services, including safer shutdown and reconnection behaviour.


eguzki (approved):
I'm finding it hard to follow all the changes happening here. Session management complexity is growing quite a bit across manager.go, session.go, and the controller-side reconnection logic.  A few pieces (session TTL, heartbeat interval, reconnect backoff, the reaper) are all independently configurable and interacting in ways that aren't obvious from any single file.

One thing that stands out: nothing reacts to connection loss on the server side. Cleanup is entirely driven by the reaper's periodic sweep, rather than anything event-driven off the stream/connection actually closing. That means worst-case cleanup latency is bounded by the janitor's schedule, not by when the disconnect actually happened.

Question: when a stream is re-established (e.g. after a transient Unavailable and retry), is the session re-validated at that point, or only implicitly via the next heartbeat/ping? It looks like Subscribe just resumes reading off the DAG update channel without re-checking ValidateSession. Want to confirm that's intentional and not a gap.

Anyway, changes are acceptable so I am giving the thumbs-up

adam-cattermole:
> I'm finding it hard to follow all the changes happening here. Session management complexity is growing quite a bit across manager.go, session.go, and the controller-side reconnection logic. A few pieces (session TTL, heartbeat interval, reconnect backoff, the reaper) are all independently configurable and interacting in ways that aren't obvious from any single file.

Agree, a bit of coupling here and also the env vars might be _too much_ configuration allowing users to shoot themselves in the foot. Perhaps I could reduce this and have a single timeout value that others are derived from, but I'll leave that as a follow up.

> One thing that stands out: nothing reacts to connection loss on the server side. Cleanup is entirely driven by the reaper's periodic sweep, rather than anything event-driven off the stream/connection actually closing. That means worst-case cleanup latency is bounded by the janitor's schedule, not by when the disconnect actually happened.

So actually, the client on the successful cleanup case closes the connection freeing it up for someone to consume straight away - on graceful exit the client sends `ReleaseSession` (`pkg/extension/controller/client.go`), which revokes the session server-side immediately - so in regular rollouts the outgoing pod frees the slot for the new pod to take over, rather than waiting on the reaper. Regarding worst-case cleanup, this is partially by design - the server has no way of knowing that the extension is gone forever vs back in a second. If the server revoked on stream disconnect, every transient error would force a full re-handshake instead of a same-session resubscribe (this is important as the handshake is going to 'sync-state' eventually). So the TTL is a grace window for transient errors. Takeover also isn't entirely gated on the reapers tick - once a session becomes stale, a handshake from a returning/new client deletes the old session itself - the reaper only clears once the client isn't returning at all (or doesn't return in the time period). The one caveat is that if the session isn't stale yet, a client dying ungracefully and restarting trying to handshake before ttl has elapsed will be rejected.

> Question: when a stream is re-established (e.g. after a transient Unavailable and retry), is the session re-validated at that point, or only implicitly via the next heartbeat/ping? It looks like Subscribe just resumes reading off the DAG update channel without re-checking ValidateSession. Want to confirm that's intentional and not a gap.

So every RPC (apart from Handshake of course) is validated to have an authenticated session token (it's handled by the interceptor), when the client re-establishes the connection:
- If the session has not yet been reaped; whether it's stale or not, the connection is accepted and the liveness is refreshed to avoid being reaped - all continues on the same session
- If the session has been reaped; the client gets `Unauthenticated` and knows that it needs to go back to `Handshake` again to be provided a new session token (if accepted)

PR #2215, opened 2026-09-01
fix: reject response-only token limit predicates

### LINKED ISSUE #2145: responseBodyJSON in when predicates is not a coherent concept for TokenRateLimitPolicy limits: should be validated and rejected
### Summary
`TokenRateLimitPolicy` limits accept `responseBodyJSON(...)` in `when` predicates today, but this is not a coherent thing to express: a limit's `when` predicates apply to both the pre-flight check and the post-response report for that limit. The check runs *before* the upstream call is made — a response, and therefore a response body, cannot exist yet. There is no reading of `responseBodyJSON` in a `when` predicate that is meaningful for this policy type; it should never be accepted, not merely "used more carefully."

### Evidence
Tested live against a TokenRateLimitPolicy with:
```yaml
when:
- predicate: request.path == "/v1/chat/completions"
- predicate: responseBodyJSON("/object") != "error"
```
Every request logs:
```
wasm log kuadrant-wasm-shim: Missing json property: /object
wasm log kuadrant-wasm-shim: Task failed: "1"
wasm log kuadrant-wasm-shim: Task failed: "2"
```
and Limitador's `authorized_calls`/`report_calls`/`limited_calls` never move — the limit is never enforced, silently. `TokenRateLimitPolicy.status` reports `Enforced: True` throughout, with nothing indicating the limit isn't active.

Even when the referenced JSON pointer happens to resolve (e.g. `responseBodyJSON("/model")`), no runtime error occurs, but the check call can only have executed *after* the response existed — i.e. after the upstream LLM call it should have gated had already completed. It degenerates into a redundant post-hoc counter bump, not a pre-flight gate. There's no working interpretation here, just different failure shapes depending on the response.

### Expected behavior
`responseBodyJSON` should be rejected outright when it appears in a `TokenRateLimitPolicy` limit's `when` predicates, at policy validation/admission time, with the policy's `Enforced` condition set to `False` and a clear message explaining why. Silent acceptance followed by silent no-op/failure at the data-plane level is the specific behavior to eliminate.

riccardomenegazzo:
## Summary

Reject `responseBodyJSON(...)` when it is used in `TokenRateLimitPolicy` `when` predicates that participate in the pre-request rate-limit guard.

A token limit's predicates are shared by both its request-phase check and its response-phase report. The request-phase check runs before an upstream response exists, so a predicate that depends on `responseBodyJSON` cannot be evaluated coherently and can result in rate limiting being silently bypassed while the policy appears enforced.

The validation is phase-aware and operates on the checked CEL AST rather than matching expression text. This means:

* `responseBodyJSON(...)` is rejected in TokenRateLimitPolicy guard predicates;
* `requestBodyJSON(...)` remains valid in those predicates;
* response-phase report expressions remain valid;
* string literals that merely contain the text `responseBodyJSON(...)` are not false positives.

No API or CRD changes are required.

## Testing

Added unit coverage for:

* rejecting a real `responseBodyJSON` call in a token-rate-limit guard predicate;
* allowing `requestBodyJSON` in the same phase;
* allowing `responseBodyJSON` in response-phase report predicates;
* avoiding false positives when the function name appears only inside a CEL string literal.

## Why this matters

`TokenRateLimitPolicy` is an enforcement API. Accepting a predicate that cannot exist during the pre-flight decision is worse than returning a validation error because it can give operators a false sense that a token budget is being enforced.

Failing the expression through the existing CEL validation pipeline also means the policy status can surface the configuration problem instead of deferring it to a runtime wasm failure.

Fixes #2145

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for TokenRateLimit policy predicates.
  * Prevented unsupported response-body data from being used in request-phase guard predicates.
  * Continued to allow request-body data for guard predicates and response-body data for report predicates.
  * Improved detection to avoid rejecting predicates when function names appear only within text strings.

### DIFF
--- a/internal/cel/kuadrant_validator.go
+++ b/internal/cel/kuadrant_validator.go
@@ -1,9 +1,12 @@
 package cel
 
 import (
+	"fmt"
+
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common/types/ref"
 	"github.com/samber/lo"
+	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
 
 	"github.com/kuadrant/kuadrant-operator/internal/wasm"
 )
@@ -96,7 +99,6 @@ func NewRootValidatorBuilder() *ValidatorBuilder {
 			// just for parsing and checking purposes, not evaluation
 			return nil
 		},
-		),
 	)
 	builder.AddFunction("responseBodyJSON", responseBodyJSON)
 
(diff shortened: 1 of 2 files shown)


eguzki:
this PR does not compile

PR #2235, opened 2026-09-08
Revert "ADD: blocking network policy"

### LINKED ISSUE #2234: Wasm plugin not able to be pulled
**Describe the bug**
Due to new deny-all https://github.com/Kuadrant/kuadrant-operator/pull/2221 policy on Openshift clusters the wasm plugin cannot be pulled by gateway causing 503 responses on any protected gateways.

**To Reproduce**
Smoke target in https://github.com/Kuadrant/testsuite fails
```sh
make smoke
```

azgabur:
To temporarily fix https://github.com/Kuadrant/kuadrant-operator/issues/2234 and to allow nightly builds to keep working, PR https://github.com/Kuadrant/kuadrant-operator/pull/2221 will be reverted until more permanent fix to the policies in in place.

This reverts commit c6c4579d37fe55a9438c093f36290e30afdc8290.

## Summary by CodeRabbit

* **Changes**
  * Default-deny ingress network policies have been removed from the operator’s Kubernetes and Helm deployment configurations. Deployments will no longer automatically apply this namespace-wide restriction.
  * Operator SDK bundle metadata now references version 1.33.0.
  * Development environment setup now installs the available OLM version instead of enforcing a specific version.


jjkiely:
I'm not sure that reverting this is the right change, we are required to have a networkpolicy for the next release, so instead of reverting it, shouldn't we just identify what is causing the nightly failures and fix it? 

If there is some network activity that is failing due to the new policy, then let's change the policy to allow that traffic instead of removing the policy altogether

azgabur:
I completely agree and if we can find the fix quickly enough that would be preferable choice now. I suggest reverting so nightly builds start working again tomorrow, while also work on the fix. The deny-all networkpolicy can be deployed as one CR for testing the fix and when the fix is merged the deny-all networkpolicy can be again merged. I just want to prevent having main broken for the time being.

jjkiely:
I think I'm okay to approve that plan, as long as you will start looking for the fix immediately

We need to have a working policy by the end of this sprint, so the sooner we have this policy working in the nightlies the better

fabikova:
I will start looking into that, considering that I don't have experience with operators, Alex will help me with the fixes

jjkiely (approved):
Cool, I will be freed up tomorrow after the release so if you need any help with the investigation let me know

PR #2241, opened 2026-09-09
Extensions SDK cleans up finalizer correctly

### LINKED ISSUE #2233: Fix extension finalizer cleanup

adam-cattermole:
This addresses the issue where a policy is marked for deletion when the extension isn't running - once the extension starts, the event received for the policy is a `Create` event not `Update`, so it would have never cleared it's finalizer.

Resolves #2233

## Summary by CodeRabbit

* **Bug Fixes**
  * Finalizer cleanup now runs consistently after reconciliation, including create and unknown events.
  * Objects already being deleted no longer receive a new finalizer.
  * Reconciliation now handles missing manager configuration safely.
  * Improved behaviour when policy-clearing requests fail.

### DIFF
--- a/pkg/extension/controller/controller.go
+++ b/pkg/extension/controller/controller.go
@@ -331,13 +331,11 @@ func (ec *ExtensionController) Reconcile(ctx context.Context, request reconcile.
 		return result, err
 	}
 
-	if eventType == EventTypeUpdate {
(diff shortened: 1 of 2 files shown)


silvi-t (approved):
LGTM. Tested on cluster with a standalone extension. Verified that a CR deleted while the extension was scaled down gets properly cleaned up when the extension comes back up.

PR #2244, opened 2026-09-09
Fix: allow WASM plugin access through deny-all NetworkPolicy

### LINKED ISSUE #2234: Wasm plugin not able to be pulled

fabikova:
Fixes #2234

## Summary
- Re-introduces the default deny-all `NetworkPolicy` (cherry-pick of #2221).
- **First commit** (`fix: do not set blockOwnerDeletion on NetworkPolicy owner refs`): set `blockOwnerDeletion: false` so allow NPs can be created on OCP without extra RBAC.
- **Follow-up** (latest commit): same root cause, matching the rest of the operator — add `*/finalizers` on the NP owners and keep `blockOwnerDeletion: true`.

Without a fix, deny-all ships but allow NPs fail with:
> `cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on`

Gateway envoy then cannot fetch WASM (`failure_policy: FAIL_RELOAD`) → **503**.

The NP reconciler was the only controller missing those finalizer rules. Owners are Deployments (operator/WASM), Authorino, and Limitador.

---

## How to reproduce (bug)

**Images (linux/amd64):**
- **broken:** `quay.io/mfabikov/kuadrant-operator:np-broken` (`blockOwnerDeletion: true`, **no** finalizers RBAC)
- **workaround:** `quay.io/mfabikov/kuadrant-operator:np-fix` (`blockOwnerDeletion: false`)
- **fix (this PR):** `quay.io/mfabikov/kuadrant-operator:np-rbac` (`blockOwnerDeletion: true` **+** finalizers RBAC)

**Steps on OCP with Istio GatewayClass + Kuadrant CR, deny-all in `kuadrant-system`:**

1. Update the operator image to the broken version:
   ```bash
   oc set image deploy/kuadrant-operator-controller-manager manager=quay.io/mfabikov/kuadrant-operator:np-broken -n kuadrant-system

```

2. Delete allow NPs if any; leave deny-all.
3. Check operator logs for errors:
```bash
oc logs deploy/kuadrant-operator-controller-manager -n kuadrant-system | grep 'cannot set blockOwnerDeletion'

```

4. Check NetworkPolicies (should only show default-deny-all):
```bash
oc get networkpolicy -n kuadrant-system

```

5. Run smoke tests:
```bash
make smoke

```

**Expected:** allow NP create forbidden; curl/smoke 503.

---

## How to see the fix (RBAC + blockOwnerDeletion: true)

ClusterRole must include the new rules (kustomize deploy from this branch, or patch `kuadrant-operator-manager-role`):

* `apps` → `deployments/finalizers` (update)
* `operator.authorino.kuadrant.io` → `authorinos/finalizers` (update)
* `limitador.kuadrant.io` → `limitadors/finalizers` (update)

**Steps:**

1. Update the operator image to the fixed version:
```bash
oc set image deploy/kuadrant-operator-controller-manager manager=quay.io/mfabikov/kuadrant-operator:np-rbac -n kuadrant-system

```

2. Force recreate so `ownerRefs` are not leftover from `np-fix` (leave `kuadrant-operator-default-deny-all`):
```bash
oc delete networkpolicy -n kuadrant-system authorino-operator dns-operator-controller-manager kuadrant-authorino kuadrant-limitador kuadrant-operator-controller-manager limitador-operator-controller-manager

```

3. Verify NetworkPolicies and ownerReferences (should show `"blockOwnerDeletion": true`):
```bash
oc get networkpolicy -n kuadrant-system
oc get networkpolicy kuadrant-operator-controller-manager -n kuadrant-system -o jsonpath='{.metadata.ownerReferences}{"\n"}'

```

4. Check operator logs (should be empty):
```bash
oc logs deploy/kuadrant-operator-controller-manager -n kuadrant-system | grep 'cannot set blockOwnerDeletion'

```

5. Verify operator NP allow rules (with a Gateway in e.g. `kuadrant`, must allow `8082` from that namespace — ingress: `8080` (any); `50051` + `8082` from `kubernetes.io/metadata.name: kuadrant`):
```bash
oc get networkpolicy kuadrant-operator-controller-manager -n kuadrant-system -o yaml

```

6. Verify WASM pull from that namespace (istio-proxy often has no curl):
```bash
oc run wasm-curl -n kuadrant --image=curlimages/curl --restart=Never -- curl -sS -o /dev/null -w "%{http_code}\n" [http://kuadrant-operator-wasm.kuadrant-system.svc.cluster.local:8082/plugin.wasm](http://kuadrant-operator-wasm.kuadrant-system.svc.cluster.local:8082/plugin.wasm)
oc wait --for=jsonpath='{.status.phase}'=Succeeded pod/wasm-curl -n kuadrant --timeout=60s
oc logs wasm-curl -n kuadrant  # should show 200
oc delete pod wasm-curl -n kuadrant

```

7. Run smoke tests:
```bash
make smoke

```

**Expected:** allow NPs created; WASM 200; auth + limitador smoke 200.


azgabur (dismissed):
tests in testsuite pass, the wasm problem was fixed

jjkiely (commented):
This fix is nice and simple, and it seems to work which is awesome.

However, [I've noticed a thread on slack](https://redhat-internal.slack.com/archives/C0BE6GY5DEW/p1788884705020179) which seems to indicate another issue with the networkPolicy reconciliation.

It seems that there is an RBAC issue which is preventing the policies from ever having the ownerRef set.

I think this is the root cause of this bug, and also something we need to fix anyway since it will affect more than the wasm-shim

I looked into it and I think we need to add RBAC rules for the deployments, authorino, and limitador finalizers. And if we do this it will also fix the wasm-shim deployment.

All of our other reconcilers have those finalizers rules, the np reconciler is the only one with them missing.

Could you try adding those rules and building the operator again and seeing if it also fixes the issue?

If that does turn out to be the root cause, there is still a separate conversation to be had about what value we want 			BlockOwnerDeletion to be set to in the reconciler.

jjkiely (commented):
Changes look good to me, thanks @fabikova 

Just waiting for a cluster to spin up so I can run through the verification steps.

FYI @mikenairn this should also fix the RBAC issues you were seeing with Jim's original policy work, as well as fixing the original bug we were seeing with the wasm plugin.

jjkiely:
Also, if we're no longer going to be using the original ownerDeletion fix in your first commit, could you remove it from this PR? It would give a cleaner git history that way

fabikova:
@jjkiely thanks for the review, please let me know if everything worked as expected after testing on a cluster. My first commit is removed.

mikenairn:
> FYI @mikenairn this should also fix the RBAC issues you were seeing with Jim's original policy work, as well as fixing the original bug we were seeing with the wasm plugin.

Q: Do you even need/want `blockOwnerDeletion` on the ownerRef (The reason it wants to add a finalizer)? What is the owner a deployment?

fabikova:
> > FYI @mikenairn this should also fix the RBAC issues you were seeing with Jim's original policy work, as well as fixing the original bug we were seeing with the wasm plugin.
> 
> Q: Do you even need/want `blockOwnerDeletion` on the ownerRef (The reason it wants to add a finalizer)? What is the owner a deployment?

Operator NPs are owned by the controller deployments; Authorino/Limitador NPs by those CRs.
blockOwnerDeletion: true is what made ocp require the finalizers RBAC. From what I understood, keeping true is to match the other reconcilers. If we only want the NP deleted when the owner is deleted, false is enough and we wouldn’t need that extra RBAC.

jjkiely:
> > FYI @mikenairn this should also fix the RBAC issues you were seeing with Jim's original policy work, as well as fixing the original bug we were seeing with the wasm plugin.
> 
> Q: Do you even need/want `blockOwnerDeletion` on the ownerRef (The reason it wants to add a finalizer)? What is the owner a deployment?

Good question, and like Martina says I don't think there's really  a strong argument for it other than convention from the other reconcilers, I also wasn't sure what specific RBAC issues you were seeing after Jim's original PR merged, so wanted to cover all bases by suggesting to add the finalizers.

If setting blockOwnerDeletion to false also fixes the RBAC issue that you were seeing @mikenairn then I'm happy to revert back to just deleting that block for simplicity.

Either fix works and I don't have a strong (or any) opinion either way. 

Sorry for all the flip-flopping on this @fabikova :sweat_smile: Looks like maybe we can revert back to the way this was when you first opened it

fabikova:
> > > FYI @mikenairn this should also fix the RBAC issues you were seeing with Jim's original policy work, as well as fixing the original bug we were seeing with the wasm plugin.
> > 
> > 
> > Q: Do you even need/want `blockOwnerDeletion` on the ownerRef (The reason it wants to add a finalizer)? What is the owner a deployment?
> 
> Good question, and like Martina says I don't think there's really a strong argument for it other than convention from the other reconcilers, I also wasn't sure what specific RBAC issues you were seeing after Jim's original PR merged, so wanted to cover all bases by suggesting to add the finalizers.
> 
> If setting blockOwnerDeletion to false also fixes the RBAC issue that you were seeing @mikenairn then I'm happy to revert back to just deleting that block for simplicity.
> 
> Either fix works and I don't have a strong (or any) opinion either way.
> 
> Sorry for all the flip-flopping on this @fabikova 😅 Looks like maybe we can revert back to the way this was when you first opened it

@mikenairn what is your opinion on this?

jjkiely (approved):
Looks good, thanks Martina

mikenairn:
> @mikenairn what is your opinion on this?

Agree, `blockOwnerDeletion` should be false IMO. When the owner is removed it will cascade delete the `NetworkPolicy` which is what you care about. I see no scenario where blocking the deletion of the owner is advisable or wanted. This is likely to cause issues.

Issue #2243, opened 2026-09-09
TelemetryPolicy: bindings accumulate across spec updates, so removed and renamed labels persist in the generated wasm config


maksymvavilov:
> **Correction (2026-09-10):** this issue was originally scoped to *removing* a label. Further testing showed the defect is broader — bindings accumulate across **every** update to the spec, so renamed and edited labels persist too. The body below has been corrected.

## What

Updating a `TelemetryPolicy`'s `spec.metrics.default.labels` never removes bindings from the generated `EnvoyFilter`; each `kubectl apply` only adds to what is already there. Labels that were removed, renamed, or had their expression changed all persist indefinitely, and the bindings from every past revision of the policy accumulate side by side. Updating a policy therefore cannot be used to undo a bad label — the only recovery is to delete the `TelemetryPolicy` and recreate it.

This compounds badly with #2242: because a single failing expression discards the whole descriptor, one stale label left behind by an earlier revision is enough to stop **all** metrics for the route, even though the current spec is perfectly valid. The intuitive fix — edit the policy to drop or correct the offending label — does not work.

It also makes the policy hard to iterate on. Successive edits while tuning a label expression silently pile up, so after a handful of applies the generated config bears little resemblance to the spec.

## Steps to reproduce

1. Apply a `TelemetryPolicy` targeting a Gateway with several labels, e.g. `workload`, `destination`, `probe_response_code`.
2. Confirm the bindings appear in the generated `EnvoyFilter`:

```sh
kubectl get envoyfilter kuadrant-kuadrant-egressgateway -n gateway-system -o json \
  | python3 -c "import json,sys,re; print(sorted(set(re.findall(r'probe_[a-z_]+', json.dumps(json.load(sys.stdin))))))"
```

3. `kubectl apply` an updated version of the same policy with one label removed, one renamed, and one expression changed.
4. Re-run the command from step 2.

## Observed

The policy spec no longer contains the removed label:

```console
$ kubectl get telemetrypolicy egress-telemetry -n gateway-system -o jsonpath='{.spec.metrics.default.labels}'
{"destination":"request.host","probe_literal":"\"egress\"","probe_method":"request.method",
 "probe_path":"request.path","probe_scheme":"request.scheme",
 "probe_source_addr":"source.address","workload":"auth.identity.username"}
```

But the `EnvoyFilter` still carries it, rechecked after 30s+ to rule out reconcile lag:

```console
[...]

Issue #2242, opened 2026-09-09
TelemetryPolicy: a label expression referencing an unavailable attribute discards the whole descriptor and silently stops all metrics


maksymvavilov:
> **Correction (2026-09-10):** this issue was originally filed as "all traffic returns HTTP 500". Further testing showed that symptom is conditional on the route's failure modes. The general, always-present symptom is silent loss of *all* metrics for the route. The body below has been corrected; the underlying defect is unchanged.

## What

A `TelemetryPolicy` label whose CEL expression references an attribute that is not available in the request phase is not dropped — it breaks the construction of the entire rate limit descriptor. **No metrics are recorded at all for the affected route**, including the labels that would otherwise resolve fine, while the policy continues to report `Enforced=True`.

Whether the request itself also fails depends on the failure modes of the services on the route:

| Route configuration | Result |
|---|---|
| `RateLimitPolicy` / `TokenRateLimitPolicy` only | HTTP 200, metrics silently lost |
| Also carries an `AuthPolicy` | HTTP 500 |
| No rate limit policy | Unaffected — no descriptor is built |

The split comes from the defaults in [`internal/wasm/utils.go:74-99`](https://github.com/Kuadrant/kuadrant-operator/blob/main/internal/wasm/utils.go#L74-L99): `AuthServiceFailureMode` defaults to `FailureModeDeny`, while `RatelimitServiceFailureMode`, `RatelimitCheckServiceFailureMode` and `RatelimitReportServiceFailureMode` all default to `FailureModeAllow`.

So there are two distinct problems here:

1. A metrics-labelling policy silently disables metrics it was created to produce, with no signal in policy status.
2. On a route that also has an `AuthPolicy`, that same policy can take down data plane traffic.

This also contradicts the documented behaviour in [`doc/overviews/telemetrypolicy.md:10`](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/overviews/telemetrypolicy.md#L10):

> Only labels with CEL expressions that resolve successfully will be included.

## Steps to reproduce

1. Have a `RateLimitPolicy` or `TokenRateLimitPolicy` in effect on a route, so Limitador metrics are produced. Add an `AuthPolicy` as well if you want to reproduce the HTTP 500 variant.
2. Apply a `TelemetryPolicy` with a mix of valid labels and one referencing a response-phase attribute:

```yaml
apiVersion: extensions.kuadrant.io/v1alpha1
kind: TelemetryPolicy
[...]

PR #2165, opened 2026-08-04
docs(egress): add distributed tracing section to observability guide

maksymvavilov:
## Summary

- Extends the egress observability guide with a distributed tracing section, based on hands-on verification against a Kind cluster with Jaeger
- Documents the egress span chain (kuadrant-filter, Authorino, Limitador), trace boundary correlation via `x-request-id`, `disableContextPropagation` for preventing trace header leaking to external services, and non-sidecar workload behavior
- Addresses CodeRabbit review feedback from #2108: scoped Telemetry selector, HTTP-only error rate label, `UPSTREAM_CLUSTER_RAW` for Istio 1.23+, corrected `rq_error` explanation
- Improves JTBD framing: goal-oriented title, user-situation introduction, common tasks navigation table, forward references

## Test plan

- [ ] Verify guide renders correctly on the docs site
- [ ] Verify all internal links and anchors resolve
- [ ] Walk through the tracing section commands on a Kind cluster with egress gateway and Jaeger

Closes #2076

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Documentation

* Expanded and retitled the egress observability guide with task-oriented navigation and clearer egress-versus-ingress guidance
* Added HTTP and proxy-level error monitoring queries
* Enhanced gateway-scoped access-log configuration and upstream failure interpretation
* Added comprehensive distributed tracing guidance, including span chains, trace boundaries, header-leak prevention, sidecar-less workloads and troubleshooting
* Updated guidance to cover monitoring egress traffic and distributed tracing
* Added links to related tracing and egress observability guidance

PR #2245, opened 2026-09-09
docs(egress): document TelemetryPolicy custom metric labels

### LINKED ISSUE #2080: Validate TelemetryPolicy on egress, write user guide
**JIRA:** [CONNLINK-1251](https://redhat.atlassian.net/browse/CONNLINK-1251)

## What
Validate that TelemetryPolicy (an OOP extension) works on an Istio egress gateway and produce a user guide with worked examples of custom metric labelling for egress traffic.

### What we expect

TelemetryPolicy is an extension that publishes data bindings consumed by the wasm-shim to add custom labels to Envoy metrics. It uses CEL expressions to extract values from request/response attributes. Since the wasm-shim operates direction-agnostically (validated in 1.4), TelemetryPolicy should work on egress without code changes.

### Key things to validate

- TelemetryPolicy attached to an egress gateway is picked up by the extension manager and reconciled
- CEL expressions resolve correctly in the egress context (available attributes may differ from ingress)
- Custom metric labels appear on Prometheus metrics for egress traffic

## Acceptance Criteria

- [ ] Validate TelemetryPolicy applies custom metric labels to egress traffic
- [ ] Validate CEL expressions work in the egress context for useful dimensions (destination, workload, response status)
- [ ] Write user guide with worked examples and sample Prometheus queries

maksymvavilov:
## Summary

Adds a **Custom Metric Labels with TelemetryPolicy** section to the egress observability guide, based on hands-on validation against a Kind cluster with an Istio egress gateway. Every command, output sample and CEL result in the section was observed on a live cluster.

TelemetryPolicy works on an egress gateway with no code changes, but several things about it are easy to get wrong, and each is documented here:

- **It labels the Kuadrant data plane metrics, not `istio_requests_total`.** Two components carry the labels: the Limitador counters (`authorized_calls`, `limited_calls`, `authorized_hits`, `report_calls`) get the complete set, and Authorino applies the same set to its own `auth_server_authconfig_*` metrics, served on `/server-metrics` rather than `/metrics`. So TelemetryPolicy does *not* fix the "`source_workload` is the gateway, not the caller" problem on the Istio metrics — it provides a parallel metric stream that does carry workload identity. The `source_workload` caveat now forward-references the new section.
- **A rate limit policy must be in effect on the route.** Labels ride on the descriptor sent to Limitador, so a gateway-wide TelemetryPolicy over a route with no `RateLimitPolicy`/`TokenRateLimitPolicy` emits nothing — while still reporting `Accepted=True` / `Enforced=True`.
- **An `auth.*` label additionally needs the rate limit policy to reference `auth.*`.** Otherwise Kuadrant orders the rate limit check ahead of authentication and the label is dropped from the descriptor with no error and no status change. Filed as #2279; documented here with the one-line workaround (an `auth.identity.username` counter on the rate limit policy).
- **Not every Authorino series carries every label.** Authorino resolves the labels at several points in its pipeline and only the last runs after identity resolution, so `workload` reaches `auth_server_authconfig_response_status` and nothing else. Aggregate those with an explicit `sum by (...)`.

Also covers: the `targetRef`-is-Gateway-only restriction, a CEL attribute availability matrix for the egress request phase, worked examples for workload and destination attribution, PromQL over the custom labels, and label cardinality guidance.

### Validated CEL attributes (egress request phase)

| Expression | Available | Example value |
|---|---|---|
| `auth.identity.username` | Yes, with an AuthPolicy **and** an auth-aware rate limit policy | `system:serviceaccount:egress-test:team-gold` |
| `request.host` | Yes | `api.ai-mock.local` |
| `request.path` / `.method` / `.scheme` | Yes | `/v1/chat/completions`, `POST`, `http` |
| `source.address` | Yes | `10.244.0.25:54146` — includes the ephemeral port |
| literal, e.g. `"egress"` | Yes | `egress` |
| `response.code` | **No** | descriptors are built before the response exists |

`source.address` creates a new time series per connection. The guide documents

```
source.address.substring(0, source.address.lastIndexOf(":")).replace("[", "").replace("]", "")
```

rather than `split(":")[0]`, which works for IPv4 but truncates an IPv6 address to `[2001`. The `lastIndexOf` form was checked in the wasm-shim (cel-rust) against `10.244.0.24:54146`, `[2001:db8::1]:8080`, `[fe80::1ff:fe23:4567:890a]:9000` and a fully uncompressed `[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:41235`. It is plain string truncation, so the address is never reformatted or re-compressed.

### Known limitations recorded in the guide

Three bugs were found during validation and filed separately; the guide references each plus its workaround:

- #2242 — a label expression referencing an unavailable attribute (for example `response.code`) discards the *whole* descriptor, so **all metrics for the route stop silently** while the policy still reports `Enforced=True`. Whether traffic also fails depends on the route's failure modes: the rate limit services default to `failureMode: allow` (200s, metrics lost), while an AuthPolicy on the same route defaults to `deny` (HTTP 500).
- #2243 — bindings accumulate across *every* spec update, not just deletions, so removing or renaming a label leaves the old one in the generated config. Because one failing expression discards the whole descriptor, a single stale label is enough to stop all metrics for the route. Only delete-and-recreate clears it.
- #2279 — an `auth.*` label is silently dropped unless the route's rate limit policy also references `auth.*`.

The inaccurate claim at `doc/overviews/telemetrypolicy.md:10` ("Only labels with CEL expressions that resolve successfully will be included") is **intentionally not fixed here** — correcting it is part of the acceptance criteria of #2242, so the doc and the behaviour change together.

### Review feedback

- Reading Limitador's `/metrics` with `kubectl exec` from the workload namespace no longer works: the `kuadrant-limitador` NetworkPolicy admits ingress only from namespaces containing a Gateway, so `egress-test` is blocked. Switched to port-forward, matching what the other observability guides do. Thanks @philbrookes for catching it.
- The tracing section's `### Prerequisites` heading collided with the page's top-level one, producing an ambiguous anchor. Renamed to `### Tracing Prerequisites`. Pre-existing, introduced by #2165; folded in here since the same file was already open.

## How to review

Single file, `doc/user-guides/egress/observability.md`, 5 commits. Suggested order:

1. `9f61127e` — the new `## Custom Metric Labels with TelemetryPolicy` section (the bulk of the diff), inserted between `## Egress Metrics` and `## Access Logging`, plus the forward reference from **Understanding `source_workload`**, the reworded **Next Steps** entry and two new **References** links.
2. `f214f56a` — review round: removed redundant `istio_*` framing, fixed a self-contradiction about egress-specific configuration, switched plain YAML to `kubectl apply -f - <<EOF` to match the sibling egress guides, replaced the IPv4-only `split`.
3. `fbc62b03` — port-forward, and the `Tracing Prerequisites` rename.
4. `4b09bacc` — the #2279 requirement.
5. `d13c9298` — the Authorino half of "Which Metrics Get Labeled".

Areas worth a close look: whether the placement alongside the metrics content reads correctly given that these are Limitador and Authorino rather than Istio metrics, and whether the known-limitations framing is the right call versus waiting for the fixes.

### Re-verified on a fresh cluster from current `main`

Full `make local-cleanup && make local-setup` on `main` @ `a716f30e`, plus `config/samples/kuadrant_v1beta1_kuadrant.yaml` and `hack/setup-egress.sh`. Route `httpbin-external` with the guide's `AuthPolicy` and a `RateLimitPolicy`:

- [x] Egress data path works: 200 with an SA token, 401 without
- [x] TelemetryPolicy extracted verbatim from the guide's own code block reports `Accepted=True` / `Enforced=True`
- [x] All three labels land on Limitador:
      `authorized_calls{workload="system:serviceaccount:egress-test:default",destination="httpbin.org",source_ip="10.244.0.19",limitador_namespace="gateway-system/httpbin-external"} 4`
- [x] #2279 reproduces and its workaround works: with `limit: 100/1m` alone the descriptor holds only `destination` and `source_ip`; adding an `auth.identity.username` counter reorders auth ahead of the rate limit check and `workload` appears
- [x] The removed `kubectl exec` command fails exactly as reported: `curl: (28) Connection timed out after 15001 milliseconds`
- [x] Control for that: the identical curl from a pod in `gateway-system` succeeds immediately, so the service and port are fine and the source namespace is the only variable
- [x] Port-forward returns real data through the NetworkPolicy
- [x] Authorino `/server-metrics` carries the labels — `destination` and `source_ip` on `auth_server_authconfig_total`, `_duration_seconds` and `_evaluator_*`; `workload` only on `_response_status`
- [x] All relative links and in-page anchors in the file resolve

### Verified earlier, on the base commit (`0e2ea665`)

Route `ai-mock-external` with an `AuthPolicy` and a `TokenRateLimitPolicy`, which is the scenario the guide's examples describe:

- [x] Labels appear on `limited_calls` once a workload exceeds its limit, verified by driving traffic past the tier limit until requests returned 429
- [x] A route with no rate limit policy produces zero series while traffic still succeeds
- [x] `response.code` reproduces the `UndeclaredReference("response")` log line and the metric loss of #2242
- [x] Editing the spec leaves stale bindings in the `EnvoyFilter`; deleting the policy clears them (#2243)

### Outstanding

- [ ] Guide renders correctly on the docs site

### Reproducing locally

```sh
# with an egress gateway and an AuthPolicy in place, and a rate limit policy
# that references auth.* — see #2279 for why the counter matters
kubectl apply -f - <<'EOF'
apiVersion: kuadrant.io/v1
kind: RateLimitPolicy
metadata:
  name: egress-ratelimit
  namespace: gateway-system
spec:
  targetRef:
    kind: HTTPRoute
    name: httpbin-external
  limits:
    global:
      rates:
        - limit: 100
          window: 1m
      counters:
        - expression: auth.identity.username
EOF

kubectl apply -f - <<'EOF'
kind: TelemetryPolicy
metadata:
  name: egress-telemetry
  namespace: gateway-system
spec:
  targetRef:
    kind: Gateway
    name: kuadrant-egressgateway
  metrics:
    default:
      labels:
        workload: auth.identity.username
        destination: request.host
        source_ip: source.address.substring(0, source.address.lastIndexOf(":")).replace("[", "").replace("]", "")
EOF

kubectl get telemetrypolicy egress-telemetry -n gateway-system \
    -o jsonpath='{range .status.conditions[*]}{.type}={.status}{"\n"}{end}'

# send authenticated traffic through the egress gateway, then:
kubectl port-forward -n kuadrant-system svc/limitador-limitador 8080:8080 >/dev/null 2>&1 &
curl -sS --max-time 15 http://localhost:8080/metrics | grep -v '^#' | grep authorized_calls
```

Note that a TelemetryPolicy must be **deleted** rather than edited between experiments — see #2243, or every run after the first is meaningless.

## Notes

Independent of #2165 (egress distributed tracing docs), which has since merged; this branch is rebased on top of it.

Closes #2080

## Summary by CodeRabbit

* **Documentation**
  * Reorganised the egress observability guide around common monitoring tasks.
  * Added PromQL examples for HTTP errors and missing responses.
  * Expanded guidance on custom metric labels, CEL attributes, IPv6-safe source IPs, cardinality, and custom-label queries.
  * Clarified requirements, failure modes, policy-edit limitations, and potential metric or response impacts.
  * Updated access-log and upstream-cluster configuration guidance.
  * Added distributed tracing configuration, span relationships, trace boundaries, header protection, sidecarless tracing, and troubleshooting guidance.
  * Updated Next Steps and References with related metrics and tracing documentation.


maksymvavilov:
@coderabbitai review

doc/user-guides/egress/observability.md
```diff
@@ -138,6 +138,180 @@ For Kubernetes queries (kubectl, log filtering), use the pod label `gateway.netw
...
+To label by response status, use the `istio_requests_total` metric and its `response_code` label instead, as shown in [PromQL examples](#promql-examples).
+
+### Label Cardinality
+
+Every distinct combination of label values creates a new Prometheus time series, so a high-cardinality expression on a busy egress gateway is expensive.
+
+`source.address` is the common trap: it includes the ephemeral source port, so **every connection produces a new series**. Strip the port with CEL:
+
+```yaml
+source_ip: source.address.split(":")[0]
```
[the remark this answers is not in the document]
maksymvavilov: addressed in the last commit

doc/user-guides/egress/observability.md
```diff
@@ -145,6 +145,187 @@ For Kubernetes queries (kubectl, log filtering), use the pod label `gateway.netw
...
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: Gateway
+    name: kuadrant-egressgateway
+  metrics:
+    default:
+      labels:
+        workload: auth.identity.username
+        destination: request.host
+        source_ip: source.address.substring(0, source.address.lastIndexOf(":")).replace("[", "").replace("]", "")
```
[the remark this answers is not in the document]
maksymvavilov: Direct address will include the port

doc/user-guides/egress/observability.md
```diff
@@ -145,6 +145,187 @@ For Kubernetes queries (kubectl, log filtering), use the pod label `gateway.netw
...
+
+```text
+Accepted=True
+Enforced=True
+```
+
+Send some authenticated traffic through the egress gateway, then read the Limitador metrics:
+
+```sh
+kubectl exec -n egress-test test-client -- \
```
philbrookes: When I tried this locally, it timed out. A NetworkPolicy (kuadrant-limitador) restricts ingress to Limitador to pods in gateway-system only. The egress-test namespace is blocked.

Fix options:
  1. Run the curl from a pod in gateway-system instead (this is what I did)
  2. Use kubectl port-forward as the suggested approach
  3. Use kubectl exec into the Limitador pod itself to query localhost

philbrookes (changes requested):

doc/user-guides/egress/observability.md
maksymvavilov: I see what happened here: main landed a commit that re-enables network policies. 
I changed it to port forwarding and verified it. 
Didn't use a pod in the gateway system, as most gateway pods [don't have curl](https://github.com/Kuadrant/kuadrant-operator/blob/main/doc/observability/tracing.md#troubleshooting) and the same reasoning behind not using the limitador pod.

doc/user-guides/egress/observability.md
```diff
@@ -145,6 +145,217 @@ For Kubernetes queries (kubectl, log filtering), use the pod label `gateway.netw
...
+| `authorized_hits` | Hits counted against a limit (for TokenRateLimitPolicy, token consumption) |
+| `report_calls` | Calls to the Limitador `Report` method, which TokenRateLimitPolicy uses to report token usage after the response |
+
+Each series also carries Limitador's own `limitador_namespace` label, which identifies the targeted route (for example, `gateway-system/ai-mock-external`). For the full set of Limitador counters and their built-in labels, see the [Limitador metrics guide](../observability/limitador-metrics.md).
+
+**Authorino** receives the same labels on any route that carries an AuthPolicy, and applies them to its own per-AuthConfig metrics, which it serves on `/server-metrics` rather than `/metrics`:
+
+| Metric | Labels applied |
+|--------|----------------|
+| `auth_server_authconfig_response_status` | All labels |
```
philbrookes, line 171: Should this also say: All labels except those derived from `auth.*` like the others below?

philbrookes: This is the related error I was seeing at the reportStatusMetric call site (line 510), which is the function that reports response_status, when trying to validate this locally:
```
  {"msg":"failed to evaluate CEL expression",
   "expression":"auth.identity.username",
   "error":"no such key: username",
```

maksymvavilov: > Should this also say: All labels except those derived from `auth.*` like the others below?

Good catch, changed!

maksymvavilov: > This is the related error I was seeing at the reportStatusMetric call site (line 510), which is the function that reports response_status, when trying to validate this locally:
> 
> ```
>   {"msg":"failed to evaluate CEL expression",
>    "expression":"auth.identity.username",
>    "error":"no such key: username",
> ```

The same query served to Limitador and Authorino, and one of them will break. If we break Limitador, the whole thing is discarded, so the choice is to "break" Authorino. The Telemetry policy needs some love: #2242 #2243 #2279  #2298

philbrookes (approved):

PR #2246, opened 2026-09-09
test: restore dev DNS provider config after KuadrantControlPlane deletion

mikenairn:
## Summary

Regression caused by #2236: deleting the `KuadrantControlPlane` now cascade-deletes every non-CRD resource it owns, including the `dns-operator-controller-env` ConfigMap. That ConfigMap carries `make local-env-setup`'s `inmemory` DNS provider override (the dns-operator chart itself renders it with no `data` at all, so normal reconciles never touch it — only a full delete+recreate cycle wipes it).

The `controlplane` integration suite has a test that deletes the KCP to verify the new cascade-delete behavior. Once it does, the ConfigMap comes back blank on redeploy, silently breaking any other integration suite in the same run (e.g. `dnspolicy`) that relies on the `inmemory` provider — causing flaky test runs.

## Fix

Extend the suite's existing `AfterEach` (which already restores the CR when a test deletes it) to also re-apply `config/dev/dns-operator-configmap.yaml` via the same `dev-setup` server-side-apply field manager used by `local-env-setup`, reading directly from that file so it can't drift from the Makefile's source of truth. This runs exactly when a test caused the KCP (and therefore the cascade) to be deleted.

## Test plan

- [x] `gofmt`, `go vet -tags integration`, `go build -tags integration ./tests/...`
- [x] `make test-integration` — confirm the `controlplane` suite no longer leaves the `inmemory` DNS provider missing for later suites

## Summary by CodeRabbit

- **Tests**
  - Improved integration test reliability by waiting for asynchronously deleted DNS configuration to be fully removed before restoring the development configuration.
  - Reapplied the configuration before recreating control-plane resources, reducing test interference and making subsequent test runs more consistent and dependable.

### DIFF
--- a/tests/common/controlplane/controlplane_controller_test.go
+++ b/tests/common/controlplane/controlplane_controller_test.go
@@ -3,6 +3,8 @@
 package controlplane
 
 import (
+	"os"
+	"path/filepath"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


Patryk-Stefanski:
Question about the restoration race: could we wait for the old `dns-operator-controller-env` ConfigMap to be fully deleted (`NotFound`) before recreating the `KuadrantControlPlane`, and then confirm that the replacement ConfigMap is not terminating and contains the expected data?

At the moment, `restoreDevEnvOverrides` considers restoration successful when apply plus a read-back sees the correct data. If that read-back is against the old ConfigMap after garbage collection has started, it can still be deleted immediately afterward, leaving later integration tests without the `inmemory` DNS override.

mikenairn:
> Question about the restoration race: could we wait for the old `dns-operator-controller-env` ConfigMap to be fully deleted (`NotFound`) before recreating the `KuadrantControlPlane`, and then confirm that the replacement ConfigMap is not terminating and contains the expected data?

Yes, and this was what coderabbit was suggesting also https://github.com/Kuadrant/kuadrant-operator/pull/2246#pullrequestreview-5154468403. I was having issues getting it to work that way, but I've realised now it was just an ordering issue. Will update.

Patryk-Stefanski (approved):

PR #2247, opened 2026-09-09
Remove developer portal enable/disable toggle from Kuadrant CR

### LINKED ISSUE #2224: Remove the ability to enable and disable the dev portal controller from the kuadrant cr
Currently the dev portal controller has to be enabled in the kuadrant cr as the Dev portal piece was tech preview. As we will be GA in october the dev portal controller can be enbled by defualt and not required in the kuadrant cr anymore

Anton-Fil:
---

What

Removes the ability to enable/disable the developer portal controller via the Kuadrant CR. The developer portal controller is now always deployed by default.

The spec.components.developerPortal.enabled field is kept in the CRD but deprecated and ignored by the operator — it is retained purely to keep OLM CRD Upgrade Safety happy (see Backwards compatibility below).

Closes #2224

Why                                                                                    The developer portal was previously tech preview, so it shipped disabled by default
and had to be opted into via the Kuadrant CR. As the feature goes GA in October,
the opt-in toggle is no longer needed — the portal should just work out of the box.

Changes

API (hand-written)                                                                                                                                                   
- api/v1beta1/kuadrant_types.go: removed the IsDeveloperPortalEnabled() helper. The Components field on KuadrantSpec and the Components / DeveloperPortal structs are kept but marked Deprecated — the operator no longer reads them.

Controller (hand-written)

- internal/controller/developerportal_reconciler.go: the developer portal Deployment is now created unconditionally (buildDeployment(ns, false)) instead of being toggled by the flag. Finalizer-based cleanup on Kuadrant CR deletion is unchanged.

Generated

- Regenerated deepcopy, CRD, OLM bundle and Helm chart via make generate manifests bundle helm-build. The components.developerPortal.enabled field is still present in all schemas (now documented as deprecated/ignored).

Tests

- api/v1beta1/kuadrant_types_test.go: removed TestIsDeveloperPortalEnabled.
- internal/controller/developerportal_reconciler_test.go: dropped the enabled param and assert the Deployment is always created; added a regression test (Reconcile ignores the deprecated enabled=false field) proving the portal is deployed even when enabled: false is set.
- tests/istio/developerportal_controller_test.go: replaced the enable→disable flow with "enabled by default"; kept the CR-deletion cleanup coverage.

Backwards compatibility

The enabled field is intentionally not removed from the CRD. OLM runs a CRD Upgrade Safety Preflight Check that raises NoExistingFieldRemoved if a field disappears from an existing served version, which would block upgrades. Instead we follow the Kubernetes deprecation process: keep the field in the schema, mark it deprecated, and make the operator ignore it. It can be removed later via a proper CRD version bump + conversion.

Users who previously set enabled: true or enabled: false need no action — the portal is always enabled; the field is silently ignored.

Testing

- make verify-all — passing (generated manifests in sync)
- make test-unit — passing (includes the new enabled=false regression test)
- make test-istio-env-integration — 46/46 specs passing (includes the developer portal controller e2e test)

---

## Summary by CodeRabbit

* **New Features**
  * The Developer Portal is now consistently deployed and enabled as a standard part of Kuadrant.

* **Bug Fixes**
  * Setting the deprecated `developerPortal.enabled` option to `false` no longer prevents the Developer Portal from being deployed or recreated.

* **Documentation**
  * Updated configuration references to clarify that Developer Portal settings are deprecated, ignored, and retained only for backwards compatibility.


eguzki:
Can we ensure that OLM does not prevent the upgrade? OLM includes a CRD Upgrade Safety Preflight Check and I am afraid of OLM detecting that the field was removed from an existing version and raise `NoExistingFieldRemoved` error.

Anton-Fil:
> Can we ensure that OLM does not prevent the upgrade? OLM includes a CRD Upgrade Safety Preflight Check and I am afraid of OLM detecting that the field was removed from an existing version and raise `NoExistingFieldRemoved` error.


Good catch, and you're right to be worried — I check it and OLM's NoExistingFieldRemoved check does fire, since we're dropping spec.components from an existing v1beta1 CRD. So removing the field outright would block upgrades for anyone already running the operator.
I'll change the approach: instead of removing the field, I'll keep spec.components.developerPortal.enabled in the CRD but mark it as deprecated/no-op. The controller just ignores it now and always deploys the developer portal. That way the schema doesn't lose any fields (preflight stays happy), and we still get what the issue asks for — there's no longer any way to actually turn the portal off from the CR. We can drop the field for real later, whenever we do a proper CRD version bump. @eguzki  what did u think ?

tests/istio/developerportal_controller_test.go
```diff
@@ -85,37 +69,63 @@ var _ = Describe("Developer Portal Controller", Serial, func() {
...
 			Eventually(func(g Gomega) {
-				err := testClient().Get(ctx, client.ObjectKeyFromObject(kuadrantCR), kuadrantCR)
-				g.Expect(err).NotTo(HaveOccurred())
-				g.Expect(kuadrantCR.Spec.Components).NotTo(BeNil())
-				g.Expect(kuadrantCR.Spec.Components.DeveloperPortal).NotTo(BeNil())
-				kuadrantCR.Spec.Components.DeveloperPortal.Enabled = false
-				err = testClient().Update(ctx, kuadrantCR)
-				g.Expect(err).NotTo(HaveOccurred())
+				kuadrantCR := getKuadrantCR(ctx, testClient())
+				kuadrantCR.Spec = savedSpec
```
jasonmadigan: Could we capture the spec before the deployment wait and guard the restore? If that wait fails, `AfterEach` writes the zero-value `savedSpec` back to the shared CR. The deletion test below already uses a saved pointer and nil check for cleanup.

tests/istio/developerportal_controller_test.go
```diff
@@ -85,37 +69,63 @@ var _ = Describe("Developer Portal Controller", Serial, func() {
...
+				}
+				g.Expect(testClient().Update(ctx, kuadrantCR)).NotTo(HaveOccurred())
 			}).WithContext(ctx).Should(Succeed())
 
+			// The Deployment must remain: enabled=false is ignored.
+			Consistently(func(g Gomega) {
+				err := testClient().Get(ctx, client.ObjectKeyFromObject(deployment), deployment)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(deployment.Labels).To(HaveKeyWithValue(kuadrant.DeveloperPortalLabel, "true"))
+			}).WithTimeout(30 * time.Second).WithContext(ctx).Should(Succeed())
```
jasonmadigan: Could we leave room for this 30s check in the overall timeout? The initial wait can consume the full two-minute `SpecTimeout`, so anything over 90s leaves this unable to pass. Changing the individual timeout from five minutes to two doesn't change that—the spec context already limits it.

jasonmadigan: For the `enabled=false` regression coverage, could we follow the [deployment recreation test](https://github.com/Kuadrant/kuadrant-operator/blob/52df17d17f13562b685ec13acd61ec87aedf0dce/tests/common/controlplane/controlplane_controller_test.go#L319-L344)? Set the field, delete the Deployment, then `Eventually` check for a replacement with a different UID. That would exercise reconciliation directly without the fixed 30s observation window.

jasonmadigan:
generally fine, just a few things i noticed

jasonmadigan (approved):

PR #2237, opened 2026-09-08
test: add MCP Gateway nightly signal

Patryk-Stefanski:
## Nightly run

1. Trigger at 02:45 UTC or manually with an optional `mcp_ref`.
2. Check out `kuadrant-operator@main` and `mcp-gateway` at the requested ref, then record both resolved commit SHAs.
3. Create an Istio-on-Kind cluster with the MCP Gateway E2E port mappings and initialize the Kuadrant environment.
4. Build and load run-specific operator, MCP controller, and MCP broker images into Kind.
5. Render and install the Kuadrant Helm chart, using the committed MCP component chart and the run-specific MCP image references.
6. Wait for the Kuadrant control plane, operator, MCP CRDs, and managed MCP controller to become ready.
7. Deploy the MCP Gateway fixtures, E2E gateways, TLS resources, and cross-namespace `ReferenceGrant`, then create the managed `MCPGatewayExtension`.
8. Verify the deployed operator, MCP controller, and MCP broker image identities and `IfNotPresent` pull policies.
9. Deploy the MCP test servers and run the existing full MCP Gateway E2E suite.
10. Publish source, image, chart, and provider metadata to the workflow summary. On failure, collect cluster and MCP diagnostics and upload the test output and logs as an artifact.

## Summary by CodeRabbit

- **Tests**
  - Added scheduled nightly end-to-end testing for MCP Gateway integrations.
  - Added optional manual test runs against a specified MCP Gateway revision.
  - Added validation of deployed component versions and image settings.
  - Added automatic collection and upload of diagnostic information when tests fail.
  - Added a concise nightly test summary for easier monitoring.

PR #2262, opened 2026-09-14
test: move MCP Gateway AuthPolicy E2E to operator

Patryk-Stefanski:
## Summary

Move the MCP Gateway AuthPolicy integration coverage from `mcp-gateway` into the operator's Istio integration suite. These tests exercise AuthPolicy, the Kuadrant operator, and MCP Gateway together rather than the MCP Gateway component in isolation.

This extends the nightly workflow introduced by #2237, which is merged. The duplicate component-side AuthPolicy test and dedicated workflow were removed in [Kuadrant/mcp-gateway#1343](https://github.com/Kuadrant/mcp-gateway/pull/1343), also merged.

## Changes

- Extend the nightly setup with Redis, the Keycloak ingress port, authentication fixtures, and AuthPolicy/Authorino failure diagnostics. After the generic MCP Gateway E2E suite cleans up, recreate the extension, ReferenceGrant, and gateway CA trust before running the operator AuthPolicy suite.
- Add `make test-mcp-authpolicy-e2e` with the required operator environment and image exports. Enable the `integration,mcp_auth_e2e` build tags and run both the focused helper regression test and the AuthPolicy Ginkgo specs. Keep the MCP suite excluded from the normal Istio PR integration target.
- Create the backend HTTPRoutes and MCPServerRegistrations, switch the extension to the trusted-header public key, and wait for rollout/readiness. Cover unauthenticated and malformed JWT requests, role-based tool/prompt filtering, permitted and denied tool calls, first-request `prompts/get`, and virtual-server filtering.
- Require the routed tool policy's HTTP **403** response with structured `"error": "Forbidden"` for the forbidden tool call. Transport failures, missing routes, gateway errors, and the broker's separate HTTP 401 method-denial response must not satisfy this assertion.
- Reject MCP tool results with `isError: true`, including HTTP 200 JSON and SSE responses, and verify the successful greeting rather than merely nonempty content. Avoid pinning the unauthenticated response's human-readable wording.
- Add cluster-independent HTTP regressions in `tests/istio/mcp_gateway_auth_helpers_test.go` for successful results with an omitted `isError` flag and tool-execution errors in JSON and SSE responses.


Patryk-Stefanski:
Addressed the actionable CodeRabbit comments in commit 224668ca: Keycloak setup now waits on deployment rollout status, and unauthorized tool-call coverage requires HTTP 401 with a Forbidden response body. The docstring-coverage item is an advisory warning, not a failing repository check.

Patryk-Stefanski:
PR #2262 is now stacked on #2237. The duplicate workflow addition is removed from the effective PR diff; #2262 only extends #2237's nightly workflow with AuthPolicy setup and adds the operator-side test. Merge #2237 first.

trepel:
@Patryk-Stefanski  Is there any reason for not introducing a PR check running these tests? If I got it right these will only be executed nightly whereas all the other tests are executed as a PR check as well.

Patryk-Stefanski:
@trepel In my opinion, running these tests on every PR would be too expensive: they build and load MCP Gateway images and provision Kind, Istio, Keycloak, AuthPolicy, TLS, and MCP fixtures. I think the nightly run is a sufficient signal for now. If others agree these should run on every PR, I have no problem switching to that approach.

mikenairn:
@Patryk-Stefanski The way this is currently setup they would be included in the PR tests via `make test-istio-env-integration` https://github.com/Kuadrant/kuadrant-operator/blob/main/.github/workflows/test.yaml#L290

If you want these excluded from PR testing, which i think for now at least would be a good idea, we will need to filter those out. I'd use labels (`mcp-gateway`, `long-running` etc..) and add a label filter  `--label-filter="!mcp-gateway"`

tests/istio/mcp_gateway_auth_test.go
```diff
@@ -0,0 +1,670 @@
...
+)
+
+var (
+	mcpAuthGatewayURL       = mcpAuthEnv("MCP_AUTH_GATEWAY_URL", mcpAuthGatewayURLDefault)
+	mcpAuthKeycloakTokenURL = mcpAuthEnv("MCP_AUTH_KEYCLOAK_TOKEN_URL", mcpAuthKeycloakTokenURLDefault)
+	mcpAuthNamespace        = mcpAuthEnv("MCP_AUTH_NAMESPACE", mcpAuthNamespaceDefault)
+	mcpAuthTestServerNS     = mcpAuthEnv("MCP_AUTH_TEST_SERVER_NAMESPACE", mcpAuthTestServerNamespaceDefault)
+)
+
+var _ = Describe("MCP Gateway AuthPolicy integration", Ordered, func() {
```
mikenairn:
```suggestion
var _ = Describe("MCP Gateway AuthPolicy integration", Ordered, Labels{"mcp-gateway"}, func() {
```

Patryk-Stefanski:
implemented your suggestion in commit `185176d0`. The AuthPolicy suite is labeled `mcp-gateway` and `long-running`, and `test-istio-env-integration` now passes `--label-filter='!mcp-gateway'` so the expensive suite stays out of PR integration runs. The nightly workflow invokes it explicitly.

trepel:
> @trepel In my opinion, running these tests on every PR would be too expensive: they build and load MCP Gateway images and provision Kind, Istio, Keycloak, AuthPolicy, TLS, and MCP fixtures. I think the nightly run is a sufficient signal for now. If others agree these should run on every PR, I have no problem switching to that approach.

All that makes assuming that there is someone watching the nightly results daily. My impression was that all the tests are executed per PR and nightly pipelines are just a backup nobody is watching actively. But I can start watching the mcp gateway related tests myself.

Patryk-Stefanski:
Yes, that's a presumptuous assumption. I do like the reporting we have by email of failed nightlies that you setup. I'm wondering if we could set up something similar through GitHub.

tests/istio/mcp_gateway_auth_test.go
```diff
@@ -0,0 +1,670 @@
...
+
+	It("rejects an unauthorized tool call", func(ctx SpecContext) {
+		headers := mcpAuthAuthorizationHeaders(ctx)
+		sessionID := mcpAuthInitializeSession(ctx, headers)
+		Expect(mcpAuthNotifyInitialized(ctx, mcpAuthGatewayURL, sessionID, headers)).To(Succeed())
+
+		status, _, err := mcpAuthCallTool(ctx, mcpAuthGatewayURL, sessionID, "test1_time", nil, headers)
+		Expect(err).To(HaveOccurred())
+		Expect(status).To(Equal(http.StatusUnauthorized))
+		Expect(err).To(MatchError(ContainSubstring("Forbidden")))
```
trepel: It used to be `Expect(status).NotTo(Equal(http.StatusOK)` (https://github.com/Kuadrant/mcp-gateway/commit/eff5a942cf2015b486b24181fde40f091ae00e3c#diff-443d22089cf044a9aa3e717183c6bba3c03b39d4e71064e9258648bb9cbc19dd) . I think that this fails because 401 will be returned but Forbidden matches 403

.github/workflows/mcp-gateway-nightly.yaml
```diff
@@ -447,6 +451,135 @@ jobs:
...
+        env:
+          MCP_AUTH_E2E: "true"
+          MCP_AUTH_GATEWAY_URL: https://mcp.mcp-gateway.local:8009/mcp
+          MCP_AUTH_KEYCLOAK_TOKEN_URL: https://keycloak.127-0-0-1.sslip.io:8002/realms/mcp/protocol/openid-connect/token
+          MCP_AUTH_NAMESPACE: kuadrant-system
+          MCP_AUTH_TEST_SERVER_NAMESPACE: mcp-test
+          GATEWAYAPI_PROVIDER: istio
+        run: |
+          mkdir -p "$RUNNER_TEMP/mcp-gateway-nightly"
+          go test -tags=integration,mcp_auth_e2e ./tests/istio \
```
trepel: Won't invoking the tests this directly make it missing some env vars that are populated automatically if running via make command, eg. CHARTS_PATH: https://github.com/Kuadrant/kuadrant-operator/blob/main/Makefile#L467

tests/istio/mcp_gateway_auth_test.go
```diff
@@ -0,0 +1,669 @@
...
+			patch := client.MergeFrom(extension.DeepCopy())
+			Expect(unstructured.SetNestedField(extension.Object, map[string]any{
+				"secretName": "trusted-headers-public-key",
+				"generate":   "Disabled",
+			}, "spec", "trustedHeadersKey")).To(Succeed())
+			Expect(testClient().Patch(ctx, extension, patch)).To(Succeed())
+		}
+		Expect(mcpAuthWaitDeploymentRollout(ctx, testClient(), deploymentKey, previousDeploymentGeneration, needsDeploymentGeneration)).To(Succeed())
+
+		By("creating MCPServerRegistrations for the authenticated test servers")
```
trepel, line 97: Do HTTPRoutes required for MCPServerRegistrations exist on cluster? I can't find where they are created after being removed by MCP Gateway e2e test suite: https://github.com/Kuadrant/mcp-gateway/blob/main/tests/e2e/suite_test.go#L144 - I think that each test cleans up the HTTPRoute at the end if it creates one.

.github/workflows/mcp-gateway-nightly.yaml
Patryk-Stefanski: Yes. `go test` bypasses `LOCAL_RUN_ENV`, but the shared Istio suite still starts an operator. It needs the same exports, including `CHARTS_PATH` and the nightly MCP image overrides.

tests/istio/mcp_gateway_auth_test.go
Patryk-Stefanski: You're right. The AuthPolicy setup doesn't restore those routes after the MCP suite's cleanup. It needs to recreate `mcp-server1-route`, `mcp-server2-route` and `everything-server-route` before creating the registrations.

.github/workflows/mcp-gateway-nightly.yaml
```diff
@@ -467,6 +497,141 @@ jobs:
...
+
+      - name: Run operator MCP AuthPolicy E2E
+        env:
+          MCP_AUTH_GATEWAY_URL: https://mcp.mcp-gateway.local:8009/mcp
+          MCP_AUTH_KEYCLOAK_TOKEN_URL: https://keycloak.127-0-0-1.sslip.io:8002/realms/mcp/protocol/openid-connect/token
+          MCP_AUTH_NAMESPACE: kuadrant-system
+          MCP_AUTH_TEST_SERVER_NAMESPACE: mcp-test
+        run: |
+          mkdir -p "$RUNNER_TEMP/mcp-gateway-nightly"
+          make test-mcp-authpolicy-e2e \
```
trepel, line 631: @Patryk-Stefanski I think that running this make target triggers SetupKuadrantOperatorForTest (https://github.com/Kuadrant/kuadrant-operator/blob/main/tests/istio/suite_test.go#L78) which is not desired. At this point there is a KIND cluster with Kuadrant Operator up & running already.

mikenairn: fyi, i added an option to start operator controllers optionally in this PR https://github.com/Kuadrant/kuadrant-operator/pull/2291/commits/c54c841b1b822279c6e7fe4e0c9c33b91196fbce

Patryk-Stefanski: Good catch. This job already has the Kuadrant Operator running in the KIND cluster, so the target must not start SetupKuadrantOperatorForTest again. I rebased this branch onto #2291 (ef9312c8), which adds the USE_EXISTING_OPERATOR=true path; I will pass that flag for this target.

tests/istio/mcp_gateway_auth_test.go
```diff
@@ -0,0 +1,712 @@
...
+				"secretName": "trusted-headers-public-key",
+				"generate":   "Disabled",
+			}, "spec", "trustedHeadersKey")).To(Succeed())
+			Expect(testClient().Patch(ctx, extension, patch)).To(Succeed())
+		}
+		Expect(mcpAuthWaitDeploymentRollout(ctx, testClient(), deploymentKey, previousDeploymentGeneration, needsDeploymentGeneration)).To(Succeed())
+
+		By("creating backend HTTPRoutes on the authenticated gateway listener")
+		for _, route := range []*gatewayapiv1.HTTPRoute{
+			mcpAuthHTTPRoute("mcp-server1-route", "server1.mcp-gateway.local", "mcp-test-server1"),
```
trepel, line 87: In order to make this runnable on OpenShift the HTTPRoutes' hostnames must be configurable (overridable via env var maybe?), stuff like `server1.mcp-gateway.local` won't work.

In my head there is that if using HTTPS Listener then the same listener needs to be used for both 2025-11-25 version hairpin requests and "normal" requests. That's why such local hostnames won't work - "normal" requests require normal hostname, not this local one on OCP. However, AI told me that I can omit hostname on the Listener so if proper certificate (for both mcp.apps.<ocp-domain> and *.mcp-gateway.local) signed by proper CA is used for that Listener both hairpin and "normal" requests will be able to use it just fine.

So maybe this can stay as is and I will just have to update the pipeline accordingly. But I will have to validate that AI's suggestion first. @maleck13 does that AI suggestion make sense to you?

trepel: It seems that the AI suggestion might work but it would require some changes to MCP Gateway itself, listeners without a hostname seems to be ignored by current implementation, see https://github.com/Kuadrant/mcp-gateway/blob/main/internal/controller/mcpgatewayextension.go#L187

It will probably be simpler to make the hostnames configurable in this PR

trepel:
@Patryk-Stefanski I think that this is good to go now. Thanks! Two non-blocking things
- the failed test you mentioned in PR description causes subsequent tests to be skipped (I just use `ginkgo.skip` to workaround that)
- `waiting for AuthPolicy to reject unauthenticated initialization` - there's no timeout there so it waits indefinitely in case of something went wrong

trepel (approved):

PR #2225, opened 2026-09-03
feat(tokenratelimit): add Kuadrant mode and policy reservation fields with Reserve/Commit reconciler branch

aniketpandey05:
# Description
Adds the cluster-wide mode switch and per-limit reservation CRD fields, plus the reconciler logic that generates Reserve/Commit wasm actions (Reservation mode) or Check/Report actions (CheckReport mode), implementing RFC 0021 for `kuadrant-operator`.

Closes #2148
Parent: #192 (Epic: Token Rate Limit Reservations)
Design: [RFC 0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md)
Depends on: [Kuadrant/wasm-shim/pull/422](https://github.com/Kuadrant/wasm-shim/issues/402) https://github.com/Kuadrant/limitador/pull/515

## Summary of Changes
- **API & CRD**:
  - `KuadrantSpec`: Added `spec.tokenRateLimiting.mode` enum (`Reservation` | `CheckReport`, default `Reservation`) and nil-safe `GetTokenRateLimitingMode()`.
  - `TokenRateLimitPolicy`: Added `spec.limits[<name>].reservation` (`amount`, `ttl` CEL expressions).
  - Regenerated deepcopy methods, CRD manifests, OLM bundle, and Helm templates.
- **WASM Actions**:
  - Registered `ratelimit-reserve-service` and `ratelimit-commit-service` in `NewServiceBuilder`.
  - Configured `IsGuard()` to exclude `ratelimit-commit-service` (response phase).
  - Dispatched reserve (request phase) and commit (response phase) gRPC actions.
- **Reconciler Branching**:
  - `wasmActionSpecsFromTokenLimit`:
    - **Reservation mode** (default): Emits `ratelimit-reserve-service` with `ratelimit.amount` (default: `uint(5000)`) and `ratelimit.ttl` (route `timeouts.backendRequest` or fallback `duration('60s')`), and `ratelimit-commit-service` with `ratelimit.hits_addend: responseBodyJSON("/usage/total_tokens")`.
    - **CheckReport mode**: Emits `ratelimit-check-service` (`hits_addend: 0`) and `ratelimit-report-service` (`hits_addend: responseBodyJSON(...)`).
    - Preserves per-limit bypass: `amount: "0"` calls `Reserve` without holding capacity.
[...]

PR #2263, opened 2026-09-14
Tokenratelimit reservation

### LINKED ISSUE #2148: Add Kuadrant/TokenRateLimitPolicy reservation fields and Reserve/Commit reconciler branch
## Summary
Add the cluster-wide `mode` switch and per-limit `reservation` CRD fields, plus the reconciler logic that generates `Reserve`/`Commit` wasm actions (Reservation mode) or today's `Check`/`Report` actions (CheckReport mode).

Design: [RFC 0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md), reference-level section "Expected work by component" (kuadrant-operator).

Depends on: Kuadrant/wasm-shim#402 (`RateLimitReserve`/`RateLimitCommit` service types).

## Acceptance Criteria
- [ ] `Kuadrant` CR field `spec.tokenRateLimiting.mode` (enum `Reservation`/`CheckReport`, default `Reservation`)
- [ ] `TokenRateLimitPolicy` per-limit `reservation` field (`amount`, `ttl` CEL expressions), including defaults/overrides merge behavior across the policy hierarchy
- [ ] Default `amount`/`ttl` CEL generation for limits that omit `reservation`: flat `uint(5000)` for amount, route's `HTTPRoute.spec.rules[].timeouts.backendRequest` for ttl
- [ ] Reconciler branch generating the correct wasm actions based on effective state:
  - [ ] Effectively active: `Reserve` action (`ServiceType::RateLimitReserve`) at request phase + `Commit` action (`ServiceType::RateLimitCommit`) at response/stream-end phase
  - [ ] The `reservation_id` returned by the `Reserve` call is captured (via an `on_reply` → `Store` action into `stored_values`) and reused in the `Commit` action's request. Per wasm-shim#402 this capture/reuse is entirely configuration-driven, so the operator must generate the wasm config that stores the id from the Reserve reply and passes it back on `Commit`.
  - [ ] Effectively inactive: today's `ratelimit-check-service` (`hits_addend=0`) + `ratelimit-report-service` (`hits_addend=<actual>`)
- [ ] Regular `RateLimitPolicy` reconciliation path is untouched (separate code, verified by existing tests still passing)
- [ ] Docs covering both modes
- [ ] E2e tests covering both modes and the upgrade-window transition (old Limitador/operator without reservation support)

## Notes for reviewers
PR description should include: how to review (suggested order: CRD field additions → default CEL generation → reconciler branching → e2e), and manual verification (apply a `TokenRateLimitPolicy` in each mode, inspect generated wasm plugin config for the expected actions).

eguzki:
# What

Implements [RFC 0021 (Token Rate Limit Reservations)](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md) for TokenRateLimitPolicy: 
* a cluster-wide mode switch on the Kuadrant CR
* per-limit reservation tuning
*  the Reserve/Commit wasm action generation that closes the check/report race window for token-based rate limiting.

Closes #2148
Design: [RFC 0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md)

Summary of Changes

- API & CRD:
  - Kuadrant.spec.tokenRateLimiting.mode (Reservation | Optimistic, default Reservation).
  - TokenRateLimitPolicy.spec.limits[<name>].reservation (amount, ttl CEL expressions).
- WASM action generation: Reservation mode emits `ratelimit-reserve-service/ratelimit-commit-service` in place of `ratelimit-check-service/ratelimit-report-service`. Default amount is 0 (no capacity reserved, so Reservation mode is behavior-neutral for any policy that doesn't set reservation.amount explicitly, matching Optimistic until it does); default ttl falls back to the route's `HTTPRoute.spec.rules[].timeouts.backendRequest`. Commit follows RFC 0021's degrade-to-Report semantics: it fires whenever a reservation was held or usage is parseable, and is skipped only when neither is true.
- CEL validation (internal/cel/kuadrant_validator.go): Reserve/Commit actions are attributed to TokenRateLimitPolicyKind, and reservation.amount/reservation.ttl are type-checked at reconcile time.
- Tests: unit coverage across internal/wasm, internal/cel, internal/controller, api/v1alpha1; integration coverage in tests/istio and tests/envoygateway.
- Docs: doc/reference/tokenratelimitpolicy.md and doc/overviews/token-rate-limiting.md updated with the reservation field and failure-handling behavior.

#  Verification Steps

## 1. Set up the local environment

```sh
make local-setup
```

## 2. Apply the Kuadrant CR (defaults to Reservation mode)

```sh
kubectl apply -n kuadrant-system -f - <<EOF
kind: Kuadrant
metadata:
  name: kuadrant
spec: {}
EOF
```

## 3. Deploy the LLM simulator behind a Gateway/HTTPRoute

```sh
kubectl create namespace llm

kubectl apply -n llm -f - <<EOF
apiVersion: apps/v1
kind: Deployment
metadata:
  name: llm-sim
  labels:
    app: llm-sim
spec:
  replicas: 1
  selector:
    matchLabels:
      app: llm-sim
  template:
    metadata:
      labels:
        app: llm-sim
    spec:
      containers:
      - name: simulator
        image: ghcr.io/llm-d/llm-d-inference-sim:v0.3.0
        args:
          - --model
          - Qwen/Qwen2.5-1.5B-Instruct
          - --port
          - "8000"
        ports:
          - containerPort: 8000
---
apiVersion: v1
kind: Service
metadata:
  name: llm-sim
spec:
  selector:
    app: llm-sim
  ports:
    - port: 80
      targetPort: 8000
      protocol: TCP
---
kind: HTTPRoute
metadata:
  name: llm-sim
spec:
  hostnames:
    - llm.example.com
  parentRefs:
      namespace: gateway-system
      sectionName: http
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: "/"
      backendRefs:
        - name: llm-sim
          port: 80
EOF
```

## 4. Apply a TokenRateLimitPolicy with an explicit reservation.amount/ttl

An explicit `reservation` block, overriding the `0`-token/route-timeout defaults.

```sh
kubectl apply -n llm -f - <<EOF
kind: TokenRateLimitPolicy
metadata:
  name: token-limits
spec:
  targetRef:
    kind: HTTPRoute
    name: llm-sim
  when:
  - predicate: request.url_path.startsWith('/v1/chat/completions')
  limits:
    global:
      rates:
        - limit: 12000
          window: 1m
      reservation:
        amount: "2000"
        ttl: 'duration("30s")'
EOF
```

Check policy status:

```sh
kubectl get tokenratelimitpolicy token-limits -n llm -o jsonpath='{.status.conditions[?(@.type=="Accepted")].message}{"\n"}{.status.conditions[?(@.type=="Enforced")].message}{"\n"}'
```

## 5. Inspect the generated EnvoyFilter's wasm config

Confirm it's using `Reserve`/`Commit`, not `Check`/`Report`.

```sh
kubectl get envoyfilter -n gateway-system -o json | \
  jq -r '.items[].spec.configPatches[] | select(.applyTo=="HTTP_FILTER") | .patch.value.typed_config.value.config.configuration.value' | \
  jq -r '.actionSets[].actions[] | select(.type=="grpc") | .service' | sort -u
```

Expected: `ratelimit-reserve-service` and `ratelimit-commit-service` present, `ratelimit-check-service`/`ratelimit-report-service` absent.

Inspect the Reserve action's message builder, to confirm the explicit `reservation.amount`/`reservation.ttl` override took effect:

```sh
kubectl get envoyfilter -n gateway-system -o json | \
  jq -r '.items[].spec.configPatches[] | select(.applyTo=="HTTP_FILTER") | .patch.value.typed_config.value.config.configuration.value' | \
  jq -r '.actionSets[].actions[] | select(.service=="ratelimit-reserve-service") | .messageBuilder'
```

Expected:

```
kuadrant.service.ratelimit.v1.ReserveRequest {
    domain: "llm/llm-sim",
    descriptors: [envoy.extensions.common.ratelimit.v3.RateLimitDescriptor { entries: ((request.url_path.startsWith('/v1/chat/completions')) ? [envoy.extensions.common.ratelimit.v3.RateLimitDescriptor.Entry { key: "tokenlimit.global__65da223d", value: string(1) }] : []) }],
    amount: uint(2000),
    ttl: duration("30s")
}
```

Inspect the Commit action's message builder, to confirm the truth-table behavior from this PR (reservation released with `actual_amount: 0` when usage can't be parsed, empty `reservation_id` when no reservation was held):

```sh
kubectl get envoyfilter -n gateway-system -o json | \
  jq -r '.items[].spec.configPatches[] | select(.applyTo=="HTTP_FILTER") | .patch.value.typed_config.value.config.configuration.value' | \
  jq -r '.actionSets[].actions[] | select(.service=="ratelimit-commit-service") | .messageBuilder'
```

Expected:

```
kuadrant.service.ratelimit.v1.CommitRequest {
    domain: "llm/llm-sim",
    descriptors: [envoy.extensions.common.ratelimit.v3.RateLimitDescriptor { entries: ((request.url_path.startsWith('/v1/chat/completions')) ? [envoy.extensions.common.ratelimit.v3.RateLimitDescriptor.Entry { key: "tokenlimit.global__65da223d", value: string(1) }] : []) }],
    reservation_id: (has(kuadrant.internal.tokenratelimit.reservation.tokenlimit_global__65da223d)) ? kuadrant.internal.tokenratelimit.reservation.tokenlimit_global__65da223d : "",
    actual_amount: ((kuadrant.internal.response.body.total_tokens) != null) ? uint(kuadrant.internal.response.body.total_tokens) : 0u
}
```

(the exact store-path suffix, e.g. `tokenlimit_global__65da223d`, is a hash derived from the policy name and limit name, so it will differ per environment.)

## 6. Send a request and confirm Limitador received one Reserve and one Commit call

Bump Limitador's log verbosity to `DEBUG` (`verbosity: 3`), which is the level its `reserve`/`commit` gRPC handlers log at (`limitador-server/src/envoy_rls/kuadrant_service.rs:214,287`):

```sh
kubectl patch limitador limitador -n kuadrant-system --type=merge --patch '{"spec": {"verbosity": 3}}'
kubectl rollout status deployment/limitador-limitador -n kuadrant-system
```

Send a request:

```sh

curl --resolve llm.example.com:$INGRESS_PORT:$INGRESS_HOST \
     -H 'Content-Type: application/json' \
     -X POST http://llm.example.com:$INGRESS_PORT/v1/chat/completions \
     -d '{"model":"Qwen/Qwen2.5-1.5B-Instruct","messages":[{"role":"user","content":"What is Kubernetes?"}],"max_tokens":100,"stream":false}'
```

Then grep the logs for exactly one `Reserve` and one `Commit`:

```sh
kubectl logs deployment/limitador-limitador -n kuadrant-system --since=30s | grep -E "Reserve request received|Commit request received"
```

Expected (one of each; `x-request-id`, `reservation_id` and `actual_amount` will differ per run):

```
DEBUG grpc{x-request-id="3c380d29-d14b-4fd9-84f1-ff5569c20b24"}:reserve: limitador_server::envoy_rls::kuadrant_service: Reserve request received: Request { ..., message: ReserveRequest { domain: "llm/llm-sim", descriptors: [...], amount: 2000, ttl: Some(Duration { seconds: 30, nanos: 0 }) }, ... }
DEBUG grpc{x-request-id="3c380d29-d14b-4fd9-84f1-ff5569c20b24"}:commit: limitador_server::envoy_rls::kuadrant_service: Commit request received: Request { ..., message: CommitRequest { domain: "llm/llm-sim", descriptors: [...], reservation_id: Some("7c093d85-b17a-4a18-846a-94137ce9c8da"), actual_amount: 28 }, ... }
```

The key things to check: `amount: 2000`/`ttl: Some(Duration { seconds: 30, ... })` on the Reserve line match the policy's explicit `reservation.amount`/`reservation.ttl`, and the Commit line's `reservation_id` is `Some("<a real uuid>")` (not empty) with `actual_amount` matching the real token count from the response.

## 7. Confirm real token usage is recorded

```sh
kubectl port-forward -n kuadrant-system service/limitador-limitador 18080:8080
```

```sh
curl --silent http://127.0.0.1:18080/counters/llm%2Fllm-sim | yq e -P
curl --silent http://127.0.0.1:18080/limits/llm%2Fllm-sim | yq e -P
```

Expected: the counter reflects the response's real `usage.total_tokens`, not `1`.

## 8. Switch to Optimistic mode and re-check the EnvoyFilter

```sh
kubectl patch kuadrant kuadrant -n kuadrant-system --type merge -p '{"spec":{"tokenRateLimiting":{"mode":"Optimistic"}}}'
```

Re-run the `jq` command from step 5.

Expected: `ratelimit-check-service`/`ratelimit-report-service` now present, `ratelimit-reserve-service`/`ratelimit-commit-service` gone.

## Cleanup

```sh
make local-cleanup
```

## Summary by CodeRabbit

- **New Features**
  - Added token reservation enforcement for token rate limits, enabled by default.
  - Added configurable reservation amounts and time-to-live values, with route-based defaults.
  - Added a cluster-wide option to choose between Reservation and Optimistic enforcement modes.
  - Reservations are reconciled with actual upstream token usage, releasing unused capacity.
  - Added validation for reservation amount and TTL expressions.

- **Documentation**
  - Updated configuration references and examples for enforcement modes and reservations.


eguzki:
@jland-redhat, tagging you on this since you've might have a better idea and help us. 

Right now `reservation.amount/.ttl` are per-TokenLimit. If a policy defines N limits that each set reservation, we generate N independent Reserve/Commit action pairs, i.e. one gRPC round-trip per limit, per request. It's not a bug, but inherent: Limitador's `ReserveRequest.amount/.ttl` are single scalars per call, so the proto can't carry two different reservation amounts in one request.

Current shape:
```yaml
limits:
  global:
    rates: [{limit: 272, window: 1m}]
    reservation: {amount: "109", ttl: 'duration("30s")'}
  premium:
    rates: [{limit: 5000, window: 1m}]
    reservation: {amount: "3000", ttl: 'duration("45s")'}  # different -> can't merge
```

Alternative shape: reservation moved to policy level, shared by all limits:
```yaml
reservation: {amount: "3000", ttl: 'duration("45s")'}
limits:
  global:
    rates: [{limit: 272, window: 1m}]
  premium:
    rates: [{limit: 5000, window: 1m}]
```

With one shared top level  `amount/.ttl,` all limits' `Reserve` calls could merge into one, the same way TokenRateLimitPolicy's own `CheckReport` mode already merges multiple limits into a single `Check` call and a single `Report` call (the same generic merge logic also collapses RateLimitPolicy's multiple limits into one `ShouldRateLimit` call). 

Worth noting: RFC 0021 doesn't seem to weigh this trade-off anywhere. `reservation` is introduced as "a new, optional per-limit block" without discussing a policy-level alternative.

Question: is there a real use case for different reservation sizes across limits within the same policy? If not, policy-level reservation gets us 1 round-trip instead of N with no real loss.

internal/wasm/action_spec.go
```diff
@@ -305,8 +312,74 @@ func (s ActionSpec) buildRateLimit(responseVar string, isGuard bool, label strin
...
+//     which Limitador treats like a plain Report (RFC 0021).
+func (s ActionSpec) buildCommit() *GrpcAction {
+	id := findReservationAttrCEL(s.ConditionalData, reservationIDAttr)
+	storePath := reservationStorePath(id)
+	hasReservation := fmt.Sprintf("has(%s)", storePath)
+
+	rawActualAmount := lookupReservationAttr(s.ConditionalData, reservationActualAmountAttr)
+	amountKnown := "false"
+	if rawActualAmount != "" {
+		amountKnown = fmt.Sprintf("(%s) != null", rawActualAmount)
```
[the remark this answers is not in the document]
eguzki: Confirmed real, traced it through wasm-shim's vendored cel-rust: `uint()`'s only overloads are `uint`/`int`/`double`/`string` (`common/types/uint.rs`), and the `string` overload does `s.parse::<u64>()`, returning `ExecutionError::FunctionError` on failure — so `uint("unknown")` does throw, and there's no `type()` builtin or safe-conversion helper registered anywhere in the stdlib or in wasm-shim's own function registrations to guard it from the CEL side. A purely CEL-level fix (a type-check predicate before `uint()`) isn't buildable with what's currently registered.

The right fix isn't a CEL guard here, though — it's already in flight as [RFC 0024](https://github.com/Kuadrant/architecture/blob/main/rfcs/0024-token-rate-limit-api-format-compatibility.md), which moves "is this value actually numeric" validation out of CEL and into wasm-shim's own extraction layer. Per its "What counts as 'resolves'" section, a candidate JSON pointer only resolves if it dereferences to a JSON number or a numeric string; anything else (including a non-numeric string like `"unknown"`) is treated identically to "no match," falling through to absent/`null` — the same semantics `!= null` here already assumes. Since `reservation.actual_amount` and `ratelimit.hits_addend` are both sourced from the same underlying extraction value, RFC 0024's wasm-shim change closes this gap for both, without any change needed in this PR's CEL generation.

Leaving this as a known, tracked limitation rather than patching around it here — RFC 0024 is the real fix.

doc/overviews/token-rate-limiting.md
```diff
@@ -24,6 +24,34 @@ This approach ensures accurate usage-based rate limiting where limits are enforc
...
 
+### Enforcement modes
+
+Because the real token cost of a request is only known *after* the upstream responds, the gateway has to decide what to do on the way in. This is controlled cluster-wide by the Kuadrant CR field `spec.tokenRateLimiting.mode`, which applies to every TokenRateLimitPolicy in the cluster:
+
+- **`Reservation`** (default): On request arrival the gateway *reserves* an estimated token amount from Limitador. If the reservation would exceed the limit, the request is rejected with `429` before it ever reaches the upstream. Once the upstream responds, the gateway *commits* the actual `usage.total_tokens` and releases the unused portion of the reservation. This closes the race window where many concurrent requests could each pass a zero-cost check before any of them reports usage. It costs one extra call to Limitador per request. The reservation is designed per RFC [0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md), and the reserved amount / hold time are tuned per limit with [`reservation`](../reference/tokenratelimitpolicy.md#reservation).
```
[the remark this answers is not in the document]
eguzki: fixed

internal/wasm/utils.go
```diff
@@ -210,6 +247,22 @@ func NewServiceBuilder(logger *logr.Logger) *ServiceBuilder {
...
+				GrpcMethod:  ptr.To(RateLimitReserveGrpcMethod),
+			},
+			RateLimitCommitServiceName: {
+				Type:        DynamicServiceType,
+				Endpoint:    kuadrant.KuadrantRateLimitClusterName,
+				FailureMode: RatelimitCommitServiceFailureMode(logger),
+				Timeout:     ptr.To(RatelimitCommitServiceTimeout()),
+				GrpcService: ptr.To(KuadrantRateLimitGrpcService),
+				GrpcMethod:  ptr.To(RateLimitCommitGrpcMethod),
+			},
```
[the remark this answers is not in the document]
eguzki: Fixed

aniketpandey05:
Thanks @eguzki  for putting this together! The implementation is very thorough, clean, and adheres closely to RFC 0021.

### Suggestions

#### 1. "Upgrades and compatibility" documentation

As discussed in #2225, adding an "Upgrades and compatibility" callout in `doc/overviews/token-rate-limiting.md` would be really helpful for cluster operators.

Because `Reservation` is the default mode, if the operator is upgraded before Limitador supports `Reserve`/`Commit` (or if Limitador is running with `--disable-reservations`), the `Reserve` calls will fail and, with the default failure mode (`allow`), requests will be let through without token counting.


#### 2. End-to-end integration test for CheckReport mode

`tokenratelimit_workflow_test.go` has solid unit test coverage for both `Reservation` and `CheckReport` modes. However, the reconciler integration test suites (`tests/envoygateway/extension_reconciler_test.go` and `tests/istio/extension_reconciler_test.go`) i believe currently only exercise the default `Reservation` mode.

It might be worth adding a test case to either suite where the `Kuadrant` CR is set to `spec.tokenRateLimiting.mode: CheckReport`, to verify end-to-end that the reconciler switches the `EnvoyExtensionPolicy` / `EnvoyFilter` from `ratelimit-reserve-service`/`ratelimit-commit-service` to `ratelimit-check-service`/`ratelimit-report-service`.

appreciate the work you have put into this thanks

jland-redhat:
> Question: is there a real use case for different reservation sizes across limits within the same policy? If not, policy-level reservation gets us 1 round-trip instead of N with no real loss.

Sorry for the delayed response, but yes there very much is a use case and for us a requirement that this be separated per limit.

From the MaaS perspective a user creates a "MaaSSubscriptoin" which has a list of models(HTTPRoutes) that they specify limits for. So a single Subscription can create limit rules for multiple Routes across the multiple Routes. 

And each of these subscriptions could have different users with very different expected request/responses. So each limit would need to be something that that Subscription's owner could define separately.

eguzki:
Thanks @jland-redhat, that makes sense and confirms per-limit reservation is needed here.

For the record, when each MaaS subscription's limit is scoped via `when` to its own subscriber traffic, there's no extra round-trip cost from keeping reservation per-limit: only that one subscriber's Reserve/Commit fires. The N-round-trip cost I raised only shows up when multiple limits' when conditions are simultaneously true for the same request.

So, keeping reservation per-limit as designed in RFC 0021. Thanks for clarifying the requirement.

eguzki:
> Thanks @eguzki for putting this together! The implementation is very thorough, clean, and adheres closely to RFC 0021.
> ### Suggestions
> #### 1. "Upgrades and compatibility" documentation
> 
> As discussed in #2225, adding an "Upgrades and compatibility" callout in `doc/overviews/token-rate-limiting.md` would be really helpful for cluster operators.
> 
> Because `Reservation` is the default mode, if the operator is upgraded before Limitador supports `Reserve`/`Commit` (or if Limitador is running with `--disable-reservations`), the `Reserve` calls will fail and, with the default failure mode (`allow`), requests will be let through without token counting.
> #### 2. End-to-end integration test for CheckReport mode
> 
> `tokenratelimit_workflow_test.go` has solid unit test coverage for both `Reservation` and `CheckReport` modes. However, the reconciler integration test suites (`tests/envoygateway/extension_reconciler_test.go` and `tests/istio/extension_reconciler_test.go`) i believe currently only exercise the default `Reservation` mode.
> 
> It might be worth adding a test case to either suite where the `Kuadrant` CR is set to `spec.tokenRateLimiting.mode: CheckReport`, to verify end-to-end that the reconciler switches the `EnvoyExtensionPolicy` / `EnvoyFilter` from `ratelimit-reserve-service`/`ratelimit-commit-service` to `ratelimit-check-service`/`ratelimit-report-service`.
> 
> appreciate the work you have put into this thanks

Thanks for both suggestions, picked up here.

1. Instead of documenting the upgrade footgun, made it self-detecting: Enforced=False (reason ReservationsDisabled) when mode is Reservation but Limitador has reservations disabled. Not concerned about compatibility issues between kuadrant and limitador. Kuadrant operator now ships limitador internally (so called umbrella operator). 
2. Added tests/istio/tokenratelimitpolicy_test.go and tests/envoygateway/tokenratelimitpolicy_test.go assert on the generated EnvoyFilter/EnvoyExtensionPolicy for both Reservation and CheckReport modes.

eguzki:
This is ready for review  @adam-cattermole

api/v1beta1/kuadrant_types.go
```diff
@@ -102,6 +111,33 @@ type KuadrantSpec struct {
...
+
+const (
+	// TokenRateLimitingModeReservation reserves an estimated token amount on
+	// request arrival and commits the actual usage on response, closing the
+	// check/report race window (see RFC 0021).
+	TokenRateLimitingModeReservation TokenRateLimitingMode = "Reservation"
+
+	// TokenRateLimitingModeCheckReport checks the limit (hits_addend=0) on
+	// request arrival and reports the actual usage on response.
+	TokenRateLimitingModeCheckReport TokenRateLimitingMode = "CheckReport"
```
adam-cattermole: We discussed this a little on the community call, I still think perhaps from the API perspective `CheckReport` is somewhat meaningless to a user (I think we've done a good job of not exposing semantics of the limitador API to users so far). I'm wondering what's a good alternative name for the 'legacy' style

adam-cattermole, line 132: We discussed this a little on the community call, I still think perhaps from the API perspective `CheckReport` is somewhat meaningless to a user (I think we've done a good job of not exposing semantics of the limitador API to users so far). I'm wondering what's a good alternative name for the 'legacy' style

doc/overviews/token-rate-limiting.md
```diff
@@ -24,6 +24,36 @@ This approach ensures accurate usage-based rate limiting where limits are enforc
...
 
+### Enforcement modes
+
+Because the real token cost of a request is only known *after* the upstream responds, the gateway has to decide what to do on the way in. This is controlled cluster-wide by the Kuadrant CR field `spec.tokenRateLimiting.mode`, which applies to every TokenRateLimitPolicy in the cluster:
+
+- **`Reservation`** (default): On request arrival the gateway *reserves* an estimated token amount from Limitador. If the reservation would exceed the limit, the request is rejected with `429` before it ever reaches the upstream. Once the upstream responds, the gateway *commits* the actual `usage.total_tokens` and releases the unused portion of the reservation. This closes the race window where many concurrent requests could each pass a zero-cost check before any of them reports usage. The reservation is designed per RFC [0021](https://github.com/Kuadrant/architecture/blob/main/rfcs/0021-token-rate-limit-reservations.md), and the reserved amount / hold time are tuned per limit with [`reservation`](../reference/tokenratelimitpolicy.md#reservation).
```
adam-cattermole, line 31: Wondering if in one of these docs we should expose the fact that default max fraction of 0.5 lets the user overshoot, looks like it's not mentioned here that default behaviour is to let the user go over the limit

api/v1beta1/kuadrant_types.go
eguzki: Agreed. Happy to accept any suggestion better than that.

doc/overviews/token-rate-limiting.md
eguzki: Good catch, but the framing needs a small correction IMO: it's not the reservation system that lets you go over the limit. What actually allows overshoot is the "check on request, report on response" accounting pattern used unconditionally. A held (non-zero) reservation system is exactly what minimizes it, when not preventing it. 

Should the doc call out this maxFraction trade-off explicitly (lower = more concurrency, more compounding risk; higher = less concurrency, smaller blast radius per gap)?

adam-cattermole: Sorry yeah I wasn't clear in my statement, what I meant was with a perfectly accurate measure of reservation vs actual usage the system lets you overshoot due to max fraction defaults - the max fraction overshoot is somewhat irrespective of the check on request/report on response. I think it'd be worth noting the trade-off plus default is to overshoot to require 2 requests minimum to consume the limit

tests/istio/extension_reconciler_test.go
```diff
@@ -254,6 +188,22 @@ var _ = Describe("Rate Limiting EnvoyFilter controller", func() {
...
+						GrpcMethod:  ptr.To(wasm.RateLimitReserveGrpcMethod),
+					},
+					wasm.RateLimitCommitServiceName: {
+						Type:        wasm.DynamicServiceType,
+						Endpoint:    kuadrant.KuadrantRateLimitClusterName,
+						FailureMode: wasm.RatelimitCommitServiceFailureMode(&logger),
+						Timeout:     ptr.To(wasm.RatelimitCommitServiceTimeout()),
+						GrpcService: ptr.To(wasm.KuadrantRateLimitGrpcService),
+						GrpcMethod:  ptr.To(wasm.RateLimitCommitGrpcMethod),
+					},
```
adam-cattermole: nit: wonder if there's a better way for us to define these so it's not repeated every test, but realistically doesn't need solving now

internal/wasm/action_spec.go
```diff
@@ -564,6 +652,37 @@ func buildAuthOnReply(name string) []Action {
...
+//     filter-local store path for the reservation_id (see reservationStorePath).
+//   - reservation.amount: tokens to reserve on request arrival (uint).
+//   - reservation.ttl: how long the reservation is held (duration).
+//   - reservation.actual_amount: actual tokens to commit on response (uint).
+const (
+	reservationIDAttr           = "reservation.id"
+	reservationAmountAttr       = "reservation.amount"
+	reservationTTLAttr          = "reservation.ttl"
+	reservationActualAmountAttr = "reservation.actual_amount"
+)
```
adam-cattermole: I'm not entirely against it but something is telling me this model of wiring this info as conditionaldata to extract it later is trying to fit this into the original box with the `ratelimit.domain`/`ratelimit.hits_addend` convention instead of modelling with current constraints. I'm wondering if you thought/tried an alternative here?

adam-cattermole: I guess I'm wondering if a typed field on the `ActionSpec` like `*ReservationSpec` would be a better means of transport for all of this information than wiring it in through the conditional data

internal/cel/kuadrant_validator.go
```diff
@@ -117,6 +119,24 @@ func ValidateWasmActionSpec(spec wasm.ActionSpec, validator *Validator) error {
...
 		}
 	}
+	if amount := spec.ReservationAmountCEL(); amount != "" {
+		ast, err := validator.Validate(pol, amount)
+		if err != nil {
+			return err
+		}
+		if ast.OutputType() != cel.UintType {
+			return fmt.Errorf("reservation amount expression must evaluate to uint, got %s", ast.OutputType())
+		}
```
adam-cattermole: nit: I think this validation can't fire because amount is wrapped by `uint()` so has to evaluate to uint or won't compile

api/v1beta1/kuadrant_types.go
adam-cattermole: Yeah not got any great ideas here; `BestEffort`, `Optimistic` or `Postpaid`? wdyt - got any suggestions for an alternative?

doc/overviews/token-rate-limiting.md
eguzki: Added clarification paragraph highlighting the trade-off.

tests/istio/extension_reconciler_test.go
eguzki: Added helper function and refactored a bit the tests

internal/wasm/action_spec.go
eguzki: Agreed. Reserve/Commit are excluded from `mergeableServices` (one `ActionSpec` always corresponds to exactly one limit). There's no need for `reservation.id`/`amount`/`ttl`/`actual_amount` to be predicate gated, per conditional block data the way `ratelimit.domain`/`hits_addend` genuinely are for. Added a typed `*ReservationSpec` field on `ActionSpec` instead

doc/overviews/token-rate-limiting.md
adam-cattermole: Looks great thanks!

tests/istio/extension_reconciler_test.go
adam-cattermole: Awesome, looks perfect - reduced a lot of code

internal/cel/kuadrant_validator.go
eguzki: agreed. No need to validate that.

api/v1beta1/kuadrant_types.go
eguzki: I like `Optimistic`. Reservation is pessimistic (hold capacity upfront). I'll go with that

internal/wasm/action_spec.go
```diff
@@ -305,8 +362,80 @@ func (s ActionSpec) buildRateLimit(responseVar string, isGuard bool, label strin
...
+//   - no reservation, usage parseable: commit with an empty reservation_id,
+//     which Limitador treats like a plain Report (RFC 0021).
+func (s ActionSpec) buildCommit() *GrpcAction {
+	var id, rawActualAmount string
+	if s.Reservation != nil {
+		id = s.Reservation.ID
+		rawActualAmount = s.Reservation.ActualAmount
+	}
+	storePath := reservationStorePath(id)
+	hasReservation := fmt.Sprintf("has(%s)", storePath)
```
adam-cattermole: I think it might be better to have this as:

```suggestion
	hasReservation := fmt.Sprintf("%s != null", storePath)
```

I think they should behave the same but after looking at the wasm-shim's source there might be a minor issue with siblings and `has()`

adam-cattermole: To be clearer, `has` works great / fine for proto fields so it's perfect for the `has(reserve_response.reservation_id)`, but I think it might have unintended behaviour for accessing the data from the store task

api/v1beta1/kuadrant_types.go
eguzki: Replaced

doc/overviews/token-rate-limiting.md
```diff
@@ -24,6 +24,38 @@ This approach ensures accurate usage-based rate limiting where limits are enforc
...
+
+  Reservation still allows some overshoot, independently of the race above: Limitador caps every `Reserve` call to at most `spec.reservations.maxFraction` of a limit's `max_value` (a Limitador CR field, default `0.5`). So no single reservation can ever claim the whole limit — it takes at least `1/maxFraction` requests' worth of held reservations (2, by default) before Limitador itself starts rejecting on capacity. Because each of those holds is only an estimate of real usage, actual consumption can land above what was reserved, so the achievable overshoot scales with the number of concurrent holders allowed under `maxFraction`. A lower `maxFraction` permits more concurrent in-flight requests before capacity-rejects kick in but widens that overshoot window; a higher `maxFraction` tightens the bound at the cost of concurrency headroom.
```
adam-cattermole, line 33: One other thought, might be worth quickly mentioning maxTTL overriding ttl in the TRLP but we can also update docs later

api/v1alpha1/tokenratelimitpolicy_types.go
```diff
@@ -222,10 +223,39 @@ type TokenLimit struct {
...
+// Reservation configures token reservation behavior for a TokenLimit, used when
+// the Kuadrant CR is in Reservation mode (see RFC 0021).
+type Reservation struct {
+	// Amount is either a literal integer number of tokens to reserve on request
+	// arrival, or a CEL expression evaluating to the number of tokens (uint).
+	// Defaults to 0 when omitted, which reserves no capacity: Limitador
+	// short-circuits amount-0 reservations, so this limit behaves like
+	// Optimistic unless amount is set explicitly to a non-zero estimate.
+	// +optional
+	// +kubebuilder:validation:XIntOrString
```
adam-cattermole, line 248: nit: I wonder if we can have a rule that enforces if integer it's >= 0

go.mod
```diff
@@ -192,3 +192,8 @@ require (
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.3.2 // indirect
 )
+
+// Pinned ahead of a tagged release to pick up the Reservations CRD field (limitador-operator#283),
+// Remove this replace once a limitador-operator release ships Reservation fields.
+// git describe --tags --match 'v*' returns v0.10.0-237-gece3f4b5b5a3, meaning v0.10.0 is the highest tag actually reachable from this commit, this is not a regression.
+replace github.com/kuadrant/limitador-operator => github.com/kuadrant/limitador-operator v0.10.1-0.20260911154343-ece3f4b5b5a3
```
adam-cattermole, line 199: Cool I think lets make sure to follow up and unpin this as soon as we cut a limitador-operator release

adam-cattermole:
Changes are looking good, thanks, just going to run verification!

internal/wasm/action_spec.go
eguzki: replaced

doc/overviews/token-rate-limiting.md
eguzki: Added a parallel paragraph for the ttl

adam-cattermole (approved):
Changes look good, thanks for addressing the feedback!

api/v1alpha1/tokenratelimitpolicy_types.go
eguzki: Filing a GitHub issue to track this properly rather than bolting on a partial fix now.

PR #2281, opened 2026-09-17
fix: add dns_refresh_rate and outlier_detection to Authorino ext_authz cluster

### LINKED ISSUE #2280: Add outlier_detection to Authorino and Limitador gRPC clusters
Companion fix to https://github.com/Kuadrant/authorino-operator/issues/362, which makes the Authorino authorization service headless so DNS returns individual pod IPs rather than a single ClusterIP VIP.

## Problem

When Authorino pods restart, their pod IPs change. Until the next DNS refresh, Envoy continues routing gRPC ext_authz streams to stale pod IPs, causing a burst of \`UNAVAILABLE\` (gRPC code 14) → HTTP 500 errors. The same applies to Limitador, which has been headless since its introduction.

## Fix

Add Envoy outlier detection to the Authorino ext_authz and Limitador rate-limit clusters. Outlier detection ejects endpoints that accumulate consecutive gateway failures (gRPC UNAVAILABLE / connection errors) between DNS refresh cycles, reducing the window where traffic is routed to a dead pod.

Scoped to \`authClusterPatch\` and \`rateLimitClusterPatch\` only — both are pure gRPC services. Not applied via the shared \`buildClusterPatch\` to avoid affecting tracing and extension clusters, which may be HTTP-based.

Note: \`dns_refresh_rate\` was considered but is a no-op (Envoy default is already 5s). \`dns_lookup_family: V4_ONLY\` was considered but rejected as it would break IPv6-capable clusters.

thomasmaas:
## Summary

- Adds \`dns_refresh_rate: 5s\` to bound the window where Envoy routes to stale pod IPs after Authorino pod restarts
- Adds \`outlier_detection\` to automatically eject dead endpoints after 5 consecutive gateway failures, rather than waiting passively for the DNS refresh
- Adds \`dns_lookup_family: V4_ONLY\` to prevent inconsistent endpoint sets on dual-stack clusters

## Scope

\`buildClusterPatch\` is shared by both the Authorino ext_authz cluster and the Limitador rate-limit cluster (\`rateLimitClusterPatch\` calls it directly). These improvements apply to both upstreams. Note that Limitador's service is already headless (\`ClusterIP: None\`) in limitador-operator — DNS already returns pod IPs for Limitador, so \`dns_refresh_rate\` and \`outlier_detection\` provide additional resilience there too.

## Context

The \`STRICT_DNS\` cluster generated for Authorino's gRPC ext_authz service previously had no controls on DNS refresh timing, endpoint health checking, or address family selection. When Authorino pods restart and get new IPs, Envoy continued routing to the old IPs until an uncontrolled DNS refresh happened, causing a burst of \`UNAVAILABLE\` (gRPC code 14) → HTTP 500 errors.

This is a companion fix to [kuadrant/authorino-operator#363](https://github.com/Kuadrant/authorino-operator/pull/363), which makes the authorization service headless so DNS returns individual pod IPs rather than a single ClusterIP VIP. Together they enable gRPC stream distribution across Authorino replicas and ensure pod IP churn during rolling updates is handled gracefully.

## Test plan

- [ ] Unit tests pass (\`go test ./internal/controller/...\`)
- [ ] Verify the generated EnvoyFilter contains the new cluster fields: \`kubectl exec -n istio-system <gateway-pod> -- curl -s localhost:15000/config_dump | jq '.configs[] | select(.["@type"] | contains("clusters")) | .dynamic_active_clusters[] | select(.cluster.name == "kuadrant-auth-service")'\`

Fixes #2280


thomasmaas:
The two `bare_k8s` test failures (`Limitador CR is ready` timeout and `monitors are not created at first`) are unrelated to this PR.

This PR only modifies `buildClusterPatch` in `internal/controller/cluster_patch_helpers.go`, which is called when generating Envoy cluster config patches for gateway EnvoyFilters. The `bare_k8s` suite runs without any gateway provider — no EnvoyFilters are created, so this code path is never exercised. The failures are timing-sensitive tests that timed out after 197s in CI.

Triggering a re-run.

thomasmaas:
The `controllers` integration test failures (24 AuthPolicy timeouts across istio/istioctl, istio/sail, envoygateway) are a CI-wide outage — the same 24 tests are timing out on unrelated PRs running concurrently (e.g. `sdk-store-action` branch). Not caused by this PR, which only modifies `buildClusterPatch` in `internal/controller/cluster_patch_helpers.go`.

internal/controller/cluster_patch_helpers.go
```diff
@@ -4,10 +4,19 @@ package controllers
...
+		"type":              "STRICT_DNS",
+		"connect_timeout":   "1s",
+		"dns_refresh_rate":  "5s",
+		"dns_lookup_family": "V4_ONLY",
+		"outlier_detection": map[string]any{
+			"consecutive_gateway_failure":           5,
+			"enforcing_consecutive_gateway_failure": 100,
+			"interval":                              "10s",
+			"base_ejection_time":                    "30s",
+			"max_ejection_percent":                  100,
```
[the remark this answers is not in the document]
thomasmaas: Disagree — these settings are not Authorino-specific. `buildClusterPatch` is intentionally shared between both the Authorino ext_authz cluster and the Limitador rate-limit cluster (`rateLimitClusterPatch` delegates to it directly). All three additions apply equally to both upstreams:

- `dns_refresh_rate: 5s` — Limitador's service has been headless (`ClusterIP: None`) from the start, so DNS already returns pod IPs. Fast refresh matters equally for Limitador pod restarts.
- `dns_lookup_family: V4_ONLY` — prevents dual-stack resolution inconsistency for both clusters.
- `outlier_detection` — generic Envoy endpoint health checking; ejects stale pod IPs after repeated failures. Beneficial for both Authorino and Limitador rolling updates.

The PR description calls this out explicitly. Splitting the config would add complexity without benefit.

internal/controller/cluster_patch_helpers.go
```diff
@@ -4,10 +4,19 @@ package controllers
 // HTTP/2 is always enabled since all current callers require it (gRPC upstreams).
 func buildClusterPatch(clusterName, host string, port int, mTLS bool) map[string]any {
```
adam-cattermole, line 5: So this effects all callers of this, ratelimiting, auth plus user configured tracing clusters and any extension registered clusters - I'm wondering whether that was the intended scope of the change

internal/controller/cluster_patch_helpers.go
```diff
@@ -4,10 +4,19 @@ package controllers
...
 	base := map[string]any{
-		"name":            clusterName,
-		"type":            "STRICT_DNS",
-		"connect_timeout": "1s",
-		"lb_policy":       "ROUND_ROBIN",
+		"name":              clusterName,
+		"type":              "STRICT_DNS",
+		"connect_timeout":   "1s",
+		"dns_refresh_rate":  "5s",
```
adam-cattermole: According to docs [here](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-dns-refresh-rate), the `dns_refresh_rate` default is 5000ms so this is a no-op?

internal/controller/cluster_patch_helpers.go
```diff
@@ -4,10 +4,19 @@ package controllers
...
+		"dns_lookup_family": "V4_ONLY",
```
adam-cattermole: Do we support IPv6 only clusters? wouldn't this break them?

internal/controller/cluster_patch_helpers.go
adam-cattermole: Similarly, looks like [the default](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/outlier_detection.proto#config-cluster-v3-outlierdetection) of `outlier_detection` already sets `interval` to 10s, `base_ejection_time` to 30s, and `consecutive_gateway_failure` to 5. Defining this block though does enable outlier detection which also silently turns on the 5xx detector (defaults 5 at 100%) which I think would both trigger at the same time or sooner than the gateway_failure trigger.

thomasmaas:
Thanks @adam-cattermole, restructuring the pr. I think only `outlier_detection` and then scoped to auth and rate limiting  will survive.

internal/controller/ratelimit_workflow_helpers.go
```diff
@@ -106,23 +103,22 @@ func RateLimitClusterName(gatewayName string) string {
...
-}
-
-// ToActionScope converts a LimitNamespace to an ActionScope.
-func (ln LimitNamespace) ToActionScope() ActionScope {
-	return ActionScope(ln)
+	patch := buildClusterPatch(kuadrant.KuadrantRateLimitClusterName, host, port, mTLS)
+	// Limitador is a pure gRPC service. Outlier detection ejects pod IPs that
+	// accumulate consecutive gateway failures between DNS refresh cycles.
+	patch["outlier_detection"] = grpcOutlierDetection()
+	return patch
```
adam-cattermole: I'm not against this outlier detection, but is this necessary? I know we moved to a headless service for authorino but didn't for limitador - what's this expected to solve? could this be detrimental for limitador without a headless service?

internal/controller/cluster_patch_helpers.go
```diff
@@ -39,6 +39,28 @@ func buildClusterPatch(clusterName, host string, port int, mTLS bool) map[string
...
 }
 
+// grpcOutlierDetection returns an Envoy outlier detection configuration for pure gRPC
+// upstreams. It ejects pod endpoints that accumulate consecutive gateway failures
+// (gRPC UNAVAILABLE / connection errors) between DNS refresh cycles, reducing the
+// window where traffic is routed to a dead pod after a rolling update or restart.
+// Only safe to use on gRPC upstreams: defining this block also activates the
+// consecutive_5xx detector (Envoy default: 5 errors, 100% enforcing), which is
+// harmless for gRPC because gRPC always returns HTTP/2 status 200 — errors are
+// carried in trailers, not HTTP status codes.
```
adam-cattermole: nit: I think envoy maps gRPC status codes to HTTP for outlier purposes, so the 5xx detector and gateway_failure detector are doing the same thing, with 5xx firing at the same point or sooner. Might just be worth making this comment accurate to avoid confusion if we change these values later?

internal/controller/ratelimit_workflow_helpers.go
thomasmaas: I was working under the premise that limitador is a headless service https://github.com/Kuadrant/limitador-operator/blob/main/pkg/limitador/k8s_objects.go#L49. Not true?

adam-cattermole: Ah you're right, my mistake here

adam-cattermole (approved):

eguzki (approved):
LGTM

nit: `tests/istio/rate_limit_cluster_reconciler_test.go` (Istio's rate-limit cluster test) wasn't given a `HaveKey("outlier_detection") `assertion the way the Istio auth cluster test was. Same pattern, just missing for symmetry/coverage

eguzki:
@thomasmaas would this change be applied on upgrade scenario? My concern would be having the authorino change (headless service) applied but not this on existing kuadrant deployments.

thomasmaas:
Yes, on operator pod restart (upgrade), every existing object goes through AddFunc → reconcile → regenerate EnvoyFilter, and the new cluster config is applied automatically. Unless you mean something different?

PR #2302, opened 2026-09-22
Extensions protocol version validation

### LINKED ISSUE #2278: Validate Extensions version
Currently the version is part of the Handshake proto, but isn't validate or passed by the client. Lets add this with an initial version number so that breaking changes on the operator-side can enforce an upgrade on the extensions side.

Look into options for exposing the version mismatch to the user

adam-cattermole:
Resolves #2278

## Summary by CodeRabbit

- **New Features**
  - Added protocol version negotiation during extension handshakes.
  - Compatible patch-level versions can now connect successfully, while incompatible versions are rejected.
  - Handshake responses now report the negotiated protocol version and provide clear rejection reasons, including incompatible version, invalid request, unauthorized access, and unavailable service.

- **Bug Fixes**
  - Improved session handling so terminal handshake failures are surfaced promptly instead of being retried indefinitely.
  - Session supervision now responds correctly to handshake failures and cancellation.


silvi-t (approved):
`protocol.Version (pkg/extension/protocol/protocol.go)` is a hand-maintained constant with nothing enforcing that it gets bumped — no CI check or test fails if `kuadrant.proto`/handshake semantics change without a matching version update. Could we at least add a short comment next to the const documenting when/how to bump it (e.g. "bump minor on any breaking change to kuadrant.proto or handshake semantics "), so it doesn't get forgotten on a future change? If you think of something stronger than a doc comment (e.g. a CI check tying proto changes to a version bump), that'd be a great addition too.

This is a nit — the PR looks good overall and I'm approving.

adam-cattermole:
> protocol.Version (pkg/extension/protocol/protocol.go) is a hand-maintained constant with nothing enforcing that it gets bumped — no CI check or test fails if kuadrant.proto/handshake semantics change without a matching version update. Could we at least add a short comment next to the const documenting when/how to bump it (e.g. "bump minor on any breaking change to kuadrant.proto or handshake semantics "), so it doesn't get forgotten on a future change? If you think of something stronger than a doc comment (e.g. a CI check tying proto changes to a version bump), that'd be a great addition too.

I think the actual proto bumping should come with a version change to the proto (we're on v1 right now) to guard that side, so it'd be moreso to enforce breaking changes to 'semantics' of the extension/operator where the wire format _doesn't_ change. It might be we don't actually need this and could drop the validation / denial on the operator side and this could become a logging/informational piece - I guess I'm just trying to include it as a precaution

pkg/extension/protocol/protocol.go
```diff
@@ -0,0 +1,24 @@
+package protocol
+
+import (
+	"fmt"
+
+	"github.com/Masterminds/semver/v3"
+)
+
+const Version = "0.1.0"
```
adam-cattermole: @silvi-t this was set to `1.0.0` before and has become `0.1.0`, so it'll need the `PipelinePolicyExtension` to re-consume the go.mod or it'll be denied on handshake. I think we'll only bump it to `1.0.0` on extensions GA

silvi-t:
> > protocol.Version (pkg/extension/protocol/protocol.go) is a hand-maintained constant with nothing enforcing that it gets bumped — no CI check or test fails if kuadrant.proto/handshake semantics change without a matching version update. Could we at least add a short comment next to the const documenting when/how to bump it (e.g. "bump minor on any breaking change to kuadrant.proto or handshake semantics "), so it doesn't get forgotten on a future change? If you think of something stronger than a doc comment (e.g. a CI check tying proto changes to a version bump), that'd be a great addition too.
> 
> I think the actual proto bumping should come with a version change to the proto (we're on v1 right now) to guard that side, so it'd be moreso to enforce breaking changes to 'semantics' of the extension/operator where the wire format _doesn't_ change. It might be we don't actually need this and could drop the validation / denial on the operator side and this could become a logging/informational piece - I guess I'm just trying to include it as a precaution

Thanks for the clarification. In that case I think just adding a comment like:
```go
  // Version is the extension protocol version negotiated during handshake.
  // Bump the minor version for any behavioral/semantic change to the
  // handshake or extension-operator contract. Bump to 1.0.0 at extensions GA.
  const Version = "0.1.0"
 ```

  would be enough. Either way, not blocking. You can leave it as-is if you prefer.

pkg/extension/protocol/protocol.go
silvi-t: I know I've already done this locally. I'll create a PR for `internal-extensions` once it's merged into the main.

adam-cattermole:
Sure thing, let me add a comment there

adam-cattermole:
Just added it now, should make it's use a little clearer, thanks for the feedback @silvi-t

pkg/extension/protocol/protocol.go
adam-cattermole: Awesome - just trying to let you know in advance while I'm breaking things

silvi-t (approved):
LGTM

